PR #70, opened 2021-04-09
feat: add `@vue/apollo` example implementation

danielroe:
This is now updated to work with new examples layout, and with a sample graphql server backend to show it in action.

We currently need to transpile `@vue/apollo-composables` as it does not provide a CJS build but only AMD or ESM (see https://github.com/vuejs/vue-apollo/issues/1081).

closes  nuxt/nuxt.js#11729

### DIFF
--- a/examples/vue-apollo/nuxt.config.js
+++ b/examples/vue-apollo/nuxt.config.js
@@ -1,0 +1,13 @@
+export default {
+  alias: {
+    '@apollo/client/core': require.resolve('@apollo/client/core/index.js'),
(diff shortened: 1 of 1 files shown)


tobiasdiez:
@pi0 @danielroe I would be interested in providing an up-to-date GraphQL example. 

What is the problem with this PR? 

I was also wondering about the best way to create the graphql server with h3. This example here uses the express middleware which seems to be overkill. Maybe its worth to create a special h3 middleware and add it to https://www.apollographql.com/docs/apollo-server/integrations/middleware/?

PR #625, opened 2021-10-02
feat(nuxt3, bridge): useRuntimeConfig

pi0:
### 🔗 Linked issue

nuxt/nuxt.js#10943 (follow up nuxt/framework#248)

### ❓ Type of change

- [ ] Documentation (updates to the documentation or readme)
- [ ] Bug fix (a non-breaking change that fixes an issue)
- [ ] Enhancement (improving an existing functionality like performance)
- [x] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

- [nitro] Deep Freeze nitro's exposes `#config` and use named exports
- [nuxt3] Use `ssrContext.runtimeConfig*` and hydrate public config as `payload.config`
- [nuxt3/bridge] Expose `useRuntimeConfig` composable   
  - Can be used within nuxt plugins and setup lifecycle
  - Frozen on server side to avoid leaking data
  - Reactive on client side to allow applying modifications on app (/cc @Tahul)

Todo later: Also expose `useRuntimeConfig` for nitro middleware for code consistency. Currently it is possible by using `import config from '#config'` but can be easier to read with something like `import { useRuntimeConfig } from '#nitro'`

### 📝 Checklist

### DIFF
--- a/packages/nitro/src/runtime/app/config.ts
+++ b/packages/nitro/src/runtime/app/config.ts
@@ -27,4 +27,5 @@ function deepFreeze (object: Record<string, any>) {
       deepFreeze(value)
     }
   }
+  return Object.freeze(object)
 }


duyleekun:
Hi @pi0, I'm currently on `nuxt3@^3.0.0-27313139.1c88580` with `ssr: true`. I'm trying to use `publicRuntimeConfig` but I think I found an unintended behavior which I'm not sure if it's an issue or not.

```js
// nuxt.config.ts
publicRuntimeConfig: {
    apiEndpoint: process.env.API_ENDPOINT || 'asdf',
  },
```

```js
// .output/server/chunks/app/render.mjs
const _runtimeConfig = {"public":{"app":{"basePath":"/","assetsPath":"/_nuxt/","cdnURL":null},"apiEndpoint":"asdf"},"private":{}};
console.log(process.env)
for (const type of ["private", "public"]) {
  for (const key in _runtimeConfig[type]) {
    _runtimeConfig[type][key] = destr(process.env[key] || _runtimeConfig[type][key]);
  }
}
```

After running `nuxi build` and looking at `.output/server/chunks/app/render.mjs`, I found out that we should pass env by running `apiEndpoint=qwer node .output/server/index.mjs` instead of `API_ENDPOINT=qwer node .output/server/index.mjs`. I tested only the former syntax works.

Is it correct that nuxt3 no longer process the `nuxt.config.ts` at runtime?

danielroe:
@duyleekun By design, we don't need `nuxt.config` any more after a Nitro build - and yes, the behaviour you're describing is by design, though I think we might improve the way of overriding runtime config. For now, I'd recommend using the following pattern:
```js
// nuxt.config.ts
publicRuntimeConfig: {
  API_ENDPOINT: process.env.API_ENDPOINT || 'asdf',
}
```

PR #773, opened 2021-10-12
feat(bridge): init function for `useState`

antfu:
nuxt/nuxt.js#11916


pi0:
Thanks! Do you mind also updating docs for `useFetch`?

PR #823, opened 2021-10-12
fix(app): filter out falsy plugins

antfu:
fix nuxt/bridge#249

### DIFF
--- a/packages/nuxt3/src/app/nuxt.ts
+++ b/packages/nuxt3/src/app/nuxt.ts
@@ -136,13 +136,15 @@ export async function applyPlugins (nuxt: NuxtApp, plugins: Plugin[]) {
 export function normalizePlugins (_plugins: Array<Plugin | LegacyPlugin>) {
   let needsLegacyContext = false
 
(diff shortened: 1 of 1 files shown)


marshallswain:
While the proposed fix by @antfu likely prevents errors, it doesn't seem like this is the correct fix, all respects intended.  When you install with npm, there are four plugins.  In these screenshots, take a look at the `_plugins` variable in the `Closure(normalizePlugins)` frame of the debugger: 

Screen Shot 2021-10-13 at 12 40 36 AM

When you install with pnpm and shamefully hoist enabled, the fourth plugin is undefined.  It's likely that this will lead to behavioral differences.

Screen Shot 2021-10-13 at 12 41 58 AM

It would probably be more useful to provide some beneficial messaging, but not before finding the cause of this missing plugin.

marshallswain:
It looks like the missing import is `/node_modules/@nuxt/nitro/dist/runtime/app/nitro.client.mjs`:

Screen Shot 2021-10-13 at 12 56 22 AM

PR #846, opened 2021-10-12
fix(bridge): add head type without circular type reference

Patrick-Remy:
### 🔗 Linked issue

nuxt/bridge#264

### ❓ Type of change

### 📚 Description

Redeclaring `ConfigScheme` does already add the props with the desired types in `NuxtConfig`
https://github.com/nuxt/framework/blob/645f8f93977a735b73b4ca40e2174246f3013cfa/packages/kit/src/types/config.ts#L19

There was a change by https://github.com/nuxt/framework/pull/844 but that does leave `head` not as good declared as before.

Resolves nuxt/bridge#264

### 📝 Checklist

- [xI have linked an issue or discussion.
- [] I have updated the documentation accordingly.

### DIFF
--- a/packages/bridge/types.d.ts
+++ b/packages/bridge/types.d.ts
@@ -1,5 +1,6 @@
 import type {} from '@nuxt/nitro'
 import type { NuxtConfig as _NuxtConfig } from '@nuxt/kit'
+import type { ConfigSchema as _ConfigSchema } from '@nuxt/kit/schema/config'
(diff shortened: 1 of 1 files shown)


pi0:
I see but extending kit types with legacy nuxt2 is probably not the best idea and we also need to extend exported `NuxtConfig` from bridge without implicit shim to kit.

Patrick-Remy:
Okay, I got it! Thanks for your fast fix.

PR #942, opened 2021-10-13
fix(kit): pass `options` object to lodash templates

Diizzayy:
### 🔗 Linked issue

Resolves [#936](https://github.com/nuxt/nuxt.js/issues/11962)

### ❓ Type of change

### 📚 Description

This change improves Developer Experience.

Simplifies passing options for a plugin to be interpolated by lodash.template

example: allows you to simply pass moduleOptions like so
```
  configKey: 'example',
  setup(options) {
    addPluginTemplate({
      src: path.resolve(__dirname, 'plugin.ts'),
      mode: 'client',
      options,
    })
  },
})
```
or even `options: { ...options }`. As opposed to **requiring it to be passed like this** `options: { options },` 

### 📝 Checklist

### DIFF
--- a/packages/kit/src/module/utils.ts
+++ b/packages/kit/src/module/utils.ts
@@ -259,7 +259,7 @@ export function addVitePlugin (plugin: VitePlugin, options?: ExtendViteConfigOpt
 }
 
 export async function compileTemplate (template: NuxtTemplate, ctx: any) {
-  const data = { ...ctx, ...template.options }
+  const data = { ...ctx, options: template.options }
   if (template.src) {
     try {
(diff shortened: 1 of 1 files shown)


pi0:
Hey @richardeschloss currently there is not an easy way to trigger rebuild when template options are changed but it is preferred if you move away from lodash templates to either:
- using non template plugins + `runtimeConfig` (example: https://stackblitz.com/edit/github-jauvhr?file=modules%2Fhello.ts)
- new `getContents` API (example: https://github.com/nuxt/framework/blob/main/packages/nuxt3/src/meta/module.ts#L33)

PR #952, opened 2021-10-13
chore: alias types for `@nuxt/kit` src

antfu:
Not really sure why, but my VS Code can't find declarations for `@nuxt/kit`, probably because #820? But either way I guess it makes sense to alias the local package for typing in developement.


pi0:
> Ah, didn't know that. Does that mean I need to run stub after every installation?

Stub basically creates placeholders with exception of kit schema which is auto-generated and need restub if updting

PR #969, opened 2021-10-13
feat: auto-import for composables

antfu:
close nuxt/nuxt.js#11862

- [x] Named exports
- [x] Default exports (use file's basename)
- [x] Warn for name collisions
- [x] Watch for changes - This would involve a few other modules for the watcher support (e.g. the types generator)


pi0:
>  I decide to not support it. It could be confusing to auto-import composables/foo.ts as foo, and force users to the files as the function name is not ideal to me.

I think yet would be a good option to support default exports in addition to named exports. This convention, similar to component nameing, can be useful to quickly find composables based on file name (kindly please use RFC or Issue discussions to agree one something together before implementation 🙏 )

pi0:
@antfu We have many conflicts. Do you mind to rework it with new hooks and mlly utils?

PR #1070, opened 2021-10-15
docs: add warning that pages need single root element

danielroe:
### 🔗 Linked issue

closes nuxt/nuxt.js#12088

### ❓ Type of change

### 📚 Description

As currently implemented, pages need a single root element or there are issues when navigating backwards/forwards throughout an app (not to mention a warning in the browser dev console).

I prefer the user to have to wrap a page in a single root element because it means they are more in control of the HTML structure of their app, but we could also change this behaviour by automatically wrapping pages in a div (though the element name and attributes will then be out of the user's control).

### 📝 Checklist


dacoto97:
When we have pages with more than one root element it doesn't work.

image

danhogan:
> Hi @danhogan But this is not a bug? Or will it be like this forever.

I think things are working as intended, there just isn't a warning or anything to clearly tell you what's going wrong. Looks like that's exactly what this PR is attempting to resolve.

PR #1176, opened 2021-10-18
feat: auto-import for composables

antfu:
Close nuxt/nuxt.js#11862

Continue nuxt/framework#969

Followup nuxt/nuxt.js#12789

- Update docs
- Add new `autoImports:dirs` hook and `autoImports: { dirs }` option
- Auto  scan `~/composables` with both default exports and named exports (using static ESM syntax analyzes)


pi0:
Followup #1273

Issue #941, opened 2021-10-13
Error when using Vite with Nuxt Bridge


DerYeger:
### Environment

- Operating System: `Windows_NT`
- Node Version:     `v16.5.0`
- Nuxt Version:     `2.16.0-27226092.034b9901`
- Package Manager:  `Yarn`
- Bundler:          `Webpack`
- User Config:      `bridge`, `target`, `head`, `css`, `plugins`, `components`, `buildModules`, `modules`, `build`, `env`, `colorMode`, `fontLoader`, `i18n`, `loading`, `pwa`, `sitemap`
- Runtime Modules:  `@nuxtjs/i18n@7.0.3`, `@nuxtjs/pwa@3.3.5`, `@nuxtjs/sitemap@2.4.0`
- Build Modules:    `@nuxtjs/stylelint-module@4.0.0`, `@nuxtjs/color-mode@2.1.1`, `nuxt-font-loader@1.1.3`, `@nuxt/bridge@3.0.0-27234503.d5127e9`


### Describe the bug

The following error occurs when running `nuxt dev` with Nuxt Bridge installed and Vite enabled (see the logs section below for the entire output):

`node_modules/@nuxt/bridge/dist/runtime/vue.mjs:1:16: error: Could not resolve "#vue" (mark it as external to exclude it from the bundle)`the 

### Reproduction

1. Install Nuxt Bridge according to the instructions
2. Use the following `nuxt.config.ts`:

  ```ts
  export default defineNuxtConfig({
    bridge: {
      vite: true,
    },
    // additional configuration
  }
  ```

3. Run `nuxt dev`
  
  Repository with reproduction: https://github.com/DerYeger/noted/tree/feat/nuxt-bridge

### Additional context

_No response_

### Logs

```shell
❯ yarn dev
yarn run v1.22.15
$ nuxt
�  Vite mode is experimental and some nuxt modules might be incompatible                                                                                                                                                                                                                                14:17:35
    If found a bug, please report via https://github.com/nuxt/vite/issues with a minimal reproduction.

 ERROR  (node:17196) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time                                                                                                                                                                                   14:17:35
(Use `node --trace-warnings ...` to show where the warning was created)


   ╭───────────────────────────────────────╮
   │                                       │
   │   Nuxt @ v2.16.0-27226092.034b9901    │
   │                                       │
   │   ▸ Environment: development          │
   │   ▸ Rendering:   server-side          │
[...]

PR #1180, opened 2021-10-18
chore: move `nuxt-vite` to bridge

antfu:
- [x] Move base from [nuxt/vite](https://github.com/nuxt/vite)
- [x] `vue-template-compiler` hoisting issue
- [x] Resolving `#vue` alias (#941)
- [x] `store.js` template broken
- [x] Fix webpack4
- [x] "./dist/index.mjs" export in "hookable" package` on vite build
- [x] Fix CI
- [x] Working client side
- [x] Working prod build

Todo later:
- [ ] Update mocks with unenv
- [ ] Opt-in legacy 
- [ ] Hide invalid component name warnings (does it happens for external package too?)
- [ ] Remove defaultexport workaround

Braking changes from `nuxt-vite`:
- Bridge and nitro need to be enabled in order to work
- Workaround removed for i18n alias


antfu:
Actually not too much, it's already working locally. I just need to make sure the CI passes and do a bit more tests.

PR #1296, opened 2021-10-20
fix(nitro): filter alias in `/_vfs` listing

antfu:
Before
image

After
image


pi0:
Thanks! I guess i finally have to revert back temporary `#` vfs aliases as well. It was due to some resolution issues between webpack and vite.

PR #1612, opened 2021-10-31
feat(nuxt3): support async function as nuxt.config

### LINKED ISSUE #1611: feature request: async function as nuxt.config
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.12.0`
- Nuxt Version:     `3.0.0-27258467.4e424d0`
- Package Manager:  `npm@8.1.0`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Describe the bug

I would like to define nuxt configs by executing an async function as is possible in Nuxt v2..!!
It is useful to fetch and set the configs asynchronously.

```ts
import { defineNuxtConfig } from "nuxt3";

export default async () => {
  const publicRuntimeConfig = await fetchConfigs();

  return defineNuxtConfig({
    publicRuntimeConfig
  });
};

async function fetchConfigs() {
  await new Promise((resolve) => setTimeout(resolve, 1000));

  return {
    MESSAGE: "HELLO WORLD"
  };
}
```

Is it acceptable for you?

### Reproduction

https://codesandbox.io/s/nuxt3-config-function-sqdep?file=/nuxt.config.ts

### Additional context

I think these are related codes.

### v2

https://github.com/nuxt/nuxt.js/blob/db11f7bf937cdbb96ff0067bdee1aa153978e4e0/packages/config/src/load.js#L61-L71

### v3

https://github.com/nuxt/framework/blob/d6ad4f3b16d124f01b21bb535a8e224020526cfb/packages/kit/src/config/load.ts#L34-L39

### Logs

_No response_

TasukuUno:
### 🔗 Linked issue

#1611

### ❓ Type of change

### 📚 Description

resolve #1611
Add new feature to support async function as nuxt.config like v2.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/config/load.ts
+++ b/packages/kit/src/config/load.ts
@@ -33,6 +33,11 @@ export async function loadNuxtConfig (opts: LoadNuxtConfigOptions): Promise<Nuxt
 
   if (nuxtConfigFile && existsSync(nuxtConfigFile)) {
     nuxtConfig = requireModule(nuxtConfigFile, { clearCache: true })
+
+    if (typeof nuxtConfig === 'function') {
+      nuxtConfig = await nuxtConfig(opts)
+    }
+
     nuxtConfig = { ...nuxtConfig }
     nuxtConfig._nuxtConfigFile = nuxtConfigFile
     nuxtConfig._nuxtConfigFiles = Array.from(scanRequireTree(nuxtConfigFile))


pi0 (approved):
Note: This might be finally deprecated in 3.0.0 since this is not a good practice to await on config loading vs using an async hook but makes sense for BW compatibility.

PR #1618, opened 2021-11-01
fix(nuxt3): allows plugins without default export

antfu:
### 🔗 Linked issue

https://github.com/antfu/unocss/issues/27

### ❓ Type of change

### 📚 Description

For example, in some cases, we might be relying on module side-effect in a plugin but does not have the explicit `default export`, this PR mocked those plugins to provide a noop function to avoid the undefined error.

```ts
// plugins/foo.js
import 'winid.css'
```

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/nuxt.ts
+++ b/packages/nuxt3/src/app/nuxt.ts
@@ -137,6 +137,9 @@ export function normalizePlugins (_plugins: Array<Plugin | LegacyPlugin>) {
   let needsLegacyContext = false
 
   const plugins = _plugins.map((plugin) => {
+    if (!plugin) {
+      return () => {}
+    }
     if (isLegacyPlugin(plugin)) {
       needsLegacyContext = true
       return (nuxtApp: NuxtApp) => plugin(nuxtApp._legacyContext!, nuxtApp.provide)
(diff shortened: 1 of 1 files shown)


danielroe (commented):
Should we instead change so we are not importing default at all?

https://github.com/nuxt/framework/blob/07086c839753f349d6b250bb80d5161b2bbb5d56/packages/nuxt3/src/core/templates.ts#L33-L44

so rather than:
```js
import asd_234 from '/my/no/default.js'

export default [
  asd_234
]
```
We get:

```js
import '/my/no/default.js'

export default []
```

antfu:
That's a good point, I guess we could use `mlly` to detect if it has `default export` right? But then my concern become that some plugins might not be on the disk by served by some virtual modules and we might not always been able to read their content?

pi0:
> That's a good point, I guess we could use mlly to detect if it has default export right? But then my concern become that some plugins might not be on the disk by served by some virtual modules and we might not always been able to read their content?

I love this idea this can help reducing runtime check overheads + removing workaround we had for bridge default export ! Well actually mlly can detect syntax from any string content (also from vfs)

pi0 (approved):
Side note: You still need to add a default export `export default {}` to the plugin with this fix. Vite throws an error otherwise since we generate `import x from 'plugin'` in template

Issue #1589, opened 2021-10-29
defineNuxtPlugin doesn't provide inside nuxtApp


liqueflies:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v14.17.0`
- Nuxt Version:     `2.16.0-27256336.293dc11a`
- Package Manager:  `Yarn`
- Bundler:          `Webpack`
- User Config:      `target`, `head`, `css`, `loading`, `publicRuntimeConfig`, `plugins`, `components`, `buildModules`, `modules`, `i18n`, `tailwindcss`, `pwa`, `sanity`, `gtm`, `sitemap`, `robots`, `build`
- Runtime Modules:  `@nuxtjs/i18n@7.1.0`, `@nuxtjs/robots@2.5.0`, `@nuxtjs/gtm@2.4.0`
- Build Modules:    `@nuxtjs/tailwindcss@4.2.1`, `@/modules/sitemapRouteGenerator`, `@nuxt/bridge@3.0.0-27257216.f5aea9f`
------------------------------

### Describe the bug

If I define a plugin that use defineNuxtPlugin I am not able to use the .provide as described in docs, but I find it inside the nuxt2Context object.

This is not great because when the project will finally move to nuxt3 another refactor will be necessary (and maybe not small cause in every template I assume to call nuxtApp.nuxt2Context.$myPlugin).

Here is a simple example: https://codesandbox.io/s/floral-paper-sn46i

### Reproduction

https://codesandbox.io/s/floral-paper-sn46i

cdwmhcc:
same

PR #1623, opened 2021-11-01
feat!(bridge): allow accessing injections from `useNuxtApp`

### LINKED ISSUE #967: improve dev experience accessing context in bridge
At the moment there's no good replacement for `useContext` from `@nuxtjs/composition-api` in Bridge.

We can instead define getters on the app so it matches all `$` utils (which is also how Nuxt 3 will work for injected context).

```js
// current workaround 
const { $store } = useNuxtApp().legacyApp

// desired, and equivalent with Nuxt 3
const { $store } = useNuxtApp()
```

related: https://github.com/nuxt/framework/discussions/964#discussioncomment-1471713

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/1589
resolves #967
context: https://github.com/nuxt/framework/discussions/964#discussioncomment-1471713

### ❓ Type of change

### 📚 Description

This PR enables accessing injected properties within Bridge directly via `useNuxtApp()` to decrease the difference between Nuxt 2/3.

Example:
```ts
const { $myInjectedProperty } = useNuxtApp()
```

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/runtime/app.plugin.mjs
+++ b/packages/bridge/src/runtime/app.plugin.mjs
@@ -14,7 +14,7 @@ export default (ctx, inject) => {
       mount: () => { },
       provide: inject,
       unmount: () => { },
-      use(vuePlugin) {
+      use (vuePlugin) {
         vuePlugin.install(this)
       },
       version: Vue.version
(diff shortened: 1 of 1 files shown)


packages/bridge/src/runtime/app.plugin.mjs
```diff
@@ -42,7 +42,16 @@ export default (ctx, inject) => {
     nuxtApp.vue2App = this
   })
 
-  setNuxtAppInstance(nuxtApp)
+  const proxiedApp = new Proxy(nuxtApp, {
```
pi0, line 45: Note: With this we are dropping IE compatibility from bridge (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#browser_compatibility)

danielroe: Indeed 👍 Your thoughts? Do we aim to keep IE compatibility for Bridge?

pi0: I don't mind. Just a side note and that by this, we can upgrade babel target and add docs :)

danielroe:
@NidMo For now, you'll need to add that plugin to `build.transpile` in nuxt.config. If you have other questions/follow-up, would you create a discussion rather than responding to this issue? 🙏

PR #1638, opened 2021-11-02
chore(lint): enable linting on docs files

clarkdo:
### ❓ Type of change

### 📚 Description

I'm not sure if there is any specific reason for disable linting on docs files, this pr is adding it back and fixing all issues so that components in doc can be update to latest vue recommended rules.

### 📝 Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,4 +3,3 @@ node_modules
 schema
 **/*.tmpl.*
 sw.js
-docs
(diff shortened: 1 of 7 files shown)


pi0:
Thanks! We only have a small (current) issue that if `docs/node_modules` are not installed, wrong deps (`vue3`) will be checked against and fails CI. I've added to import ignore rules.

PR #1646, opened 2021-11-02
fix(vite): remove relative vite workaround

### LINKED ISSUE #1424: `swiper` Missing "./vue/swiper-vue.js" export
### Environment

Nuxt CLI v3.0.0-27251455.110b2b2                                                                                                                            07:15:02
RootDir: C:\Users\Andy\Desktop\test\lombakativ                                                                                                             07:15:03
Nuxt project info:                                                                                                                                          07:15:03

- Node Version:     `v16.11.0`
- Nuxt Version:     `3.0.0-27251455.110b2b2`
- Package Manager:  `npm`
- Bundler:          `Vite`
- User Config:      `vite`, `build`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Describe the bug

fresh install , problem with only run dev...

```
 WARN  [SSR] Error transforming ./node_modules/swiper/vue/swiper-vue.js: Error: Missing "./vue/swiper-vue.js" export in "swiper" package                      07:18:15

✔ Vite server built in 2694ms                                                                                                                                 07:18:15
✔ Nitro built in 248 ms                                                                                                                                       07:18:16

 ERROR  [uncaughtException] r.writeLatin1String is not a function        
```

 3.0.0-27246961.9d40a27 works good.

### Reproduction

https://stackblitz.com/edit/github-ar5zza-xbrnjt?file=app.vue

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #1593: jsx not working
### Environment

Nuxt CLI v3.0.0-27254511.41ff3b8                                                                                   01:24:33
RootDir: /home/xhe/project/ioclub                                                                                  01:24:33
Nuxt project info:                                                                                                 01:24:33

- Operating System: `Linux`
- Node Version:     `v14.17.4`
- Nuxt Version:     `3.0.0-27254511.41ff3b8`
- Package Manager:  `Yarn`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

👉 Report an issue: https://github.com/nuxt/framework/issues/new                                                   01:24:33

Done in 1.19s.

### Describe the bug

Just create whatever jsx file.

```
[nuxt-dev] [Vue Router warn]: uncaught error during route navigation:
[nuxt-dev] TypeError: __vite_ssr_import_3__.ssrRegisterHelper is not a function
[nuxt-dev]     at $id_4aec7587 (file:///.../nuxt/dist/server/server.mjs:3095:23)
[nuxt-dev]     at async __instantiateModule__ (file:///.../.nuxt/dist/server/server.mjs:5124:3)
```

### Reproduction

```
diff --git a/node_modules/@nuxt/vite-builder/dist/index.mjs b/node_modules/@nuxt/vite-builder/dist/index.mjs
index ffc21b7..0648c46 100644
--- a/node_modules/@nuxt/vite-builder/dist/index.mjs
+++ b/node_modules/@nuxt/vite-builder/dist/index.mjs
@@ -191,7 +191,7 @@ async function transformRequest(opts, id) {
   }
   if (id && id.startsWith("/@fs/")) {
     id = id.slice("/@fs".length);
-  } else if (!id.includes("entry") && id.startsWith("/")) {
+  } else if (!id.includes("entry") && id.startsWith("/") && !id.includes('vue-jsx')) {
     id = "." + id;
   }
   if (await isExternal(opts, id)) {
```

This patch will fix the problem. Since `plugin-vue-jsx` will use `id === '/__vue-jsx...'` to match the id https://github.com/vitejs/vite/blob/70fd32c1569c3d93a231577b573dbd2b34da4de2/packages/plugin-vue-jsx/index.js#L72, but vite-builder adds an extra `.` before the id, it can not resolve the internal module.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #1424
resolves #1593

### ❓ Type of change

### 📚 Description

This PR sets the id to a fully resolved path rather than a relative one (but only if it exists, which should protect us from vite plugins that have their own aliases).

### 📝 Checklist

### DIFF
--- a/packages/vite/src/dev-bundler.ts
+++ b/packages/vite/src/dev-bundler.ts
@@ -1,4 +1,6 @@
 import { pathToFileURL } from 'url'
+import { resolve } from 'pathe'
+import { existsSync } from 'fs-extra'
 import * as vite from 'vite'
 import { ExternalsOptions, isExternal as _isExternal, ExternalsDefaults } from 'externality'
 import { hashId, uniq } from './utils'
@@ -57,7 +59,10 @@ async function transformRequest (opts: TransformOptions, id: string) {
     id = id.slice('/@fs'.length)
   } else if (!id.includes('entry') && id.startsWith('/')) {
     // Relative to the root directory
-    id = '.' + id
+    const resolvedPath = resolve(opts.viteServer.config.root, '.' + id)
+    if (existsSync(resolvedPath)) {
+      id = resolvedPath
+    }
   }
 
   if (await isExternal(opts, id)) {
--- a/packages/vite/src/dev-bundler.ts
+++ b/packages/vite/src/dev-bundler.ts
@@ -1,6 +1,6 @@
 import { pathToFileURL } from 'url'
 import { resolve } from 'pathe'
-import { existsSync } from 'fs-extra'
+import fse from 'fs-extra'
 import * as vite from 'vite'
 import { ExternalsOptions, isExternal as _isExternal, ExternalsDefaults } from 'externality'
 import { hashId, uniq } from './utils'
@@ -60,7 +60,7 @@ async function transformRequest (opts: TransformOptions, id: string) {
   } else if (!id.includes('entry') && id.startsWith('/')) {
     // Relative to the root directory
     const resolvedPath = resolve(opts.viteServer.config.root, '.' + id)
-    if (existsSync(resolvedPath)) {
+    if (fse.existsSync(resolvedPath)) {
       id = resolvedPath
     }
   }
--- a/packages/vite/src/dev-bundler.ts
+++ b/packages/vite/src/dev-bundler.ts
@@ -1,6 +1,6 @@
 import { pathToFileURL } from 'url'
+import { existsSync } from 'fs'
 import { resolve } from 'pathe'
-import fse from 'fs-extra'
 import * as vite from 'vite'
 import { ExternalsOptions, isExternal as _isExternal, ExternalsDefaults } from 'externality'
 import { hashId, uniq } from './utils'
@@ -60,7 +60,7 @@ async function transformRequest (opts: TransformOptions, id: string) {
   } else if (!id.includes('entry') && id.startsWith('/')) {
     // Relative to the root directory
     const resolvedPath = resolve(opts.viteServer.config.root, '.' + id)
-    if (fse.existsSync(resolvedPath)) {
+    if (existsSync(resolvedPath)) {
       id = resolvedPath
     }
   }


packages/vite/src/dev-bundler.ts
```diff
@@ -57,7 +59,10 @@ async function transformRequest (opts: TransformOptions, id: string) {
     id = id.slice('/@fs'.length)
   } else if (!id.includes('entry') && id.startsWith('/')) {
     // Relative to the root directory
-    id = '.' + id
+    const resolvedPath = resolve(opts.viteServer.config.root, '.' + id)
+    if (fse.existsSync(resolvedPath)) {
```
pi0, line 63: According to the last behavior I spotted, absolute ids are like `/node_modules/...` and `root` of vite server is set to `srcDir` (#1018) while `node_modules` resides on `rootDir`. We probably need a better workaround to fix this

danielroe: Curiously, when there is a custom `srcDir` the paths are all fully resolved already by this point and so this condition isn't hit.

pi0: Ids not related to srcDir (other packages) with current change will still be resolved and checked against srcDir/node_modules (root=srcDir)

danielroe: what about just `resolve('.' + id')` and default to `process.cwd()`?

danielroe: > Ids not related to srcDir (other packages) with current change will still be resolved and checked against srcDir/node_modules (root=srcDir)

Could you explain more? I believe vite fully resolves them to absolute paths in this case. Maybe @antfu can shed light on this?

pi0: Imagine:

```
rootDir
  srcDir (=vite.config.root)
  node_modules
    swiper
```

And input id is `./node_modules/swiper/vue/swiper-vue.js` (#1424) here we set `resolvedPath` to `srcDir/node_modules/swiper` that does not escists

danielroe: I am saying that vite will not provide that input id to `transformRequest` in this case. It fully resolves the path to an absolute path because it is outside of `root`.

antfu: This seems to be a valid fix to me. When vite doing `/node_modules` is should be saying that `node_modules` is under the vite's root. If it's outside, it will be converted to `/@fs/` with absolute path automatically.

pi0: Actually, vite makes a request to `./node_modules` for some reason when using in external projects. And it is not valid to use `./src/node_modules`.

danielroe: Do you have an example? I'd love to see what's causing it. I've tested this externally too with a custom `srcDir` (i.e. not within the framework repo), and was unable to reproduce that.

pi0: #1424. Regardless this change is _wrong_ in a combination with using `root: srcDir`.

danielroe: I think we're talking at cross-purposes. It could just be that I'm not understanding you. This is my understanding of how the request is processed:

1. A request comes in.
2. If this is within `root` (`srcDir`) then it is will be relative, e.g. `/app.vue`. In this case we can rightly resolve it relative to `root`.
3. If this is _not_ within `root` (`srcDir`), for example, if it is in `node_modules` and not within `srcDir`, then it will be fully resolved, e.g. `/@fs/Users/me/my/project/node_modules/vue/swiper-vue.js`, in which case we don't hit this case.

I have tested the project in https://github.com/nuxt/framework/issues/1424 with this fix, and it works, both with and without a custom `srcDir`.

pi0: Ok lets have the fix :)

PR #1648, opened 2021-11-02
fix: sanitize rollup-generated filenames

danielroe:
### 🔗 Linked issue

resolves nuxt/nuxt.js#12606
reverts nuxt/framework#477

### ❓ Type of change

### 📚 Description

This PR removes characters from rollup filenames that are excluded from valid URLs (see [RFC 2396](https://datatracker.ietf.org/doc/html/rfc2396)). As rollup-produced files may be consumed by browsers, this is one approach to provide broader ecosystem safety without requiring URL encoding.

I've also proposed this in a [PR directly to rollup](https://github.com/rollup/rollup/pull/4262) but there may be legitimate reasons for closing it, as this is relevant specifically to browsers rather than the node ecosystem. If accepted we can remove the implementation in this PR.

(Having said that I have also implemented for nitro as I feel most of these characters may also pose problems for filesystems, and in a worker target the filenames may also become URLs.)

### 📝 Checklist


pi0:
This is a nice idea and we need sanitizers both for generated file names and export identifiers (instead of unreadable hashes we use). What do you think to move them to mlly?

PR #1662, opened 2021-11-03
fix(bridge): can't install vue plugin

kazupon:
### 🔗 Linked issue

not applicable in issues.
but, I have a reproduction link
https://github.com/kazupon/nuxt-i18n-next/tree/repro-vue-plugin-install

### ❓ Type of change

### 📚 Description

As described in the vue official docs, the vue plugin is a function whose first argument is a Vue constructor.
https://vuejs.org/v2/guide/plugins.html#Writing-a-Plugin

The nuxt bridge implements `vueApp` that is compatible with the Vue constructor.
However, the first argument of `install`, which is executed in `use`, is not passed a Vue constructor.

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/runtime/app.plugin.mjs
+++ b/packages/bridge/src/runtime/app.plugin.mjs
@@ -15,7 +15,7 @@ export default (ctx, inject) => {
       provide: inject,
       unmount: () => { },
       use (vuePlugin) {
-        vuePlugin.install(this)
+        vuePlugin.install(Vue)
       },
       version: Vue.version
     },


packages/bridge/src/runtime/app.plugin.mjs
```diff
@@ -15,7 +15,7 @@ export default (ctx, inject) => {
       provide: inject,
       unmount: () => { },
       use (vuePlugin) {
-        vuePlugin.install(this)
```
pi0, line 18: I think this was intentional to pass Vue 3 compatible interface as Vue (`this`) to plugins but it is missing some Vue 2 options

pi0:
`vueApp` in bridge trying to mimic Vue 3's app interface. For i18n-next plugin, I would suggest using normal Vue 2 import and installing if i18n-bridge also expects Vue 2. For nuxt3, support, we can find a solution to switch to the Vue 3 plugin.

PR #1671, opened 2021-11-03
feat(nitro): improve externals

### LINKED ISSUE #1670: import externals with full path for nitro dev
In development, we generate `.nuxt/nitro/index.mjs` with externals implicitly pointing to `node_modules` (like `import destr from 'destr';`) . This makes two major issues:
- Node.js has different search path for `node_modules` than the one configured for rollup and as a result hoisting issues
- With pnp (#581) we need explicit dependencies while imports like `h3` are actually not a dependency of the project

pi0:
### 🔗 Linked issue

resolves #1670

### ❓ Type of change

### 📚 Description

- [x] Resolve externals with full path
- [ ] Use externality

### 📝 Checklist

### DIFF
--- a/packages/nitro/src/rollup/config.ts
+++ b/packages/nitro/src/rollup/config.ts
@@ -57,6 +57,7 @@ export const getRollupConfig = (nitroContext: NitroContext) => {
       '@vue/compiler-core': 'unenv/runtime/mock/proxy',
       '@vue/compiler-dom': 'unenv/runtime/mock/proxy',
       '@vue/compiler-ssr': 'unenv/runtime/mock/proxy',
+      '@vue/devtools-api': 'unenv/runtime/mock/proxy',
       ...nitroContext.alias
     }
   }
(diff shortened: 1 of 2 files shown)


packages/nitro/src/rollup/plugins/externals.ts
```diff
@@ -60,8 +63,16 @@ export function externals (opts: NodeExternalsOptions): Plugin {
         }
       }
 
+      const resolved = await this.resolve(originalId, importer, { ...options, skipSelf: true })
+      if (!existsSync(resolved.id)) {
```
pi0, line 67: rollup-plugin-resolve is preferred unless it does not resolve to full path

packages/nitro/src/rollup/config.ts
```diff
@@ -273,7 +273,13 @@ export const getRollupConfig = (nitroContext: NitroContext) => {
         base: '/',
         processCwd: nitroContext._nuxt.rootDir,
         exportsOnly: true
-      }
+      },
+      exportConditions: [
```
danielroe, line 277: Should we pass these to node resolve too?

pi0: Yes. Actually we do in nitro config 😁

pi0:
This change is rather risky. Doing it later with nitropack

PR #1712, opened 2021-11-05
fix(vite): swap `<link>` style on dev ssr

antfu:
### ❓ Type of change

### 📚 Description

Currently we have the css entries in the manifest and injected as <link ref="stylesheet"> by the renderer. The problem here is that during dev, vite's css are actually a JS module that controls a <style> for HMR. This makes client as duplicated of styles

For example:
image

Since the style from <link> can't be changed without a full page reload. This makes the old style persist and potentially conflict with the updated ones.

This PR makes the dev css module swap the `<link>` version with the same id to prevent duplication.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/plugins/dev-ssr-css.ts
+++ b/packages/vite/src/plugins/dev-ssr-css.ts
@@ -1,0 +1,23 @@
+import { Plugin } from 'vite'
+import { isCSS } from '../utils'
+
+export function devStyleSSRPlugin (rootDir: string): Plugin {
+  return {
+    name: 'nuxt:dev-style-ssr',
+    apply: 'serve',
+    enforce: 'post',
+    transform (code, id) {
+      if (!isCSS(id) || !code.includes('import.meta.hot')) {
+        return
+      }
+
+      let moduleId = id
+      if (moduleId.startsWith(rootDir)) {
+        moduleId = moduleId.slice(rootDir.length)
+      }
+
+      // When dev `<style>` is injected, remove the `<link>` styles from manifest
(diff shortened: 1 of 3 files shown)


packages/vite/src/plugins/dev-ssr-css.ts
```diff
@@ -0,0 +1,23 @@
...
+        return
+      }
+
+      let moduleId = id
+      if (moduleId.startsWith(rootDir)) {
+        moduleId = moduleId.slice(rootDir.length)
+      }
+
+      // When dev `<style>` is injected, remove the `<link>` styles from manifest
+      return code + `\ndocument.querySelector(\`link[href="/_nuxt${moduleId}"]\`).forEach(i=>i.remove())`
```
pi0, line 20: Can we please use `options.app.assetsPath` in place of `/_nuxt`? 

An alternative idea, we can inject hash of moduleId as an attr by renderer and select it here by unique module hash

pi0: (merging PR to test asap. can do this later)
```suggestion
      // TODO: Use `app.assetsPath` or unique hash
      return code + `\ndocument.querySelector(\`link[href="/_nuxt${moduleId}"]\`).forEach(i=>i.remove())`
```

pi0 (approved):

JamieCurnow:
@pi0 @antfu I think this introduced a bug that I'm seeing with a brand new nuxt project. Shouldn't `document.querySelector` actually be `document.querySelectorAll` ?

Have made an issue: #1731 
And if this is the case a PR: #1734 

Thanks for your great work 🙌

PR #1723, opened 2021-11-05
refactor(vite): reuse common modules

antfu:


pi0:
@antfu  Can you please check failing tests?

packages/bridge/src/vite/utils/wpfs.ts
```diff
@@ -0,0 +1,7 @@
+import { join } from 'pathe'
+import fsExtra from 'fs-extra'
+
+export const wpfs: any = {
```
antfu: I am not sure what is happening here, revert to the old workaround but I guess we'd better solve this in the long term.

pi0 (approved):

packages/bridge/src/vite/utils/wpfs.ts
pi0: Sure. It will probably be gone soon :)

PR #1735, opened 2021-11-05
fix(nuxt3): add missing auto imports

### LINKED ISSUE #1706: isReactive is not auto import on bridge
### Discussed in https://github.com/nuxt/framework/discussions/1683

<div type='discussions-op-text'>

<sup>Originally posted by **baixiaoyu2997** November  4, 2021</sup>
use `import {isReactive} from "#app"` is succeed

in `auto-imports.d.ts` file, i can't find isReactive</div>

repo : https://codesandbox.io/s/chs4m

danielroe:
### 🔗 Linked issue

resolves #1706

### ❓ Type of change

### 📚 Description

This PR adds some missing auto-imports in Nuxt 3. (We may want to tweak this list further - please suggest away - as a lot of the more technical ones probably don't need to be auto-imported.) It also disables auto imports in Nuxt Bridge that aren't supported by the Vue 2 Composition API. And it adds tests so we will know in future if there is a new composable introduced that we should add to the list.

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/auto-imports.ts
+++ b/packages/bridge/src/auto-imports.ts
@@ -1,7 +1,9 @@
 import { installModule, useNuxt } from '@nuxt/kit'
+import * as CompositionApi from '@vue/composition-api'
 import autoImports from '../../nuxt3/src/auto-imports/module'
 
 const UnsupportedImports = new Set(['useAsyncData', 'useFetch'])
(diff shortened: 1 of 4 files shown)


pi0 (approved):

pi0:
Thanks this seems a nice step forward! I would suggest later on maintain a single DB of auto imports to also include possibly auto import by install from vue-use.

PR #1769, opened 2021-11-07
fix(bridge): Recognize {... as default } construction in vite plugin

tobiasdiez:
### ❓ Type of change

### 📚 Description

Currently, the construction `export { xyz as default }` is not recognized as a valid default export by the default-export vite plugin. Thus, a second default export is added, which leads to an error.
An example is 
https://github.com/posva/pinia/blob/a76e34387419e86154a63985ad038eefeb2bc9fd/packages/nuxt/src/templates/plugin.ts#L50
which is transpiled as `export { PiniaNuxtPlugin as default };` in the distributed build. (cc @posva)

This is fixed by adding a regex-based check.

### 📝 Checklist


pi0:
Thanks for pull request @tobiasdiez. An alternative solution merged to use [mlly](https://github.com/unjs/mlly) utils. See #1774

tobiasdiez:
Yes, that looks like a cleaner solution indeed!

PR #1779, opened 2021-11-08
fix(kit): normalize plugin imports paths

eakhadov:
### 🔗 Linked issue

posva/pinia#767

### ❓ Type of change

### 📚 Description

Fixing the slash of paths when importing plugins for windows. 

Example error (@pinia/nuxt) latest

```
WARN  [SSR] Error transforming virtual:C:/Users/Emil/Desktop/nuxt-app/src/.nuxt/plugins/server.mjs: 
Error: Failed to resolve import "C:UsersEmilDesktopnuxt-appnode_modules@pinianuxtdistplugin.mjs" from "virtual:C:\Users\Emil\Desktop\nuxt-app\src\.nuxt\plugins\server.mjs". Does the file exist?`

Cannot read properties of undefined (reading 'map')
  at Module.normalizePlugins (file:///C:/Users/Emil/Desktop/nuxt-app/src/.nuxt/dist/server/server.mjs:180:28)   
  at $id_c8abffd4 (file:///C:/Users/Emil/Desktop/nuxt-app/src/.nuxt/dist/server/server.mjs:28:39)  
  at async __instantiateModule__ (file:///C:/Users/Emil/Desktop/nuxt-app/src/.nuxt/dist/server/server.mjs:996:3 ) 
 ```

### 📝 Checklist

### CODE REVIEW

**Review by danielroe — APPROVED** [review]

**Review by danielroe — COMMENTED** [review]
Actually, I think we should probably do this here instead:
https://github.com/nuxt/framework/blob/52cce01962cd6ba2ba89a0e0b122dca13e7c4f2a/packages/kit/src/module/utils.ts#L81-L103

File: packages/nuxt3/src/core/templates.ts [hunk 1]
```diff
@@ -33,11 +34,14 @@ export const cssTemplate = {
 export const clientPluginTemplate = {
   filename: 'plugins/client.mjs',
   getContents (ctx: TemplateContext) {
-    const clientPlugins = ctx.app.plugins.filter(p => !p.mode || p.mode !== 'server')
+    const clientPlugins = ctx.app.plugins
+      .filter(p => !p.mode || p.mode !== 'server')
+      .map(p => normalize(p.src))
```
antfu, line 39: Shall we normalize it in the first place in plugins resolving?

eakhadov: In general, when importing paths, we must somehow normalize the paths, because an error will occur in different os due to the slash. With @danielroe  help, I changed the code a bit elsewhere.

packages/kit/src/module/utils.ts
```diff
@@ -88,6 +88,8 @@ export function normalizePlugin (plugin: NuxtPlugin | string): NuxtPlugin {
 
   if (!plugin.src) {
     throw new Error('Invalid plugin. src option is required: ' + JSON.stringify(plugin))
+  } else {
+    plugin.src = normalize(plugin.src)
   }
```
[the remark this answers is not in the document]
eakhadov: OK. I'll fix it. But if else, in my opinion, is better, because the code is executed immediately after passing the check than before exiting the block.

pi0 (approved):
Thanks! LGTM however this doesn't fix latest pinia module since they reverted back from using kit. We also need to normalize in template utils

eakhadov:
Good. 
But I would argue about the "pinia" error. This fix fixes the "pinia" error (and not only it), which I indicated in the description.

pi0:
> But I would argue about the "pinia" error. This fix fixes the "pinia" error (and not only him), which I indicated in the description.

Partially :) For nuxt3, if a module using legacy ModuleContainer (like pinia), we still apply kit utils and thus normalizing but not applied to bridge.

Issue #1707, opened 2021-11-05
scrollBehavior setting doesn't work


hecateball:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.11.0`
- Nuxt Version:     `3.0.0-27267528.292b524`
- Package Manager:  `npm@8.0.0`
- Bundler:          `Vite`
- User Config:      `buildModules`, `css`, `meta`
- Runtime Modules:  `-`
- Build Modules:    `nuxt-windicss@^2.0.3`
------------------------------

### Describe the bug

`scrollBehavior` setting (which is available in Nuxt2) doesn't work.
I tried both nuxt.config.ts `router.scrollBehavior` and putting `router.scrollBehavior.js` in project route but they have no effect.

I couldn't find any document on scrollBehavior at Nuxt3 docs website... Is there any way to make it work? 

### Reproduction

nuxt.config.ts

```
import { defineNuxtConfig } from 'nuxt3'

export default defineNuxtConfig({
  // ......
  router: {
    scrollBehavior: () => ({ x: 0, y: 0 }),
  },
})
```

~/router.scrollBehavior.js
```
export default () => {
  console.log('here') // never called
  return { x: 0, y: 0 }
}

```

danielroe:
You can set this in a plugin:
```js
export default defineNuxtPlugin((nuxtApp) => {
  nuxtApp.$router.options.scrollBehavior = () => {
    return { left: 0, top: 0 }
  }
})
```

hecateball:
@danielroe Thank you!
Is that  a standard way to configure router in Nuxt 3?

dergunovs:
> You can set this in a plugin:
> 
> ```js
> export default defineNuxtPlugin((nuxtApp) => {
>   nuxtApp.$router.options.scrollBehavior = () => {
>     return { left: 0, top: 0 }
>   }
> })
> ```

This fix works not so smooth. White blink spoils the impression.

K-Honkawa:
I'm sorry.
I have reopened...

Gyurmatag:
The plugin solution is not working smooth. Any idea when will it be fixed?

madsh93:
Another thing I noticed is if you add anything with router.push it will scroll to top, even though it may not be what you want to do.

```js
    router.push({
      query: {
        ...route.query,
        equipment: reAddedEquipment.map((x) => x.id)
      }
    })
```

danielroe:
[...]

PR #1781, opened 2021-11-08
bug: add missing $router property typing

nandi95:
### ❓ Type of change

### 📚 Description

In `defineNuxtPlugin` I did not see a router property. I tried to use `useRouter()` but that returned `undefined`. On inspection of the given `nuxt` to the callback, it in fact has a `$router` property.
It seem this should be documented to the user as per this suggestion: https://github.com/nuxt/framework/issues/1707#issuecomment-961983939

### 📝 Checklist


danielroe:
Would you raise this as an issue instead? Would be nice to identify the regression that's preventing https://github.com/nuxt/framework/pull/1603 from working.

nandi95:
Moved this to an [issue](https://github.com/nuxt/framework/issues/1794) as per request.

Issue #1751, opened 2021-11-06
It looks like #config isn't defined for TypeScript?


danielroe:
### Discussed in https://github.com/nuxt/framework/discussions/1745

<div type='discussions-op-text'>

<sup>Originally posted by **jeremeevans** November  5, 2021</sup>
tsconfig.json has: `"extends": "./.nuxt/tsconfig.json"` but I still get this error. It does build and work... so not breaking things, just wondering if I need to update something.

![image](https://user-images.githubusercontent.com/1747006/140578014-53d97d25-3adb-4f47-a9ad-bb47683adaf0.png)
</div>

tobiasdiez:
It would be nice if it could infer the type based on what is specified for public/privateRuntimeConfig in the nuxt config. The same applies to `useRuntimeConfig`.

By the way, `#config` doesn't seem to be documented yet.

PR #1783, opened 2021-11-08
fix(nitro, nuxi): add runtimeConfig types (for `#config` and `useRuntimeConfig()`)

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/1751

### ❓ Type of change

### 📚 Description

This PR:
* adds support for shimming `#config` in Nitro routes
* adds initial docs for `useRuntimeConfig()`
* adds support for typing `privateRuntimeConfig` & `publicRuntimeConfig`

Note, the way I've implemented this allows us in future to automatically type runtime config.

### 📝 Checklist


docs/content/3.docs/1.usage/5.runtime-config.md
```diff
@@ -0,0 +1,71 @@
...
+### Typing runtime config
+
+It is possible to manually type your runtime config.
+
+```ts [nuxt.config.js|ts]
+import { defineNuxtConfig } from 'nuxt3'
+
+declare module '@nuxt/nitro' {
+  interface PublicRuntimeConfig {
+    testConfig: string
```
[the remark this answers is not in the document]
danielroe: Agreed - auto-generating types would be ideal. I've left that for a subsequent PR.

docs/content/3.docs/1.usage/5.runtime-config.md
```diff
@@ -0,0 +1,71 @@
+# Runtime configuration
+
+You can use dynamic config and environment variables in your Nuxt app and Nitro server routes.
+
+## Exposing runtime config
+
+In addition to any ambient environment variables, uf you have a `.env` file in your project root directory, it will be automatically loaded into `process.env` and accessible within your `nuxt.config` file.
```
tobiasdiez, line 7:
```suggestion
In addition to any ambient environment variables, if you have a `.env` file in your project root directory, it will be automatically loaded into `process.env` and accessible within your `nuxt.config` file.
```

packages/bridge/src/runtime/composables.ts
```diff
@@ -16,7 +17,7 @@ export const useRuntimeConfig = () => {
   if (!nuxtApp.$config) {
     nuxtApp.$config = reactive(nuxtApp.nuxt2Context.app.$config)
   }
-  return nuxtApp.$config
+  return nuxtApp.$config as RuntimeConfig
```
tobiasdiez, line 20:
```suggestion
  return nuxtApp.$config as PublicRuntimeConfig
```
At least to the added docs, only the public config is available on the client side

packages/nitro/types/shims.d.ts
```diff
@@ -9,3 +9,10 @@ declare module '#assets' {
   export function statAsset (id: string): Promise<AssetMeta>
   export function getKeys() : Promise<string[]>
 }
+
+declare module '#config' {
+  import type { PublicRuntimeConfig, PrivateRuntimeConfig, RuntimeConfig } from '@nuxt/kit'
+  export const privateConfig: PrivateRuntimeConfig
+  export const publicConfig: PublicRuntimeConfig
+  export default RuntimeConfig
```
tobiasdiez, line 17:
```suggestion
  export default const config: RuntimeConfig
```
Otherwise the interface is exported if I'm not mistaken.

packages/bridge/src/runtime/composables.ts
danielroe: `RuntimeConfig` is a merged type with private config marked as optional (as this composable also runs on server side)

packages/nitro/types/shims.d.ts
danielroe: You're right that this needs to be tweaked:
```suggestion
  const runtimeConfig: RuntimeConfig
  export default runtimeConfig
```

packages/nitro/types/shims.d.ts
```diff
@@ -9,3 +9,11 @@ declare module '#assets' {
   export function statAsset (id: string): Promise<AssetMeta>
   export function getKeys() : Promise<string[]>
 }
+
+declare module '#config' {
+  import type { PublicRuntimeConfig, PrivateRuntimeConfig, RuntimeConfig } from '@nuxt/kit'
+  export const privateConfig: PrivateRuntimeConfig
+  export const publicConfig: PublicRuntimeConfig
+  const runtimeConfig: RuntimeConfig
```
tobiasdiez, line 17: @danielroe shouldn't this then be 
```suggestion
  const runtimeConfig: PublicRuntimeConfig & PrivateRuntimeConfig
```
since as you say above `RuntimeConfig` has the private config as optional (which would lead to unnecessary null checks server-side if you use `#config` there.

danielroe: Indeed you are right :+1

danielroe, line 17:
```suggestion
  import type { PublicRuntimeConfig, PrivateRuntimeConfig } from '@nuxt/kit'
  export const privateConfig: PrivateRuntimeConfig
  export const publicConfig: PublicRuntimeConfig
  const runtimeConfig: PrivateRuntimeConfig & PublicRuntimeConfig
```

danielroe: [`ce90b23` (#1783)](https://github.com/nuxt/framework/pull/1783/commits/ce90b2386812c2bd966bc3083f9985e764f6cbb9)

tobiasdiez:
I hope it's okay to use this PR for a related question: how does one use `#config` in jest tests? If one of the tests (indirectly) tries to import `#config` then this obviously fails without any further setup.

danielroe:
@tobiasdiez you would likely want to use  `moduleNameMapper` to point it to a stub.

pi0 (approved):

PR #1849, opened 2021-11-11
feat(bridge, nuxt3): mock vue-demi

### LINKED ISSUE #1737: Mock vue-demi
Some libraries might use `vue-demi` and we are already mocking vue3 interface for bridge. Mocking `vue-demi` by name, can prevent any potential issues with `@vuejs/composition-api` compatibility.

antfu:
### 🔗 Linked issue

Close #1737 

### ❓ Type of change

### 📚 Description

I think it's safe to only mock for bridge since in Vue 3 `vue-demi` is simply some redirection without side effects.

Todo: (@pi0)
- [x] refactor alias in nuxt3
- [x] mock `@vuejs/composition-api` for nuxt3
- [x] #app and auto imports from vue-demi utils (`isVue2` and `isVue3`)

### 📝 Checklist


pi0:
It would be nice to implement for Nuxt3 before closing linked issue since:
- vue-demi as I could understand uses a postinstall hook to detect version and copy impl which can go wrong in some edge cases (like when using both nuxt2 and nuxt3 in monorepo) while in Nuxt 3 during build we are absolutely sure it is vue3 or vue2
- Same flags can be available even as global auto imports since it is going to be a common asset for module authors and vue-demi idea is already amazing only we could make it more stable in Nuxt's context

packages/bridge/src/runtime/vue2-bridge.mjs
```diff
@@ -1,7 +1,14 @@
...
 export const isFunction = fn => fn instanceof Function
 
 export { Vue as default }
+
+// mock for vue-demi
+export const Vue2 = Vue
+export const isVue2 = true
+export const isVue3 = false
+export const install = () => {}
+export const version = Vue.version
```
pi0, line 14: Should we also provide named set/get exports?

antfu: get/set is already provided by `@vue/composition-api`. The shims is only needed for Vue 3

antfu:
Added the mock for Nuxt 3, not sure what's the best place to set the alias, feel free to modify.

Also converting to draft for now, I will need to do some tests for it.

pi0 (approved):
Please ping when done testing @antfu  (we can always improve by trying edge releases too :D)

antfu:
Love your changes!

However I guess this is a bit tricky, in my testing, our mock does not take effects for `@vueuse/core` since it's been pre-optimized by vite (so the aliasing is not working), would only work when I exclude them:

```ts
optimizeDeps: {
  exclude: [
    '@vueuse/core',
    '@vueuse/shared',
    'vue-demi'
  ],
}
```

I don't think it's flexible as we can't possibly list all the packages that use `vue-demi`. Some idea:

- Disable the bundling feature of vite, which would be fine for most of the modern packages that only ship one single js file, but would cost the perf when user importing things like `lodash`.
- We let this to library integration library to extend the vite config then exclude themselves (I can do in VueUse and I think Pinia could do as well).

pi0:
Good points @antfu!

Well indeed this mocking solution only works for transpired dependencies to mitigate compatibility issues and certainly, we cannot prevent all based cases.

It makes sense to include `vue-demi` to `optimizeDeps.exclude` array by default and let module authors add by need. 

Regarding pinna's Nuxt module, I made a rewrite PR (https://github.com/posva/pinia/pull/791) that splits the vue2 plugin from vue3. But I can see for `@vueuse/*` case, it is almost impossible however I can see you are [prebundling](https://unpkg.com/browse/@vueuse/core@6.8.0/index.mjs) so I guess the impact of including it in `optimizeDeps.exclude` would below.

Related: Basically, the way to config a library being inlined, is by pushing to `options.build.transpile` that affects all different places to ensure lib is bundled and transpired. With a bridge, we exclude transpile by default :)

antfu:
That makes sense. Then I think we could merge it and I will do the exclusion on VueUse.

Issue #646, opened 2021-10-05
Client only components


pi0:
Related: #315, #648, #413

- [ ] Support `<ClientOnly>` component similar to Nuxt 2 (we can inspire from [egoist/vue-client-only](https://github.com/egoist/vue-client-only))
- [ ] Support `Foo.client.vue` to be removed from SSR bundle

Discussion: When using `<Foo>` without a `<ClientOnly>` wrapper, should we still support placeholder attribute and slot similar to `<ClientOnly>`? If not we are basically going to use a null element until hydration that can significantly affect client performance and make layout shifts.

andysay:
ClientOnly not work now ?

bountyx:
> ClientOnly not work now ?

Yeah, it seems to be the case that its not yet working.

Atinux:
It is not yet implemented

webfansplz:
mark

webfansplz:
@pi0 Hi, do we have any plans to support it recently ?

Gyurmatag:
Hey! When this will be available?

birdlavv:
Hey @Gyurmatag, it looks like it may be already available in the newer releases.
Check out: https://v3.nuxtjs.org/docs/directory-structure/components/#clientonly-component

PR #1853, opened 2021-11-11
feat: built-in client only component

antfu:
### 🔗 Linked issue

#646

### ❓ Type of change

### 📚 Description

Add built-in `<ClientOnly/>` component (taken [from VitePress](https://github.com/vuejs/vitepress/blob/49ddb1f120f1c66c6c8d1a7e388e2f49bdf1c552/src/client/app/components/ClientOnly.ts)).

### 📝 Checklist


packages/nuxt3/src/core/templates.ts
```diff
@@ -49,9 +51,11 @@ export const serverPluginTemplate = {
     const serverPlugins = ctx.app.plugins.filter(p => !p.mode || p.mode !== 'client')
     return [
       "import preload from '#app/plugins/preload.server'",
+      "import clientOnly from '#app/plugins/client-only'",
```
antfu, line 54: I am not sure what's the best way to add built-in plugins, or should I make a module for this?

pi0: Since this is a component, we can move it to `packages/nuxt3/src/app/components` and register as an auto import component to be included on bundle when used. We can use (https://github.com/nuxt/framework/blob/main/packages/nuxt3/src/core/nuxt.ts#L55) to register this component (this should probably later being refactored to `nuxt3/app.ts` module alongside with few other app related hooks

packages/nuxt3/src/app/plugins/client-only.ts
```diff
@@ -0,0 +1,16 @@
+import { ref, onMounted, defineComponent } from 'vue'
+import { defineNuxtPlugin } from '#app'
+
+export const ClientOnly = defineComponent({
+  setup (_, { slots }) {
+    const show = ref(false)
+    onMounted(() => {
+      show.value = true
+    })
+    return () => (show.value && slots.default ? slots.default() : null)
```
pi0: Would be nice if we can also support `placeholder` slot/props similar to `vue-client-only` (https://github.com/egoist/vue-client-only#placeholder)

packages/nuxt3/src/core/templates.ts
antfu: Thanks a lot for the pointers! Just updated. One last question - could we use `.ts` or should we use `.mjs`?

pi0: Both work but probably mjs is easier for internal testing to work with stub (alternatives: using .vue or import without ext)

packages/nuxt3/src/app/plugins/client-only.ts
antfu: Done. I named it `fallback` to align with `Suspense` or do you prefer `placeholder`?

packages/nuxt3/src/core/templates.ts
antfu: Changed to mjs. Probably not `.vue` since it has a manual render function

antfu:
I was thinking to add it to the docs, but I didn't find a way to place it. Do yo think we should create a new page like https://nuxtjs.org/docs/features/nuxt-components?

packages/nuxt3/src/app/plugins/client-only.ts
pi0: I have no preference and indeed it makes sense to be aligned with vue 3 namings. But can we maybe support both as aliases? It can help improving backward compatibility and for people that prefer placeholder over fallback.

pi0:
? I was thinking to add it to the docs, but I didn't find a way to place it. Do yo think we should create a new page like https://nuxtjs.org/docs/features/nuxt-components?

A "Built-in components" page might be good idea

packages/nuxt3/src/app/plugins/client-only.ts
antfu: Done.

pi0 (approved):

PR #1857, opened 2021-11-11
feat: add `useState` overrides

nandi95:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/1803

### ❓ Type of change

### 📚 Description
This adds better typing support to `useState`

This could be somewhat breaking for people already using this (type system wise). They may have to add non null assertions now after updating to nuxt that includes this change.
### 📝 Checklist


packages/nuxt3/src/app/composables/state.ts
```diff
@@ -1,5 +1,6 @@
+/* eslint-disable no-redeclare */
```
nandi95, line 1: Is this rule meant to be a thing in a typescript project?

docs/content/3.docs/1.usage/2.state.md
```diff
@@ -32,8 +32,17 @@ export default defineNuxtPlugin((nuxt) => {
 })
 ```
 
+```ts [composables/index.ts]
+import { InjectionKey } from "vue";
+
+export const stateKey: InjectionKey<'state'> = Symbol('state')
```
pi0, line 38: I'm wondering can we instead infer from string key types? /cc @danielroe

packages/nuxt3/src/app/composables/state.ts
danielroe: yes, we need to disable this for the function signature merging below

packages/nuxt3/src/app/composables/state.ts
```diff
@@ -8,9 +9,13 @@ import { useNuxtApp } from '#app'
  * @param key a unique key ensuring that data fetching can be properly de-duplicated across requests
  * @param init a function that provides initial value for the state when it's not initiated
  */
-export const useState = <T> (key: string, init?: (() => T)): Ref<T> => {
+export function useState<T> (key: InjectionKey<T>): Ref<T | undefined>
+export function useState<T = any> (key: string): Ref<T | undefined>
+export function useState<T> (key: InjectionKey<T>, init: (() => T)): Ref<T>
+export function useState<T = any> (key: string, init: (() => T)): Ref<T>
+export function useState<T = any> (key: InjectionKey<T> | string, init?: (() => T)): Ref<T> {
```
nandi95, line 16: Maybe these `any` types should be set to unknown to encourage type assertions and/or usage of symbols (That also ensures uniqueness).

docs/content/3.docs/1.usage/2.state.md
danielroe: I think it would be more consistent to use strings, particularly as the symbols will have to be stringified _anyway_, which means we lose the benefit of symbols being unique.

nandi95: infer from strings?
how would you maintain such level of state within the type system?

danielroe: Yes, you can infer from string literals. 👍 It also forces users to choose the string, and avoids any symbol minification issues in production.

nandi95: I do need help. I still understand how we could tie things to strings unless we have 
`const myKey = 'myKey' as const;` and the overloads checking for `'myKey'`?
You could update the PR or perhaps message me on discord (same name).

docs/content/3.docs/1.usage/2.state.md
```diff
@@ -32,12 +34,39 @@ export default defineNuxtPlugin((nuxt) => {
...
 </script>
 
 <template>
   Current locale: {{ locale }}
 </template>
 ```
+
+### Typing states
+
+In addition to using a `stateKey` to infer the type of `useState`, as above, you can also declare the type of the states throughout in your application:
```
danielroe, line 62:
```suggestion
Instead of using a `stateKey` to infer the type of `useState`, as above, you can also declare the type of the states throughout in your application:
```

docs/content/3.docs/1.usage/2.state.md
danielroe: I've pushed an implementation. Would value your thoughts. Particularly whether we need both global `NuxtStates` and `stateKey` options. cc: @pi0

packages/nuxt3/src/app/composables/state.ts
```diff
@@ -1,14 +1,26 @@
...
  * Create a global reactive ref that will be hydrated but not shared across ssr requests
  *
  * @param key a unique key ensuring that data fetching can be properly de-duplicated across requests
  * @param init a function that provides initial value for the state when it's not initiated
  */
-export const useState = <T> (key: string, init?: (() => T)): Ref<T> => {
+
+export function useState<T = unknown, K extends string = string, S = K extends keyof NuxtStates ? NuxtStates[K] : T> (key: K | StateKey<T>): Ref<S | undefined>
+export function useState<T = unknown, K extends string = string, S = K extends keyof NuxtStates ? NuxtStates[K] : T> (key: K | StateKey<T>, init: (() => K extends keyof NuxtStates ? NuxtStates[K] : T)): Ref<S>
+export function useState (key: string, init?: (() => any)): Ref<any> {
```
nandi95, line 23: Added NuxtStates key hint, replaced duplicate type for init method return value, removed redundant any type
```suggestion
export function useState<T = unknown, K extends string | keyof NuxtStates = string, S = K extends keyof NuxtStates ? NuxtStates[K] : T> (key: K | StateKey<T>): Ref<S | undefined>
export function useState<T = unknown, K extends string | keyof NuxtStates = string, S = K extends keyof NuxtStates ? NuxtStates[K] : T> (key: K | StateKey<T>, init: (() => S)): Ref<S>
export function useState (key: string, init?: (() => any)): Ref {
```

docs/content/3.docs/1.usage/2.state.md
```diff
@@ -32,12 +34,39 @@ export default defineNuxtPlugin((nuxt) => {
...
+import { stateKey } from '#app'
+
+export const localeKey = stateKey<string>('locale')
+```
+
 ```vue
 <script setup>
+import { localeKey } from '~/utils/state-keys'
+
+// this will be typed `string | undefined`
```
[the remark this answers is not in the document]
danielroe: It depends on whether you have `strict: true` set in your `tsconfig.json` or not...

There are not _yet_ type tests but we should create them 👍

packages/nuxt3/src/app/composables/state.ts
danielroe: the duplicate type for init method return value is necessary for TS to complain if you return the wrong type from the init - otherwise it silently reshapes the type.

I also don't think adding `| keyof NuxtStates` gives us any key hint, but please let me know if it's different for you.

nandi95: didn't realise about the duplicate 👍 

https://user-images.githubusercontent.com/41805560/141422385-10fb50cc-2592-4db7-afa6-68638b0ccd5b.mov

That's in jetbrains IDE

docs/content/3.docs/1.usage/2.state.md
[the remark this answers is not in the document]
nandi95: This is in this project, which is set to strict

https://user-images.githubusercontent.com/41805560/141423104-4b399794-77f4-4739-8f80-2d3cc5510c90.mov

Again, this is jetbrains IDE

docs/content/3.docs/1.usage/2.state.md
```diff
@@ -36,12 +38,39 @@ export default defineNuxtPlugin((nuxt) => {
...
 
+```ts [utils/state-keys.ts]
+import { stateKey } from '#app'
+
+export const localeKey = stateKey<string>('locale')
+```
+
 ```vue
 <script setup>
+import { localeKey } from '~/utils/state-keys'
```
pi0, line 49: If users are supposed to create `utils/` and import a key to type global state, they can already create a typed composable (and even enjoy auto-import)

```ts [composables/states.ts]
import { stateKey } from '#app'

export const useLocaleState = () => useState<string>('locale')
```

```vue [app.vue]
<script setup>
const locale = useLocaleState('locale')
</script>

<template>
  Current locale: {{ locale }}
</template>
```

pi0:
Hi @nandi95. Thanks for your time to work on this PR. 

Initially using Symbols was an attractive approach but finally looking at changes, I think we are adding too much complexity to achieve better typing while as an alternative, we can make use of auto imported compostables that can guarantee type and key validation with less boilerplate and more implicit typing. (please check #1945)

I'm closing this PR but please feel free to share your feedback on #1803 about the new pattern. We can always think of improving Typescript DX .

PR #1861, opened 2021-11-11
feat(nuxt3, bridge): add lazy helpers (`useLazyAsyncData` and `useLazyFetch`)

### LINKED ISSUE #1688: Add `useLazyAsyncData` and `useLazyFetch`
Currently, it is possible to use `useAsyncData` both as an awaited composable and not awaiting with `defer: true` and adding `pending` guard in template

Using an explicit alias:
- Allows easier explanation in docs about cons/pros of each method and why swr-like non-blocking navigation is better
- Allows backporting non-blocking lazy version to Nuxt Bridge
- Allows decoupling implementation in the future to improve blocking and non blocking composables individually

### LINKED ISSUE #1831: No typescript support anymore for imports from Vue
### Environment

- Node Version:     `v17.0.1`
- Nuxt Version:     `2.16.0-27273615.ab1c6cb4`
- Bundler:          `Webpack`
- User Config:      `target`, `bridge`, `vite`, `alias`, `nitro`, `ssr`, `head`, `css`, `plugins`, `components`, `tailwindcss`, `buildModules`, `modules`, `serverMiddleware`, `watch`, `privateRuntimeConfig`, `build`, `storybook`, `createRequire`
- Runtime Modules:  `-`
- Build Modules:    `@nuxtjs/tailwindcss@4.2.1`, `portal-vue/nuxt`, `@nuxt/postcss8@1.1.3`, `@pinia/nuxt@0.1.0`, `@nuxt/bridge@3.0.0-27276049.89aa7aa`

### Describe the bug

After importing `import Vue from 'vue'` `Vue` has type `any`. 

I think the issue is that 
https://github.com/nuxt/framework/blob/6ec518ff244a17c1986a092d2d28ac4c0d6182d5/packages/bridge/src/runtime/vue2-bridge.mjs#L1
generates a `vue2-bridge.d.ts` file that overwrites `from 'vue'` essentially with 'from 'vue2'`, which doesn't provide type info.

Or to be more precise, in the generated tsconfig under `paths` there is the entry
```
      "vue": [
        "node_modules/@nuxt/bridge/dist/runtime/vue2-bridge"
      ],
```
Deleting this brings back the typescript info for `Vue`.

### Reproduction

https://github.com/JabRef/JabRefOnline/pull/480

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #1688, resolves #1831

### ❓ Type of change

### 📚 Description

This PR:
 * renames `defer` to `lazy` within the `useAsyncData` and `useFetch` composables
 * adds new `useLazyAsyncData` and `useLazyFetch` as aliases for `lazy: true` versions of these composables
 * exposes these within Nuxt Bridge
 * adds `onServerPrefetch` to `useAsyncData` for Nuxt 2 support
 * adds documentation for `default` option
 * adds composition api migration docs

~~**Note**: there is a (separate) issue with typing for these composables within Bridge as `Ref` is being imported from `vue` - still thinking of how to solve 🤔~~

### 📝 Checklist

- [x] solve typing in Nuxt Bridge


docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -12,19 +12,20 @@ Within your pages, components and plugins you can use `useAsyncData` to get acce
...
 )
 ```
 
 * **key**: a unique key to ensure that data fetching can be properly de-duplicated across requests
 * **fn** an asynchronous function that returns a value.
 * **options**:
-  * _defer_: whether to load the route before resolving the async function (defaults to `false`)
+  * _lazy_: whether to resolve the async function after loading the route, instead of blocking navigation (defaults to `false`)
+  * _default_: a factory function to set the default value of the data, before the async function resolves - particularly useful with the `lazy: true` option
   * _server_: whether the fetch the data on server-side (defaults to `true`)
```
DamianGlowala, line 24: * _server_: whether **to** fetch the data on server-side (defaults to `true`)

DamianGlowala (changes requested):

docs/content/3.docs/1.usage/1.data-fetching.md
danielroe:
```suggestion
  * _server_: whether to fetch the data on server-side (defaults to `true`)
```

pi0 (approved):

PR #1876, opened 2021-11-12
fix(nitro): skip non existing externals

tobiasdiez:
### ❓ Type of change

### 📚 Description

Some packages don't have a `package.json` file in their root. For example, Prisma has the package.json in `.prisma/client/package.json` and not under `.prisma`. This leads to the following problem at the end of `nuxt generate`
```
 Nuxt Fatal Error                                                                                               │
   │                                                                                                                    │
   │   Error: ENOENT: no such file or directory, stat 'D:\Programming\JabRefOnline\node_modules\.prisma\package.json' 
```
This is fixed by silently ignore those packages (for prisma this is not a problem as `.prisma/client` is also imported, and thus the `package.json` file is still copied).

I wasn't sure about the coding style since the file uses an interesting mix of `await/async` and `Promise.then`.

### 📝 Checklist

### DIFF
--- a/packages/nitro/src/rollup/plugins/externals.ts
+++ b/packages/nitro/src/rollup/plugins/externals.ts
@@ -87,7 +87,7 @@ export function externals (opts: NodeExternalsOptions): Plugin {
 
         const writeFile = async (file) => {
           // Skip symlinks that are included in fileList
-          if (await fse.stat(file).then(i => i.isDirectory())) {
+          if (await fse.stat(file).then((i) => i.isDirectory()).catch(() => true)) {
             return
           }
           const src = resolve(opts.traceOptions.base, file)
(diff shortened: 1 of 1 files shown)


packages/nitro/src/rollup/plugins/externals.ts
```diff
@@ -87,7 +87,7 @@ export function externals (opts: NodeExternalsOptions): Plugin {
 
         const writeFile = async (file) => {
           // Skip symlinks that are included in fileList
-          if (await fse.stat(file).then(i => i.isDirectory())) {
+          if (await fse.stat(file).then(i => i.isDirectory()).catch(() => true)) {
```
pi0, line 90: We can probably make a helper function for this. `err.code` need to be tested against `ENOENT` and I guess we an only exclude `package.json` as safe-to-ignore. Generally hiding errors is probably not best idea and covers other bugs

pi0: fee8ed4978444d16a3baa9e66043072e08ceb94e

PR #1919, opened 2021-11-15
feat(components): `.client.vue` / `.server.vue` convention

antfu:
### 🔗 Linked issue

#646 

Continues from #1853

### ❓ Type of change

### 📚 Description

This PR bring `.client.vue` and `.server.vue` conversion for components directly. This PR is focus on `.client.vue` client only components with optional `.server.vue` as SSR fallback. Server-only components will be left to future PRs.

###### How it works

When we do the component scanning, components with suffixes will be treated as one component entry with two file paths. Then we create a virtual component using templates with the wrapper by `<ClientOnly>`. This way we could use the component as normal but be able to do environment-specific logic. 

###### Progress

- [x] Scanning for `.server.vue` and `.client.vue`
- [x] Virutal component
- [x] Optimize bundling

### 📝 Checklist


packages/kit/src/types/components.ts
```diff
@@ -9,6 +11,10 @@ export interface Component {
   prefetch: boolean
   preload: boolean
   global?: boolean
+  envPaths?: {
```
antfu, line 14: Not sure how should we call it, I was thinking we can call it "EnvComponents" for "Environment-specific components"

packages/nuxt3/src/components/module.ts
```diff
@@ -65,6 +65,15 @@ export default defineNuxtModule({
         return
       }
 
+      components.filter(i => i.envPaths).forEach((component) => {
+        component.filePath = `#build/c-${component.pascalName}.vue`
+        app.templates.push({
+          filename: `c-${component.pascalName}.vue`,
```
antfu, line 71: I guess you might have better ideas of how should we name it and where to put. (It's weird I can't specify dir paths in the template 🤔)

packages/nuxt3/src/components/templates.ts
```diff
@@ -22,6 +22,20 @@ const createImportMagicComments = (options: ImportMagicCommentsOptions) => {
...
+import Client from '${component.envPaths.client}'
+${component.envPaths.server ? `import Server from '${component.envPaths.server}'` : ''}
+</script>
+
+<template>
+  <ClientOnly>
+    <Client v-bind="$attrs"><slot/></Client>
+    ${component.envPaths.server ? '<template #fallback><Server v-bind="$attrs"><slot/></Server></template>' : ''}
+  </ClientOnly>
+</template>
```
antfu, line 36: TODO: optimize bundle to have different imports in client / server.

pi0:
Hi @antfu do you mind to fix merge conflicts?

antfu:
For the bundle optimization, for now, I can see two ways of doing it:

- We create two "virtual components" that and generate two `components` plugin templates for client / server
- We create a stub virtual component and add two transform plugins for server/client to transform the stub into env-specific components

pi0:
Some improvement suggestions:
- We can simplify component interface with `path: { [env: 'client' | 'server' | 'default']: string } `
- Instead of always using a virtual component wrapping with `<ClientOnly>`, we can configure aliases for server/client bundlers differently that resolves to env specific version (unless there is missing pair of component for env)

antfu:
> We can simplify component interface with path: { [env: 'client' | 'server' | 'default']: string }

My initial idea is that `filePath` could serve the path virtual module so the rest of the code could remove as-is. Also it would be easier to check if a component has multiple versions by `if(component.envPaths)`. This interface you proposal could also do, but is that ok as I guess it will be a breaking change?

> Instead of always using a virtual component wrapping with <ClientOnly>, we can configure aliases for server/client bundlers differently that resolves to env specific version (unless there is missing pair of component for env)

I'd love to, but I guess that's more like the initial approach that we faced the hydration errors - and that's why we introduced the `<ClientOnly>` component. I have given another try based on the work of this PR on branch https://github.com/nuxt/framework/tree/feat/client-server-components-redirects , but unfortunately, the hydration does not seem to like it.

image

antfu:
Guess what, I made it work! 🚀https://github.com/nuxt/framework/commit/d093fa92b917beecc0a5df1121bd2f5db27bc01b

packages/nuxt3/src/components/templates.ts
antfu: Done!

andysay:
hello, when it will be merged??

packages/nuxt3/src/app/components/client-only.mjs
```diff
@@ -17,3 +17,19 @@ export default defineComponent({
     }
   }
 })
+
+export function wrapClientOnly (component, mode) {
+  return defineComponent({
+    name: 'ClientOnlyWarpper',
```
WoH, line 23:
```suggestion
    name: 'ClientOnlyWrapper',
```

WoH, line 23:
```suggestion
    name: 'ClientOnlyWrapper',
```

antfu:
Global mode probably won't work due to the latest refactor of how components been resolved. Working with @danielroe on this.

antfu:
Resolved :)

Atinux:
New conflicts again 😅 

What do we do about this PR? If agree, I think next step if to update the docs in https://v3.nuxtjs.org/docs/directory-structure/components#clientonly-component in order to merge.

pi0:
I think will finally have to rework this PR. Idea is nice but implementation needs to be revised with simpler component options.

pi0:
Thank you again @antfu for working on this. Since PR is stalled and I expect simpler interface for implementation, let's track this via #646. We are also closer with nitro to support server-only components 🤞🏼

Issue #1818, opened 2021-11-10
Allow optional build-time type checking


pi0:
We can enable built-time type checking behind a flag (`typescript.typeCheck: true`) using [fork-ts-checker-webpack-plugin
](https://www.npmjs.com/package/fork-ts-checker-webpack-plugin) (for webpack 5), [@rollup/plugin-typescript](https://www.npmjs.com/package/@rollup/plugin-typescript) for rollup (nitro and vite prod) and [vite-plugin-typechecker](https://github.com/fi3ework/vite-plugin-checker).

PR #1940, opened 2021-11-15
feat: `typescript` options

pi0:
### ❓ Type of change

### 📚 Description

- Support `options.typescript`
  - `strict` (default is false)
  - `tsConfig` (allow overriding generated tsconfig defaults)
  - [placeholder] `typeCheck` (#1818)
- Strict mode is disabled by default allowing users to progressively update their codebase and learn ts
- Add note about strict typechecking in concepts > typescript
 
### 📝 Checklist

### DIFF
--- a/packages/kit/src/config/schema/typescript.ts
+++ b/packages/kit/src/config/schema/typescript.ts
@@ -1,0 +1,14 @@
+export default {
+  /**
+   * TypeScript comes with certain checks to give you more safety and analysis of your program.
+   * Once you’ve converted your codebase to TypeScript, you can start enabling these checks for greater safety.
+   * [Read More](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html#getting-stricter-checks)
+   */
+  strict: false,
+
+  /**
+   * You can extend generated `.nuxt/tsconfig.json` using this option
+   * @typedef {Awaited<ReturnType<typeof import('pkg-types')['readPackageJSON']>>}
+   */
+  tsConfig: {}
+}
--- a/packages/nuxi/src/utils/prepare.ts
+++ b/packages/nuxi/src/utils/prepare.ts
@@ -1,19 +1,20 @@
 import { promises as fsp } from 'fs'
 import { join, relative, resolve } from 'pathe'
 import { Nuxt, TSReference } from '@nuxt/kit'
+import defu from 'defu'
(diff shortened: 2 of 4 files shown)


packages/kit/src/config/schema/typescript.ts
```diff
@@ -0,0 +1,14 @@
...
+  /**
+   * TypeScript comes with certain checks to give you more safety and analysis of your program.
+   * Once you’ve converted your codebase to TypeScript, you can start enabling these checks for greater safety.
+   * [Read More](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html#getting-stricter-checks)
+   */
+  strict: false,
+
+  /**
+   * You can extend generated `.nuxt/tsconfig.json` using this option
+   * @typedef {Awaited<ReturnType<typeof import('pkg-types')['readPackageJSON']>>}
```
pi0, line 11: /cc @danielroe This was only possible could find to type with the current version of untyped :D `typeof import('pkg-types).TSConfig` apparently is not working since is a type-only export

packages/nuxi/src/utils/prepare.ts
```diff
@@ -1,19 +1,20 @@
...
 
-  const tsConfig: TSConfig = {
+  const tsConfig: TSConfig = defu(nuxt.options.typescript?.tsConfig, {
     compilerOptions: {
       target: 'ESNext',
       module: 'ESNext',
       moduleResolution: 'Node',
       skipLibCheck: true,
-      strict: true,
+      strict: nuxt.options.typescript.strict,
```
pi0, line 17: Hidden fix: I tried to put it as defaults in `typescript.compilerOptions` but untyped had strange behavior with `$resolve`. Have to check more in-depth making reprodution

packages/kit/src/config/schema/typescript.ts
nandi95: why not type this as `TSConfig` from `import type { TSConfig } from 'pkg-types'` with typescript?
Then `@typedef` isn't even that important anymore?

nandi95:
I know it's a bit late to the party but I think the default `strict` should be true. And people who progressively upgrade should turn this off if they need to. That is to encourage strictness by default in the community and not have a "breaking" change if and when this is turned back to true later down the line.

PR #1949, opened 2021-11-16
fix(vite): external modules under`/esm/` pattern

antfu:
Co-authored-by: Ben <bencodezen@users.noreply.github.com>

### ❓ Type of change

### 📚 Description

Found during pairing with @bencodezen 🙌

To reproduce the bug (Vite):

```html
<script setup lang="ts">
import format from 'date-fns/format'

const date = format(new Date(), 'yyyy/MM/dd')
</script>

<template>
  <div>
    {{ date }}
  </div>
</template>
```

which will cause the error of `export` is not a valid syntax. Searching down, the root cause is that the internal module of `date-fns` been externalized but in a `.js` format of ESM in `server.mjs`

This PR adds the conversion to externalize modules under `/esm/` which I believe is a common pattern (`date-fns` is using is at least).

https://unpkg.com/browse/date-fns@2.25.0/esm/_lib/protectedTokens/index.js

### 📝 Checklist

### DIFF
--- a/packages/vite/src/server.ts
+++ b/packages/vite/src/server.ts
@@ -42,6 +42,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       noExternal: [
         ...ctx.nuxt.options.build.transpile,
         // TODO: Use externality for production (rollup) build
+        /\/esm\//,
         /\.(es|esm|esm-browser|esm-bundler).js$/,
         '/__vue-jsx',
         '#app',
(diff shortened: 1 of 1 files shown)


packages/vite/src/server.ts
```diff
@@ -42,6 +42,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       noExternal: [
         ...ctx.nuxt.options.build.transpile,
         // TODO: Use externality for production (rollup) build
+        /\/esm\//,
```
pi0, line 45: This check seems too broad. 
```suggestion
        /\/esm\/.*\.js$/,
```

pi0:
Also made https://github.com/unjs/mlly/issues/23 to resolve for mlly/externality

PR #1740, opened 2021-11-05
feat(kit, nuxt3): add `pages:extend` and `extendPages`

danielroe:
### 🔗 Linked issue

resolves #1739

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This allows usage of `extendPages` in kit (and adds `extendPages` hook in Nuxt 3).

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #1966, opened 2021-11-16
fix(kit): support chainable `extendRoutes` in container

### LINKED ISSUE #1965: `extendRoutes` from modules broken with Bridge
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.18.0`
- Nuxt Version:     `2.16.0-27282256.ab1c6cb4`
- Bundler:          `Webpack`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://codesandbox.io/s/modest-currying-8liys?file=/pages/test.vue

### Describe the bug

Using `this.extendRoutes()` from a regular module (`async function () { console.log(this) }`) is broken since latest versions of Bridge.

I couldn't identify which version tag is breaking it exactly, but it seem not to be more than a week old (or two).

### Additional context

I spotted that by debugging an issue coming from `@nuxtjs/i18n`.

The error was that the additional routes generated by this module were not added to the routes list.

Console was spammed with these error messages:
```
[vue-router] Route with name 'test___fr' does not exist
```

Here are the code references:

The place where `this.extendRoutes` is called:
https://github.com/nuxt-community/i18n-module/blob/main/src/index.js#L87

The function that never gets executed:
https://github.com/nuxt-community/i18n-module/blob/555e66e183f5188e61477b7bb6f10ba872bf5714/src/core/hooks.js#L22

The patch used to resolve the issue:
https://github.com/docusgen/app/blob/dev/patches/%40nuxtjs%2Bi18n%2B7.2.0.patch

### Logs

```shell
There is no logs to be provided, as this issue does not seem to throw a specific error, apart from the one being thrown by the modules that relies on that feature.

You can find more context in the reproduction though.
```

pi0:
### 🔗 Linked issue

ref: #1740, #1965

### ❓ Type of change

### 📚 Description

- Allow chaining routes when using kit's legacy ModuleContainer 
- Separate `extendRoutes` and `extendPages` implementations

Caveats: For Nuxt3, chaining is still not supported anymore even when using legacy ModuleContainer (A warning will be emitted to show when skipping)

### 📝 Checklist

### DIFF
--- a/packages/kit/src/module/container.ts
+++ b/packages/kit/src/module/container.ts
@@ -107,7 +107,12 @@ export function createModuleContainer (nuxt: Nuxt) {
       if (isNuxt2(nuxt)) {
         nuxt.options.router.extendRoutes = chainFn(nuxt.options.router.extendRoutes, fn)
       } else {
-        nuxt.hook('pages:extend', fn)
+        nuxt.hook('pages:extend', async (pages, ...args) => {
+          const maybeRoutes = await fn(pages, ...args)
+          if (maybeRoutes) {
+            console.warn('[kit] [compat] Using `extendRoutes` in Nuxt 3 needs to directly modify first argument instead of returning updated routes. Skipping extended routes.')
+          }
+        })
       }
     },
 
--- a/packages/kit/src/module/container.ts
+++ b/packages/kit/src/module/container.ts
@@ -108,7 +108,7 @@ export function createModuleContainer (nuxt: Nuxt) {
       if (isNuxt2(nuxt)) {
         nuxt.options.router.extendRoutes = chainFn(nuxt.options.router.extendRoutes, fn)
       } else {
-        extendPages(routes => fn(routes, resolveAlias))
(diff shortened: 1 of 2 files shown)


packages/kit/src/module/container.ts
```diff
@@ -104,7 +104,16 @@ export function createModuleContainer (nuxt: Nuxt) {
...
+      if (isNuxt2(nuxt)) {
+        nuxt.options.router.extendRoutes = chainFn(nuxt.options.router.extendRoutes, fn)
+      } else {
+        nuxt.hook('pages:extend', async (pages, ...args) => {
+          const maybeRoutes = await fn(pages, ...args)
+          if (maybeRoutes) {
+            console.warn('[kit] [compat] Using `extendRoutes` in Nuxt 3 needs to directly modify first argument instead of returning updated routes. Skipping extended routes.')
+          }
+        })
+      }
```
danielroe, line 115: I don't think we need to chain routes.

PR #1976, opened 2021-11-16
chore(ci): run test on active lts nodejs versions

clarkdo:
### ❓ Type of change

### 📚 Description

- Run test and bridge test on active lts node versions
- Use latest lts version to run lint an release

### 📝 Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node: [14]
+        node: [lts/*]
 
     steps:
       - uses: actions/checkout@v2
@@ -39,7 +39,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node: [14]
+        node: [lts/fermium, lts/gallium]
 
     steps:
       - uses: actions/checkout@v2
@@ -66,7 +66,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node: [14]
+        node: [lts/fermium, lts/gallium]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


danielroe (approved):

pi0:
Thanks for PR @clarkdo but for few reasons I rather to keep it as is at least for now:
- Minimum Node.js version of Nuxt, is not exactly **latest LTS** (while recommended by docs) (see discussion https://github.com/nuxt/framework/discussions/1234) and this needs to be a manual bump
- At least for now that we are running a big test matrix against `[nuxt3,bridge]x[each platform]x[vite, webpack]`[win, linux]` running each on more than one version increases test time and github limit usage (for windows mainly). So even if we are going to test against all supported LTS (14, 16) we should wait for https://github.com/nuxt/framework/discussions/1690 that moved platform out of the matrix

clarkdo:
The reason I made pr for running test (at least build) on multiple nodejs versions is that I happened to see `yarn build` is broken in a lower version nodejs 14, so I'm wondering if any pr in future may break one of the supported major nodejs versions. What do you think of at least running yarn build as health checking on multiple envs ?

pi0:
>  see yarn build is broken in a lower version nodejs 14,

We both enforce the minimum version using the "engines" field in all `package.jsons` and a runtime warning for users.  Can you please describe the scenario you had the issue?

clarkdo:
It’s error about import CommonJS default exports as named exports because mkdirp package doesn’t support esm, I remembered that I was using 14.8 and seeing error by just running yarn build.

clarkdo:
I think this has been fixed in 14.13, but it makes me think that testing multiple versions makes sense

pi0:
`engines.node` in package.json (monorepo) is `^14.16.0 || ^16.11.0 || ^17.0.0`. `yarn install` should disallow this but seems not. I guess we can handle this check in `unbuild` then to fail CI if using unsupported Node.js version (moved to https://github.com/unjs/unbuild/issues/22)

clarkdo:
Nice, I’ll take a look

PR #1988, opened 2021-11-17
feat(kit): add more specific typings for schema

danielroe:
### ❓ Type of change

### 📚 Description

This PR:
* adds some more specific types for the nuxt config schema
* moves resolution handling for components and meta -> schema

### 📝 Checklist

- [x] https://github.com/unjs/untyped/pull/25
- [x] https://github.com/unjs/untyped/pull/26


packages/kit/src/config/schema/_adhoc.ts
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * Configure Nuxt component auto-registration.
+   *
+   * Any components in the directories configured here can be used throughout your
+   * pages, layouts (and other components) without needing to explicitly import them.
+   *
+   * @default {{ dirs: [`~/components`] }}
+   * @see [Nuxt 3](https://v3.nuxtjs.org/docs/directory-structure/components) and
+   * [Nuxt 2](https://nuxtjs.org/docs/directory-structure/components/) documentation
+   * @type {boolean | typeof import('../src/types/components').ComponentsOptions | typeof import('../src/types/components').ComponentsOptions['dirs']}
```
pi0, line 11: This should directly refer to `@nuxt/kit`

danielroe: The issue is that because this schema is used as _source_ for the `@nuxt/kit` entrypoint, it is a circular reference. We need to keep this relative, and it will be compiled out in the final `index.d.ts` build.

pi0: Good idea

pi0: Only issue is that `kit/schema` is being published to npm with not existing imports which is guess we can remove it later on

pi0 (approved):

PR #1996, opened 2021-11-17
chore(bridge): Improve typing for vueApp

tobiasdiez:
### ❓ Type of change

### 📚 Description

The methods of `nuxtApp.vueApp` were of type `any` for me. Instead of using the `VueAppCompat` interface, I propose to directly use `VueConstructor`.

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/runtime/app.ts
+++ b/packages/bridge/src/runtime/app.ts
@@ -1,6 +1,5 @@
 import type { Hookable } from 'hookable'
-// @ts-ignore
-import type { Vue } from 'vue/types/vue'
+import type { VueConstructor } from 'vue'
 import type { ComponentOptions } from 'vue'
 import { defineComponent, getCurrentInstance } from './composables'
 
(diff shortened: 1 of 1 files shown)


packages/bridge/src/runtime/app.ts
```diff
@@ -11,26 +9,11 @@ export const defineNuxtComponent = defineComponent
...
-  use: Vue['use']
-  version: string
-}
-
 export interface NuxtAppCompat {
   nuxt2Context: Vue
   vue2App: ComponentOptions<Vue>
 
-  vueApp: VueAppCompat
+  vueApp: VueConstructor
```
pi0, line 16: `vueApp` is a `vue@3` a partial shim (while `VueConstructor` imported from `vue` in Nuxt2/bridge is from Vue2). Maybe we can use `VueConstructor` and keep `VueAppCompat`?

tobiasdiez: Can you maybe clarify the purpose of vueApp in the context of bridge. All plugins and other libraries are based on vue2 anyway. Or is it meant as a migration path to use vue3-based libraries with nuxt/bridge?

pi0: Indeed. it mimic vue3 insterface for integrations with vue3-only libs like `@vueuse/head`. Otherwise on bridge you should rely on `vue2App`

tobiasdiez: Okay thanks for the clarification. In this case, this PR doesn't make any sense. Let me see if I find a nice solution for VueAppCompat and I'll open a new PR then.

PR #2001, opened 2021-11-17
feat(nuxt3): automatically inject returns from plugins

### LINKED ISSUE #1881: automatic plugin injection
### Discussed in https://github.com/nuxt/framework/discussions/1864

<div type='discussions-op-text'>

<sup>Originally posted by **danielroe** November 11, 2021</sup>
## Background

We now have a number of automatic features:
* automatic component registration & type recognition (via `GlobalComponents`)
* automatic composables/helpers registration & type declarations
* automatic server-api typings
* automatic response handling for direct returns from server middleware

This means that the typing of injected properties from plugins (via `nuxtApp.provide`) is one of the few areas left where users still have to manually augment types of their application. See issues (for example, https://github.com/nuxt/framework/issues/1863), and [current docs](https://v3.nuxtjs.org/docs/directory-structure/plugins#typing-plugins).

## Proposal

Users can return an object from a plugin to inject these values into the context, with the keys of the object being the keys injected. 

A key benefit of this approach is that we can automatically generate the types for these injected properties (including `$router`, for example - see https://github.com/nuxt/framework/issues/1794).

### Example
`~/plugins/auth.ts`:
```js
export default defineNuxtPlugin(() => {
  return {
    auth: {
      login: () => 'success'
    }
  }
})
```

`index.vue`
```vue
<template>
  {{ $auth.login() }}
</template>

<script setup>
// automatically typed
const { $auth } = useNuxtApp()
</script>
```

danielroe:
### 🔗 Linked issue

resolves  #1881

### ❓ Type of change

### 📚 Description

This PR implements automatic plugin injections. It also converts `Plugin` type to generic, in a non-breaking way, so we can infer return types.

These types are not defined on the `NuxtApp` type itself (which would lead to circular reference as the plugins receive this type) but on the return of `useNuxtApp()` which can be used in plugins, if necessary, to get typesafe access to injected properties.

### 📝 Checklist


docs/content/3.docs/2.directory-structure/10.plugins.md
```diff
@@ -24,22 +24,62 @@ export default defineNuxtPlugin(nuxtApp => {
...
+```
+
+## Typing plugins
+
+If you return your helpers from the plugin, they will be typed automatically; you'll find them typed for the return of `useNuxtApp()` and within your templates.
 
+::alert
+If you need to use a provided helper _within_ another plugin, you can call `useNuxtApp()` to get the typed version. But in general this should be avoided due to issues with ordering of plugins.
+::
+
```
danielroe, line 63:
```suggestion
```

packages/nuxt3/src/app/nuxt.ts
```diff
@@ -117,9 +119,14 @@ export function createNuxtApp (options: CreateOptions) {
   return nuxtApp
 }
 
-export function applyPlugin (nuxtApp: NuxtApp, plugin: Plugin) {
+export async function applyPlugin (nuxtApp: NuxtApp, plugin: Plugin) {
   if (typeof plugin !== 'function') { return }
-  return callWithNuxt(nuxtApp, () => plugin(nuxtApp))
+  const result = await callWithNuxt(nuxtApp, () => plugin(nuxtApp))
```
pi0, line 124: - maybe `const injections`? 
- I missed this in the initial RFC review but I think would be safer if we return `{ provide: {}` from plugins. If later plugins want to expose other things like hooks we are not locked-in :)

packages/nuxt3/src/app/nuxt.ts
```diff
@@ -56,7 +56,7 @@ export interface NuxtApp {
 
 export const NuxtPluginIndicator = '__nuxt_plugin'
 export interface Plugin<Injections extends Record<string, any> = Record<string, any>> {
-  (nuxt: NuxtApp): Promise<void> | Promise<Injections> | void | Injections
+  (nuxt: NuxtApp): Promise<void> | Promise<{ provide?: Injections }> | void | { provide: Injections }
```
pi0, line 59:
```suggestion
  (nuxt: NuxtApp): Promise<void> | Promise<{ provide?: Injections }> | void | { provide?: Injections }
```

packages/nuxt3/src/app/nuxt.ts
```diff
@@ -121,10 +121,10 @@ export function createNuxtApp (options: CreateOptions) {
 
 export async function applyPlugin (nuxtApp: NuxtApp, plugin: Plugin) {
   if (typeof plugin !== 'function') { return }
-  const result = await callWithNuxt(nuxtApp, () => plugin(nuxtApp))
-  if (result && typeof result === 'object') {
-    for (const key in result) {
-      nuxtApp.provide(key, result[key])
+  const { provide } = await callWithNuxt(nuxtApp, () => plugin(nuxtApp)) || {}
+  if (provide && typeof provide === 'object') {
```
pi0, line 125: This condition also passes by `provide: null` if type safety should take into account but I'm wondering do we need the overhead of check if partially handling edge cases? 

```suggestion
  if (provide) {
```

packages/nuxt3/src/core/templates.ts
```diff
@@ -74,3 +76,32 @@ export const appViewTemplate = {
...
+import type { Plugin } from '#app'
+
+type Decorate<T extends Record<string, any>> = { [K in keyof T as K extends string ? \`\${K}\` : never]: T[K] }
+
+type InjectionType<A extends Plugin> = A extends Plugin<infer T> ? Decorate<T> : unknown
+
+type PluginReturns = \n  ${tsImports.map(p => `InjectionType<typeof import('${p}').default>`).join(' &\n  ')}
+
+declare module '#app' {
+  interface NuxtAppInjections extends PluginReturns { }
```
pi0, line 97: @danielroe I'm wondering why would we need to extend `NuxtAppInjections` instead of (same as advanced section) extending `interface Nuxt App extends PluginReturns` ?)

pi0:
Note: (to do in same PR or later) Bridge types support seems missing

packages/nuxt3/src/core/templates.ts
danielroe: This will cause a circular reference: https://github.com/nuxt/framework/pull/2001/commits/5e3d94a78028bdeb67f656a1a43ca5ec921cbc71

pi0: How? `NuxtApp` type is also being used **without `useNuxtApp`** (like in other plugins). We need to find another way then like an internal `_NuxtApp` to allow extending anywhere.

danielroe: > If you need to use a provided helper _within_ another plugin, you can call `useNuxtApp()` to get the typed version. But in general this should be avoided unless you are certain of the plugins' order.

The issue is that `NuxtApp` type is used _within_ defineNuxtPlugin, so we can't then augment it based on the result of `defineNuxtPlugin`. I think the option implemented here is a good solution, but happy to talk more. Implementing an internal `_NuxtApp` would be fine, but we also couldn't provided the injected types to `defineNuxtPlugin` or to other plugins for the same reason of circularly referenced types.

packages/nuxt3/src/app/nuxt.ts
```diff
@@ -25,13 +25,13 @@ export interface RuntimeNuxtHooks {
   'meta:register': (metaRenderers: Array<(nuxt: NuxtApp) => NuxtMeta | Promise<NuxtMeta>>) => HookResult
 }
 
-export interface NuxtApp {
+export interface _NuxtApp {
```
pi0, line 28: To avoid confusion of usage, maybe keep it unexported?

```suggestion
interface _NuxtApp {
```

packages/nuxt3/src/core/templates.ts
```diff
@@ -74,3 +76,32 @@ export const appViewTemplate = {
...
+import type { Plugin } from '#app'
+
+type Decorate<T extends Record<string, any>> = { [K in keyof T as K extends string ? \`$\${K}\` : never]: T[K] }
+
+type InjectionType<A extends Plugin> = A extends Plugin<infer T> ? Decorate<T> : unknown
+
+type NuxtAppInjections = \n  ${tsImports.map(p => `InjectionType<typeof import('${p}').default>`).join(' &\n  ')}
+
+declare module '#app' {
+  interface NuxtApp extends NuxtAppInjections { }
```
pi0, line 97: Just to ensure this seems okay now?

danielroe: Yes, this is working in my tests 👍.

Issue #1795, opened 2021-11-09
loadNuxt not found


jbmolle:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.13.0`
- Nuxt Version:     `3.0.0-27274229.29599f0`
- Package Manager:  `npm@8.1.0`
- Bundler:          `Vite`
- User Config:      `publicRuntimeConfig`, `buildModules`, `intlify`
- Runtime Modules:  `-`
- Build Modules:    `@intlify/nuxt3@0.1.8`, `nuxt-windicss@2.0.12`

### Describe the bug

nuxt-windicss use @nuxt/kit: latest as a dependency.
As soon as I download the package nuxt-windicss, my app cannot start and throws an error "loadNuxt not found".

In packages/nuxi/src/utils/kit.ts, we have the function loadKit.
This function is called in packages/nuxi/src/commands/dev.ts at line 36
```
const { loadNuxt, buildNuxt } = await loadKit(rootDir)
```
The problem is that loadKit returns a default key. So we should either do 
```
const { default as defaultKit } = await loadKit(rootDir)
const { loadNuxt, buildNuxt } = defaultKit
```
Or change packages/nuxi/src/utils/kit.ts with
```
import { importModule } from './cjs'

export const loadKit = async (rootDir: string): Promise<typeof import('@nuxt/kit')> => {
  try {
    const module = await importModule('@nuxt/kit', rootDir) as typeof import('@nuxt/kit')
    return module.default
  } catch (e) {
    if (e.toString().includes("Cannot find module '@nuxt/kit'")) {
      throw new Error('nuxi requires `@nuxt/kit` to be installed in your project. Try installing `nuxt3` or `@nuxt/bridge` first.')
    }
    throw e
  }
}
```

### Reproduction

The problem comes from the last version of nuxt-windicss which uses @nuxt/kit: latest.
Reproduction:
npx nuxi init nuxt3-app

Then add in package.json, devDependencies -> "nuxt-windicss": "^2.0.12"

npm run dev fails with "Cannot start nuxt:  loadNuxt is not a function"

jbmolle:
Hi @pi0 @danielroe 
Actually the problem arises when I use @nuxt/kit
I don't know why but the lib uses CJS so when I start the app I get the warning message: "You are using CommonJS version of `@nuxt/kit` that is an ESM-only package. CJS support will be dropped in stable release."
And loadKit returns a default key.
[...]

PR #2008, opened 2021-11-18
fix: Nuxi cjs importModule with jiti

jbmolle:
### 🔗 Linked issue

#1795

### ❓ Type of change

- [X] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

Solves #1795
We couldn't loadNuxt in CJS because import would return a default key.
By using Jiti we ensure that both in CJS and ESM we get { loadNuxt: Function[] } when calling loadKit.
And we can then destructure const { loadNuxt } = loadKit()

### 📝 Checklist

- [X] I have linked an issue or discussion.

### DIFF
--- a/packages/nuxi/src/utils/cjs.ts
+++ b/packages/nuxi/src/utils/cjs.ts
@@ -1,6 +1,6 @@
 import { createRequire } from 'module'
-import { pathToFileURL } from 'url'
 import { normalize, dirname } from 'pathe'
+import jiti from 'jiti'
 
 export function getModulePaths (paths?: string | string[]): string[] {
   return [].concat(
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/utils/cjs.ts
```diff
@@ -31,7 +31,7 @@ export function requireModule (id: string, paths?: string | string[]) {
 
 export function importModule (id: string, paths?: string | string[]) {
   const resolvedPath = resolveModule(id, paths)
-  return import(pathToFileURL(resolvedPath).href)
+  return jiti(resolvedPath, { interopDefault: true })(resolvedPath)
```
pi0, line 34: I'm not sure this is a right fix. `importModule` is intended to use [native ESM import](https://v3.nuxtjs.org/concepts/esm)

pi0:
> We couldn't loadNuxt in CJS because import would return a default key.

`@nuxt/kit` should be imported using `await import('@nuxt/kit')` in the CJS context.

jbmolle:
Yes I understand.
The problem is that `importModule` called in `loadKit` in `framework/packages/nuxi/src/utils/kit.ts` returns `{ loadNuxt, ... }` in ESM and `{ default: { loadNuxt, ... } }` in CJS. So when using @nuxt/kit: latest, we can't start Nuxt.

pi0:
CJS proxy removed from `@nuxt/kit` stub (https://github.com/nuxt/framework/issues/1795#issuecomment-972749713)

PR #2010, opened 2021-11-18
fix(bridge): add docs links and warnings for data composables

### LINKED ISSUE #1993: Docs (bridge): clarify migration of defineNuxtPlugin
On one hand, the docs https://github.com/nuxt/framework/blob/main/docs/content/1.getting-started/4.bridge-composition-api.md#definenuxtplugin recommend to simply replace defineNuxtPlugin by a function export. On the other hand, https://github.com/nuxt/framework/blob/main/docs/content/3.docs/2.directory-structure/10.plugins.md and https://github.com/nuxt/framework/blob/7cc526ce69a541782f16023deb5104dc51068615/packages/bridge/src/runtime/capi.legacy.mjs#L82 advocate to still use `defineNuxtPlugin` imported from `#app`.

Moreover, I'm not sure if the advised typing information (`import type { Plugin } from '@nuxt/types'`) makes sense; at least I cannot get a `nuxtApp` from the context: The context from `@nuxt/types` doesn't have a `$_nuxtApp` variable; but maybe one is supposed to use `useNuxtApp` anyway?

danielroe:
### 🔗 Linked issue

resolves #1993

### ❓ Type of change

### 📚 Description

This adds docs links to capi warnings/deprecations and re-enables global deprecation warning now that we have `useMeta` & data fetching composables.

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/runtime/capi.legacy.mjs
+++ b/packages/bridge/src/runtime/capi.legacy.mjs
@@ -214,6 +214,7 @@ export const defineComponent = (options) => {
 }
 
 export const useMeta = (init) => {
+  warnOnce('useMeta', 'You are using `useMeta`, which has a replacement provided by Nuxt Bridge. See https://v3.nuxtjs.org/getting-started/bridge-composition-api#usemeta).')
   const vm = getCurrentInstance()
   const refreshMeta = () => vm.$meta().refresh()
 
@@ -241,6 +242,7 @@ export const useMeta = (init) => {
 
 // Wrapped properties
 export const wrapProperty = (property, makeComputed = true) => () => {
+  warnOnce('wrapProperty', 'You are using `wrapProperty`, which is deprecated. See https://v3.nuxtjs.org/getting-started/bridge-composition-api#wrapproperty).')
   const vm = getCurrentInstance()
   return makeComputed ? computed(() => vm[property]) : vm[property]
 }
--- a/packages/bridge/src/runtime/capi.legacy.mjs
+++ b/packages/bridge/src/runtime/capi.legacy.mjs
@@ -62,7 +62,7 @@ export { ref }
 
 // Common deprecation utils
 // TODO: Add migration guide docs to @nuxtjs/composition-api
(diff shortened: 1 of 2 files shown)


packages/bridge/src/runtime/capi.legacy.mjs
```diff
@@ -73,24 +73,23 @@ const warnOnce = (id, message) => {
...
 
-// TODO: Add warning back once we had replacement for useAsync and useMeta
 // Warn in case of having any imports from `@nuxtjs/composition-api`
-// warnOnce('import', `\`@nuxtjs/composition-api\` is deprecated. ${checkDocsMsg}`)
+warnOnce('import', `\`@nuxtjs/composition-api\` is deprecated. ${checkDocsMsg}`)
 
 // Stub functions that provided type support
-export const defineNuxtMiddleware = unsupported('You are using `defineNuxtMiddleware`, which is not supported. You can remove wrapper function to keep using Nuxt 2 middleware.')
-export const defineNuxtPlugin = unsupported('You are using `defineNuxtPlugin`, which can be replaced with `defineNuxtPlugin` (import from `#app`).')
+export const defineNuxtMiddleware = unsupported('You are using `defineNuxtMiddleware`, which is not supported. See https://v3.nuxtjs.org/getting-started/bridge-composition-api#definenuxtmiddleware).')
```
pi0, line 80: Maybe a refactor like this is worth reducing bundle size and maintenance?

```suggestion
export const defineNuxtMiddleware = unsupported(`You are using \`defineNuxtMiddleware\`, which is not supported. See ${DOCS_CAPI_MIGRATION}#definenuxtmiddleware).`)
```

(alternatively we can just replace DOCS_CAPI_MIGRATION in warnOnce

danielroe: Or we could just link to the page rather than the specific section, and just append to all warnings...

pi0: That would be also nice 👍

PR #2013, opened 2021-11-18
fix(bridge): exclude bridge alias types to support Volar

### LINKED ISSUE #1920: Volar support for bridge
When generating `tsconfig.json` for Nuxt bridge, we shall set `experimentalCompatMode` that helps work Volar with Vue 2/Bridge out of the box. 

To test: Not sure if volar also resolves `tsconfig.json` with `extend` for `experimentalCompatMode` support or not. /cc @johnsoncodehk

Bonus: Would be nice DX if we can somehow also eliminate requirement of `@vue/runtime-dom` being installed

Refs: (I could find)
- https://github.com/johnsoncodehk/volar/commit/647391f6d72ba7c6d2154c541d5151a0e11b5364
- https://github.com/nuxt/modules/commit/cd127cd879d08be934e0e88e9ec5d634be5d3c6e

antfu:
### 🔗 Linked issue

#1920

### ❓ Type of change

### 📚 Description

Currently, the Volar support on Bridge is still broken. I think the root cause is that Volar will use the types from `@vue/runtime-core` to resolve the types but we are forced to alias it. 

image

I can confirm locally that removing the selected items will make it work.

Repro: 
- https://github.com/antfu/vitesse-nuxt-bridge
- https://github.com/nuxt/modules

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/utils/prepare.ts
+++ b/packages/nuxi/src/utils/prepare.ts
@@ -39,7 +39,13 @@ export const writeTypes = async (nuxt: Nuxt) => {
     '#assets': '@nuxt/nitro'
   }
 
+  // Do not alias types for Vue 3 packages to support Volar
+  const excludedAlias = [/^vue$/, /^@vue\/.*$/]
+
   for (const alias in aliases) {
+    if (excludedAlias.some(re => re.test(alias))) {
+      continue
+    }
     const relativePath = relative(nuxt.options.rootDir, aliases[alias]).replace(/(?<=\w)\.\w+$/g, '') /* remove extension */ || '.'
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/utils/prepare.ts
```diff
@@ -39,7 +39,13 @@ export const writeTypes = async (nuxt: Nuxt) => {
     '#assets': '@nuxt/nitro'
   }
 
+  // Do not alias types for Vue 3 packages to support Volar
```
pi0, line 42:
```suggestion
  // Do not bridge alias types to support Volar
```

PR #2019, opened 2021-11-18
Update 12.server.md

Sascha-Block:
Clarification that the directory needs to be created first

### ❓ Type of change

### 📝 Checklist


danielroe (commented):
I think this will be true for all of the directories in the list, and shouldn't be called out specifically for this one.

Sascha-Block:
@danielroe Understood, but then - just in relation to the circumstance and those reference - it would make sense and seems to be important to explain this overarching determination in a sentence at a higher level and also to resolve the "redundancies"...

PR #1960, opened 2021-11-16
docs: add examples section

Atinux:
Will be used later on for examples section 👀 

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

- [x] Add stackblitz component (pros: super fast)
- [ ] Add codesandbox component (pros: mobile support and typescript)
- [ ] Basic design for `/examples` section (sidebar and page)
- [ ] Auto generate from `/examples` to `/docs/content/examples`
- [ ] Add instructions to clone and play locally using `npx nuxi init`

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #2057, opened 2021-11-21
docs(examples): add example for runtime route extend

Atinux:
### ❓ Type of change

### 📚 Description

Since vue-router 4 has the `addRoute` method, I believe adding routes at runtime is quite interesting.
We can discuss about merging this example with https://github.com/nuxt/framework/tree/main/examples/module-extend-pages to showcase the two possibilities.

Once done, and adding the examples section (#1960), we will be able to improve the documentation about routing 😊 

### 📝 Checklist

- [ ] There is an hydration error happening when hitting the dynamically added route coming from SSR. Any help on this would be much appreciated @posva 

https://user-images.githubusercontent.com/904724/142760337-b35ee82b-901d-459b-a43f-3f25bd460a41.mp4

To run the example locally (read https://github.com/nuxt/framework#-development first):
```bash
yarn example extend-routes
```

### DIFF
--- a/examples/extend-routes/README.md
+++ b/examples/extend-routes/README.md
@@ -1,6 +1,6 @@
 # Adding routes dynamically
 
-This example showcase how to add routes at runtime using [`router.addRoute()`](https://next.router.vuejs.org/guide/advanced/dynamic-routing.html#adding-routes) method from [vue-router](https://next.router.vuejs.org).
+This example showcases how to add routes at runtime using [`router.addRoute()`](https://next.router.vuejs.org/guide/advanced/dynamic-routing.html#adding-routes) method from [vue-router](https://next.router.vuejs.org).
 
 Play online on [StackBliz](https://stackblitz.com/github/nuxt/framework/tree/main/examples/extend-routes)
 
--- a/examples/extend-routes/README.md
+++ b/examples/extend-routes/README.md
@@ -1,0 +1,11 @@
+# Adding routes dynamically
+
+This example showcase how to add routes at runtime using [`router.addRoute()`](https://next.router.vuejs.org/guide/advanced/dynamic-routing.html#adding-routes) method from [vue-router](https://next.router.vuejs.org).
+
+Play online on [StackBliz](https://stackblitz.com/github/nuxt/framework/tree/main/examples/extend-routes)
+
+## Setup
+
+```bash
+npx nuxi init nuxt3-app -t nuxt/framework/examples/extend-routes
+```
--- a/examples/extend-routes/README.md
+++ b/examples/extend-routes/README.md
@@ -2,7 +2,7 @@
 
 This example showcases how to add routes at runtime using [`router.addRoute()`](https://next.router.vuejs.org/guide/advanced/dynamic-routing.html#adding-routes) method from [vue-router](https://next.router.vuejs.org).
 
-Play online on [StackBliz](https://stackblitz.com/github/nuxt/framework/tree/main/examples/extend-routes)
+Play online on [StackBlitz](https://stackblitz.com/github/nuxt/framework/tree/main/examples/extend-routes).
 
 ## Setup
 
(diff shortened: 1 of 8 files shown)


examples/extend-routes/plugins/extend-routes.ts
```diff
@@ -0,0 +1,6 @@
+export default defineNuxtPlugin(({ $router }) => {
+  $router.addRoute({
```
pi0, line 2: It would be nice if we add `addRoute` auto import from pages module. Making this simpler and less prune to errors

pi0:
Can you please use documentation to better explain purpose is for runtime dynamic routes addition? Current example is adding a route in all conditions, lacking a real world use

examples/extend-routes/README.md
```diff
@@ -0,0 +1,11 @@
+# Adding routes dynamically
+
+This example showcase how to add routes at runtime using [`router.addRoute()`](https://next.router.vuejs.org/guide/advanced/dynamic-routing.html#adding-routes) method from [vue-router](https://next.router.vuejs.org).
```
DamianGlowala, line 3: This example showcase**s** (...)

Atinux:
One example is the ability to add routes for localised website, I think that's what @loick was looking for to use in BackMarket website.

examples/extend-routes/plugins/extend-routes.ts
Atinux: I was looking at how we could have it but cannot find, do you have a clue where/how to add this?

I tester and `useRouter()` returns `undefined` in Nuxt plugins.

Atinux:
Some context in Nuxt v2: https://github.com/nuxt/nuxt.js/issues/9594

pi0:
Seems a nice example using either localized condition, feature flags or context user.

examples/extend-routes/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Play online on [StackBliz](https://stackblitz.com/github/nuxt/framework/tree/main/examples/extend-routes)
```
DamianGlowala, line 5: - StackBliz -> StackBli**t**z
- Missing full stop ending the sentence

DamianGlowala (changes requested):

examples/extend-routes/README.md
pi0: Thanks! Small tip you can use <kbd>ctrl+g</kbd> to directly add change suggestion:

image

pi0:
```suggestion
Play online on [StackBlitz](https://stackblitz.com/github/nuxt/framework/tree/main/examples/extend-routes)
```

examples/extend-routes/README.md
pi0:
```suggestion
This example showcases how to add routes at runtime using [`router.addRoute()`](https://next.router.vuejs.org/guide/advanced/dynamic-routing.html#adding-routes) method from [vue-router](https://next.router.vuejs.org).
```

examples/extend-routes/README.md
DamianGlowala: Oh, that will make it more convenient, didn't know that - thanks! (:

examples/extend-routes/plugins/extend-routes.ts
posva: it looks like you are adding the route after the router is used by app with `app.use(router)` which triggers the initial navigation on the client to a still non existent route.
That is okay, but ideally, some navigation guard should detect that the route must be added dynamically, add the route, and then trigger a new navigation with `next(to.fullPath)` or by returning the location.

Nirlah:
For reference, the plugin `addRoute` approach causes an hydration error #2475

pi0:
Moving to #3600 to add example when feature lands.

PR #2058, opened 2021-11-21
docs(examples): add `integrations-supabase`

Atinux:
### ❓ Type of change

### 📚 Description

This example showcases how to use [Supabase](https://supabase.com) with [Nuxt](https://v3.nuxtjs.orgs).

## Features

- `useSupabase()` composable
- User sessions with SSR support
- Handle auth changes (and removes the session)


examples/integrations-supabase/nuxt.config.ts
```diff
@@ -0,0 +1,16 @@
+import { defineNuxtConfig } from 'nuxt3'
+
+export default defineNuxtConfig({
+  publicRuntimeConfig: {
+    supabase: {
+      url: process.env.SUPABASE_URL,
```
pi0, line 6: Key shall be same as env name to allow overriding un runtime

examples/integrations-supabase/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "example-integrations-supabase",
+  "private": true,
+  "devDependencies": {
+    "@nuxthq/ui": "npm:@nuxthq/ui-edge@latest",
```
pi0, line 5: TODO: we need to move it to nuxt/ui first

examples/integrations-supabase/plugins/supabase.server.ts
```diff
@@ -0,0 +1,11 @@
+import { useCookies } from 'h3'
+
+export default defineNuxtPlugin(async (nuxtApp) => {
+  const { user, auth } = useSupabase()
+  const req = nuxtApp.ssrContext!.req
+
+  // Supabase needs to read from req.cookies
+  req.cookies = useCookies(req)
+  // Check authenticated user during SSR
+  user.value = (await auth.api.getUserByCookie(req)).user
```
pi0, line 10: This logic can be handle on demand in composable without using a global plugin (unless global guard is the purpose)

examples/integrations-supabase/server/api/supabase/session.ts
```diff
@@ -0,0 +1,19 @@
...
+export default async (req, res) => {
+  const body = await useBody(req)
+
+  // Mock for supabase API expecting to use Express 👀
+  // TODO: open an issue for Supabase gotrue-js
+  req.body = body
+  req.cookies = useCookies(req)
+  res.status = () => ({ json: () => ({}) })
+
+  supabase.auth.api.setAuthCookie(req, res)
```
pi0, line 16: I believe this can also be handled in app namespace (ssr only) instead of adding an additional api route

examples/integrations-supabase/README.md
```diff
@@ -0,0 +1,48 @@
+# Nuxt + Supabase
+
+This example showcases how to use [Supabase](https://supabase.com) with [Nuxt](https://v3.nuxtjs.orgs).
+
+## Features
+
+- `useSupabase()` composable
+- User sessions with SSR support
+- Handle auth changes (and removes the session)
```
DamianGlowala, line 9: I would suggest rewording to:

Handle auth changes (and remov**ing** the session)

examples/integrations-supabase/README.md
```diff
@@ -0,0 +1,48 @@
...
+```bash
+yarn install
+cp .env.example .env
+```
+
+Create an account and a project on [Supabase](https://supabase.com) and add the `SUPABASE_URL` and `SUPABASE_KEY` in `.env`.
+
+## Development
+
+Start the development server on http://localhost:3000
```
DamianGlowala, line 34: Missing colon at the end:

Start the development server on http://localhost:3000:

DamianGlowala (changes requested):

examples/integrations-supabase/README.md
DamianGlowala, line 34:
```suggestion
Start the development server on http://localhost:3000:
```

examples/integrations-supabase/README.md
DamianGlowala, line 9: I would suggest either of the following:
- Handle**s** auth changes (and removes the session)
- Handl**ing** auth changes (and remov**ing** the session)

examples/integrations-supabase/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+[![Open in StackBlitz](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://stackblitz.com/github/nuxt/framework/tree/main/examples/integrations/supabase)
+
+## Setup
+
+```bash
+npx nuxi init nuxt3-app -t nuxt/framework/examples/integrations/supabase
+```
+
+Make sure to install the dependencies and setup the `.env`.
```
DamianGlowala, line 23:
```suggestion
Make sure to install the dependencies and setup the `.env`:
```

DamianGlowala (changes requested):

examples/integrations-supabase/server/api/supabase/session.ts
Atinux: Well actually when doing the login it needs to also call our server to store the cookie as secured. How to would want to handle it without calling back the server?

Or we need to have a saveSession() exposed from Nuxt to easily create HTTP only cookies for secured sessions experience from day 1

examples/integrations-supabase/plugins/supabase.server.ts
Atinux: Well actually having it by default offers a lot of features directly (feature flags per user, redirects, etc).

But if Nuxt exposes a smarter way to cachify() the user to speed up the process that would be neat.

I need to time the time execution of the plugin.

Something that Nuxt itself should expose (through hooks) so we can display them in the devtools (performance tabs)

examples/integrations-supabase/package.json
Atinux: Let's get an estimate date for it :)

examples/integrations-supabase/nuxt.config.ts
Atinux: Not sure to understand, what feature does it bring?

examples/integrations-supabase/README.md
Atinux: It feels weird to me to add the : at the end

Atinux: What do you think @danielroe ?

sduduzog:
This is so awesome. I've been trying to experiment with ways to try add SSR support to https://github.com/supabase-community/nuxt-supabase with no luck but I'll definitely take learnings here and work on a PR there.

examples/integrations-supabase/nuxt.config.ts
pi0: Overriding env variables at **runtime**. Current config basically adds env on build-time to bundle. Check this: https://github.com/nuxt/framework/blob/d68318f9d5ee25f6377edb866cd1b5846e8bfdce/packages/nitro/src/runtime/app/config.ts#L7

examples/integrations-supabase/package.json
pi0: If not ready soon, we shall remove UI components, same as other examples but I guess initial version should be ready by next week.

examples/integrations-supabase/plugins/supabase.server.ts
pi0: Not quite sure to understand what you mean but yes we can cache promise within a composable/context :)

examples/integrations-supabase/server/api/supabase/session.ts
pi0: Sure we probaly need `useSession`. My main point is that it can be in a composable and yet running in server-only to set secure cookie. This reduces boilerplate and allows logic working with rest of ssr context.

examples/integrations-supabase/README.md
pi0: This convention can also potentially make formatting issues. If we do decide to add `:`, should use explicit markdown syntax for generated link or add a space after 3000

examples/integrations-supabase/server/api/supabase/session.ts
Atinux: You need to try the demo to understand, when using Supabase, you basically no need to create API routes.

So I don't see how you want to implement this without an API call to works.

examples/integrations-supabase/nuxt.config.ts
Atinux: What would be the purpose of changing these keys at runtime? Do you have an example?

pi0: it is runtime config :) Changing config without need to rebuild

examples/integrations-supabase/server/api/supabase/session.ts
pi0: Same server logic just in composable :) Will make refactor

examples/integrations-supabase/composables/useSupabase.ts
```diff
@@ -0,0 +1,39 @@
...
+  const user = useState<User | null>('supabase_user', () => null)
+  const nuxtApp = useNuxtApp()
+  const config = useRuntimeConfig()
+
+  const supabase = createClient(config.supabase.url, config.supabase.key, config.supabase.options)
+
+  if (process.client && !authStateListerner) {
+    // Once Nuxt app is mounted
+    authStateListerner = nuxtApp.hook('app:mounted', () => {
+      // Check if user is authenticated on client-side
```
larbish, line 16:
```
      // Check if user is authenticated on client-side
      const session = supabase.auth.session()
      if (session) {
        // Set cookie on server-side
        setServerSession('SIGNED_IN', session)
        user.value = supabase.auth.user()
      }
```

This part of code can be removed. Cookie are already set on server side by `supabase.auth.onAuthStateChange()`

larbish (changes requested):

eMeRiKa13:
Thanks a lot for this code. It's very helpful. 

Could you add an example of how to query Supabase in a page, (other to the authenticated user) please?

`supabase.from("my_table").select('*').eq(...)`

I'm trying to use the useSupabase() composable to do that but I can't make it work.

pi0:
Finally moving this to a Nuxt module. https://github.com/larbish/supabase-module/issues/8

PR #2116, opened 2021-11-23
feat(nitro): pass Rollup options from Nuxt config (#2091)

### LINKED ISSUE #2091: Nitro Watcher Needs Polling Config
### Environment

Docker, Windows, WSL

### Reproduction

Run the most minimal Nuxt3 boilerplate setup via the docs in a simple docker container.

### Describe the bug

For all typical/known Docker+WSL file watcher issues:

Vite polling works with `usePolling: true`.

Webpack works with chokidar or polling config options set.

**Nitro has no (documented) support for the equivalent.**

Polling option needs to be added, or if it's already supported, the docs need to be much clearer on how to enable polling.

~~(And development should have better testing for Docker+WSL.)~~

### Additional context

_No response_

### Logs

_No response_

peteromano:
### 🔗 Linked issue

#2091 

### ❓ Type of change

### 📚 Description

Specifically, WSL 1 users still require polling on the Nitro server to trigger rebuilds on file changes. Other build components support this option, but seems to have been missed for Nitro.

`RollupOptions` have been added as `NuxtConfig.build.rollup`, such that `chokidar.usePolling` may be used for Nitro for example.

Resolves #2091 

### 📝 Checklist

### DIFF
--- a/packages/nitro/src/rollup/config.ts
+++ b/packages/nitro/src/rollup/config.ts
@@ -79,6 +79,7 @@ export const getRollupConfig = (nitroContext: NitroContext) => {
   const runtimeAppDir = join(nitroContext._internal.runtimeDir, 'app')
 
   const rollupConfig: RollupConfig = {
+    ...nitroContext.rollupConfig,
     input: resolvePath(nitroContext, nitroContext.entry),
     output: {
       dir: nitroContext.output.serverDir,
--- a/packages/schema/src/config/build.ts
+++ b/packages/schema/src/config/build.ts
@@ -28,6 +28,14 @@ export default {
   analyze: false,
 
   /**
+   * Pass through Rollup options. Useful, for example, if polling for rebuild is required for some environments.
+   *
+   * @see [Rollup Options](https://rollupjs.org/guide/en/#big-list-of-options)
+   * @type {RollupOptions}
+   */
+  rollup: {},
+
+  /**
    * Enable the profiler in webpackbar.
    *
    * It is normally enabled by CLI argument `--profile`.
--- a/docs/content/3.docs/2.directory-structure/12.server.md
+++ b/docs/content/3.docs/2.directory-structure/12.server.md
@@ -72,3 +72,31 @@ export default async (req: IncomingMessage, res: ServerResponse) => {
   req.someValue = true
 }
 ```
+
+## Server Options
+
+You may include Rollup's [big list of options](https://rollupjs.org/guide/en/#big-list-of-options) in the Nuxt config as:
+
+```js
+export default defineNuxtConfig({
+  build: {
+    rollup: {
+      // ...
+    }
+  }
+})
+```
+
+if you need support for file watch polling for the server directory, for example:
+
+```js
+export default defineNuxtConfig({
+  build: {
+    rollup: {
+      chokidar: {
+        usePolling: true
+      }
+    }
+  }
+})
+```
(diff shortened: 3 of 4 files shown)


packages/schema/src/config/build.ts
```diff
@@ -27,6 +27,14 @@ export default {
    */
   analyze: false,
 
+  /**
+   * Pass through Rollup options. Useful, for example, if polling for rebuild is required for some environments.
+   *
+   * @see [Rollup Options](https://rollupjs.org/guide/en/#big-list-of-options)
+   * @type {RollupOptions}
```
danielroe, line 34:
```suggestion
   * @type {typeof import('rollup').RollupOptions}
```

packages/nitro/src/rollup/config.ts
```diff
@@ -79,6 +79,7 @@ export const getRollupConfig = (nitroContext: NitroContext) => {
   const runtimeAppDir = join(nitroContext._internal.runtimeDir, 'app')
 
   const rollupConfig: RollupConfig = {
+    ...nitroContext.rollupConfig,
```
danielroe, line 82: we should likely be merging these configs rather than simply spreading at the beginning.

pi0:
Thanks for PR @peteromano but:

1. It doesn't exactly resolve #2091 to automatically enable watcher pooling in incompatible environments
2. We already allow using `{ nitro: { rollupConfig: {} } }` from nuxt.config. `build.rollup` is unnecessary and ambiguous as vite uses rollup for production build as well and not for same purpose.

packages/schema/src/config/build.ts
peteromano: I originally tried that, it doesnt work.. the build cant resolve it.

peteromano:
@pi0 Nitro config in nuxt config is undocumented.. But also, this:

```
  nitro: {
    rollupConfig: {
      chokidar: {
        usePolling: true
      }
    }
  },
  ```
  
doesn't work.

Vite config also doesn't work automatically (vite config needs explicit polling option for WSL 1 to work.)

Also, config to rollup.rollup isnt actually using rollupConfig from nuxt config, and rollup config isn't in the config type defs.. Are you sure we don't need this? 

Putting it under `build` is arguable -- i didn't know the ebst key to put it under and was waiting for feedback on that, but i think its still needed to put it _somewhere_ on the nuxt config.

packages/nitro/src/rollup/config.ts
peteromano: That's effectively what this does, no? I use this pattern all the time for simple merging.

packages/schema/src/config/build.ts
peteromano: Does it need `typeof` though?

pi0:
Your change on `rollup.config` is a nice fix 👍  (we missed option merging probably in a previous refactor). And sure providing option as workaround is nice (but not solving the root cause which we might later look into automated solution with a background self-test)

packages/nitro/src/rollup/config.ts
pi0: We might use [defu](https://github.com/unjs/defu) to apply defaults from user config :)

peteromano: @pi0 ah, interesting

peteromano:
@pi0 So what changes are needed to get this PR approved?

BTW, i just installed a WSL 2 ubuntu and the nitro watcher still doesn't work! Very strange..

peteromano:
Closing in favor of env var `CHOKIDAR_USEPOLLING =1` (see https://github.com/nuxt/framework/issues/2091#issuecomment-980791094)

Issue #2061, opened 2021-11-21
Tried to install Nuxt3 using the documentation site, failed. because of schema-edge


timothymarois:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v14.18.1`
- Nuxt Version:     `latest`
- Package Manager:  `yarn@1.22.5`
- Bundler:          `Webpack`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

It's easy, go here: 

https://v3.nuxtjs.org/getting-started/installation

run `npx nuxi init .`

then `npm install` 

and it failed. 

### Describe the bug

```
npm ERR! code ETARGET
npm ERR! notarget No matching version found for @nuxt/schema-edge@3.0.0-27291865.d68318f.
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
npm ERR! notarget 
npm ERR! notarget It was specified as a dependency of 'nuxt3'
npm ERR! notarget 
```

timothymarois:
`yarn install` also fails, but does give you this massive list of version builds to choose from **which is ridiculous.**

<img width="554" alt="Screen Shot 2021-11-21 at 11 36 29 AM" src="https://user-images.githubusercontent.com/2489333/142770788-f3d802a9-18a1-4d74-b16d-0bb6b1d9f49f.png">

yarn error: 

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

Zibbp:
I'm having the same issue as well when re-installing node modules for an existing project.
![image](https://user-images.githubusercontent.com/21207065/142772543-5e2eeae9-3927-442e-8c4b-8a38e93492e5.png)

weleo:
I'm having the same issue as well when installing node modules for a new project.
![Screenshot 2021-11-21 at 18 28 57](https://user-images.githubusercontent.com/57549150/142772638-0c52f0e7-c9f5-4b24-8941-3df9906f6bf5.png)

Niron68:
Same issue here, i temporary installed using this version and it works for people who want to install: 3.0.0-27291847.07c7a20

pkej:
Use overrides to work around it. The error I got was:

```shell
[...]

PR #2120, opened 2021-11-24
docs: add section about upgrading nuxt version

gangsthub:
### 🔗 Linked issue

#2061

### ❓ Type of change

### 📚 Description

After seeing this comment (https://github.com/nuxt/framework/issues/2061#issuecomment-975002663) on one issue, I thought it would be nice to add this information into the docs. Let me know what you think about the wording or if it could fit better somewhere else. It's obviously open for edits as well.

I based the explanation on the current implementation of the command: https://github.com/nuxt/framework/blob/5eb8e07/packages/nuxi/src/commands/upgrade.ts

### 📝 Checklist

### DIFF
--- a/docs/content/1.getting-started/2.installation.md
+++ b/docs/content/1.getting-started/2.installation.md
@@ -62,6 +62,19 @@ npm run dev -- -o
 Well done! A browser window should automatically open for <http://localhost:3000>
 ::
 
+## Upgrade
+
+If you want to upgrade Nuxt to the latest version, you can use `nuxi` as well:
+
+```bash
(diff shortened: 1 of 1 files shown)


danielroe (commented):
How about including this in `/getting-started/commands` instead?

gangsthub:
Oh, thanks! I swear I searched for it but couldn't find it! I see there is a section with the upgrade command already on it.

https://v3.nuxtjs.org/getting-started/commands#upgrade-nuxt3-version

Probably I'll just add the `--force` thing in there, for now.

My apologies!

PR #2129, opened 2021-11-24
docs: remove dangling comma from sample `tsconfig.json`

wiese:
Dangling comma is not valid JSON.
Given the whole section is added, it does not look like `experimentalCompatMode` is added as one of many properties, but it's the only one and then should not have a comma.

### DIFF
--- a/docs/content/1.getting-started/3.bridge.md
+++ b/docs/content/1.getting-started/3.bridge.md
@@ -129,7 +129,7 @@ If you are using TypeScript, you can edit your `tsconfig.json` to benefit from a
     ...
   },
 + "vueCompilerOptions": {
-+   "experimentalCompatMode": 2,
++   "experimentalCompatMode": 2
 + }
 }
 ```


danielroe (approved):
Seems a good change to make for the sake of consistency, but interestingly, a `tsconfig.json` does support dangling commas and even comments.

wiese:
Maybe a bit late to ask but is this line really necessary anyway given it is already part of [`.nuxt/tsconfig.json`](https://github.com/nuxt/framework/blob/0a74940bf619d33e9163e40b305105cb1dfb3001/packages/bridge/src/app.ts#L60) which we get instructed to extend [above](https://v3.nuxtjs.org/getting-started/bridge/#update-tsconfigjson)?

danielroe:
@wiese It is necessary because volar doesn't yet support reading this value from an extended config: see https://github.com/nuxt/framework/issues/1920#issuecomment-969056786

PR #2130, opened 2021-11-24
fix(nuxt3): handle hydration for `error` and `pending` state

### LINKED ISSUE #2032: pending status of `useLazyAsyncData` or `useLazyFetch` with SSR is incorrect
### Environment

- Operating System: `Linux`
- Nuxt Version:     `2.16.0-27288016.ab1c6cb4`
- Bundler:          `Webpack`
- User Config:      `buildModules`
- Runtime Modules:  `-`
- Build Modules:    `@nuxt/bridge@3.0.0-27288761.d3fb512`

### Reproduction

code sandbox: https://codesandbox.io/s/heuristic-kowalevski-elrth

- click **picked users** (= with CSR)
  - all `pending` will be false
- Reload preview (=with SSR)
  - Some of the pending may turn out to be true

### Describe the bug

When using `useLazyAsyncData` or `useLazyFetch` in a component that is called multiple times in a page, `pending` may not be resolved correctly.

The `pending` remains `true` even though the correct response is obtained. This is only seen in components with SSR, and does not seem to occur when loading after a page transition.

Also, the network requests seem to be fetched on the client side unnecessarily, even though the data retrieved on the server side is hydrated.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #2122: useAsyncData does not pass errors from server to client-side
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-27294839.7e5e50b`
- Bundler:          `Vite`

### Reproduction

https://stackblitz.com/edit/github-ygz9df?file=app.vue

### Describe the bug

On server side the error content is rendered  as expected while the client side says it's null.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2122
resolves #2032

### ❓ Type of change

### 📚 Description

This PR includes pending/error state within the payload. Otherwise a page that (for example) checked `v-if="error"` would fail to hydrate between client/server. Because we're including pending state as well, we can use that to detect whether we need to refetch data on client side.

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/runtime/app.plugin.mjs
+++ b/packages/bridge/src/runtime/app.plugin.mjs
@@ -64,9 +64,7 @@ export default (ctx, inject) => {
     nuxtApp.vue2App = this
   })
 
-  ctx.app.mounted.push(function () {
-    nuxtApp.isHydrating = false
-  })
+  ctx.app.mounted.push(() => { nuxtApp.isHydrating = false })
 
   const proxiedApp = new Proxy(nuxtApp, {
     get (target, prop) {
--- a/packages/bridge/src/runtime/app.plugin.mjs
+++ b/packages/bridge/src/runtime/app.plugin.mjs
@@ -5,6 +5,7 @@ import { setNuxtAppInstance } from '#app'
 // Reshape payload to match key `useLazyAsyncData` expects
 function proxiedState (state) {
   state._asyncData = state._asyncData || {}
+  state._errors = state._errors || {}
   return new Proxy(state, {
     get (target, prop) {
       if (prop === 'data') {
@@ -39,7 +40,7 @@ export default (ctx, inject) => {
     globalName: 'nuxt',
     payload: proxiedState(process.client ? ctx.nuxtState : ctx.ssrContext.nuxt),
     _asyncDataPromises: [],
-    isHydrating: ctx.isHMR,
+    isHydrating: true,
     nuxt2Context: ctx
   }
 
@@ -51,12 +52,20 @@ export default (ctx, inject) => {
     ctx.app.created = [ctx.app.created]
   }
 
+  if (!Array.isArray(ctx.app.mounted)) {
+    ctx.app.mounted = [ctx.app.mounted]
+  }
+
   if (process.server) {
     nuxtApp.ssrContext = ctx.ssrContext
   }
 
   ctx.app.created.push(function () {
     nuxtApp.vue2App = this
+  })
+
+  ctx.app.mounted.push(function () {
+    nuxtApp.isHydrating = false
   })
 
   const proxiedApp = new Proxy(nuxtApp, {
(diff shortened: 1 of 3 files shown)


pi0:
By changing flatten structure of payload into `{ data, error?, pending? }` for each key, we are increasing HTML size. 

- `pending` state is really a runtime specific flag. It should not be hydrated and  `useLazy*` composables need to set it to `true` by default. (It can be in `nuxtApp._payloadState` that is not hydrated.

- `error` should be explicitly consumed by by user (to display in component or log). Adding it to payload, not only reduces performance but also adds security risk of exposing server errors to public.

packages/bridge/src/runtime/app.plugin.mjs
```diff
@@ -51,6 +52,10 @@ export default (ctx, inject) => {
     ctx.app.created = [ctx.app.created]
   }
 
+  if (!Array.isArray(ctx.app.mounted)) {
```
pi0, line 55: Any clues why it can be a nonarray?

packages/bridge/src/runtime/app.plugin.mjs
```diff
@@ -59,6 +64,10 @@ export default (ctx, inject) => {
     nuxtApp.vue2App = this
   })
 
+  ctx.app.mounted.push(function () {
+    nuxtApp.isHydrating = false
```
pi0, line 68: Maybe an arrow function since not using `this`?

packages/bridge/src/runtime/app.plugin.mjs
danielroe: By default it's a single function: https://github.com/nuxt/nuxt.js/blob/9d33456e69a3bc74734461fa2eb866d2ceac490f/packages/vue-app/template/App.js#L120

pi0: Ohh sorry it was for bridge plugin 🤦

pi0 (approved):
LGTM (didn't locally test hydration changes hope is good :D)

nndnha:
> * `error` should be explicitly consumed by by user (to display in component or log). Adding it to payload, not only reduces performance but also adds security risk of exposing server errors to public. PS: A client-side retry strategy can be used to obtain payload for errors like timeouts.

@pi0 
```javascript
const { error } = await useFetch('https://jsonplaceholder.typicode.com/404');
```
This line of code can run on both server and client with the same result and I can say that the server error for this action is already public so it was completely unrelated to Nuxt. I just want to get the  http status code(404 in this case) and those recent commits transformed it from `FetchError` to `Boolean` then how can I get the status code on client side?https://stackblitz.com/edit/github-ygz9df-hnnre8?file=app.vue

packages/bridge/src/runtime/app.plugin.mjs
```diff
@@ -39,7 +40,7 @@ export default (ctx, inject) => {
     globalName: 'nuxt',
     payload: proxiedState(process.client ? ctx.nuxtState : ctx.ssrContext.nuxt),
     _asyncDataPromises: [],
-    isHydrating: ctx.isHMR,
+    isHydrating: true,
```
voltane, line 43: This change also fixes bug, where `useLazyAsyncData` was called again on client side during hydration.

PR #2131, opened 2021-11-24
fix(schema): extend `NuxtOptions` for nitro & bridge types

### LINKED ISSUE #2126: `NuxtConfig` missing `nitro` type
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.16.1`
- Nuxt Version:     `3.0.0-27294839.7e5e50b`
- Package Manager:  `pnpm@6.17.1`
- Bundler:          `Vite`
- User Config:      `nitro`, `env`, `alias`, `srcDir`, `dev`, `build`, `publicRuntimeConfig`, `buildModules`, `meta`, `css`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `@vueuse/core/nuxt@6.9.2`, `@pinia/nuxt@0.1.5`

### Reproduction

-

### Describe the bug

image

Maybe it’s what caused by :

https://github.com/nuxt/framework/commit/4fe1fc881c06a7f1b61e922c42792c26355fccd1

https://github.com/nuxt/framework/commit/4fe1fc881c06a7f1b61e922c42792c26355fccd1

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2126

### ❓ Type of change

### 📚 Description

Follow-up on migration to `@nuxt/schema`. I've also changed NuxtConfig to extend NuxtOptions. But this may have been  intentional; let me know if so (cc: @pi0).

### 📝 Checklist

### DIFF
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -1,11 +1,11 @@
 import { ConfigSchema } from '../../schema/config'
 
-/** Normalized Nuxt options */
+/** Normalized Nuxt options available as `nuxt.options.*` */
 export interface NuxtOptions extends ConfigSchema { }
 
 type DeepPartial<T> = T extends Record<string, any> ? { [P in keyof T]?: DeepPartial<T[P]> | T[P] } : T
 
-/** User configuration (may be different from normalized NuxtOptions) */
+/** User configuration in `nuxt.config` file */
 export interface NuxtConfig extends DeepPartial<ConfigSchema> { }
 
 
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -4,7 +4,7 @@ export interface NuxtOptions extends ConfigSchema { }
 
 type DeepPartial<T> = T extends Record<string, any> ? { [P in keyof T]?: DeepPartial<T[P]> | T[P] } : T
 
-export interface NuxtConfig extends DeepPartial<ConfigSchema> { }
+export interface NuxtConfig extends DeepPartial<NuxtOptions> { }
 
 
 export interface PublicRuntimeConfig extends Record<string, any> { }
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -1,10 +1,12 @@
 import { ConfigSchema } from '../../schema/config'
 
+/** Normalized Nuxt options */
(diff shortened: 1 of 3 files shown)


packages/schema/src/types/config.ts
```diff
@@ -4,7 +4,7 @@ export interface NuxtOptions extends ConfigSchema { }
 
 type DeepPartial<T> = T extends Record<string, any> ? { [P in keyof T]?: DeepPartial<T[P]> | T[P] } : T
 
-export interface NuxtConfig extends DeepPartial<ConfigSchema> { }
+export interface NuxtConfig extends DeepPartial<NuxtOptions> { }
```
pi0, line 7: Why changing? This was intentional. Both config and internal options extend same base schema but user input is different than normalized options (and module options should be typed by kit wrapper)

danielroe: I'll add a comment to explain. 👍

danielroe: The reason for the PR is that `ConfigSchema` is not exposed and can't be augmented at the moment. If the intent is to augment `ConfigSchema` I can fix that too; just let me know.

pi0: The intent is to augment NuxtConfig to type hint `nuxt.config` file. Modules on the other hand should not depend on `nuxt.options` types for their module options but use resolved options type.

packages/schema/src/types/config.ts
```diff
@@ -1,9 +1,11 @@
 import { ConfigSchema } from '../../schema/config'
 
+/** Normalized Nuxt options */
 export interface NuxtOptions extends ConfigSchema { }
 
 type DeepPartial<T> = T extends Record<string, any> ? { [P in keyof T]?: DeepPartial<T[P]> | T[P] } : T
 
+/** User configuration (may be different from normalized NuxtOptions) */
```
pi0, line 8:
```suggestion
/** User configuration in `nuxt.config` file */
```

packages/schema/src/types/config.ts
```diff
@@ -1,9 +1,11 @@
 import { ConfigSchema } from '../../schema/config'
 
+/** Normalized Nuxt options */
```
pi0, line 3:
```suggestion
/** Normalized Nuxt options available as `nuxt.options.*` */
```

PR #2132, opened 2021-11-24
feat(nuxi): add `typecheck` command using `vue-tsc`

### LINKED ISSUE #1825: `nuxi typecheck` shortcut
for `npx vue-tsc --noEmit`

_Originally posted by @pi0 in https://github.com/nuxt/framework/pull/1822#discussion_r746579636_

danielroe:
### 🔗 Linked issue

resolves #1825

### ❓ Type of change

### 📚 Description

This adds support for using `vue-tsc` in any nuxt project. It will first generate the types and then run `vue-tsc` on the project.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/typecheck.ts
+++ b/packages/nuxi/src/commands/typecheck.ts
@@ -1,0 +1,32 @@
+import { execa } from 'execa'
+import { resolve } from 'pathe'
+import { tryResolveModule } from '../utils/cjs'
+
+import { loadKit } from '../utils/kit'
+import { writeTypes } from '../utils/prepare'
+import { defineNuxtCommand } from './index'
+
+export default defineNuxtCommand({
+  meta: {
+    name: 'typecheck',
+    usage: 'npx nuxi typecheck [rootDir]',
+    description: 'Runs `vue-tsc` to check types throughout your app.'
+  },
+  async invoke (args) {
+    process.env.NODE_ENV = process.env.NODE_ENV || 'production'
+    const rootDir = resolve(args._[0] || '.')
+
+    const { loadNuxt } = await loadKit(rootDir)
+    const nuxt = await loadNuxt({ rootDir, config: { _prepare: true } })
+
+    // Generate types and close nuxt instance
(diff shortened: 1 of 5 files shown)


packages/nuxi/src/commands/typecheck.ts
```diff
@@ -0,0 +1,32 @@
...
+  async invoke (args) {
+    process.env.NODE_ENV = process.env.NODE_ENV || 'production'
+    const rootDir = resolve(args._[0] || '.')
+
+    const { loadNuxt } = await loadKit(rootDir)
+    const nuxt = await loadNuxt({ rootDir, config: { _prepare: true } })
+
+    // Generate types and close nuxt instance
+    await writeTypes(nuxt)
+    await nuxt.close()
```
pi0, line 24: I wish we could reuse `prepare` command logic and just invoke it here but let's refactor later :)

packages/nuxi/src/commands/typecheck.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    // Prefer local install if possible
+    const hasLocalInstall = tryResolveModule('typescript') && tryResolveModule('vue-tsc/package.json')
+    const npxArgs = hasLocalInstall ? '' : '-p vue-tsc -p typescript '
+
+    await execa('npx', `${npxArgs}vue-tsc --noEmit`.split(' '), { stdio: 'inherit' })
```
pi0, line 30: Isn't execa supposed to pick local installs? https://www.npmjs.com/package/execa#preferlocal

danielroe: Yes. But `npx` will too.

pi0: And we also check it too in L27-28 :D One more thing to consider, we should directly pass cwd/rootDir to args of execa (and `resolve` utils. `nuxi` and `kit` are not necessarily from user's rootDir and it can make incisntent resolution)

pi0 (approved):

PR #2147, opened 2021-11-25
feat(useCookie): support default value

antfu:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/composables/cookie.ts
+++ b/packages/nuxt3/src/app/composables/cookie.ts
@@ -11,7 +11,7 @@ type _CookieOptions = Omit<CookieSerializeOptions & CookieParseOptions, 'decode'
 export interface CookieOptions<T=any> extends _CookieOptions {
   decode?(value: string): T
   encode?(value: T): string;
-  default?: T
+  default?: () => T
 }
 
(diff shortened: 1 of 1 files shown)


danielroe (approved):
Looks nice 👍

packages/nuxt3/src/app/composables/cookie.ts
```diff
@@ -7,9 +7,11 @@ import destr from 'destr'
 import { useNuxtApp } from '#app'
 
 type _CookieOptions = Omit<CookieSerializeOptions & CookieParseOptions, 'decode' | 'encode'>
+
 export interface CookieOptions<T=any> extends _CookieOptions {
   decode?(value: string): T
   encode?(value: T): string;
+  default?: T
```
pi0, line 14: Can we please use the factory function? It would help to avoid state sharing issues and consistency with `useState`

pi0 (approved):
LGTM. However, there is a tricky behavior now. When we clear cookie `cookie.value = null`, we don't fall back to default value anymore...

antfu:
> When we clear cookie cookie.value = null, we don't fall back to default value anymore...

I could fix it if you prefer that behavior.

Issue #1273, opened 2021-10-20
composable auto import improvements


pi0:
Follow up #1176

- [ ] Cache parsing results when timestamps are not changed
  - pending for general caching to be applied for components discovery)
- [x] Allow overriding composables by user instead of warning + disable
  - pending for standardized `extend` (nuxt.config) and `level` (components dir) 
- [x] #2167
  - allows using composables in non-sfc files or situations auto-import is not possible reusing existing infrastructure
 - [x] #2170

danielroe:
@Norel Auto imports don't currently work within the server scope. They just work within the Vue portion of your app.

PR #2148, opened 2021-11-25
fix(types): expose meta composable types

antfu:
### 🔗 Linked issue

Working on https://github.com/vueuse/vueuse/pull/967

```ts
import { useCookie, useMeta } from 'nuxt3'
//       ^ OK       ^ No types error
```

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/types.d.ts
+++ b/packages/nuxt3/types.d.ts
@@ -1,3 +1,4 @@
 import type { } from '@nuxt/nitro'
 export * from './dist/index'
 export * from './dist/app/types/index'
+export * from './dist/meta/runtime/index'


danielroe (commented):
I actually think we should go the other way and remove the composable type exports, which is causing erroneous suggestions in IDEs for importing the helpers from `nuxt3` rather than `#app`.

pi0:
Agree. We shall expose them as `#imports` (#1273) and only type it on one import to avoid confusing IDE using different aliases like `#app` and `#meta` (as well as `nuxt3`) but adding import guards that if one by mistake imported from wrong package, is safe.

pi0:
PS: We should move app and meta types to `@nuxt/schema` :D

PR #2153, opened 2021-11-25
perf(nuxt3): don't include null errors in payload

danielroe:
### ❓ Type of change

### 📚 Description

By setting explicitly to null when there's no error we also add asyncData _key_ to payload which can't be deduplicated as the null value can. This PR avoids that cost by only setting errors if they exist.

### DIFF
--- a/packages/nuxt3/src/app/composables/asyncData.ts
+++ b/packages/nuxt3/src/app/composables/asyncData.ts
@@ -104,7 +104,9 @@ export function useAsyncData<
       .finally(() => {
         asyncData.pending.value = false
         nuxt.payload.data[key] = asyncData.data.value
-        nuxt.payload._errors[key] = !!asyncData.error.value
+        if (asyncData.error.value) {
+          nuxt.payload._errors[key] = !!asyncData.error.value
+        }
         delete nuxt._asyncDataPromises[key]
       })
     return nuxt._asyncDataPromises[key]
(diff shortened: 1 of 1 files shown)


packages/nuxt3/src/app/composables/asyncData.ts
```diff
@@ -104,7 +104,9 @@ export function useAsyncData<
       .finally(() => {
         asyncData.pending.value = false
         nuxt.payload.data[key] = asyncData.data.value
-        nuxt.payload._errors[key] = !!asyncData.error.value
+        if (asyncData.error.value) {
+          nuxt.payload._errors[key] = !!asyncData.error.value
```
danielroe, line 108:
```suggestion
          nuxt.payload._errors[key] = true
```

packages/nuxt3/src/app/composables/asyncData.ts
```diff
@@ -104,7 +104,9 @@ export function useAsyncData<
       .finally(() => {
         asyncData.pending.value = false
         nuxt.payload.data[key] = asyncData.data.value
-        nuxt.payload._errors[key] = !!asyncData.error.value
+        if (asyncData.error.value) {
+          nuxt.payload._errors[key] = true
```
danielroe, line 108: **Note**: If we wanted to stringify errors (or subset of error, such as status code & message?) then by adopting this approach we only add length to HTML if the error exists...

pi0 (approved):

Issue #2170, opened 2021-11-26
Support auto imports in template


antfu:

antfu:
Did some research, well I think this is trickier than it looks.

Assuming we are talking about `<script setup>` (since auto importing in options API does not make that much sense.

![image](https://user-images.githubusercontent.com/11247099/143600262-b578cfa4-2d13-4da7-baf8-9aa803178636.png)

As we can see in the playground, when we use something in the template, the SFC compiler actually looked up the existing identifiers in `<script setup>` and use them directly. When it's not presented, it will fallback to use the component context (`this` object) - so that's the reason auto-importing won't work in the template. So I guess even tho we could replace `_ctx.xx` with auto importing, I think it could be a dangerous move since the user might actually want the `xx` in the runtime context. Introducing another plugin that runs before Vue's might be possible to inject the entry in `<script setup>` before SFC compilation - but I am not sure if the complexity is worth it (since we have #2168).

pi0:
It makes sense that now we have `#imports` we can simply document workaround. This way type hinting should work and we won't have context similarity conflicts.

PR #2171, opened 2021-11-26
fix(auto-import): improve auto import updates

antfu:
After #2170

### ❓ Type of change

### 📚 Description

Currently, the global mode does not include composables and does not react on auto import entries changes

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/auto-imports/module.ts
+++ b/packages/nuxt3/src/auto-imports/module.ts
@@ -1,4 +1,4 @@
-import { addVitePlugin, addWebpackPlugin, defineNuxtModule, addTemplate, resolveAlias, addPluginTemplate, useNuxt, addPlugin } from '@nuxt/kit'
+import { addVitePlugin, addWebpackPlugin, defineNuxtModule, addTemplate, resolveAlias, useNuxt, addPlugin } from '@nuxt/kit'
 import type { AutoImportsOptions } from '@nuxt/schema'
 import { isAbsolute, join, relative, resolve, normalize } from 'pathe'
 import { TransformPlugin } from './transform'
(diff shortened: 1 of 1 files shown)


packages/nuxt3/src/auto-imports/module.ts
```diff
@@ -79,9 +65,29 @@ export default defineNuxtModule<AutoImportsOptions>({
       await nuxt.callHook('autoImports:extend', ctx.autoImports)
       // Update context
       updateAutoImportContext(ctx)
+
+      // Support for importing from '#imports'
```
pi0, line 69: This can/should be called once outside of update. `getContents` will be always called on rebuild

packages/nuxt3/src/auto-imports/module.ts
```diff
@@ -79,9 +65,29 @@ export default defineNuxtModule<AutoImportsOptions>({
...
+      addTemplate({
```
pi0, line 70: This can/should be called once outside of update. `getContents` will be always called on rebuild

antfu: Wow, did it? So I guess the entire PR is meaningless then?

pi0: [here it does it](https://github.com/nuxt/framework/blob/main/packages/kit/src/template.ts#L16)

I think to resolve issue, we should register `composables` dir to trigger a nuxt rebuild or manually call `builder:generateApp` hook

antfu: Or do you think we could introduce a `updateTemplate('imports.mjs')` util?

pi0: Selective template re-generation is not yet supported in Nuxt3 since underlying tools are not fully supporting vfs cache. A regenerate would be fast and only rebuild vfs templates.

pi0: But we could support `invalidateTemplates`  maybe as a shortcut for hook calling? It can later on support filters :)

pi0: Moved to https://github.com/nuxt/framework/issues/2215

pi0:
I guess https://github.com/nuxt/framework/pull/2725 should fixed this. /cc @danielroe

danielroe:
I'm still encountering this issue but would need to dive in to find out what the cause is.

Issue #1042, opened 2021-10-14
proxy client headers to useFetch in sider-side


antoineanorra:
### Environment

Nuxt CLI v3.0.0-27236898.4a34c2a                                            18:16:03
RootDir: /home/centos/nuxt3-app/nuxt                                        18:16:04
Nuxt project info:                                                          18:16:04

------------------------------
- Operating System: `Linux`
- Node Version:     `v14.17.6`
- Nuxt Version:     `3-3.0.0-27236898.4a34c2a`
- Package Manager:  `npm`
- Bundler:          `Webpack`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Describe the bug

When I get to the page, useFetch makes an internal call to the /server/api/health.js file
However, no client header is present, this is a big problem

Chrome request headers :
![NqyjL0q4RI](https://user-images.githubusercontent.com/48593504/137357048-c8a8c226-ce37-4c10-b476-bdf7c63bb630.png)

app.vue :
```vue
<script setup>
const { data } = await useFetch("/api/health");
console.log(data);
</script>

<template>
  <div>Page visits: {{ data }}</div>
</template>
```

/server/api/health.js :
```
export default async (req, res) => {
    console.log(req.headers);
    return {
        health: true
    }
}
```

Logs : 
`{ host: undefined }`

### Reproduction

https://codesandbox.io/s/sad-hamilton-ip4yc?file=/app.vue

danielroe:
When you run `useFetch` or `$fetch` on the server side, it doesn't actually make a network request. Rather, it directly calls the api endpoint locally, mocking the req/res. So there won't be any headers other than those provided by you. There's a bit more info [here](https://v3.nuxtjs.org/concepts/server-engine#direct-api-calls). PRs to improve the docs are welcome 🙃

ilouHD:
Are there any plans to implement this in the next few days or weeks? This is too basic to postpone it to a later stage of development

ilouHD:
As described in https://v3.nuxtjs.org/docs/usage/6-cookies#handling-cookies-in-api-routes one can use `useCookie` to read in cookies and `setCookie` to set cookies. Both functions work when navigating around the app, but still not when SSR renders the page and does local fetch.

Are there any plans to realize this in the near future? In the list of all missing features for nuxt 3 this is the most important one for me.

mindtechmedia:
So usecookie is not reload safe?

[...]

PR #2173, opened 2021-11-26
feat(nuxt3, bridge): `useReqHeaders` composable

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/1042

### ❓ Type of change

### 📚 Description

This PR adds a new composable for bridge/nuxt3: `useReqHeaders`. It returns the current request's headers, which allows proxying headers via a local/isomorphic fetch call:

```js
const { data } = useFetch('/api/me', {
  headers: useReqHeaders()
})
```

On client-side, it returns an empty object.

It also supports an options object with exclude/include options to narrow the returned headers.

### 📝 Checklist


packages/nuxt3/src/app/composables/ssr.ts
```diff
@@ -0,0 +1,28 @@
...
+  const nuxtApp = useNuxtApp()
+  const headers: Record<string, string> = nuxtApp.ssrContext?.req.headers ?? {}
+  if (!opts) {
+    return headers
+  }
+
+  opts.exclude = opts.exclude?.map(key => key.toLowerCase())
+  opts.include = opts.include?.map(key => key.toLowerCase())
+
+  return Object.fromEntries(Object.entries(headers).filter(([key]) => {
```
pi0, line 18: Can we please refactor obj key filter to a helper function? (even if internetnal and same file for now should be fine)

pi0:
Would be nice if we mention `useReqHeaders` from "data fetching section" with an example of cookie proxy to resolve linked issue/q

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -153,6 +153,20 @@ const { data: mountain } = await useFetch('/api/mountains/everest', { pick: ['ti
 </template>
 ```
 
+## Authenticated requests
```
pi0, line 156:
```suggestion
## User headers in server-side
```

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -153,6 +153,20 @@ const { data: mountain } = await useFetch('/api/mountains/everest', { pick: ['ti
...
+
+If you need to access request headers or cookies within your Nitro endpoint, you can use [`useReqHeaders`](/docs/usage/ssr) to proxy them. (Without this, you won't be able to access them on the initial server-side render.) For example:
```
pi0, line 158: "within your Nitro endpoint" i guess it applies to any kind of request (including cross-origin API requests

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -153,6 +153,20 @@ const { data: mountain } = await useFetch('/api/mountains/everest', { pick: ['ti
...
+
+```vue
+<script setup>
+const { data } = useFetch('/api/me', {
+  headers: useReqHeaders()
```
pi0, line 163: This is an unsafe example unless we use sanity default for exclude list (https://github.com/nuxt-community/axios-module/blob/main/lib/module.js#L82)

danielroe: I think it's safe for a request to your own Nitro endpoint, but we'd need to make the distinction clear.

pi0: Not necessarily for nitro as well if  it makes nested subsequent requests.

danielroe: I may be misunderstanding, but I don't think they would proxied for any nested subsequent requests (if you're referring to requests made by the Nitro endpoint).

pi0: It might be hard for me to explain further on this topic. but generally please let's not set special exceptions on `$fetch` calls with direct SSR calls which are not happening. It will get confusing.

pi0:
Update: I've finally decided to simplify API. `useRequestHeaders` can be used without any arguments to get all original request headers or with an array to get allowed ones. This improves both perf/implementation and reduces chance of misusing `exclude`.

PR #2183, opened 2021-11-27
feat(auto-import): allow explicit transform exclusion patterns

### LINKED ISSUE #2180: allow auto-imports inside node_modules
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.11.1`
- Nuxt Version:     `2.16.0-27299536.ab1c6cb4`
- Bundler:          `Webpack`
- User Config:      `bridge`, `components`, `autoImports`
- Runtime Modules:  `-`
- Build Modules:    `@vueuse/core/nuxt@7.1.2`, `@nuxt/bridge@3.0.0-27296423.f3082ca`

### Reproduction

-

### Describe the bug

The auto-imports feature excludes by default all files inside 'node_modules' from being transformed. As I am currently developing a library, I would have liked to have the possibility to write auto-imported composables that can be overwritten by the user. If I want them to be overwritable, I cannot set an explicit import statement but it is currently impossible to explicitly transform a given file/folder inside node_modules. I hope this makes sense, here's an example use case:

### Example:
```ts
// node_modules/library/composables/bar.ts
export function useBar() {
    return 'bonjour'
}
```
```vue
// node_modules/library/components/Foo.vue
<script setup lang="ts">
const foo = useBar()
</script>

<template>
  <div>{{ foo }}</div>
</template>
```
```vue
// pages/index.vue
<script setup lang="ts">
import Foo from 'library/components/Foo.vue'
</script>


<template>
  <Foo /> <!-- should be <div>bonjour</div> -->
</template>
```
After overwriting my composable:
```ts
// composables/bar.ts
export function useBar() {
    return 'salut'
}
```

```vue
// pages/index.vue
<script setup lang="ts">
import Foo from 'library/components/Foo.vue'
</script>


<template>
  <Foo /> <!-- should be <div>salut</div> -->
</template>
```

### Additional context

_No response_

### Logs

_No response_

yassilah:
### 🔗 Linked issue

#2180

### ❓ Type of change

### 📚 Description

Resolves #2180

### 📝 Checklist


pi0 (approved):

pi0:
Thanks for PR @yassilah and sorry for the late review. 

When creating npm packages, I would recommend using new `#imports` virtual instead of implicitly depending on auto imports. This also helps for end-user performance to avoid transforming files inside `node_modules` as much as possible

BTW having an option to modify transform behavior makes sense. I have made a small refactor to move it to `transform: { exclude }`  and also allow completely overriding it.

PR #2193, opened 2021-11-28
feat(nuxt3): add `autoRegisterPlugins` option to config

### LINKED ISSUE #2163: Allow disabling automatic plugin imports
### Environment

- Nuxt Version:     `3.0.0-27296423.f3082ca`
- Package Manager:  `npm@7.24.2`
- Bundler:          `Webpack`
- User Config:      `buildModules`, `plugins`, `nitro`
- Runtime Modules:  `-`
- Build Modules:    `./modules/build/ModuleFederationModule.ts`

### Reproduction

In any project, try using `~/plugins` for plugins which don't follow the Nuxt structure.
**DO NOT** include such plugins in `nuxtConfig.plugins`

### Describe the bug

Currently Nuxt tries to auto-import plugins from `~/plugins`. However, this way, all the code placed inside this folder is considered a Nuxt plugin and therefore can't be used as anything else or disabled.

Toying around with `nuxtConfig.autoImports` doesn't seem to disable automatic imports, which in my setup is a destructive rather than useful feature.

### Additional context

_No response_

### Logs

```shell
ERROR in ./.nuxt/plugins/client.mjs 17:2-15
export 'default' (imported as 'auth_8a8742aa') was not found in '<omitted>/plugins/remote/auth.ts' (possible exports: <omitted>)
```

ms-fadaei:
### 🔗 Linked issue

Resolve #2163 

### ❓ Type of change

### 📚 Description

I added an option to the `nuxt.config` with `autoRegisterPlugins` name and default value of `true`. So if anyone (like me) doesn't want plugins auto registering, can disable it.

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/2.directory-structure/10.plugins.md
+++ b/docs/content/3.docs/2.directory-structure/10.plugins.md
@@ -9,7 +9,7 @@ head.title: Plugins directory
 Nuxt will automatically read the files in your `plugins` directory and load them. You can use `.server` or `.client` suffix in the file name to load a plugin only on the server or client side.
 
 ::alert{type=warning}
-All plugins in your `plugins/` directory are auto-registered, so you should not add them to your `nuxt.config` separately.
+All plugins in your `plugins/` directory are auto-registered, so you should not add them to your `nuxt.config` separately unless set `autoRegisterPlugins: false` in your `nuxt.config`.
 ::
 
 ## Creating plugins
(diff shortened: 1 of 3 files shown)


pi0:
Thanks for PR @ms-fadaei 💚 But I think we should avoid adding back feature flags like this in nuxt 3 which bring inconsistent behavior especially when something is expected by a theme or module. We should support ignore patterns to resolve linked issue.

ms-fadaei:
> Thanks for PR @ms-fadaei 💚 But I think we should avoid adding back feature flags like this in nuxt 3 which bring inconsistent behavior especially when something is expected by a theme or module. We should support ignore patterns to resolve linked issue.

Thanks, @pi0. Can you tell me more about this so I can do it for you? Do you mean we should have an option in the config like `ignoreAutoRegisterPattern`?

danielroe:
@ms-fadaei We need support for `.nuxtignore`: see https://github.com/nuxt/framework/issues/2163#issuecomment-981612377

PR #2249, opened 2021-12-01
fix(nitro): respect app.cdnURL when set

### LINKED ISSUE #1467: how to build with .mjs/.js/.css hosting on cdn?
### Environment

Nuxt CLI v3.0.0-27252573.ad7e755                                                                                                                                           19:54:26
RootDir: /Users/lookis/projects/nuxt3-app                                                                                                                                  19:54:27
Nuxt project info:                                                                                                                                                         19:54:27

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0-27252573.ad7e755`
- Package Manager:  `Yarn`
- Bundler:          `Vite`
- User Config:      `modules`, `build`, `proxy`
- Runtime Modules:  `@nuxtjs/proxy@2.1.0`
- Build Modules:    `-`

### Describe the bug

build.publicPath only affect image file, all css/.mjs file still use relativePath "/_nuxt/"

### Reproduction

https://codesandbox.io/s/throbbing-sky-7mmfe

yarn build
yarn start

### Additional context

codesandbox may run only under dev mode, i was tried production mode with no luck, 

image file got the full path when .mjs/.css still relative path:
```HTML
<!DOCTYPE html>
<html data-head-attrs="">

<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <meta name="head:count" content="2">
  <link rel="modulepreload" href="/_nuxt/entry-0e85cf82.mjs" as="script">
  <link rel="modulepreload" href="/_nuxt/index-4a5df546.mjs" as="script">
  <link rel="preload" href="/_nuxt/assets/index.045b46d2.css" as="style">
  <link rel="prefetch" href="/_nuxt/assets/googlelogo.5776cd87.png">
  <link rel="stylesheet" href="/_nuxt/assets/index.045b46d2.css">
</head>

<body data-head-attrs="">
  <div id="__nuxt">
    <!--[-->
    <div><img src="https://static.example.com/assets/googlelogo.5776cd87.png"></div>
    <!--]-->
  </div>
  <script>window.__NUXT__ = (function (a) { return { data: {}, state: {}, serverRendered: true, config: { app: { basePath: a, assetsPath: a, cdnURL: "https:\u002F\u002Fstatic.example.com\u002F" } } } }("\u002F"))</script>
  <script type="module" src="/_nuxt/entry-0e85cf82.mjs" defer></script>
  <script type="module" src="/_nuxt/index-4a5df546.mjs" defer></script>
</body>

</html>
```

### Logs

_No response_

### LINKED ISSUE #2242: 404 error everywhere after setting build.publicPath to CDN URL
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-27304790.cad09fe`
- Package Manager:  `npm@6.14.15`
- Bundler:          `Vite`
- User Config:      `build`, `css`, `publicRuntimeConfig`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://codesandbox.io/s/new-wood-hx840?file=/nuxt.config.ts

_Notes:_
- The sandbox is configured in a way that the code is ran in production mode (`npm run build && npm run start`)
- The CDN URL used is a BunnyCDN Pull Zone. The pull zone is set to fetch files from the sandbox URL:
ss215-skrypt

### Describe the bug

After `build.publicPath` in `nuxt.config.ts` is changed to any URL, the whole application (pages and assets) will return a 404 error. Changing it to a single folder name, no errors occur.

### Additional context

As of my understanding, changing `build.publicPath` to a CDN URL would make all assets go into the `.output/public` folder instead of the `.output/public/_nuxt` folder. The assets are actually there, but navigating to them also returns a 404 error.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #1467
resolves #2242

### ❓ Type of change

### 📚 Description

`cdnURL` wasn't being taken into account by bundle-renderer. In addition, when serving static files from `/` we were never allowing a fallback to rendering the vue app. I don't think it's safe simply to disable `serveStatic` when a CDN is in use, but I would certainly suggest to users that they disable serving static files when using a CDN that does not rely on 'pulling' files from their site:

```js

  nitro: {
    serveStatic: false
  }
})
```

**Note**: We didn't reserve `.app` namespace in Nuxt 2 (see https://github.com/nuxt/nuxt.js/pull/9075) in case it's in use so we need to check `._app` existence first.

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/vite/manifest.ts
+++ b/packages/bridge/src/vite/manifest.ts
@@ -46,7 +46,7 @@ export async function prepareManifests (ctx: ViteBuildContext) {
 export async function generateBuildManifest (ctx: ViteBuildContext) {
   const rDist = (...args: string[]): string => resolve(ctx.nuxt.options.buildDir, 'dist', ...args)
 
-  const publicPath = ctx.nuxt.options.app.assetsPath // Default: /nuxt/
+  const publicPath = ctx.nuxt.options._app.cdnURL || ctx.nuxt.options._app.assetsPath // Default: /nuxt/
   const viteClientManifest = await fse.readJSON(rDist('client/manifest.json'))
   const clientEntries = Object.entries(viteClientManifest)
 
--- a/packages/bridge/src/vite/manifest.ts
+++ b/packages/bridge/src/vite/manifest.ts
@@ -46,7 +46,7 @@ export async function prepareManifests (ctx: ViteBuildContext) {
 export async function generateBuildManifest (ctx: ViteBuildContext) {
   const rDist = (...args: string[]): string => resolve(ctx.nuxt.options.buildDir, 'dist', ...args)
 
-  const publicPath = ctx.nuxt.options._app.cdnURL || ctx.nuxt.options._app.assetsPath // Default: /nuxt/
+  const publicPath = ctx.nuxt.options.app.cdnURL || ctx.nuxt.options.app.assetsPath // Default: /nuxt/
   const viteClientManifest = await fse.readJSON(rDist('client/manifest.json'))
   const clientEntries = Object.entries(viteClientManifest)
 
--- a/packages/nitro/src/runtime/server/static.ts
+++ b/packages/nitro/src/runtime/server/static.ts
@@ -27,7 +27,7 @@ export default async function serveStatic (req, res) {
   }
 
   if (!asset) {
-    if (id.startsWith(PUBLIC_PATH) && !id.startsWith(STATIC_ASSETS_BASE)) {
+    if (PUBLIC_PATH !== '/' && id.startsWith(PUBLIC_PATH) && !id.startsWith(STATIC_ASSETS_BASE)) {
       throw createError({
         statusMessage: 'Cannot find static asset ' + id,
         statusCode: 404
--- a/packages/nitro/src/runtime/server/static.ts
+++ b/packages/nitro/src/runtime/server/static.ts
@@ -27,7 +27,7 @@ export default async function serveStatic (req, res) {
   }
 
   if (!asset) {
-    if (PUBLIC_PATH !== '/' && id.startsWith(PUBLIC_PATH) && !id.startsWith(STATIC_ASSETS_BASE)) {
+    if (id.startsWith(PUBLIC_PATH) && !id.startsWith(STATIC_ASSETS_BASE)) {
       throw createError({
         statusMessage: 'Cannot find static asset ' + id,
         statusCode: 404
--- a/packages/bridge/src/nitro.ts
+++ b/packages/bridge/src/nitro.ts
@@ -1,7 +1,7 @@
 import { promises as fsp } from 'fs'
 import fetch from 'node-fetch'
 import { addPluginTemplate, useNuxt } from '@nuxt/kit'
-import { stringifyQuery } from 'ufo'
+import { hasProtocol, isRelative, joinURL, stringifyQuery } from 'ufo'
 import { resolve } from 'pathe'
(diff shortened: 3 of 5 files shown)


packages/bridge/src/vite/manifest.ts
```diff
@@ -46,7 +46,7 @@ export async function prepareManifests (ctx: ViteBuildContext) {
 export async function generateBuildManifest (ctx: ViteBuildContext) {
   const rDist = (...args: string[]): string => resolve(ctx.nuxt.options.buildDir, 'dist', ...args)
 
-  const publicPath = ctx.nuxt.options.app.assetsPath // Default: /nuxt/
+  const publicPath = ctx.nuxt.options.app.cdnURL || ctx.nuxt.options.app.assetsPath // Default: /nuxt/
```
pi0, line 49: Using cdnURL here doesn't make sense. Should nitro/ssr-renderer pick the the correct path from runtimeConfig? (fix in nitro `renderer.ts` file)

packages/nitro/src/runtime/server/static.ts
```diff
@@ -27,7 +27,7 @@ export default async function serveStatic (req, res) {
   }
 
   if (!asset) {
-    if (id.startsWith(PUBLIC_PATH) && !id.startsWith(STATIC_ASSETS_BASE)) {
+    if (PUBLIC_PATH !== '/' && id.startsWith(PUBLIC_PATH) && !id.startsWith(STATIC_ASSETS_BASE)) {
```
pi0, line 30: How it would be possible `PUBLIC_PATH` to be `/`? When using a CDN, this handler obviously shouldn't be in use or enabled anymore.

danielroe: The use-case is when a CDN is set to 'pull' from the deployed Nuxt instance (so we do need to serve files for the CDN) from it.

In the case of using a CDN + serveStatic, `assetsPath` (which is what publicPath is set to) defaults to `/`:

https://github.com/nuxt/framework/blob/c347b6cc8afe49068ddc0aaf48aa3cf8c83c4e68/packages/schema/src/config/_app.ts#L33-L43

How about we default it to `false` when using a CDN, but assume some users may override `serveStatic` for 'pull' CDN usage?

packages/bridge/src/vite/manifest.ts
danielroe: Nitro will override from `runtimeConfig`, so this line simply sets the fallback `publicPath`. I think it makes sense to include, as long as we're including the publicPath in the generated build manifest at all.

packages/nitro/src/runtime/server/static.ts
pi0: When assets are served on the same domain and handler, one is practically using reverse-proxy's cache to serve `/_nuxt` (publicPath) -- kinda pull. And when `cdnURL` is specified, it means they are to be directly served by CDN not nuxt handler.

Can you please give an example setup that both needs cdnURL (and external URL) and using server-static at the same time?

packages/bridge/src/vite/manifest.ts
pi0: Considering `runtimeConfig` is also bundled (and we preserve the build-time value of `cdnURL`) this value is not even picked I assume (or if picked it will be inconsistent from runtime vs build-time)

webfansplz:
Hi, may i ask the progress of this PR ?  It's so helpful for me , Thanks 😊

newbeea:
anything update?

pi0:
Closing in the favor of #2655

PR #2275, opened 2021-12-03
feat(kit, schema)!: finalize nuxt 3 module spec and utils

pi0:
This PR is attempting to finalize Nuxt 3 modules spec and `@nuxt/kit` utils to be ready and documented.

## Module format

A plain Nuxt module, same as Nuxt 2, is a simple function accepting inline options and (new) `nuxt` arguments replacing `this.nuxt` usage of Nuxt 2.  It is totally up to the module author how to handle the rest of logic including options handling when using the plain format. This simplicity gives full freedom to modules and a solid long-term API.

Starting with Nuxt3, this format can benefit all `@nuxt/kit` utilities since a composition-api style implicit nuxt context is available via `useNuxt`

```js
// module.mjs
export default async (inlineOptions, nuxt) => { }

// In nuxt.config
  modules: [
    async (inlineOptions, nuxt) => { }
  ]
})
```

This function can additionally, have two optional properties attached into: (new in Nuxt 3 spec)
- `getOptions`: Returning a promise that resolves to merged options of the module
- `getMeta`: Returning a promise that resolves to module meta (replacing `meta` and `pkg` in Nuxt 2 spec)

## `defineNuxtModule`

A new `defineNuxtModule` utility from `@nuxt/kit`, provides a convenient and standard API to define nuxt modules:

- Supporting `defaults`, `schema` and `meta.configKey` for automatically merging module options
- Type hinting and inference
- Add shims for (basic) Nuxt 2 compatibility even without using `@nuxt/bridge`
- Ensure module get's installed once using a unique key of `meta.name` or `meta.configKey`
- Automatically register `hooks` if provided
- Automatically check for compatibility issues based on provided meta
- Expose `getOptions` and `getMeta`
- Ensuring backward and upward compatibility as long as module is using `defineNuxtmodule` from latest version of `@nuxt/kit`
- Integration with build tooling (provided soon) to build standard dist

```js
  meta: { name: 'myModule', compatibility: { nuxt: '^3.0.0'  } },
  defaults: {},
  schema: {},
  hooks: {},
  async setup(mergedOptions, nuxt) {}
})
```

## Compatibility Framework

Both core/bridge and modules, can request for compatibility checks. There are 3 async utilities for this:
- `checkNuxtCompatibility`
- `assertNuxtCompatibility`
- `hasNuxtCompatibility`

In addition to basic Nuxt version checking, modules have the possibility to hook into `kit:compatibility` and do extended compatibility checks. Nuxt can also internally leverage this hook to decentralize checks outside of kit.

## Changes

**Note:** An initial POC of `defineNuxtModule` was added a few months back but while we intentionally didn't document usage of `@nuxt/kit` due to potential breaking changes, some modules opted-in to unofficial API and depending on `@nuxt/kit`. This PR tries to keep breaking changes as few as possible and even adding workarounds to be more compatible with these modules until all use the latest spec. With all of this, there are still chances of breaking changes if a project or module was using the undocumented spec.

- Split `defaults` from `schema` (which is handled by untyped)
   - `defaults` is only applied with defu (not untyped with `$resolve` support)
   - `defaults` can be also a function for dynamic defaults based on nuxt options
- Input of `defineNuxtModule` is `ModuleDefinition` and output a normalized `NuxtModule`
- Expose `.getMeta` and `.getOptions` from normalized modules to access module info in other contexts [testing]
- Move meta under explicit meta key in the definition
   - `name`, `configKey`, `version` and `requires` are under `meta` key in `defineNuxtModule`
   - Compatibility added for top level `configKey` for unofficial modules
- `installModule` signature updated from `(nuxt, installOpts)` to `(input, inlineOptions?, nuxt?)`
   - Input only accepts a Nuxt module or string path to a module with module as default export
   - A compatibility layer is kept in module container utils for legacy `{ src, handler }` format
   - Compatibility added for old signature usage with a warning
- input of `defineNuxtModule` should be directly definition object and cannot be a factory function `nuxt => nuxtModule` 

Updated compatibility check utils:
- Renamed `ensureNuxtCompatibility` to `assertNuxtCompatibility` [async]
  - `addComponent` and `addComponent` utilities are async too
- Renamed `checkNuxtCompatibilityIssues` to `checkNuxtCompatibility` [async]
- Renamed `NuxtCompatibilityConstraints` to `NuxtCompatibility` with additional `requires` key [async]

### 📝 TODO

- [x] Implement `compatibility`
- [x] Extendable compatibility checkers 
- [x] Bridge constraint
- [x] Allow `useNuxt` with bridge on all modules


packages/kit/src/module/define.ts
```diff
@@ -12,96 +12,120 @@ import { templateUtils, compileTemplate } from '../internal/template'
...
-  let mod: NuxtModule<OptionsT>
-
-  function wrappedModule (inlineOptions: OptionsT) {
-    // Get nuxt context
-    const nuxt: Nuxt = this.nuxt || useNuxt()
+export function defineNuxtModule<OptionsT extends ModuleOptions> (definition: ModuleDefinition<OptionsT>): NuxtModule<OptionsT> {
+  // Normalize definition and meta
+  if (!definition.meta) { definition.meta = {} }
+  if (!definition.meta.configKey) {
+    // @ts-ignore TODO Remove non-meta allbacks in RC
```
danielroe, line 19:
```suggestion
    // @ts-ignore TODO Remove non-meta fallbacks in RC
```

packages/kit/src/module/define.ts
```diff
@@ -12,96 +12,120 @@ import { templateUtils, compileTemplate } from '../internal/template'
...
-    if (typeof input === 'function') {
-      mod = input(nuxt)
-    } else {
-      mod = input
+  // Resolves module options from inline options, [configKey] in nuxt.config, defaults and schema
+  function getOptions (inlineOptions?: OptionsT, nuxt: Nuxt = useNuxt()) {
+    let _options: OptionsT
+    const configKey = definition.meta.configKey || definition.meta.name
+    const _defaults = typeof definition.defaults === 'function' ? definition.defaults(nuxt) : definition.defaults
+    _options = defu(inlineOptions, nuxt.options[configKey], _defaults) as OptionsT
```
danielroe, line 30:
```suggestion
    const configKey = definition.meta.configKey || definition.meta.name
    const _defaults = typeof definition.defaults === 'function' ? definition.defaults(nuxt) : definition.defaults
    let _options = defu(inlineOptions, nuxt.options[configKey], _defaults) as OptionsT
```

packages/kit/src/module/install.ts
```diff
@@ -1,63 +1,34 @@
...
-    src = installOpts
+export async function installModule (nuxtModule: string | NuxtModule, inlineOptions?: any, nuxt: Nuxt = useNuxt()) {
+  // Detect if `installModule` used with older signuture (nuxt, nuxtModule)
+  // TODO: Remove in RC
+  // @ts-ignore
+  if (nuxtModule?._version || nuxtModule?.version || nuxtModule?.constructor?.version || '') {
+    // @ts-ignore
+    [nuxt, nuxtModule] = [nuxtModule, inlineOptions]
+    inlineOptions = {}
+    console.warn(new Error('`installModule` is being called with old signuture!'))
```
danielroe, line 16:
```suggestion
    console.warn(new Error('`installModule` is being called with old signature!'))
```

packages/schema/build.config.ts
```diff
@@ -27,6 +27,6 @@ export default defineBuildConfig({
     'vite',
     // Implicit
     '@vue/compiler-core',
-    '@vue/shared ',
+    '@vue/shared '
```
danielroe, line 30: This isn't caused by this PR - but surely:
```suggestion
    '@vue/shared'
```

danielroe (commented):
I love it! 🚀

Really nice improvements to the spec.

harlan-zw:
Nice job @pi0 

Can we get an update on the stubs with this change

PR #2300, opened 2021-12-06
docs: modify the comment description of `buildModules`

whylost:
### ❓ Type of change

### 📚 Description

From the file `packages/schema/src/config/_common.ts`, it can be seen that there are many places where `modules` is used to describe the content of `buildmodules`. This is consistent with the comment content of `modules` described above

Many of the `buildmodules` in the document are described as `modules`, and **`modules` are also used in the configuration examples of `buildmodules`** . This is particularly confusing!

### DIFF
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -232,21 +232,21 @@ export default {
   modules: [],
 
   /**
-   * Modules that are only required during development and build time.
+   * BuildModules that are only required during development and build time.
    *
-   * Modules are Nuxt extensions which can extend its core functionality and add endless integrations
+   * BuildModules are Nuxt extensions which can extend its core functionality and add endless integrations
    *
-   * Each module is either a string (which can refer to a package, or be a path to a file), a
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for the PR @whylost. With Nuxt 2, we had to introduce a new concept of build-modules to separate from ones also hook into production runtime but in Nuxt 3 / Nitro, practically all modules are build-modules. ie `modules` and `buildModules` behave exactly same now. Because of this, I believe it would be simpler if we keep referencing to them as `Modules` and remove `buildModules` from documentation at some point.

whylost:
> Thanks for the PR @whylost. With Nuxt 2, we had to introduce a new concept of build-modules to separate from ones also hook into production runtime but in Nuxt 3 / Nitro, practically all modules are build-modules. ie `modules` and `buildModules` behave exactly same now. Because of this, I believe it would be simpler if we keep referencing to them as `Modules` and remove `buildModules` from documentation at some point.
### In the automatic import of components ([library author](https://v3.nuxtjs.org/docs/directory-structure/components#library-authors)), is it better to add an appropriate description?
image

PR #2307, opened 2021-12-07
fix(useCookie): handle undefined options

antfu:
### ❓ Type of change

### 📝 Checklist


packages/nuxt3/src/app/composables/cookie.ts
```diff
@@ -25,7 +25,7 @@ export function useCookie <T=string> (name: string, _opts?: CookieOptions<T>): C
   const opts = { ...CookieDefaults, ..._opts }
   const cookies = readRawCookies(opts)
 
-  const cookie = ref(cookies[name] ?? opts.default?.())
+  const cookie = ref(cookies[name] ?? opts?.default?.())
```
danielroe, line 28: I think shouldn't be possible... We define `opts` just above.

pi0: Indeed we do always initialize `opts` object

PR #2347, opened 2021-12-11
fix(nuxt3): ignore tests within components directory

### LINKED ISSUE #2343: Test files are bundled into production builds
### Environment

- Operating System: `Darwin`
- Node Version:     `v17.0.1`
- Nuxt Version:     `3.0.0-27319101.3e82f0f`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

I've created a reproduction with instructions in this repo:
https://github.com/nixstrom/nuxt3-test-issue

### Describe the bug

In Nuxt 3, if a project has test files inside a Nuxt-directory (e.g. `components/Button.test.ts`, that file will be bundled into the production build. This can cause an error when serving the app.

In the case of my example, the error is `describe is undefined`. This is because Jest sets global properties, which are only available in a test context.

## Things I've tried
According to JS Doc, as well as the documentation, the test files should be ignored by default (so long as they follow the name patterns `**/*.spec*`, `**/*.test.*` or `**/-*.*`). Regardless, I've tried to:

1.  Rename the test file, to `Button.test.ts`, `Button.spec.ts` or `-Button.test.ts`.
2. Explicitly ignore the test files on the `ignore` property in `nuxt.config.ts`.
3. Explicitly ignore the test files by adding a `.nuxtignore`
4. Importing the Jest global methods directly into the test files. This is a poor solution for multiple reasons – also it doesn't work in my case.
5. Overriding the globals manually. Again, a very poor solution ... which doesn't work.

## A temporary solution
What worked for me in the end, was to move the files to a separate `__test__` directory directly in the root. This works, but I would prefer to colocate my unit tests together with the code.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2343

### ❓ Type of change

### 📚 Description

This adds tests to the existing list of exclusions in the components module. Ultimately we are going to want to implement `.nuxtignore` support: https://github.com/nuxt/framework/issues/2163#issuecomment-981612377.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/components/module.ts
+++ b/packages/nuxt3/src/components/module.ts
@@ -47,6 +47,7 @@ export default defineNuxtModule<ComponentsOptions>({
           ignore: [
             '**/*.stories.{js,ts,jsx,tsx}', // ignore storybook files
             '**/*{M,.m,-m}ixin.{js,ts,jsx,tsx}', // ignore mixins
+            '**/*.{spec,test}.{js,ts,jsx,tsx}', // ignore tests
             '**/*.d.ts', // .d.ts files
             // TODO: support nuxt ignore patterns
             ...(dirOptions.ignore || [])


menthol:
if someone have a problem with cypress or other testing library after `nuxi build`, try add following to your `nuxt.config.js`
```js
  components: {
    dirs: [{
      path: '~/components',
      ignore: ['**/*.{spec,test}.{js,ts,jsx,tsx}'],
    }],
  },

```

pi0 (approved):

PR #2382, opened 2021-12-14
feat(nitro): specify packages to copy to `.output/server/node_modules`

danielroe:
### ❓ Type of change

### 📚 Description

This PR allows manually specifying additional packages that will be traced and copied across to `.output/server/node_modules`, for example if `@vercel/nft` isn't able to detect the requirement.

### DIFF
--- a/packages/nitro/src/rollup/plugins/externals.ts
+++ b/packages/nitro/src/rollup/plugins/externals.ts
@@ -71,7 +71,7 @@ export function externals (opts: NodeExternalsOptions): Plugin {
       if (opts.trace !== false) {
         for (const pkgName of opts.traceInclude || []) {
           const path = await this.resolve(pkgName)
-          if (path.id) {
+          if (path?.id) {
             trackedExternals.add(path.id)
           }
         }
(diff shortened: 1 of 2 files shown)


packages/nitro/src/rollup/plugins/externals.ts
```diff
@@ -10,6 +10,8 @@ export interface NodeExternalsOptions {
   trace?: boolean
   traceOptions?: NodeFileTraceOptions
   moduleDirectories?: string[]
+  /** additional packages to include in `.output/server/node_modules` */
+  include?: string[]
```
pi0, line 14: Since this option is only relevant to trace mode, maybe should be inside `traceOptions` or something like `traceInclude` ?

packages/nitro/src/rollup/plugins/externals.ts
```diff
@@ -67,6 +69,12 @@ export function externals (opts: NodeExternalsOptions): Plugin {
     },
     async buildEnd () {
       if (opts.trace !== false) {
+        for (const pkgName of opts.traceInclude || []) {
+          const path = await this.resolve(pkgName)
+          if (path.id) {
```
pi0, line 74: Resolve result can be possibly `null`
```suggestion
          if (path?.id) {
```

pi0 (approved):

PR #2388, opened 2021-12-14
perf(bridge): allow skipping legacy composition api support

danielroe:
### ❓ Type of change

### 📚 Description

Previously, even if users had fully migrated to new Composition API, we are still loading in support for keying `ssrRef`, etc. There's a small performance win users can realise by disabling legacy support.

```js
  bridge: {
    capi: {
      legacy: false 
    }
  }
})
```

### DIFF
--- a/packages/bridge/src/capi.ts
+++ b/packages/bridge/src/capi.ts
@@ -4,7 +4,7 @@ import { resolve } from 'pathe'
 import { distDir } from './dirs'
 import { KeyPlugin } from './capi-legacy-key-plugin'
 
-export function setupCAPIBridge (_options: any) {
+export function setupCAPIBridge (options: { legacy?: boolean }) {
   const nuxt = useNuxt()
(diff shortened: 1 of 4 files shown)


Atinux:
Can you also update the docs? https://v3.nuxtjs.org/getting-started/bridge#feature-flags

Atinux (approved):

packages/bridge/src/capi.ts
```diff
@@ -4,7 +4,7 @@ import { resolve } from 'pathe'
 import { distDir } from './dirs'
 import { KeyPlugin } from './capi-legacy-key-plugin'
 
-export function setupCAPIBridge (_options: any) {
+export function setupCAPIBridge (options: { legacy?: boolean }) {
```
pi0, line 7: Can we please reuse from `types.d.ts` for options type?

PR #2395, opened 2021-12-15
fix(vite): respect users config

antfu:
When user-defined

```ts
{
  vite: {
    logLevel: 'info' 
  }
}
```

it won't work, which makes it a bit hard to debug or gather info from vite.

### DIFF
--- a/packages/bridge/src/vite/vite.ts
+++ b/packages/bridge/src/vite/vite.ts
@@ -17,15 +17,16 @@ async function bundle (nuxt: Nuxt, builder: any) {
     p.src = nuxt.resolver.resolvePath(resolve(nuxt.options.buildDir, p.src))
   }
 
+  const userViteConfig = nuxt.options.vite as any || {}
   const ctx: ViteBuildContext = {
     nuxt,
     builder,
     config: vite.mergeConfig(
-      nuxt.options.vite || {},
+      userViteConfig,
       {
         root: nuxt.options.rootDir,
         mode: nuxt.options.dev ? 'development' : 'production',
-        logLevel: 'warn',
+        logLevel: userViteConfig.logLevel || 'warn',
         define: {
           'process.dev': nuxt.options.dev
         },
--- a/packages/bridge/src/vite/vite.ts
+++ b/packages/bridge/src/vite/vite.ts
@@ -17,7 +17,6 @@ async function bundle (nuxt: Nuxt, builder: any) {
     p.src = nuxt.resolver.resolvePath(resolve(nuxt.options.buildDir, p.src))
   }
 
-  const userViteConfig = nuxt.options.vite as any || {}
   const ctx: ViteBuildContext = {
     nuxt,
     builder,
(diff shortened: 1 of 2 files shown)


packages/bridge/src/vite/vite.ts
```diff
@@ -17,15 +17,16 @@ async function bundle (nuxt: Nuxt, builder: any) {
...
     nuxt,
     builder,
     config: vite.mergeConfig(
-      nuxt.options.vite || {},
+      userViteConfig,
       {
         root: nuxt.options.rootDir,
         mode: nuxt.options.dev ? 'development' : 'production',
-        logLevel: 'warn',
+        logLevel: userViteConfig.logLevel || 'warn',
```
danielroe, line 29: Is this an issue we need to address in vite's implementation of `mergeConfig`?

Atinux (approved):

packages/bridge/src/vite/vite.ts
danielroe: Okay, we've implemented it the wrong way round then. I've opened a PR upstream to make it less likely this happens in future: https://github.com/vitejs/vite/pull/6144

For solution, we should simply move `userViteConfig` to second argument of `vite.mergeConfig` - I believe the second argument here is meant to be overrided by userConfig.

antfu: I thought it was intended to force the options. Updated

packages/vite/src/vite.ts
```diff
@@ -93,7 +93,8 @@ export async function bundle (nuxt: Nuxt) {
             ]
           }
         }
-      } as ViteOptions
+      } as ViteOptions,
+      userViteConfig
```
danielroe, line 97:
```suggestion
      nuxt.options.vite as any || {}
```

packages/vite/src/vite.ts
```diff
@@ -22,10 +22,10 @@ export interface ViteBuildContext {
 }
 
 export async function bundle (nuxt: Nuxt) {
+  const userViteConfig = nuxt.options.vite as any || {}
```
danielroe, line 25:
```suggestion
```

packages/bridge/src/vite/vite.ts
```diff
@@ -17,11 +17,11 @@ async function bundle (nuxt: Nuxt, builder: any) {
     p.src = nuxt.resolver.resolvePath(resolve(nuxt.options.buildDir, p.src))
   }
 
+  const userViteConfig = nuxt.options.vite as any || {}
```
danielroe, line 20:
```suggestion
```

packages/bridge/src/vite/vite.ts
```diff
@@ -79,7 +79,8 @@ async function bundle (nuxt: Nuxt, builder: any) {
           jsxPlugin(),
           defaultExportPlugin()
         ]
-      } as ViteOptions
+      } as ViteOptions,
+      userViteConfig
```
danielroe, line 83:
```suggestion
      nuxt.options.vite || {}
```

danielroe (commented):
minor suggestion - feel free to ignore

packages/bridge/src/vite/vite.ts
danielroe: Great! I think this will actually solve a few other issues too.

danielroe (approved):

Issue #2394, opened 2021-12-15
Improve Vite initial load


antfu:
Currently, Nuxt will stuck on the first run probably because Vite is optimizing the deps. We should provide some messages and auto refresh the page after Vite is finished the optimization.

antfu:
When I was about the send a PR to fix it, I found it's already fixed in https://github.com/vitejs/vite/pull/5452

I think we just need to bump Vite to 2.7 and also a few complement PR:
- #2396
- #2395

PR #2396, opened 2021-12-15
feat(vite): pre-optimize vite deps

antfu:
Part of #2394

### DIFF
--- a/packages/bridge/src/vite/vite.ts
+++ b/packages/bridge/src/vite/vite.ts
@@ -86,11 +86,15 @@ async function bundle (nuxt: Nuxt, builder: any) {
   await ctx.nuxt.callHook('vite:extend', ctx)
 
   if (nuxt.options.dev) {
-    ctx.nuxt.hook('vite:serverCreated', (server: vite.ViteDevServer) => {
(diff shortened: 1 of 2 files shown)


packages/bridge/src/vite/vite.ts
```diff
@@ -86,11 +86,15 @@ async function bundle (nuxt: Nuxt, builder: any) {
   await ctx.nuxt.callHook('vite:extend', ctx)
 
   if (nuxt.options.dev) {
-    ctx.nuxt.hook('vite:serverCreated', (server: vite.ViteDevServer) => {
+    ctx.nuxt.hook('vite:serverCreated', async (server: vite.ViteDevServer) => {
       const start = Date.now()
       warmupViteServer(server, ['/.nuxt/entry.mjs']).then(() => {
         consola.info(`Vite warmed up in ${Date.now() - start}ms`)
       }).catch(consola.error)
```
danielroe, line 93: How should `optimizeDeps` relate to our warmup task?

antfu: Oh, I just found the when creating a Vite server, it already does that (just no log when we have logLevel: 'warn'). I guess this is not necessary then.

PR #2408, opened 2021-12-16
feat(nuxt3, bridge): add `vue:setup` hook

### LINKED ISSUE #1017: add support for `app.vue` or global setup in nuxt bridge
### Environment

Nuxt CLI v3.0.0-27234347.3759926                                                                                                                                                                              
- Operating System: `Darwin`
- Node Version:     `v16.8.0`
- Nuxt Version:     `2.16.0-27226092.034b9901`
- Package Manager:  `Yarn`
- Bundler:          `Webpack`
- User Config:      `srcDir`, `ssr`, `dev`, `image`, `generate`, `head`, `publicRuntimeConfig`, `privateRuntimeConfig`, `css`, `loading`, `loadingIndicator`, `plugins`, `components`, `buildModules`, `buildOptimisations`, `modules`, `apollo`, `proxy`, `axios`, `vuetify`, `build`
- Runtime Modules:  `@nuxt/image@0.6.0`, `@nuxtjs/axios@5.13.6`, `@nuxtjs/apollo@4.0.1-rc.5`, `@nuxtjs/proxy@2.1.0`, `nuxt-social-meta@1.0.0`
- Build Modules:    `nuxt-storm@1.1.2`, `@nuxt/typescript-build@2.1.0`, `@nuxtjs/composition-api/module@0.29.3`, `@nuxtjs/router@1.6.1`, `nuxt-compress@5.0.0`, `@nuxtjs/vuetify@1.12.1`, `nuxt-build-optimisations@1.0.7`

### Describe the bug

I need to use onGlobalSetup to use vue apollo graphql composables
like this

```
import { provide, onGlobalSetup, defineNuxtPlugin } from '@nuxtjs/composition-api'
import { DefaultApolloClient } from '@vue/apollo-composable'

export default defineNuxtPlugin(({ app }) => {
    onGlobalSetup(() => {
        provide(DefaultApolloClient, app.apolloProvider?.defaultClient)
    })
})
```

But it seems like onGlobalSetup is missing.

Thank you

### Reproduction

import { provide, onGlobalSetup, defineNuxtPlugin } from '@nuxtjs/composition-api'

what is the counterpart of this on nuxt bridge?

Thanks

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #2075: `vue:setup` hook
As a replacement of `onGlobalSetup` (#1017) we shall support a new hook for `nuxtApp`:

```js
useNuxtApp().hook('vue:setup', () => { })
```

For nuxt 3, this hook can be called on [nuxt-root](https://github.com/nuxt/framework/blob/main/packages/nuxt3/src/app/components/nuxt-root.vue) component. /cc @danielroe for ideas re supporting bridge

danielroe:
### 🔗 Linked issue

resolves #2075, resolves #1017

### ❓ Type of change

### 📚 Description

This PR adds a `vue:setup` hook that is called within the root `setup()` function of the app. This also provides a full replacement for `onGlobalSetup` in the legacy composition API module.

**Note**: it will not be awaited because we are 'outside' of the root `<Suspense>`.

### 📝 Checklist


Atinux:
I love it!

packages/bridge/src/runtime/capi.plugin.mjs
```diff
@@ -1,6 +1,15 @@
...
 
 Vue.use(VueCompositionAPI.default || VueCompositionAPI)
 
-export default function () {}
+export default defineNuxtPlugin((nuxtApp) => {
+  const _originalSetup = nuxtApp.nuxt2Context.app.setup
+
+  nuxtApp.nuxt2Context.app.setup = function (...args) {
+    const result = _originalSetup instanceof Function ? _originalSetup(...args) : {}
+    nuxtApp.callHook('vue:setup')
```
pi0, line 12: Shouldn't we avoid on setup? (if capi's setup does not supports, at least need to handle promise's catch or warn user about async if promise returned)

packages/nuxt3/src/app/components/nuxt-root.vue
```diff
@@ -3,3 +3,13 @@
     <App />
   </Suspense>
 </template>
+
+<script>
+import { useNuxtApp } from '#app'
+
+export default {
+  setup () {
+    useNuxtApp().callHook('vue:setup')
```
pi0, line 12: We should await on this call

danielroe: If we await this call we will need _another_ root component with suspense above this one. I think it's reasonable to require `vue:setup` hook to have only syncronous callbacks.

pi0: Your reasoning makes sense but in this case, we need to handle two cases:
- In server-side, strictly disallow returning promise as it can potentially resolve after rendering lifecycle
- Handle rejected promises

Maybe as a simple solution simply detect and warn about async usage?

danielroe: Sounds good to me 👍  Will implement.

Jamiewarb:
Super keen for this 🤩

packages/nuxt3/src/app/components/nuxt-root.vue
danielroe: pushed an implementation but will open a PR to add support for sync calling of hooks in hookable (https://github.com/unjs/hookable/issues/34)

pi0 (approved):

PR #2409, opened 2021-12-16
fix(nuxi,nuxt3,bridge): generate all templates with `nuxi prepare`

### LINKED ISSUE #2401: nuxi prepare doesn't create all `.d.ts` files
### Reproduction

reproducible in playground
```bash
❯ npx nuxi prepare && find .nuxt
Nuxt CLI v3.0.0-27324955.23397e6
.nuxt
.nuxt/nuxt.d.ts
.nuxt/tsconfig.json
```

### Describe the bug

Lots of files are missing, including files referenced within `nuxt.d.ts`. So type checking and IDE type support will fail.

**TODO**:
- [ ] handle generating types in Nuxt 3 - https://github.com/nuxt/framework/pull/2409
- [ ] handle generating types in Bridge - https://github.com/nuxt/framework/pull/2412
- [x] handle generating types in Nitro - https://github.com/nuxt/framework/pull/2413

danielroe:
### 🔗 Linked issue

resolves #2401

### ❓ Type of change

### 📚 Description

This PR skips bundling when running `nuxi prepare` but otherwise follows the normal pattern of generating templates, etc., so that all `.d.ts` files will be generated appropriately in bridge & nuxt3.

### 📝 Checklist


Atinux (approved):

pi0:
After this refactor, it kinda makes sense to merge back builder and app together (i prefer using `builder` to reduce confusion with runtime app)

pi0:
As an alternative, we could also provide an option to bypass [bundle call](https://github.com/nuxt/framework/blob/5ab18162ddb5a91cb43c0e6e7a2c529f6e9247f3/packages/nuxt3/src/core/builder.ts#L23). We can keep using builder entity and if by any chance module's behavior is depending on `build:before`, `build:after` hooks, they keep working.

danielroe:
Note that this will build the app if calling `nuxi prepare` in a Bridge project. We can add a similar test in Nuxt 2.16 that will enable parallel behaviour and mean we don't need to merge https://github.com/nuxt/framework/pull/2412.

pi0:
For Nuxt2/bridge, we could also mock `bundleBuilder.build` within module but if not possible, seems a nice idea backporting.

packages/bridge/src/app.ts
```diff
@@ -10,6 +10,13 @@ export function setupAppBridge (_options: any) {
   nuxt.options.alias['#app'] = resolve(distDir, 'runtime/index.mjs')
   nuxt.options.alias['#build'] = nuxt.options.buildDir
 
+  // Mock `bundleBuilder.build` to support `nuxi prepare`
+  if (nuxt.options._prepare) {
+    nuxt.hook('builder:prepared', (builder) => {
+      builder.bundleBuilder.build = () => {}
```
pi0, line 16:
```suggestion
      builder.bundleBuilder.build = () => Promise.resolve(builder.bundleBuilder)
```

pi0: We might need to call `build:before`, `build:prepared` and `build:done` hooks after mock I guess there was some issues with storybook module /cc @farnabaz

pi0 (approved):

packages/bridge/src/app.ts
farnabaz: As I remember we need to call  `webpack:config` too

PR #2412, opened 2021-12-16
fix(bridge): add `builder:generateApp` hook implementation

danielroe:
### 🔗 Linked issue

#2401 (for bridge)

### ❓ Type of change

### 📚 Description

This adds a `builder:generateApp` compatible hook that will generate a `.nuxt` directory (for use in `nuxi prepare`)

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/app.ts
+++ b/packages/bridge/src/app.ts
@@ -96,10 +96,10 @@ export function setupAppBridge (_options: any) {
   })
 
   nuxt.hook('builder:generateApp', async () => {
-    const { Builder } = await import('@nuxt/builder')
+    const { getBuilder } = await import('nuxt')
     ;(nuxt.options.build as any).createRoutes = () => []
 
-    const builder = new Builder(nuxt)
+    const builder = getBuilder(nuxt)
     await nuxt.callHook('builder:prepared', builder, nuxt.options.build)
     await builder.generateRoutesAndFiles()
   })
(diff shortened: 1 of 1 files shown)


packages/bridge/src/app.ts
```diff
@@ -94,4 +94,13 @@ export function setupAppBridge (_options: any) {
       })
     }
   })
+
+  nuxt.hook('builder:generateApp', async () => {
+    const { Builder } = await import('@nuxt/builder')
+    ;(nuxt.options.build as any).createRoutes = () => []
+
+    const builder = new Builder(nuxt)
```
pi0, line 102: We can use `getBuilder` utility from main `nuxt` or `nuxt-edge` distribution instead. Current nuxt2 installation is via `nuxt-edge` for bridge so this is necessary.

packages/bridge/src/app.ts
```diff
@@ -94,4 +94,13 @@ export function setupAppBridge (_options: any) {
       })
     }
   })
+
+  nuxt.hook('builder:generateApp', async () => {
+    const { getBuilder } = await import('nuxt')
```
pi0, line 99: - Import path should be considered (from user's rootDir)
- `nuxt-edge` fallback should be considered

Maybe we can refactor functionality into `importNuxt` in kit? (https://github.com/nuxt/framework/blob/main/packages/kit/src/loader/nuxt.ts)

pi0: https://github.com/nuxt/framework/issues/2491

pi0:
Closing in favor of #2409 to generate templates

PR #2413, opened 2021-12-16
feat(nitro, nuxt3, bridge): expose `writeTypes` and call this within `builder:generateApp`

danielroe:
### 🔗 Linked issue

#2401

### ❓ Type of change

### 📚 Description

This exports `writeTypes` from Nitro, allowing types to be written without building a Nitro server. This allows us to call `writeTypes` within bridge & nuxt3 to generate `nitro.d.ts`.

I would suggest entirely moving the _calling_ of `writeTypes` out of `@nuxt/nitro` and into a Nuxt hook, as it seems more like the right place for the job.

I've created this as a separate PR from the other two related .d.ts PRs as we may pend this until `nitropack` lands. And this is purely additive; that is, the other PRs work fine without it, they just don't generate `nitro.d.ts`.

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/nitro.ts
+++ b/packages/bridge/src/nitro.ts
@@ -3,7 +3,7 @@ import fetch from 'node-fetch'
 import { addPluginTemplate, useNuxt } from '@nuxt/kit'
 import { stringifyQuery } from 'ufo'
 import { resolve } from 'pathe'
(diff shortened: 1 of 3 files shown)


pi0 (approved):
Future refactor: We shall change functionality of `writeTypes` from nitro into a utility that returns either stringified types or a meta object to generate them

PR #2417, opened 2021-12-16
fix(nuxt3): cache `useCookie` ref on nuxtApp instance

### LINKED ISSUE #2416: `useCookie` state is stored independently
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

```vue
<template>
  <div>
    <h1>
      {{ cookie2 }}
    </h1>
    <button @click="cookie1 = (Math.random() * 100).toFixed()">
      Change cookie
    </button>
  </div>
</template>

<script setup>
const cookie1 = useCookie('test')
const cookie2 = useCookie('test')
</script>
```

### Describe the bug

Multiple invocations of `useCookie` are not synced, and so setting one will not trigger a change to others. This is not a bug per-se, but it's inconsistent compared to `useState`, for example...

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2416

### ❓ Type of change

### 📚 Description

This PR caches `useCookie` refs so that they will be reactive across invocations of `useCookie`. There's a memory overhead for this as the refs will persist in memory but cookies are normally small and few in number, so I think the benefit outweights the risk.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/composables/cookie.ts
+++ b/packages/nuxt3/src/app/composables/cookie.ts
@@ -22,6 +22,12 @@ const CookieDefaults: CookieOptions<any> = {
 }
 
 export function useCookie <T=string> (name: string, _opts?: CookieOptions<T>): CookieRef<T> {
+  const nuxtApp = useNuxtApp()
+  nuxtApp._cookies = nuxtApp._cookies || {}
+  if (nuxtApp._cookies[name]) {
+    return nuxtApp._cookies[name]
+  }
+
   const opts = { ...CookieDefaults, ..._opts }
   const cookies = readRawCookies(opts)
 
@@ -39,6 +45,8 @@ export function useCookie <T=string> (name: string, _opts?: CookieOptions<T>): C
       }
     })
   }
+
+  nuxtApp._cookies[name] = cookie
 
   return cookie as CookieRef<T>
 }


pi0:
We probably need to only use cache if options are not provided or create separate cache keys. `useCookie` ref behavior can be different when options are provided. 

BTW as a best practice, as mentioned in docs, one can create a shared composable and using auto imports. This helps preserving ref, types and using same cookie options

danielroe:
It's definitely best practice to create a shared composable - and caching the ref does certainly provide a foot-gun opportunity. But if we don't implement caching for refs, then each composable will need to re-implement it themselves.

Some options:
* we can implement cache and leave user to handle potential issues due to passing incompatible options to different instances
* we could add a `cache` _option_ for opt-in usage
* we could add a `useSharedValue` composable that handles caching a value on the nuxt app instance, which could be used in creating the kind of shared composable we have in mind

For the last option implementation could be something like:

```js
export const useMyCookie = () => {
  return useSharedValue('my-cookie', () => useCookie('my-cookie'))
}
```

pi0:
Maybe we could simply consider a hash of opts in the cache key? In the future would be nice to refactor composable caching mechanism out

pi0:
I wish we could have a native API to watch cookie changes but even browsers don't have a proper spec for this. Maybe we could use a hook to notify other `useCookie` instances reading cookie again?

BTW I think still using cache but with proper cache-key makes sense. Any approach we do, there is an overhead for each instance creating using incrementor or hook listener. And this is a common issue for composables to solve for caching multiple calls.

danielroe:
Let's avoid caching refs but use a hook to ensure cookies stay in sync.

Issue #2337, opened 2021-12-10
useMeta removes existing tags and then adds them all again


nndnha:
### Environment

- Operating System: `Linux`
- Node Version:     `v14.16.0`
- Nuxt Version:     `3-3.0.0-27237303.6acfdcd`
- Package Manager:  `npm`
- Bundler:          `Webpack`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-jmj87r?file=app.vue

https://user-images.githubusercontent.com/12961516/145565306-7df3339a-3a11-4268-ab07-cc0fb9240d24.mp4

### Describe the bug

The body's background color is set to `#000` in the `style` tag when the page has loaded, when client hydrated that `style` is removed firstly then added again, this makes the screen flickering.

I added an extra script that uses `MutationObserver` to logs the add/remove element events. It shows that the `style` element is removed firstly then added again.  I think `useMeta` shouldn't touch these existing tags.

danielroe:
Resolved with the latest version of `@vueuse/head`.

PR #2419, opened 2021-12-16
fix(nuxt3): don't immediately trigger update dom

nndnha:
### 🔗 Linked issue

#2337 
vueuse/head#45

### ❓ Type of change

### 📚 Description

`watchEffect` is an eager watch by immediately executing the DOM update. At the first DOM update after the page has been loaded, vueuse/head library will remove all existing attributes and then set them all to the same old values again, for some attribute like `class`, `style`... this toggle will make the screen flickering. I think `watch` is a proper choice in this case.

### 📝 Checklist


fabiwlf:
Why was it closed?

nndnha:
> Why was it closed? This issue still exists and your fix seems to work fine

In think the root of this problem is here https://github.com/vueuse/head/issues/45 so this wasn't Nuxt's issue.

PR #2437, opened 2021-12-17
add components.d.ts as a reference in nuxt.d.ts

acidjazz:
hopefully this helps ide's pick up auto-imported local components

### ❓ Type of change

### 📝 Checklist


danielroe:
@acidjazz Thank you! Could you provide an issue with reproduction?

I'm up for changing things if there's an issue, but this is currently what we're doing:

https://github.com/nuxt/framework/blob/3eed5cb193ca1b57c96215295281047cd5aca72f/packages/nuxt3/src/components/module.ts#L83-L87

So if there's an issue it might be broader than just adding `components.d.ts`

acidjazz:
@danielroe this helps PHPStorm and WebStorm identify components.. but its still a bit dodgy, and like we discussed on discord it might be best to let the IDE's fix their Vue+TS compatibility,

PR #2442, opened 2021-12-18
fix(nuxt3): `useRoute` reflecting route changes

### LINKED ISSUE #2441: `useRoute` is not updated after initially being set
### Environment

- Operating System: `Linux`
- Node Version:     `v14.18.2`
- Nuxt Version:     `3.0.0-27328916.daec432`
- Bundler:          `Vite`
- User Config:      `meta`, `css`, `buildModules`, `unocss`, `vite`, `publicRuntimeConfig`
- Runtime Modules:  `-`
- Build Modules:    `@vueuse/nuxt@7.4.0`, `@unocss/nuxt@0.16.4`

### Reproduction

Reproduction simply requires multiple (`/a and /b` ) routes to be created in the `pages` directory.

navigate to `/a` then check useRoute().path, it's value will be `/a` as expected

however though, upon navigating to `/b` then checking useRoute().path. You'll notice that the current route instance has not been updated and still reflects the route that was initially navigated to, ie: `/a`

### Describe the bug

[#2406](https://github.com/nuxt/framework/pull/2406)

the changes to useRoute() made in the above PR prevents the current route from being correctly reflected after it is initially set.

### Additional context

_No response_

### Logs

_No response_

Diizzayy:
### 🔗 Linked issue

resolves #2441 

### ❓ Type of change

### 📚 Description

This PR allows useRoute() to properly reflect the updated current route after it's initially set.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/pages/runtime/composables.ts
+++ b/packages/nuxt3/src/pages/runtime/composables.ts
@@ -8,9 +8,6 @@ export const useRouter = () => {
 
 export const useRoute = (): RouteLocationNormalizedLoaded => {
   const nuxtApp = useNuxtApp()
-  if (nuxtApp._route) {
-    return nuxtApp._route
-  }
 
   const currentRoute = (nuxtApp.$router as Router).currentRoute
 


packages/nuxt3/src/pages/runtime/composables.ts
```diff
@@ -8,9 +8,6 @@ export const useRouter = () => {
 
 export const useRoute = (): RouteLocationNormalizedLoaded => {
   const nuxtApp = useNuxtApp()
-  if (nuxtApp._route) {
-    return nuxtApp._route
-  }
```
danielroe, line 13: This means we'd be creating a new computed object on each call of `useRoute`, which does resolve the issue (but at a performance cost). It's unclear why this composable as written isn't working, but I've put in an alternative approach in #2443 which resolves the issue by simply creating the reactive route at the same time the router is creatd.

Diizzayy: Ahh, this makes perfect sense. After overviewing your changes I see the problem of the performance cost you pointed out.

PR #2456, opened 2021-12-20
fix(schema): handle null/undefined values in `runtimeConfig`

### LINKED ISSUE #2423: Environment variables in privateRuntimeConfig do not work in production
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0-27324955.23397e6`
- Package Manager:  `yarn@1.21.1`
- Bundler:          `Vite`
- User Config:      `privateRuntimeConfig`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Reproduction repo: https://github.com/cyruscollier/nuxt3-demos/tree/private-runtime-config

Reproduction steps:

- Run `PRIVATE_TEST=test yarn dev`
- Visit http://localhost:3000. The environment variable / private runtime config value correctly displayed
- Run `yarn build`
- Run `PRIVATE_TEST=test node .output/server/index.mjs`
- Visit http://localhost:3000. The private runtime config value is missing. 

Looking inside of `.output/server/chunks/app/render.mjs` this code is generated:

```
const _runtimeConfig = {"public":{"app":{"basePath":"/","assetsPath":"/_nuxt/","cdnURL":null}},"private":{}};
for (const type of ["private", "public"]) {
  for (const key in _runtimeConfig[type]) {
    _runtimeConfig[type][key] = destr(process.env[key] || _runtimeConfig[type][key]);
  }
}
```

It appears that because the `private` key is empty, the loading mechanism of the environment variables is lost.

### Describe the bug

When environment variables are used in the private runtime config, they are not loaded during a production build. They do get statically copied into the bundle when a `.env` file is used, but not when using live environment variables.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2423

### ❓ Type of change

### 📚 Description

By default `defu` doesn't process values that are null or undefined at the top level:

https://github.com/unjs/defu/blob/main/src/defu.ts#L22-L24

But we likely want to preserve these in `runtimeConfig` as this enables the runtime check for the existence of these environment variables.

I've also switched to using `devalue` for Nitro's runtimeConfig as this enables us to preserve `undefined` values.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -714,6 +714,6 @@ export default {
    * @version 3
    */
   publicRuntimeConfig: {
-    $resolve: (val = {}, get) => ({ ...val, app: defu(val.app, get('app') })
+    $resolve: (val = {}, get) => ({ ...val, app: defu(val.app, get('app')) })
   },
 }
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -714,6 +714,6 @@ export default {
    * @version 3
    */
   publicRuntimeConfig: {
-    $resolve: (val, get) => defu(val, { app: get('app') })
+    $resolve: (val, get) =>  Object.assign({}, val, defu(val, { app: get('app') }))
   },
 }
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -714,6 +714,6 @@ export default {
    * @version 3
    */
   publicRuntimeConfig: {
-    $resolve: (val, get) =>  Object.assign({}, val, defu(val, { app: get('app') }))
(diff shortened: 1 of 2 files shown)


packages/schema/src/config/_common.ts
```diff
@@ -714,6 +714,6 @@ export default {
    * @version 3
    */
   publicRuntimeConfig: {
-    $resolve: (val, get) => defu(val, { app: get('app') })
+    $resolve: (val, get) => Object.assign({}, val, defu(val, { app: get('app') }))
```
pi0, line 717:
```suggestion
    $resolve: (val = {}, get) => ({ ...val, app: defu(val.app, get('app') })
```

danielroe: much cleaner

PR #2481, opened 2021-12-21
fix(nuxt3): don't mutate options when unsetting cookie

### LINKED ISSUE #2480: `useCookie` can't set cookies after clearing it
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

```js
const cookie = useCookie('test-cookie')
cookie.value = null

function setCookie() {
  cookie.value = 'some value'
}
```

### Describe the bug

we are mutating cookie options here:
https://github.com/nuxt/framework/blob/6a25d3e245a88d3c838dafa643a5210dba625d73/packages/nuxt3/src/app/composables/cookie.ts#L61-L63

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2480

### ❓ Type of change

### 📚 Description

When unsetting the cookie, we were mutating the options which mean that any subsequent setting of cookie value would have a `maxAge` of `-1`, which would effectively unset the cookie immediately.

https://github.com/nuxt/framework/blob/6a25d3e245a88d3c838dafa643a5210dba625d73/packages/nuxt3/src/app/composables/cookie.ts#L61-L63

This PR clones the options in the serialize function so we don't mutate the original options object.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/composables/cookie.ts
+++ b/packages/nuxt3/src/app/composables/cookie.ts
@@ -57,7 +57,8 @@ function readRawCookies (opts: CookieOptions = {}): Record<string, string> {
   }
 }
 
-function serializeCookie (name: string, value: any, opts: CookieSerializeOptions = {}) {
+function serializeCookie (name: string, value: any, _opts: CookieSerializeOptions = {}) {
+  const opts = { ..._opts }
   if (value === null || value === undefined) {
     opts.maxAge = -1
   }
--- a/packages/nuxt3/src/app/composables/cookie.ts
+++ b/packages/nuxt3/src/app/composables/cookie.ts
@@ -57,10 +57,9 @@ function readRawCookies (opts: CookieOptions = {}): Record<string, string> {
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/nuxt3/src/app/composables/cookie.ts
```diff
@@ -57,7 +57,8 @@ function readRawCookies (opts: CookieOptions = {}): Record<string, string> {
   }
 }
 
-function serializeCookie (name: string, value: any, opts: CookieSerializeOptions = {}) {
+function serializeCookie (name: string, value: any, _opts: CookieSerializeOptions = {}) {
+  const opts = { ..._opts }
```
pi0, line 61: Perf: Maybe we could set a flag like `const removeCookie = value === null || value === undefined` and only destructure+merge when is set?

pi0 (approved):

PR #2559, opened 2021-12-28
docs: add Stackblitz links for examples

antfu:
### 🔗 Linked issue

In the future we could also have some meta info for each example in a JSON file, with a short descriptions and pointers to docs

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,17 @@
+# Nuxt 3 Examples
+
+| Example | Source | Playground |
+|---|---|---|
+| `hello-world` | [GitHub](https://github.com/nuxt/framework/tree/main/examples/hello-world) | [Play Online](https://stackblitz.com/github/nuxt/framework/tree/main/examples/hello-world?file=app.vue) |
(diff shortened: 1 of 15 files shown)


pi0:
Love it! Thanks!

pi0:
Some comments:
- We need to fix lint issue (probably also with auto-update script)
- We probably don't need to maintain a `.stackblitzrc` file and reduce boilerplate
  - `installDependencies` can be part of `startCommand`
  - `startCommand` can be embedded as `?command` query string
- Maybe making a symlink from docs to also display table somewhere? (~> #1960)

antfu:
> startCommand can be embedded as ?command query string

That sounds like a good idea. But it does not seem to work and I didn't find it documented anywhere either, did I missed anything?

pi0:
My fault! Query is `terminal=`. Example:

- https://stackblitz.com/github/nuxt/framework/tree/main/examples/use-cookie?terminal=dev

antfu:
Where did you learn that magic? 🤣

Updated

pi0 (approved):

PR #2574, opened 2021-12-30
perf(ssr): disable vite's pre-trasnfroming to improve perf

antfu:
### 🔗 Linked issue

In Vite 2.6, it introduced a mechanism to pre-transform the direct imports before the browser requests landed. It could improve the response time of waterflow requests. However, when in SSR, we have our custom logic to handle the externalization, which the pre-transform might start transforming the externalized modules, causing some computation power waste.

Refer to https://github.com/vitest-dev/vitest/issues/229#issuecomment-1002685027

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/vite/src/server.ts
+++ b/packages/vite/src/server.ts
@@ -69,6 +69,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       }
     },
     server: {
+      // https://github.com/vitest-dev/vitest/issues/229#issuecomment-1002685027
       preTransformRequests: false
     },
     plugins: [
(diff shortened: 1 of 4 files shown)


packages/vite/src/server.ts
```diff
@@ -68,6 +68,9 @@ export async function buildServer (ctx: ViteBuildContext) {
         }
       }
     },
+    server: {
+      preTransformRequests: false
```
pi0, line 72:
```suggestion
      // https://github.com/vitest-dev/vitest/issues/229#issuecomment-1002685027
      preTransformRequests: false
```

PR #2587, opened 2022-01-03
fix(vite): only remove version query

antfu:
### 🔗 Linked issue

With the fix of #2521, it removes all queries regardless. Which breaks the usage of Vite's `?raw`, `?url`, `?worker` etc. This PR changes the behavior by only removing the `?v=123` query.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/vite/src/dev-bundler.ts
+++ b/packages/vite/src/dev-bundler.ts
@@ -3,7 +3,6 @@ import { existsSync } from 'fs'
 import { resolve } from 'pathe'
 import * as vite from 'vite'
 import { ExternalsOptions, isExternal as _isExternal, ExternalsDefaults } from 'externality'
-import { parseURL } from 'ufo'
 import { hashId, uniq } from './utils'
 
 export interface TransformChunk {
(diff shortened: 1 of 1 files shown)


packages/vite/src/dev-bundler.ts
```diff
@@ -70,11 +69,12 @@ async function transformRequest (opts: TransformOptions, id: string) {
     }
   }
 
-  const { pathname } = parseURL(id)
-
-  if (await isExternal(opts, pathname)) {
+  // Vite will add ?v=123 to bypass browser cache
+  // Remove for externals
+  const withoutVersionQuery = id.replace(/\?v=\w+$/, '')
```
danielroe, line 74: Do we ever get multiple params? e.g. `?v=123&worker` or `?worker&v=123`

antfu: In that case, it shouldn't be externalized and fallback to vite's resolving and it understand it. So the only case that matters is the raw `?v=123`

danielroe (approved):

pi0 (approved):

PR #2611, opened 2022-01-05
fix(bridge): install bridge module at the end

### LINKED ISSUE #2501: "autoimports:sources" hook is called before module installation
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.11.1`
- Nuxt Version:     `2.16.0-27334095.777a4b7f`
- Bundler:          `Webpack`
- User Config:      `target`, `srcDir`, `dir`, `bridge`, `head`, `components`, `generate`, `buildModules`, `modules`, `alias`, `vite`, `hooks`, `layouts`, `autoImports`
- Runtime Modules:  `normalizedModule()`
- Build Modules:    `@nuxt/bridge@3.0.0-27334958.ca9761d`, `@vueuse/nuxt@7.4.0`, `unplugin-icons/nuxt`, `@pinia/nuxt@0.1.6`

### Reproduction

Install @vueuse/nuxt.

### Describe the bug

The "autoImports:sources" is called before the `@vueuse/nuxt` package is installed, so the hook isn't triggered and autoImport sources are not added. My suggestion is to install the `autoImports` module inside a `modules:done` hook to make sure all other modules relying on it actually work.

### Additional context

_No response_

### Logs

_No response_

antfu:
Fix #2501

### ❓ Type of change

### 📚 Description

I am not very sure if there is a reason for the initial decision to append the module at the beginning, but having that at the beginning caused the other modules relying on the hooks provided by the bridge fail to work.

e.g. in @vueuse/nuxt

We have
```ts
// @vueuse/nuxt
{
  setup() {
    nuxt.hook('autoImports:sources', () => { /* register the auto imports */ })
  }
}
```

Where if the bridge module has been installed before, the hook has already been called, causing the auto-import registration in VueUse never called.

As a temporary workaround, explicitly set the bridge module at the end

```ts
// nuxt.config.js
{
  buildModules: [
    '@vueuse/nuxt',
    '@nuxt/bridge'
  ]
}
```

### 📝 Checklist

### DIFF
--- a/packages/bridge/module.cjs
+++ b/packages/bridge/module.cjs
@@ -13,6 +13,7 @@ module.exports.defineNuxtConfig = (config = {}) => {
       config.buildModules = []
     }
     if (!config.buildModules.find(m => m === '@nuxt/bridge' || m === '@nuxt/bridge-edge')) {
+      // Ensure other modules register their hooks before
       config.buildModules.push('@nuxt/bridge')
     }
   }
--- a/packages/bridge/module.cjs
+++ b/packages/bridge/module.cjs
@@ -13,7 +13,7 @@ module.exports.defineNuxtConfig = (config = {}) => {
       config.buildModules = []
     }
     if (!config.buildModules.find(m => m === '@nuxt/bridge' || m === '@nuxt/bridge-edge')) {
-      config.buildModules.unshift('@nuxt/bridge')
+      config.buildModules.push('@nuxt/bridge')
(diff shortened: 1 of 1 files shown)


yassilah:
I believe this should also fix https://github.com/nuxt/framework/issues/2501?

antfu:
> I believe this should also fix #2501?

Yes exactly, thanks for the reference

PR #2636, opened 2022-01-07
perf(vite): do not resolve `module` field for packages

antfu:
### ❓ Type of change

### 📚 Description

https://github.com/vitejs/vite/blob/04016df8beb7110ad2087ada8e1abbefdbd70254/packages/vite/src/node/constants.ts#L3-L7

We have confirmed this works for Vitest and sloved many issues we had.

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/vite/server.ts
+++ b/packages/bridge/src/vite/server.ts
@@ -41,7 +41,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       alias,
       // By default Vite resolves `module` field, which not always a native ESM module
       // Setting this option can bypass that and fallback to cjs version
-      mainFields: ['main']
+      mainFields: ctx.nuxt.options.dev ? ['main'] : undefined
     },
     ssr: {
       external: [
--- a/packages/bridge/src/vite/server.ts
+++ b/packages/bridge/src/vite/server.ts
@@ -38,7 +38,10 @@ export async function buildServer (ctx: ViteBuildContext) {
     },
     cacheDir: resolve(ctx.nuxt.options.rootDir, 'node_modules/.cache/vite/server'),
     resolve: {
-      alias
+      alias,
+      // By default Vite resolves `module` field, which not always a native ESM module
+      // Setting this option can bypass that and fallback to cjs version
+      mainFields: []
     },
     ssr: {
       external: [
--- a/packages/bridge/src/vite/server.ts
+++ b/packages/bridge/src/vite/server.ts
@@ -41,7 +41,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       alias,
       // By default Vite resolves `module` field, which not always a native ESM module
       // Setting this option can bypass that and fallback to cjs version
-      mainFields: []
+      mainFields: ['main']
     },
     ssr: {
       external: [
--- a/packages/vite/src/server.ts
+++ b/packages/vite/src/server.ts
@@ -38,7 +38,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       },
       // By default Vite resolves `module` field, which not always a native ESM module
       // Setting this option can bypass that and fallback to cjs version
-      mainFields: []
+      mainFields: ['main']
     },
     ssr: {
       external: [],
--- a/packages/vite/src/server.ts
+++ b/packages/vite/src/server.ts
@@ -38,6 +38,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       },
       // By default Vite resolves `module` field, which not always a native ESM module
       // Setting this option can bypass that and fallback to cjs version
+      // TODO: Fix nitro/unenv issue with `main` field
       mainFields: ctx.nuxt.options.dev ? ['main'] : undefined
     },
     ssr: {
--- a/packages/vite/src/server.ts
+++ b/packages/vite/src/server.ts
@@ -35,7 +35,10 @@ export async function buildServer (ctx: ViteBuildContext) {
         '@vue/shared': _resolve(`@vue/shared/dist/shared.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`),
         'vue-router': _resolve(`vue-router/dist/vue-router.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`),
         vue: _resolve(`vue/dist/vue.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`)
-      }
+      },
+      // By default Vite resolves `module` field, which not always a native ESM module
+      // Setting this option can bypass that and fallback to cjs version
+      mainFields: []
     },
     ssr: {
       external: [],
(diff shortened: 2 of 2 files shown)


packages/bridge/src/vite/server.ts
```diff
@@ -38,7 +38,10 @@ export async function buildServer (ctx: ViteBuildContext) {
     },
     cacheDir: resolve(ctx.nuxt.options.rootDir, 'node_modules/.cache/vite/server'),
     resolve: {
-      alias
+      alias,
+      // By default Vite resolves `module` field, which not always a native ESM module
+      // Setting this option can bypass that and fallback to cjs version
+      mainFields: []
```
pi0, line 44: Maybe explicitly list exports to [module, main]? 

https://github.com/rollup/plugins/tree/master/packages/node-resolve#mainfields

packages/vite/src/server.ts
```diff
@@ -35,7 +35,10 @@ export async function buildServer (ctx: ViteBuildContext) {
         '@vue/shared': _resolve(`@vue/shared/dist/shared.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`),
         'vue-router': _resolve(`vue-router/dist/vue-router.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`),
         vue: _resolve(`vue/dist/vue.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`)
-      }
+      },
+      // By default Vite resolves `module` field, which not always a native ESM module
+      // Setting this option can bypass that and fallback to cjs version
+      mainFields: []
```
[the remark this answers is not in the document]
pi0: Ref: rollup plugin option: https://github.com/rollup/plugins/tree/master/packages/node-resolve#mainfields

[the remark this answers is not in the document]
antfu: I guess it will always fall back to `main`? I see no issue to leave it empty, but sure we could have it to be explict

packages/bridge/src/vite/server.ts
antfu, line 44:
```suggestion
      mainFields: ['main']
```

packages/bridge/src/vite/server.ts
```diff
@@ -38,7 +38,10 @@ export async function buildServer (ctx: ViteBuildContext) {
     },
     cacheDir: resolve(ctx.nuxt.options.rootDir, 'node_modules/.cache/vite/server'),
     resolve: {
-      alias
+      alias,
+      // By default Vite resolves `module` field, which not always a native ESM module
+      // Setting this option can bypass that and fallback to cjs version
+      mainFields: ctx.nuxt.options.dev ? ['main'] : undefined
```
pi0, line 44: Re comment: main field is not necessarily pointing to a cjs file but usually does (mjs extension or type:module changes that)

Does it makes sense to create incosnsiy behavior between dev and prod resolution? Some library might stop working only in production

antfu: > main field is not necessarily pointing to a cjs file but usually does (mjs extension or type:module changes that)

Yes, I think the more correct statement is "resolving" to valid Node.js modules, both cjs and esm are fine, the main problem was `module` field commonly pointing to ESM syntax with `.js` extension

> Does it makes sense to create incosistent behavior between dev and prod resolution? Some library might stop working only in production

In production build, we use rollup to bundle the deps so `.esm.js` are acceptable, while in dev we are doing "unbundle" and rely on native Node. I think if there are misalignments in a package's CJS/ESM versions, it's more likely the responsibility of the lib author right? IIRC, the ecosystem has been like this for a long while, like you bundle your app with webpack but external some deps (some resolve to `.esm.js` version, and those externaled will be resolve to cjs version)

antfu: Or maybe we could have this as an experiment (since we are on beta) and revert if it really breaks something?

pi0: Well, we also apply externalization checks in next steps of resolution to avoid externalizing nonstandard esm entries. Ignoring `module` field in favor of `main` (in case of `exports` lacking) makes sense for performance but two different package resolution logic for dev and production is not safe IMO. (currently both dev and prod use similar external check logic)

antfu: - `exports` always have the highest priority in resolving, which is the standard and the future. I believe more and more packages are going to adapt it so we will have less to be concerned about.
- The resolutions in dev and build are already different, since Vite doesn't bundle them, and even has deps optimization that completely rewrites the structure of deps. I don't really see this being a problem as long as library authors ships multiple versions that behaviors the same (otherwise I will consider that being a bug of the lib)
- This is indeed a trade-off, just like all the decisions. But I see that Webpack might lean more toward the consistency between dev/build, while Vite is trying to address the DX while keeping things just work. In this case, one side is huge performance gain (you can skip the transformation, and file IO for a huge amount of deps), one the other side is something that "might" cause misalignment (but not yet confirmed) - I personally think this trade-off is totally worth it (and we can revert it anytime)

So this is my path of thinking. If you are still not convinced, feel free to close it with no bad feelings :)

pi0: All seem good other than ignoring `main` field. Does always defaulting to `['main']` and ignoring `['module']` seems good to you?

antfu: Always defaulting to `main` sounds good, but it cause Nitro not happy with it (see previous CI errors). I might need to investigate what causes that - since you are much more familiar with how Nitro work, might you can share some insight? Otherwise, I can dig deeper into it.

packages/vite/src/server.ts
```diff
@@ -35,7 +35,10 @@ export async function buildServer (ctx: ViteBuildContext) {
         '@vue/shared': _resolve(`@vue/shared/dist/shared.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`),
         'vue-router': _resolve(`vue-router/dist/vue-router.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`),
         vue: _resolve(`vue/dist/vue.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`)
-      }
+      },
+      // By default Vite resolves `module` field, which not always a native ESM module
+      // Setting this option can bypass that and fallback to cjs version
```
pi0, line 40:
```suggestion
      // Setting this option can bypass that and fallback to cjs version
      // TODO: Fix nitro/unenv issue with `main` field
```

packages/bridge/src/vite/server.ts
pi0: Ahh didn't notice that. For now added a TODO regarding nitro/unenv issue to dig deeper later.

pi0 (approved):

pi0:
Changes temporary reverted because of regression discussed in #2679

antfu:
Ah sorry for that. I think it also requires some additional logic to interop the default, I will include that in the upcoming refactor of using on-demand module evaluation

PR #2656, opened 2022-01-10
docs: add examples of multi filed components

markteekman:
For our design system we use a folder for every component (which sometimes contains other component parts). To get this to work in our nuxt.js export we had to create an array of directories and loop over it to add them all to the export. I thought maybe others would be helped with it :) We also didn't want to use a prefix so I added an example for that as well. I also changed the import order, I got an error if I didn't import pathe after defineNuxtModule so I switched them around. I used @nuxt/kit-edge in this example, as I got an error when I used @nuxt/kit.

### 🔗 Linked issue

I wanted to add the issue myself but I don't really know how to reproduce a minimal reproduction using our private design system.. I don't know how to include a private NPM package in a CodeSanbox. I hope that's okay.

### ❓ Type of change

- [x ] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

This fixes a couple of things. Pathe is need before defineNuxtModule is imported, so I switched those two around. Next is that I get errors when using @nuxt/kit, @nuxt/kit-edge does work however. The biggest thing I wanted to add is how you can define your library components if they are put in their on separate folders. You need to create and array and loop over each item to add the components from those folders. I also added an example for when you don't want the prefix, we don't use that for example. This is a use case of how we are currently importing our design system components into a Nuxt 3 application.

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/2.directory-structure/4.components.md
+++ b/docs/content/3.docs/2.directory-structure/4.components.md
@@ -140,8 +140,8 @@ Imagine a directory structure like this:
 Then in `awesome-ui/nuxt.js` you can use the `components:dirs` hook:
 
 ```js
-import { join } from 'pathe'
-import { defineNuxtModule } from '@nuxt/kit'
+import { defineNuxtModule } from ‘@nuxt/kit-edge’
+import { join } from ‘pathe’
 
 export default defineNuxtModule({
   hooks: {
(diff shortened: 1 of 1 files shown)


docs/content/3.docs/2.directory-structure/4.components.md
```diff
@@ -176,3 +176,45 @@ export default {
 ```
 
 It will automatically import the components only if used and also support HMR when updating your components in `node_modules/awesome-ui/components/`.
+
+If you have library components that are stored in their own folders you can define and loop over an array to add them to the export:
+
+```js
+import { defineNuxtModule } from ‘@nuxt/kit-edge’
```
pi0, line 183: (we should use `@nuxt/kit` with npm alias)

```suggestion
import { defineNuxtModule } from '@nuxt/kit'
```

docs/content/3.docs/2.directory-structure/4.components.md
```diff
@@ -176,3 +176,45 @@ export default {
...
+import { join } from ‘pathe’
```
pi0, line 184:
```suggestion
import { join } from 'pathe'
```

pi0:
Hi @markteekman and thanks for the explained PR. I'm wondering if your goal is to guide making a module extending components dir or configure the project itself? For second case, `components:dirs` won't be necessary. We can directly use top level `components.dirs` option.

markteekman:
Hi @pi0, thanks for your reply! First of all, I think I made a mistake. The solution I added to the docs is something me and my team implemented back in November, because back then we would get errors because of this multi filed components structure. This was our solution to it and only recently I thought "Hey, shouldn't this be in the docs?", without realizing it might be fixed along the way. I tested it again with how the docs tell you to use it and it works*.. so that's my bad for not checking whether it did work now as apposed to back in November before making this PR.. 😔 So with that this PR can be ignored, sorry for the inconvenience and thanks for looking into it!

*_I do however still get an error when I use @nuxt/kit instead of @nuxt/kit-edge so I've submitted an issue for that 😄 https://github.com/nuxt/framework/issues/2683_

PR #2663, opened 2022-01-11
feat(nitro): Add Cache-Control and Etag  headers to html response for cache server

newbeea:
### 🔗 Linked issue

[Add Cache-Control and Etag headers for cache server]#2662

### ❓ Type of change

### 📚 Description

Describe your changes in detail 
I add etag header and cache-control header to html response in renderMiddleware

Why is this change required? What problem does it solve?
In ssr mode, the html response doesn't has etag header, so the cdn or browser cannot get 304 when the html is same.
This feature will reduce transporting of html file 

### 📝 Checklist

### DIFF
--- a/packages/nitro/src/runtime/app/render.ts
+++ b/packages/nitro/src/runtime/app/render.ts
@@ -121,10 +121,10 @@ export async function renderMiddleware (req, res: ServerResponse) {
     res.setHeader('Content-Type', 'text/html;charset=UTF-8')
     
     // Add Cache-Control and Etag headers for cache server
-    if (publicConfig.app.cacheControl) {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks so much for this PR @newbeea 💚  and sorry it took long to respond.

Since we are moving to nitropack (#3161) and implementing route caching rules, it would be easier to implement this feature and with wider usability for static and API routes if we support etag outside of ssr renderer but directly for all server routes.

Although the project is still private, I've made an issue in order to keep track, and rest assured neither feature or your credits for idea won't be discarded.

I'm closing this PR for maintenance. Thanks for your patience and understanding.

Issue #2074, opened 2021-11-22
Moving `<NuxtLayout>` to a built-in component


pi0:
(internally discussed within team meeting)

Allow using `NuxtLayout` directly in `app.vue`:

```vue
<!-- app.vue -->
<NuxtLayout>
  <NuxtPage/>
</NuxtLayout>
```

For transition support, NuxtPage needs to communicate with layout. For this we can make use of a global layout state:

```js
const layout = useLayout()
setLayout('welcome') // Only supported before rendering
// Same as useLayout().value = 'welcome'
```

The limitation is that `layout` can be only modified **before** or **after** rendering lifecycle (in plugins or client-side after mount)

pi0:
#3011 partially resolved this issue.

PR #2678, opened 2022-01-11
feat(nuxt3): add support for `definePageMeta` macro

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/2468

### ❓ Type of change

### 📚 Description

This PR is the first step in features enabled by 'extracting' metadata from components. I've included implementations of transition props and layouts via `definePageMeta` (though happy to extract the layouts change to another PR).

**NOTE**: at the moment HMR isn't working on route meta.

#### Transition props

It is now possible to define transition props which will be applied to the `<transition>` wrapping the page component (or pass false to disable it entirely):

https://github.com/nuxt/framework/blob/07df17137a22d9f100cc071b87d5d3479f1b0d96/packages/nuxt3/src/pages/runtime/page.vue#L4-L8

#### Layouts

This PR changes how layouts are defined. Previously they _had_ to be defined via a component option on the top-level page (ie. not in a child route). Now, they can be defined in a parent or nested route, via `definePageMeta({ layout })`. This also now accepts a computed property or a ref (for reactive layout changes within a page).

```vue
<template>
  <div>
    <button @click="enableCustomLayout">Update layout</button>
  </div>
</template>

<script setup>
const route = useRoute()
function enableCustomLayout () {
  // Note: because it's within a ref, it will persist if 
  // you navigate away and then back to the page.
  route.layout.value = "custom"
}
definePageMeta({
  layout: ref(false),
});
</script>
```

#### Middleware

I will be following up shortly with a middleware implementation.

#### Route key

I will be following up shortly with a route key implementation.

### 📝 Checklist


tobiasdiez:
I like it!

For the record, there is https://github.com/dword-design/nuxt-route-meta which provides similar features (also for nuxt2). Inspired by this library, maybe its worthwhile to also support the following syntax
```js
export default defineNuxtComponent({
   pageMeta: {
      layout: 'custom',
      auth: true
   }
} 
```
to set the page metadata.

What I'm missing for the docs is how this is used with typescript, i.e. how to define a custom PageMeta object to make sure that one only uses 'known' metadata objects. For this the current implementation
```ts
export interface PageMeta {
  [key: string]: any
   ...
}
```
might lead to problems. For example, if I define `CustomPageMeta extends PageMeta` then it still accepts arbitrary keys.

packages/nuxt3/src/pages/macros.ts
```diff
@@ -0,0 +1,94 @@
...
+  '[': ']',
+  '(': ')',
+  '<': '>',
+  '"': '"',
+  "'": "'"
+}
+
+function extractObject (code: string) {
+  // Strip comments
+  code = code.replace(/\/\/.*$/gm, '')
```
pi0, line 80: This can worngly remove strings:

```js
extractObject('{ a: "http://google.com" }') // '{ a: "http:'

File: packages/nuxt3/src/pages/macros.ts [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+      const scriptImport = findStaticImports(code).find(i => parseQuery(i.specifier.replace('?macro=true', '')).type === 'script')
+      if (scriptImport) {
+        const specifier = withQuery(scriptImport.specifier.replace('?macro=true', ''), { macro: 'true' })
+        return `export { meta } from "${specifier}"`
+      }
+
+      // [webpack] Export named exports rather than the default (component)
+      const defaultExport = code.match(/export \{ default \} from ['"]([^'"]+)['"]/)
+      if (defaultExport) {
+        return defaultExport[0].replace('{ default }', '*')
```
pi0, line 42: Maybe insert an additional explicitt export statement only for `options.macros` keys?

pi0:
In general Looks good to me 👍 But some notes:

- I think using AST to extract micro object seems to be a much safer choice than regex extraction. (how does vue extracts for built-in micros does it use regex too?)
- Considering we plan to extract Layout as a built-in component (https://github.com/nuxt/framework/issues/2074) does it still makes sense to define it in page meta micro?
   -  Can we make micro implementation more generic (`defineComponentMeta`) and extend with a special `definePageMeta` case?

pi0:
@tobiasdiez Good point! I might be wrong but I guess when extending interface with types, typescript checks it over generic `[key: string]: any`:

```ts
interface PageMeta {
  [key: string]: any
}

interface CustomPageMeta extends PageMeta {
  custom: number
}

// Type 'string' is not assignable to type 'number'.ts(2322)
const meta: CustomPageMeta = { custom: 'string' }
```

But it might be also good idea if we use something like `definePageMeta({ custom: {} })` for user-land config and more type safetely (also ensuring keys won't overlap if we introduce a new option that users already have custom use of!)

tobiasdiez:
I agree that the more narrow types overwrite the generic keys. What I meant is that in your example 
```
const meta: CustomPageMeta = { custom: 'string', someotherkey: 'which I don't want' }
```
doesn't lead to a typescript error. Thus one cannot easily say "only the following keys are allowed" (or at least I don't know how to exclude the `[key: string]: any` part from the type).

pi0:
If we include a sub-namespace of `custom: { }` top-level keys can be narrowed down to only known keys for typing without generic `string: any`.

packages/nuxt3/src/pages/macros.ts
danielroe: Nice catch. I'll amend to only match comments alone on a line, which is probably better than nothing. (We're guarding against the danger of object delimiters within a comment.) Your comments about AST parsing are fair though - these are always a little fragile.

danielroe:
A couple of comments - the macro plugin will allow us to define other macros, like `defineComponentMeta` or `definePageMeta` - or even something like `defineLayout` if we wished... The concept at the moment is that this metadata is set directly on `$route.meta`.

By doing this, we allow `vue-router` to handle the merging of metadata. But it also doesn't make sense to set route meta on components. So I'd suggest we can keep `defineComponentMeta` a separate macro for another use-case. It's easily enabled whenever we need it using this same plugin, though we'd likely want to extract it from the pages module at that point.

Another benefit is that we can declare types for these keys on `RouteMeta` which will be picked up by users using `$route.meta` throughout the app:

https://github.com/vuejs/vue-router-next/blob/c8430ea6c6033619041aadaa96527db6024ec66d/src/types/index.ts#L250-L266

Personally, I think the user DX of accessing something from e.g. `route.meta.title` is better than `route.meta.custom.title` and also avoids some edge-cases in terms of how `vue-router` merges the meta data. That is, the `custom` object on the child would override the entire `custom` object from the parent, meaning we don't get the benefit of this feature - so setting a `titleTemplate` on a parent route wouldn't make it into the child.

I'm happy to remove the layout implementation 👍 However, I think we will still need something it even with `<NuxtLayout>` extracted from `<NuxtPage>` if we want layout values to be possibly defined per-route. Probably needs some engagement with the issue so we can make a decision on that before implementing...

packages/nuxt3/src/pages/macros.ts
```diff
@@ -0,0 +1,94 @@
...
+      // [webpack] Re-export any imports from script blocks in the components
+      // with workaround for vue-loader bug: https://github.com/vuejs/vue-loader/pull/1911
+      const scriptImport = findStaticImports(code).find(i => parseQuery(i.specifier.replace('?macro=true', '')).type === 'script')
+      if (scriptImport) {
+        const specifier = withQuery(scriptImport.specifier.replace('?macro=true', ''), { macro: 'true' })
+        return `export { meta } from "${specifier}"`
+      }
+
+      // [webpack] Export named exports rather than the default (component)
+      const defaultExport = code.match(/export \{ default \} from ['"]([^'"]+)['"]/)
```
pi0, line 40: Can we reuse `fineExports` used in L46 for finding default export?

packages/nuxt3/src/pages/macros.ts
```diff
@@ -0,0 +1,94 @@
...
+        if (match) {
+          code = code.replace(match, `/*#__PURE__*/ false && ${match}`)
+        }
+      }
+
+      if (!parseQuery(search).macro) { return code }
+
+      // [webpack] Re-export any imports from script blocks in the components
+      // with workaround for vue-loader bug: https://github.com/vuejs/vue-loader/pull/1911
+      const scriptImport = findStaticImports(code).find(i => parseQuery(i.specifier.replace('?macro=true', '')).type === 'script')
```
pi0, line 33: Is removing `?micro=true` necessary if we only care about parsed type check?

packages/nuxt3/src/pages/runtime/composables.ts
```diff
@@ -8,3 +9,29 @@ export const useRouter = () => {
...
+const warnRuntimeUsage = (method: string) =>
+  console.warn(
+    `${method}() is a compiler-hint helper that is only usable inside ` +
+      '<script setup> of a single file component. Its arguments should be ' +
+      'compiled away and passing it at runtime has no effect.'
+  )
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const definePageMeta = (meta: PageMeta): void => {
+  if (!process.dev) {
```
pi0, line 34: I guess you meant `if (process.dev)` to eliminate from prod bundle?

```suggestion
  if (process.dev) {
```

packages/nuxt3/src/pages/macros.ts
danielroe: It doesn't match this syntax, alas. (A todo.)

packages/nuxt3/src/pages/runtime/composables.ts
danielroe: This is following the pattern of Vue's macros - warning only in production.

packages/nuxt3/src/pages/macros.ts
danielroe: It is working around a vue-loader bug...

packages/nuxt3/src/pages/macros.ts
danielroe: Actually, I'm wrong. `mlly` _does_ identify code but I want to distinguish a re-export rather than just an `export default defineComponent`.

pi0: Maybe extend EXPORT_NAMED_RE in mlly to match from segment? (thinking of this to avoid multiple regex match calls also keep it well tested in one place)

danielroe: https://github.com/unjs/mlly/pull/30

pi0 (approved):

packages/nuxt3/src/pages/utils.ts
```diff
@@ -225,10 +225,18 @@ export async function resolveLayouts (nuxt: Nuxt) {
...
-export function addComponentToRoutes (routes: NuxtRoute[]) {
-  return routes.map(route => ({
-    ...route,
-    children: route.children ? addComponentToRoutes(route.children) : [],
-    component: `{() => import('${route.file}')}`
-  }))
+export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> = new Set()): { imports: Set<string>, routes: NuxtPage[]} {
+  return {
+    imports: metaImports,
+    routes: routes.map((route) => {
```
richardeschloss, line 231: This can cause issues if people have components that don't export meta. Specifically, people will run into "The requested module ...' does not provide an export named 'meta'. That means every component will need to be updated to `export const meta = {}` at the least, which might not be great DX.

danielroe: We provide a default meta export (equalling undefined) via the transform plugin.

richardeschloss: I think this PR should probably be revisited, as it introduced unintended effects, such as this one and failing to honor the `layouts` property in the component. That layouts prop now needs to be moved into the meta section. If that's the design, then the docs should be updated to alert people to the breaking changes.

danielroe: If you are experiencing a problem, would you be able to raise an issue?

richardeschloss: Specifically, if I use the `extendPages` method, even if I specify meta to `{}`, I encounter this error:

> ""The requested module ...' does not provide an export named 'meta'"
 
Can you please start integrating the automated tests into the list of checks performed in PRs?

PR #2737, opened 2022-01-15
fix(kit): don't ignore `@` alias

### LINKED ISSUE #2677: @ alias does not work on Nuxt 3 + Components
- Operating System: `Darwin`
- Node Version:     `v14.18.0`
- Nuxt Version:     `3.0.0-27356801.e9128f3`
- Package Manager:  `npm@8.1.1`
- Bundler:          `Vite`
- User Config:      `css`, `components`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

`npx nuxi init project`

`cd project && npm i`

create:

`assets/css/index.css`
`components/atoms/Button.vue`

in `nuxt.config.ts` add:

```ts

  css: ['@/assets/css/index.css'], // @ works fine
  components: ['@/components/atoms'] // @ does not work
})
```

>  WARN  Components directory not found: @/components/atoms

and I split my components into folders, `atoms`, `molecules`, `organisms`, `templates`

in nuxt 2 I just need to do:

```ts
components: [{ path: '@/components', pathPrefix: false }],
```

in nuxt 3 i receiver this error: 

>  WARN  Components directory not found: @/components

already tried:

```ts
components: [{ path: '~/components', pathPrefix: false }]
```

> [Vue warn]: Failed to resolve component: Button

It only works with the verbose code below:

```ts
components: [
  '~/components/atoms',
  '~/components/molecules',
  '~/components/organisms',
  '~/components/templates'
]
```

### Describe the bug

@ alias does not work on Nuxt 3 + Components

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2677

### ❓ Type of change

### 📚 Description

We're currently completely ignoring the `@` alias to avoid resolving `@foo/bar` incorrectly. This ensures we'd don't skip the `@` alias in other situations.

**Aside**:

It might be worth considering _only_ resolving aliases with a directory separator afterwards as I imagine there could be some wrong resolutions from the current behaviour (e.g. consider aliases `~` and `~someotheralias`). This would of course be a breaking change.

Alternatively, perhaps we could resolve in reverse length order?

### 📝 Checklist

### DIFF
--- a/packages/kit/src/resolve.ts
+++ b/packages/kit/src/resolve.ts
@@ -85,9 +85,9 @@ function existsSyncSensitive (path: string, files?: string[]) {
  */
 export function resolveAlias (path: string, alias: ResolveOptions['alias']) {
   for (const key in alias) {
-    if (key === '@') { continue } // Don't resolve @foo/bar
+    if (key === '@' && !path.startsWith('@/')) { continue } // Don't resolve @foo/bar
     if (path.startsWith(key)) {
-      path = alias[key] + path.substr(key.length)
+      path = alias[key] + path.slice(key.length)
     }
   }
   return path


pi0:
Aliasing with `@` is dangerous by nature and even the linked issue is using `@/assets` in the reproduction so we are not exactly fixing that currently. What about checking directory existence for `@/` aliases for extra validation?

danielroe:
I may be misunderstanding, but I think this does resolve the issue with the linked repro. This PR continues to ignore `@` - the only change is if it is immediately followed by `/` (e.g. `@/assets`).

PR #2742, opened 2022-01-15
feat(nitro): ssl nitro support in production

### LINKED ISSUE #1701: production Node.js server with HTTPS
Currently, the default target [node server](https://v3.nuxtjs.org/docs/deployment/presets/server) generates a standalone node HTTP server. ([code](https://github.com/nuxt/framework/blob/main/packages/nitro/src/runtime/entries/server.ts))

While in %99 of cases we should always run this sever behind a reverse proxy like nginx or Cloudflare that terminates SSL, there might be cases it is handy to directly terminate on node handler.

For this, we might introduce `NITRO_SSL_CERT` and `NITRO_SSL_KEY` environment variables to be picked on runtime.

An alternative would be supporting `options.nitro.node.ssl` but is probably a dangerous operation since this way we are adding sensitive data to the production bundle and if image or CI artifcat leaks, attackers can use to do MITM.

danielroe:
### 🔗 Linked issue

resolves #1701

### ❓ Type of change

### 📚 Description

This adds support for initialising an HTTPS server (instead of HTTPS) if `NITRO_SSL_CERT` and `NITRO_SSL_KEY` are set. The provisos in issue #1701 stand.

If we're happy with this in general terms I'll go ahead and add documentation for these env variables to the docs on the server preset.

To test:

```bash
openssl genpkey -out nitro.key -algorithm RSA -pkeyopt rsa_keygen_bits:2048
openssl req -text -in nitro.csr -noout
openssl x509 -req -days 365 -in nitro.csr -signkey nitro.key -out nitro.crt
yarn nuxi build
NITRO_SSL_CERT=$(cat nitro.crt) NITRO_SSL_KEY=$(cat nitro.key) node .output/server/index.mjs
```

### 📝 Checklist

### DIFF
--- a/packages/nitro/src/runtime/entries/server.ts
+++ b/packages/nitro/src/runtime/entries/server.ts
@@ -1,9 +1,13 @@
 import '#polyfill'
-import { Server } from 'http'
+import { Server as HttpServer } from 'http'
+import { Server as HttpsServer } from 'https'
 import destr from 'destr'
 import { handle } from '../server'
 
-const server = new Server(handle)
+const cert = process.env.NITRO_SSL_CERT
+const key = process.env.NITRO_SSL_KEY
+
+const server = cert && key ? new HttpsServer({ key, cert }, handle) : new HttpServer(handle)
 
 const port = (destr(process.env.NUXT_PORT || process.env.PORT) || 3000) as number
 const hostname = process.env.NUXT_HOST || process.env.HOST || 'localhost'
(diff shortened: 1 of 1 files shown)


pi0:
> If we're happy with this in general terms I'll go ahead and add documentation for these env variables to the docs on the server preset.

Documentation would be nice addition if we first suggest to use reverse proxies and add notes about securely keeping cert files.

PR #2751, opened 2022-01-17
docs(examples): prop "showTips" should be a boolean instead of a string

bperel:
### ❓ Type of change

### 📚 Description

Solves `Invalid prop: type check failed for prop "showTips". Expected Boolean, got String with value "true".`

### 📝 Checklist

### DIFF
--- a/examples/use-async-data/app.vue
+++ b/examples/use-async-data/app.vue
@@ -3,7 +3,7 @@ const { data, refresh, pending } = await useAsyncData('/api/hello', () => $fetch
 </script>
 
 <template>
-  <NuxtExampleLayout example="use-async-data" show-tips="true">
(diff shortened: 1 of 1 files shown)


examples/use-async-data/app.vue
```diff
@@ -3,7 +3,7 @@ const { data, refresh, pending } = await useAsyncData('/api/hello', () => $fetch
 </script>
 
 <template>
-  <NuxtExampleLayout example="use-async-data" show-tips="true">
+  <NuxtExampleLayout example="use-async-data" :show-tips="true">
```
danielroe, line 6: Great catch. Or perhaps for simplicity's sake?

```suggestion
  <NuxtExampleLayout example="use-async-data" show-tips>
```

danielroe (approved):
Thanks! ❤️

PR #2773, opened 2022-01-18
feat(nuxt3): allow disabling vue type shims

danielroe:
### 🔗 Linked issue

https://github.com/vuejs/create-vue/pull/26

### ❓ Type of change

### 📚 Description

This PR removes the default '*.vue'* shim so that Volar can provide better type inference, such as clicking through directly to the component definition rather than the shim. (It may still be necessary for users of other IDEs  to add the shim themselves.)

### 📝 Checklist


pi0:
I love this! But since take-over-mode needs special configuration, maybe it would be better idea to keep generating shim and instead provide a flag (untype `typescript` configuration) to opt-in and recommend in typescript section to do?

PR #2774, opened 2022-01-18
fix(nuxt3): import top-level files and subdirectory indices

### LINKED ISSUE #2531: Composables dir, scanned only top files level
### Environment

- Operating System: `Darwin`
- Node Version:     `v17.2.0`
- Nuxt Version:     `3.0.0`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Create files in **composables** dir with few levels, for example: 

> composables

>>admin

>>>useFetch.js

>>cora.js

>>useAdmin.js

### Describe the bug

Files which deep then top level not autoloaded

[Screen](http://dl4.joxi.net/drive/2021/12/24/0016/1034/1111050/50/d9ce8ab2f5.png)

Problem in **globby**, i guess

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2531

### ❓ Type of change

### 📚 Description

This PR proposes a change to how auto registration works for composables & plugins. Rather than detecting only top level composables (as current) or all files in infinite recursion (as plugins), we adopt a new pattern that we register all top level files, and then only indices for any subdirectories.

For example:

```
plugins
 | - myPlugin.ts
 | - myOtherPlugin
 | --- supportingFile.ts
 | --- componentToRegister.vue
 | --- index.ts
```

Only `myPlugin.ts` and `myOtherPlugin/index.ts` would be registered. This would allow for better code organisation within the plugins directory.

Similarly, for composables:

```
composables
 | - useFoo.ts
 | - useBar
 | --- supportingFile.ts
 | --- index.ts
```

Only `useFoo.ts` and `useBar/index.ts` would be searched for imports - and if the latter is a default export, it would be registered as `useBar` rather than `index`.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/auto-imports/composables.ts
+++ b/packages/nuxt3/src/auto-imports/composables.ts
@@ -1,5 +1,5 @@
 import { promises as fsp, existsSync } from 'fs'
-import { parse as parsePath, join } from 'pathe'
+import { parse as parsePath } from 'pathe'
 import { findExports } from 'mlly'
 import { camelCase } from 'scule'
 import { AutoImport } from '@nuxt/schema'
--- a/packages/nuxt3/src/auto-imports/composables.ts
+++ b/packages/nuxt3/src/auto-imports/composables.ts
@@ -1,37 +1,42 @@
 import { promises as fsp, existsSync } from 'fs'
 import { parse as parsePath, join } from 'pathe'
-import { globby } from 'globby'
 import { findExports } from 'mlly'
 import { camelCase } from 'scule'
 import { AutoImport } from '@nuxt/schema'
+import { resolveFiles } from '@nuxt/kit'
 import { filterInPlace } from './utils'
 
 export async function scanForComposables (dir: string, autoImports: AutoImport[]) {
   if (!existsSync(dir)) { return }
 
-  const files = await globby(['*.{ts,js,tsx,jsx,mjs,cjs,mts,cts}'], { cwd: dir })
+  const files = await resolveFiles(dir, [
+    '*.{ts,js,mjs,cjs,mts,cts}',
+    '*/index.{ts,js,mjs,cjs,mts,cts}'
+  ])
 
   await Promise.all(
-    files.map(async (file) => {
-      const importPath = join(dir, file)
-
+    files.map(async (path) => {
       // Remove original entries from the same import (for build watcher)
-      filterInPlace(autoImports, i => i.from !== importPath)
+      filterInPlace(autoImports, i => i.from !== path)
 
-      const code = await fsp.readFile(join(dir, file), 'utf-8')
+      const code = await fsp.readFile(path, 'utf-8')
       const exports = findExports(code)
       const defaultExport = exports.find(i => i.type === 'default')
 
       if (defaultExport) {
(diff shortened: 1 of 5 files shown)


packages/nuxt3/src/auto-imports/composables.ts
```diff
@@ -1,37 +1,42 @@
 import { promises as fsp, existsSync } from 'fs'
 import { parse as parsePath, join } from 'pathe'
```
danielroe, line 2:
```suggestion
import { parse as parsePath } from 'pathe'
```

packages/nuxt3/src/auto-imports/composables.ts
```diff
@@ -1,37 +1,42 @@
...
+import { resolveFiles } from '@nuxt/kit'
 import { filterInPlace } from './utils'
 
 export async function scanForComposables (dir: string, autoImports: AutoImport[]) {
   if (!existsSync(dir)) { return }
 
-  const files = await globby(['*.{ts,js,tsx,jsx,mjs,cjs,mts,cts}'], { cwd: dir })
+  const files = await resolveFiles(dir, [
+    '*.{ts,js,mjs,cjs,mts,cts}',
+    '*/index.{ts,js,mjs,cjs,mts,cts}'
```
pi0, line 14: This seems inconsistency to me if we are skipping any nested directory but only make exception for `dir/index` (also considering ESM has no implicit `/index` resolution support)

peteromano:
This PR just broke literally everything I built off Nuxt3 in the past 2 months.

peteromano:
Why is this even a feature? All other Nuxt directories do not behave this way...

misaon:
> This PR just broke literally everything I built off Nuxt3 in the past 2 months.

You can solve your problem via custom Nuxt module and addPlugin method :)

peteromano:
@misaon Yea, sure... We need to write more code on top of nuxt to do the things we want, until we just dont need nuxt anymore.

If we continued that way, we would just compete with nuxt -- that is why we're so heated about this change (and that's where this is headed.)

peteromano:
It's a bad departure, plain and simple. EVERYTHING in nuxt dirs uses **/*. This PR just undermines that pattern for.. what reason? Developers can't import from other files in different directories (which is how basic node works)?

peteromano:
This is how the pattern should be:

/plugins/myPlugin.ts

Have a `lib` folder or whatever that has all the other arbitrary imports that myPlugin needs. This WASNT an issue.

peteromano:
Look, here's my auth plugin (which no longer works now, because this is actually registered as a symlink as /plugins/packages/myPackage/auth.ts)

```
import { defineNuxtPlugin } from '#app'
import useAuth, { init, setAuthenticated, setUser, setToken } from '@myPackage/composables/auth'
import { getAccessToken, getClaims } from '@myPackage/lib/auth.ts'
  ^
  ^
  ^ WHAT IS HARD/WRONG ABOUT DOING THIS?

export default defineNuxtPlugin(async ({ ssrContext }) => {
```

So, this PR breaks an existing nuxt dir pattern of **/*, AND it SOLVES A PROBLEM THAT WASNT A PROBLEM.

PR #2803, opened 2022-01-19
fix(nuxt3): scan all files in plugins/composables directories

### LINKED ISSUE #2780: nested plugins are no longer auto imported
### Environment

Nuxt CLI v3.0.0-27369360.33ebb01                                                                                                                                                                                  

- Operating System: `Linux`
- Node Version:     `v14.17.6`
- Nuxt Version:     `3.0.0-27369360.33ebb01`
- Package Manager:  `unknown`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

^ This is NOT the breaking version; one of the latest versions between this version and latest is breaking.

### Reproduction

No idea.

### Describe the bug

I have 0 leads as to what is wrong... I have a companion Storybook app where urql and auth is all working correctly. After a fresh npm install today, Nuxt COMPLETELY blows up:

Can't find $urql, function on my $auth are undefined (these errors come from SSR-side), the layout doesn't appear to be working (my header component doesnt load and NuxtWelcome doesnt load, but the slot/page renders), and 0 client errors.

Rolling back to v3.0.0-27369360.33ebb01 and it works again..

Did somebody break layouts or providers?

IMO: **THIS NEEDS TO BE INVESTIGATED AND FIXED ASAP.** Literally all of Nuxt is broken.

Looking through PRs, but triaging is virtually impractical since tons of PRs were merged yesterday.

**UPDATE:** Real bug: `$auth` provider is now `undefined` after latest releases, so provider registration doesn't work anymore.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2780
partially reverts https://github.com/nuxt/framework/pull/2774

### ❓ Type of change

### 📚 Description

On reflection, I think we can better achieve the aim of #2774 with adding back ignore support: see https://github.com/nuxt/framework/issues/2163#issuecomment-981612377

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/2.directory-structure/10.plugins.md
+++ b/docs/content/3.docs/2.directory-structure/10.plugins.md
@@ -12,23 +12,6 @@ Nuxt will automatically read the files in your `plugins` directory and load them
 All plugins in your `plugins/` directory are auto-registered, so you should not add them to your `nuxt.config` separately.
 ::
 
-## Which files are registered
-
-Only files at the top level of the `plugins/` directory (or index files within any subdirectories) will be registered as plugins.
-
-For example:
-
-```bash
-plugins
- | - myPlugin.ts
- | - myOtherPlugin
- | --- supportingFile.ts
- | --- componentToRegister.vue
- | --- index.ts
-```
-
-Only `myPlugin.ts` and `myOtherPlugin/index.ts` would be registered.
-
 ## Creating plugins
 
 The only argument passed to a plugin is [`nuxtApp`](/docs/usage/nuxt-app).
--- a/docs/content/3.docs/2.directory-structure/5.composables.md
+++ b/docs/content/3.docs/2.directory-structure/5.composables.md
@@ -8,22 +8,6 @@ head.title: Composables directory
 
 Nuxt 3 supports `composables/` directory to automatically import your Vue composables into your application using auto-imports!
 
-## How files are scanned
-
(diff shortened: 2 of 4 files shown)


peteromano:
@pi0 Can review?

pi0:
We probably won't revert this change but provide improved documentation for (not conventional) usage with nested plugins directory.

peteromano:
@pi0 Really? How would nested plugin levels work without **/* then? Is it possible to revert until the community agrees with this officially?

I still maintain that:

If you support arbitrary files inside plugins directory, then why not arbitrary files in components directory? Or server/api directory? Forgetting my specific case for a moment, this approach is at the least inconsistent with nuxt itself and basically undermines the whole idea of having designated directories in the first place..

Why is/was there not even a community conversation of RFC about this, before having merged a major change?

Can you at the very least explain to me why my opinion isnt suitable for nuxt, vs the justification for allowing arbitrary files in these directories? 

Also, if you have a workaround for what i had working, can you please just share that here then?

peteromano:
@pi0 Just saw your comment in the other thread btw

danielroe:
We'll pause on merging this for the moment. Here's a link to the workaround, for anyone else who encounters this: https://github.com/nuxt/framework/issues/2780#issuecomment-1016969323

PR #2819, opened 2022-01-19
docs: update documentation for configs

P4sca1:
### ❓ Type of change

### 📚 Description

It was not clear to me, how to change the runtime config via env variables at runtime.
I also added the composition API methods for accessing the config.

I don't know if the same replacing logic applies to `privateRuntimeConfig`. If that's the case, it would be good to add an explanation there too. Also, it might be good to mention that current env values are compiled into the script during build time when using `apiKey: process.env.API_KEY` and are not overwritten during runtime when setting `API_KEY`. Instead one has to set `apiKey`.

### 📝 Checklist


danielroe:
Many thanks for this! The logic applies to the privateRuntimeConfig as well. Here's the source:

https://github.com/nuxt/framework/blob/d07d572263b45108e2a98a9924bf0d8dcba902fa/packages/nitro/src/runtime/app/config.ts#L8-L12

P4sca1:
@danielroe I have updated the information about `privateRuntimeConfig` and updated the example.
Also, I decided to remove the information about the Nuxt payload, because I feel like it leads to more confusion and the new information about runtime replacement is sufficient.

By the way, thank you for your great talk at Vue.js Nation yesterday, I really enjoyed it. Keep up the good work 👏

danielroe (commented):
only a small typo - otherwise looks good to go 🚀

danielroe (approved):

pi0:
Thanks!

PR #2855, opened 2022-01-21
chore(examples): add `plugins` example

smarroufin:
### 🔗 Linked issue

None

### ❓ Type of change

### 📚 Description

Adds an example about plugins, showing the basic implementation of a plugin and how it is auto-imported.

### 📝 Checklist

### DIFF
--- a/examples/with-plugins/app.vue
+++ b/examples/with-plugins/app.vue
@@ -1,0 +1,5 @@
+<template>
+  <NuxtExampleLayout example="with-plugins" show-tips>
+    <div>{{ $myPlugin.data }}</div>
+  </NuxtExampleLayout>
+</template>
--- a/examples/with-plugins/app.vue
+++ b/examples/with-plugins/app.vue
@@ -1,5 +1,5 @@
 <template>
   <NuxtExampleLayout example="with-plugins" show-tips>
-    <div>{{ $myPlugin.data }}</div>
+    <div>{{ $myPlugin() }}</div>
   </NuxtExampleLayout>
 </template>
--- a/examples/with-plugins/plugins/my-plugin.ts
+++ b/examples/with-plugins/plugins/my-plugin.ts
@@ -1,9 +1,7 @@
-import { defineNuxtPlugin } from '#imports'
-
 export default defineNuxtPlugin(() => {
   return {
     provide: {
-      myPlugin: { data: 'Data coming from my auto-imported plugin' }
+      myPlugin: () => 'String generated from my auto-imported plugin!'
     }
   }
 })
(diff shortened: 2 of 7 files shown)


examples/with-plugins/plugins/my-plugin.ts
```diff
@@ -0,0 +1,8 @@
+import { defineNuxtPlugin } from '#imports'
+
+export default defineNuxtPlugin((nuxtApp) => {
+  nuxtApp.provide(
+    'myPlugin',
+    { data: 'Data coming from my auto-imported plugin' }
+  )
```
danielroe, line 7:
```suggestion
  return {
    provide: {
      myPlugin: { data: 'Data coming from my auto-imported plugin' }
    }
  }
```

examples/with-plugins/plugins/my-plugin.ts
```diff
@@ -0,0 +1,9 @@
+import { defineNuxtPlugin } from '#imports'
+
+export default defineNuxtPlugin(() => {
+  return {
+    provide: {
+      myPlugin: { data: 'Data coming from my auto-imported plugin' }
```
pi0, line 6: Maybe use a more realistic example for `provide` usage like inject a counter function?

examples/with-plugins/plugins/my-plugin.ts
```diff
@@ -0,0 +1,9 @@
+import { defineNuxtPlugin } from '#imports'
```
pi0, line 1: This is not required. We might comment it out for readability or remove.

```suggestion
// import { defineNuxtPlugin } from '#imports'
```

examples/with-plugins/plugins/my-plugin.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export default defineNuxtPlugin(() => {
```
pi0, line 3: Main point of using plugin entry points is availability of `nuxtApp` context:

```suggestion
export default defineNuxtPlugin((nuxtApp) => {
```

examples/with-plugins/app.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+  <NuxtExampleLayout example="with-plugins" show-tips>
+    <div>{{ $myPlugin.data }}</div>
```
pi0, line 3:
```suggestion
    <div>{{ $myPlugin() }}</div>
```

examples/with-plugins/plugins/my-plugin.ts
smarroufin: I can't add it and not use it since the CI Lint fails [1fc3403](https://github.com/nuxt/framework/pull/2855/commits/1fc3403da496e4ae7d403418d0cc8794826cfa3a)

smarroufin: I could eventually write something like:
```js
export default defineNuxtPlugin((/* nuxtApp */) => {
```

pi0:
Thanks!

Issue #2309, opened 2021-12-07
Changing child routes will rerender parent


thomasowow:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.13.0`
- Nuxt Version:     `3.0.0-27304790.cad09fe`
- Package Manager:  `yarn@1.19.2`
- Bundler:          `Vite`
- User Config:      `build`, `components`, `css`, `publicRuntimeConfig`, `srcDir`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Watch console logging setup whenever the route is changing to child route and back.

https://codesandbox.io/s/relaxed-grothendieck-dvsjz

### Describe the bug

As seen in the reproduction the parent of child routes is rendered every time the route changes. This could lead to a lot of unnecessary overhead in the parent.

I suspect both [`Page.vue`](https://github.com/nuxt/framework/blob/694e95b2b436d23a492d207a894801f7e58b6cbb/packages/nuxt3/src/pages/runtime/page.vue) and [`Child.vue`](https://github.com/nuxt/framework/blob/694e95b2b436d23a492d207a894801f7e58b6cbb/packages/nuxt3/src/pages/runtime/child.vue) to be the cause. Both of them are applying a `:key="$route.path"` to its dynamic component. Removal of this would solve this problem immidiately (tested this).

What would have been the reason to apply this `:key="$route.path"`? Would a PR that removes this code be rejected?

Harkonnen13:
Ye i noticed this before. Thats wierd behavior, using same route.path key for every view. In normal case, it should replace only child component, not everything including parent. 
For example, if my parent view has menu with some local state (open/close), it will be lost.

Atinux:
The CodeSandBox does not work for me, I duplicated on StackBlitz: https://stackblitz.com/edit/nuxt-starter-7eeck7?file=pages%2Findex.vue

thomasowow:
> The CodeSandBox does not work for me, I duplicated on StackBlitz: https://stackblitz.com/edit/nuxt-starter-7eeck7?file=pages%2Findex.vue

Shouldn't we remove the `app.vue`?

Atinux:
Should be good

PR #2859, opened 2022-01-21
fix(nuxt3): allow customising page keys

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/2309

### ❓ Type of change

### 📚 Description

This PR removes the default keying of routes to the full path, meaning that parent routes won't rerender when child routes change. It also adds support for setting custom keys for child routes, either a string or a function that receives the current route.

The key can either be set in `definePageMeta` or via a prop for `<NuxtNestedPage />`.

Contrary to [my earlier RFC](https://github.com/nuxt/framework/discussions/2468), I now think it makes sense to set the page key via meta as that way we can implement logic in a key function that will handle both child routes and deeply nested child routes (see following note).

**Note**: nested routes which have children themselves [behaved differently in Nuxt 2](https://github.com/nuxt/nuxt.js/blob/a2495a68736f9af96cd25209a948beb1ca665d4d/packages/vue-app/template/components/nuxt.js#L47-L72). I can restore this behaviour if desired.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/pages/runtime/nested-page.vue
+++ b/packages/nuxt3/src/pages/runtime/nested-page.vue
@@ -1,11 +1,26 @@
 <template>
-  <RouterView v-slot="{ Component }">
-    <component :is="Component" :key="$route.path" />
+  <RouterView v-slot="{ Component, route }">
+    <component :is="Component" :key="getKeyFor(childKey || route.meta.key, route)" />
   </RouterView>
 </template>
 
-<script>
+<script lang="ts">
+import type { RouteLocationNormalizedLoaded } from 'vue-router'
+
+const getKeyFor = (key: string | ((route: RouteLocationNormalizedLoaded) => string), route?: RouteLocationNormalizedLoaded) => {
+  return key && typeof key === 'function' ? key(route) : key
+}
+
 export default {
-  name: 'NuxtNestedPage'
+  name: 'NuxtNestedPage',
+  props: {
+    childKey: {
+      type: [Function, String] as unknown as () => string | ((route: RouteLocationNormalizedLoaded) => string),
+      default: null
+    }
+  },
+  setup () {
+    return { getKeyFor }
+  }
 }
 </script>
--- a/packages/nuxt3/src/pages/runtime/nested-page.vue
+++ b/packages/nuxt3/src/pages/runtime/nested-page.vue
@@ -21,7 +21,7 @@ export default {
     const route = useRoute()
     const key = computed(() => {
       const source = props.childKey ?? route.meta.key
-      return source && typeof source === 'function' ? source(route) : source
+      return typeof source === 'function' ? source(route) : source
     })
     return {
       key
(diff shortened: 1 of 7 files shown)


packages/nuxt3/src/pages/runtime/nested-page.vue
```diff
@@ -1,11 +1,26 @@
...
+  <RouterView v-slot="{ Component, route }">
+    <component :is="Component" :key="getKeyFor(childKey || route.meta.key, route)" />
   </RouterView>
 </template>
 
-<script>
+<script lang="ts">
+import type { RouteLocationNormalizedLoaded } from 'vue-router'
+
+const getKeyFor = (key: string | ((route: RouteLocationNormalizedLoaded) => string), route?: RouteLocationNormalizedLoaded) => {
```
pi0, line 10: Code style: Can't we merge logic of L3 + `getKeyFor` as a computed prop in the setup?

packages/nuxt3/src/pages/runtime/page.vue
```diff
@@ -3,7 +3,7 @@
     <NuxtLayout v-if="Component" :name="layout || route.meta.layout">
       <NuxtTransition :options="route.meta.transition ?? { name: 'page', mode: 'out-in' }">
         <Suspense @pending="() => onSuspensePending(Component)" @resolve="() => onSuspenseResolved(Component)">
-          <component :is="Component" :key="route.path" />
+          <component :is="Component" />
```
pi0, line 6: Does it means (main) page components are not cached anymore in case of client-side revisit?

toooni, line 6: @danielroe I think there should be a value like: `:key="$route.meta.key"` here too. In our case we are changing the locale and this should result in a rerendering. An optional key would give us this possibility.

danielroe: would you raise a feature-request issue with an example of your use-case? :pray

toooni: I'll do that 👍

misaon: @danielroe when you removed `:key="route.path"` from page component, my page file
`./pages/[code].vue` is loaded only once, when i go from url eg. `/test-1` to `/test-2`, logic in file `./pages/[code].vue` not trigged, but router change url in adress bar. So page not redraw with fresh content from data fetch mothods. Its ok?

misaon: https://github.com/nuxt/framework/issues/2903

toooni: Yeah, it seems like this issue is bigger than my case. Having an optional `$route.meta.key` could be used to solve this issue. But with the routing strategy nuxt3 uses, there must be another/better solution which works out of the box.

misaon: You can explicitly define key on `<NuxtPage :key="route.path" />` but page transitions wont work.

PR #2904, opened 2022-01-25
feat(nuxt3): add support for `<KeepAlive>`

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/2426
https://github.com/nuxt/framework/discussions/2468

### ❓ Type of change

### 📚 Description

This adds support for `<KeepAlive>` and resolves an issue with layout transitions if there was no layout enabled. It also simplifies the logic in `<NuxtPage>` by moving it to a render function, which means we need fewer components to render the page.

**NOTE**: There still seem to be issues with KeepAlive not behaving as expected (ie. not caching component state properly). However, it does not throw an error, which means this can now be tested in Nuxt 3, and raised upstream, if appropriate, in the Vue repo.

### 📝 Checklist


packages/nuxt3/src/pages/runtime/page.ts
```diff
@@ -0,0 +1,51 @@
...
+    const nuxtApp = useNuxtApp()
+    const route = useRoute()
+
+    return () => {
+      // We avoid rendering layout transition if there is no layout to render
+      const hasLayout = props.layout ?? route.meta.layout ?? 'default' in layouts
+
+      return h(RouterView, {}, {
+        default: ({ Component }: RouterViewSlotProps) => Component && wrapIf(Transition, hasLayout && (route.meta.layoutTransition ?? defaultLayoutTransition), {
+          default: () => wrapIf(NuxtLayout, hasLayout && { layout: props.layout ?? route.meta.layout }, {
```
pi0, line 29: I love `wrapIf` idea! Maybe if it returns a render function by itself and assumes last arg is default slot we could both simplify code and avoid extra call? (it can directly return child)

pi0: (merging as hotfix lets refactor later)

danielroe: For later info - the issue to face is that KeepAlive doesn't take a default slot but a child directly.

PR #2937, opened 2022-01-26
fix(nuxt3): add `app:suspense:resolve` hooks

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/2938

### ❓ Type of change

### 📚 Description

Our asyncData implementation was dependent on a hook being called within the pages module. This adds a similar hook at `app` level meaning that `asyncData` will work as expected on client-side.

Ideas for better hook naming welcome.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/bootstrap.ts
+++ b/packages/nuxt3/src/app/bootstrap.ts
@@ -48,7 +48,7 @@ if (process.client) {
     await nuxt.hooks.callHook('app:created', vueApp)
     await nuxt.hooks.callHook('app:beforeMount', vueApp)
 
-    nuxt.hooks.hookOnce('page:finish', () => {
+    nuxt.hooks.hookOnce('app:finish', () => {
       nuxt.isHydrating = false
     })
 
--- a/packages/nuxt3/src/app/bootstrap.ts
+++ b/packages/nuxt3/src/app/bootstrap.ts
@@ -48,7 +48,7 @@ if (process.client) {
     await nuxt.hooks.callHook('app:created', vueApp)
     await nuxt.hooks.callHook('app:beforeMount', vueApp)
 
-    nuxt.hooks.hookOnce('app:finish', () => {
+    nuxt.hooks.hookOnce('app:suspense:resolve', () => {
       nuxt.isHydrating = false
     })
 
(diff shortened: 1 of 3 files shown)


packages/nuxt3/src/app/components/nuxt-root.vue
```diff
@@ -14,6 +14,10 @@ export default {
     if (process.dev && results && results.some(i => i && 'then' in i)) {
       console.error('[nuxt] Error in `vue:setup`. Callbacks must be synchronous.')
     }
+    return {
+      onPending: () => nuxtApp.callHook('app:start'),
```
pi0, line 18: Maybe prefix with `app:suspense:start` to make to more explicit?

danielroe: in a normal case, `app:start` will also not be called - so maybe just one hook called `app:resolve`?

pi0: What about `app:suspense:pending`?

danielroe: I think it needs to have more of a final sound - e.g. resolve. It only gets called once in the nuxt app lifecycle.

packages/nuxt3/src/app/bootstrap.ts
```diff
@@ -48,7 +48,7 @@ if (process.client) {
     await nuxt.hooks.callHook('app:created', vueApp)
     await nuxt.hooks.callHook('app:beforeMount', vueApp)
 
-    nuxt.hooks.hookOnce('page:finish', () => {
+    nuxt.hooks.hookOnce('app:finish', () => {
```
pi0, line 51:
```suggestion
    nuxt.hooks.hookOnce('app:suspense:resolve', () => {
```

packages/nuxt3/src/app/components/nuxt-root.vue
pi0: Agreed it would be nice if we introduce `app:resolve` (or ready/done/hydrated?) hook that is called after app is resolved/ready but this one is technically being called on suspense pending event.

PR #2945, opened 2022-01-27
feat(nuxt3): middleware type generation improvements

danielroe:
### ❓ Type of change

### 📚 Description

This PR makes a number of small changes to improve middleware DX:
* Middleware types weren't being generated on creation of middleware files - this now includes the middleware directory in the regexp for regenerating app.
* We exclude global middleware from the named middleware object (and type completion) as if it's already running, it seems needless to specify it within a page...
* We now add dir.middleware to supported v3 types
* We escape values wherever we're creating RegExps in nuxt3 templates (for example edge cases, a user might pass a nested dir or include an invalid char in their dir name)

### 📝 Checklist


packages/nuxt3/src/utils.ts
```diff
@@ -0,0 +1 @@
+export const escapeRE = (str: string) => str.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&')
```
pi0, line 1: Would it be overkill to use https://www.npmjs.com/package/escape-string-regexp? Seems well-tested and simple enough (https://github.com/sindresorhus/escape-string-regexp/blob/main/index.js)

danielroe: Seems reasonable 👍

packages/nuxt3/src/pages/module.ts
```diff
@@ -84,12 +91,13 @@ export default defineNuxtModule({
       async getContents () {
         const middleware = await resolveMiddleware()
         await nuxt.callHook('pages:middleware:extend', middleware)
-        const middlewareObject = Object.fromEntries(middleware.map(mw => [mw.name, `{() => import('${mw.path}')}`]))
         const globalMiddleware = middleware.filter(mw => mw.global)
+        const namedMiddleware = middleware.filter(mw => !mw.global)
+        const namedMiddlewareObject = Object.fromEntries(namedMiddleware.map(mw => [mw.name, `{() => import('${mw.path}')}`]))
```
pi0, line 96: tip: we might use genDynamicImport in sanetize refactor

packages/nuxt3/src/pages/module.ts
```diff
@@ -118,6 +128,7 @@ export default defineNuxtModule({
       getContents: async () => {
         const composablesFile = resolve(runtimeDir, 'composables')
         const layouts = await resolveLayouts(nuxt)
+        await nuxt.callHook('pages:layouts:extend', layouts)
```
pi0, line 131: What about moving hook call to `resolveLayouts`?

packages/nuxt3/src/pages/module.ts
danielroe: I believe we do 😁

packages/nuxt3/src/pages/module.ts
danielroe: Good idea. And also I think there's space for improving this resolve behaviour. See here for an alternative approach: https://github.com/nuxt/framework/blob/11becef7355c9da668603a1e0129024d22a74ab1/packages/nuxt3/src/auto-imports/module.ts#L78-L90

pi0 (approved):

PR #2949, opened 2022-01-27
fix(nuxt3): hotfix for missing components declaration

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/2947

### ❓ Type of change

### 📚 Description

There have been a number of issues with lacking a reference to `components.d.ts` which is likely to do with some inconsistency between when these hooks are being called (ie. we may not have the components fully resolved by the time we're writing the type file). This PR always adds the file to `nuxt.d.ts`, as a hotfix.

### 📝 Checklist


pi0:
I think makes sense to generate types even if there are no components 👍

PR #2976, opened 2022-01-28
fix(nuxt3)!: key routes by interpolated path

### LINKED ISSUE #2903: allow customising top level page keys
### Environment

- Operating System: `Linux`
- Node Version:     `v17.4.0`
- Nuxt Version:     `3.0.0-27383920.81ee59c`
- Bundler:          `Vite`
- User Config:      `buildModules`
- Runtime Modules:  `-`
- Build Modules:    `normalizedModule()`, `normalizedModule()`

### Reproduction

Create simple page:
`pages/[code].vue`

Visit `http://localhost:3000/test-1`
Go to `http://localhost:3000/test-2` via NuxtLink component

### Describe the bug

After this MR https://github.com/nuxt/framework/pull/2859 and this [change](https://github.com/nuxt/framework/pull/2859/files/917809cef54b8980fe08757b60b455f7cb9e55f9#diff-20895eaf69b2b6f25c4fce199443ac54ad3c6a8fa70985168d50dc86172ef713L6), page not redraw between routes, but url in address bar is changed.

### Additional context

You can explicitly define key on `<NuxtPage :key="route.path" />` but page transitions wont work.
https://github.com/nuxt/framework/pull/2861

image

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2903

### ❓ Type of change

### 📚 Description

This PR attempts to normalise behaviour between `<NuxtPage>` and `<NuxtNestedPage>`.

1. Both now take a `pageKey` prop for overriding the key of the route component within them. This is a change from the `childKey` that `<NuxtNestedPage>` previously accepted. This allows complete customisation of the keying behaviour.

2. If a `pageKey` isn't provided, we will next go to the `key` from the page/route meta of the component being displayed within the current `<RouterView>`. This is a change from the previous behaviour, where a key would affect _every_ component matched to a given route.

3. Finally, we will default to the interpolated path of the component being displayed.

   For example, in a nested route where the parent is `~/pages/parent.vue` and the child is `~/pages/parent/[slug].vue`, upon visiting `/parent/child`, the key for the parent component will be `/parent` (and will not reload when navigating to `/parent/other-child`, etc). The key for the child component would be `/parent/child` and so would trigger a refresh when navigating to `/parent/other-child`, even though the same page component renders the route.

### 👉 Migration steps

1. Update any `child-key` props to `page-key` instead.
2. If you are relying on the previous default unkeyed behaviour of dynamic routes,  you can manually re-enable this:
   ```js
   definePageMeta({ key: false })
   ```
3. If you are using `<NuxtNestedPage>` or `<NuxtChildPage>` you should now replace these with `<NuxtPage` directly.

### 📝 Checklist


packages/nuxt3/src/pages/runtime/page.ts
```diff
@@ -14,30 +16,38 @@ export default defineComponent({
...
   },
   setup (props) {
     const nuxtApp = useNuxtApp()
-    const route = useRoute()
 
-    return () => {
-      // We avoid rendering layout transition if there is no layout to render
-      const hasLayout = props.layout ?? route.meta.layout ?? 'default' in layouts
+    return () => h(RouterView, {}, {
+      default: ({ Component, route }: RouterViewSlotProps) => {
```
pi0, line 29: Refactor: instead of destructuring props of router view, we could directly pass it as reference to getKey util

packages/nuxt3/src/pages/runtime/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { VNode } from 'vue'
+import { RouteLocationMatched, RouteLocationNormalizedLoaded } from 'vue-router'
+
+const interpolatePath = (route: RouteLocationNormalizedLoaded, match: RouteLocationMatched) => {
+  return match.path.replace(/:\w+/g, r => route.params[r.slice(1)].toString() || r)
+}
+
+export const getKey = (override: string | ((route: RouteLocationNormalizedLoaded) => string), route: RouteLocationNormalizedLoaded, Component: VNode) => {
```
pi0, line 8: Refactor: getRouteKey or another clear name

packages/nuxt3/src/pages/runtime/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+  const matchedRoute = route.matched.find(m => m.components.default === Component.type)
+  const source = override ?? matchedRoute?.meta.key ?? interpolatePath(route, matchedRoute)
```
pi0, line 10: We are generating this key purely based on route and not query params. Maye should have an option at least?

packages/nuxt3/src/pages/runtime/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { VNode } from 'vue'
+import { RouteLocationMatched, RouteLocationNormalizedLoaded } from 'vue-router'
+
+const interpolatePath = (route: RouteLocationNormalizedLoaded, match: RouteLocationMatched) => {
+  return match.path.replace(/:\w+/g, r => route.params[r.slice(1)].toString() || r)
```
pi0, line 5: Does it handle optional params?

misaon:
I can not wait! ❤️

dev-fredericfox:
@misaon Depending on your use-case, I believe you can already fix the issue if you add a key to your layout. At least it seemed to work for me since this merge (https://github.com/nuxt/framework/pull/3011)


```
//default.vue
<script setup lang="ts">
const route = useRoute();
</script>
<template>
  <div>
      <div :key="route.path"> //if you need the layout to re-render
      <TheHeader/>
      <NuxtPage :key="route.path" /> //if you need the page to re-render
      <TheFooter/>
      </div>
  </div>
</template>
```

misaon:
> @misaon Depending on your use-case, I believe you can already fix the issue if you add a key to your layout. At least it seemed to work for me since this merge (#3011)
> 
> ```
> //default.vue (layout)
> <script setup lang="ts">
> const route = useRoute();
> </script>
> <template>
>   <div>
>       <div :key="route.path"> //if you need the layout to re-render
>       <TheHeader/>
>       <NuxtPage :key="route.path" /> //if you need the page to re-render
>       <TheFooter/>
>       </div>
>   </div>
> </template>
> ```

@dev-fredericfox thank you very much for the tip, we already use this method, which you describe (see https://github.com/nuxt/framework/issues/2903 additional context). However, translations do not work naturally and we are still waiting for this PR to be resolved.

packages/nuxt3/src/pages/runtime/utils.ts
danielroe: Hm. What do you recommend? We previously didn't key routes by query in Nuxt 2, and [recommended watching query](https://nuxtjs.org/docs/features/data-fetching#listening-to-query-string-changes), e.g.:

```vue
<script setup>
const route = useRoute()
const { data, refresh } = useFetch(/** */)
watch(() => route.query, () => refresh())
</script>
```

packages/nuxt3/src/pages/runtime/page.ts
```diff
@@ -1,24 +1,28 @@
...
-type InstanceOf<T> = T extends new (...args: any[]) => infer R ? R : never
-type RouterViewSlotProps = Parameters<InstanceOf<typeof RouterView>['$slots']['default']>[0]
+import { generateRouteKey, RouterViewSlotProps, wrapIf, wrapInKeepAlive } from './utils'
+import { useNuxtApp } from '#app'
 
 export default defineComponent({
   name: 'NuxtPage',
-  setup () {
+  props: {
+    pageKey: {
```
pi0, line 10: Just to ensure, simply usage if `key` is not possible right?

packages/nuxt3/test/pages.test.ts
```diff
@@ -191,3 +192,72 @@ describe('pages:generateRoutesFromFiles', () => {
...
+  it('should key dynamic routes with array params', () => {
+    expect(generateRouteKey(undefined, getRouteProps({
+      path: '/:array+',
+      meta: {}
+    }))).to.equal('/a,b')
+    expect(generateRouteKey(undefined, getRouteProps({
+      path: '/test/:array*',
+      meta: {}
+    }))).to.equal('/test/a,b')
+    expect(generateRouteKey(undefined, getRouteProps({
```
pi0, line 258: Refactor: it might be more maintainable to use a generator pattern (an object/s containing test cases).

pi0 (approved):

packages/nuxt3/src/pages/runtime/page.ts
danielroe: Yes - key is a reserved name.

packages/nuxt3/test/pages.test.ts
danielroe: I personally prefer the `describe` pattern as I think it makes it more understandable to contributors to see the description, and avoids early abstraction of test cases. But if you want me to refactor before merge, I can do so.

pi0: We can combine two sections with descriptions btw. Testing at this low level will be already too much hassle in long term...

danielroe: https://i.imgur.com/DuSFRKv.gif

packages/nuxt3/src/pages/runtime/composables.ts
```diff
@@ -14,7 +14,7 @@ export interface PageMeta {
   [key: string]: any
   pageTransition?: boolean | TransitionProps
   layoutTransition?: boolean | TransitionProps
-  key?: string | ((route: RouteLocationNormalizedLoaded) => string)
+  key?: boolean | string | ((route: RouteLocationNormalizedLoaded) => string)
```
pi0, line 17: @danielroe Is this right? 

```suggestion
  key?: false | string | ((route: RouteLocationNormalizedLoaded) => string)
```

pi0 (approved):

PR #2980, opened 2022-01-28
docs(modules): fix npm script name

owlnai:
### ❓ Type of change

### 📚 Description
[The module docs ](https://v3.nuxtjs.org/docs/advanced/modules#quick-start) points you to use `npm run dev` to start the module playground, but according to the starter said npm script isn't defined and it seems to be `npm run play` instead.  See: https://github.com/nuxt/starter/blob/module/package.json#L19


pi0:
Thanks for noticing! I've updated the template to use `dev` script in order to make it consistent for `nuxi init` guides (https://github.com/nuxt/starter/commit/7c10e3207cda32b26fc40aa881c785e94b23a986)

PR #3007, opened 2022-01-31
feat(nitro)!: support overriding runtime config at a second level

danielroe:
### ❓ Type of change

### 📚 Description

Currently we can only override at the top level of runtime config. This now allows for overriding at a second (nested) level, for example, for modules which want to allow runtime configuration. At the same time, we also normalise to upper-snake-case (`baseURL` -> `BASE_URL`)

Consider this example:

```js
  publicRuntimeConfig: {
    // this previously required overriding with `baseURL=other node .output/server/index.mjs`
    // but is now normalised to `BASE_URL`
    baseURL: 'test',
    // this was previously impossible to override piece-meal
    // but can now be overriden with `MY_MODULE_SOME_KEY=false node .output/server/index.mjs`
    myModule: {
      someKey: true
    }
  }
})
```

#### Notes

1. Note that we have special handling for app config - perhaps we can abandon this and default to (e.g.) APP_CDN_URL instead of NUXT_APP_CDN_URL?

   https://github.com/nuxt/framework/blob/a971aab2728fb412f865140a8faa47542aa913ab/packages/nitro/src/runtime/app/config.ts#L24-L28

2. ~~If we're happy with this approach, I will update the documentation.~~

### 👉 Migration

* If you are depending on overriding with a **mixed-case** environment variable, you should convert it to upper-snake-case instead (`baseURL` -> `BASE_URL`).

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/1.usage/5.runtime-config.md
+++ b/docs/content/3.docs/1.usage/5.runtime-config.md
@@ -33,12 +33,16 @@ In addition to any process environment variables, if you have a `.env` file in y
 ```sh [.env]
 BASE_URL=https://nuxtjs.org
 API_SECRET=api_secret_token
+MY_MODULE_KEY=overridden_value
 ```
 
 ```ts [nuxt.config.ts]
 export default defineNuxtConfig({
   publicRuntimeConfig: {
-    BASE_URL: process.env.BASE_URL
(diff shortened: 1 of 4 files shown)


pi0:
(PR is merged to nitropack relesing soon)

PR #3011, opened 2022-01-31
refactor!: extract `<NuxtLayout>` from `<NuxtPage>`

### LINKED ISSUE #2074: Moving `<NuxtLayout>` to a built-in component

danielroe:
### 🔗 Linked issue

resolves #2074

### ❓ Type of change

### 📚 Description

This PR extracts `<NuxtLayout>` from `<NuxtPage>` while preserving its behaviour. (Though we do move the transition into the layout component.)

**Note**: This will mean once we have [aligned the behaviour of route keys between pages/nested pages](https://github.com/nuxt/framework/pull/2976), we can actually replace `<NuxtNestedPage>` with `<NuxtPage>` entirely, which will improve UX for transitions between nested pages.

### 👉 Migration

* Add `<NuxtLayout>` to `app.vue` if you are using layouts:
   ```diff
    <template>
   +  <NuxtLayout>
        <NuxtPage />
   +  </NuxtLayout>
    </template>
   ```

* If you were passing layout manually to `<NuxtPage>`, move this to `<NuxtLayout>` instead:
   ```diff
    <template>
   +  <NuxtLayout name="my-layout">
   -     <NuxtPage layout="my-layout" />
   +     <NuxtPage />
   +  </NuxtLayout>
    </template>
   ```

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/pages/runtime/layout.ts
+++ b/packages/nuxt3/src/pages/runtime/layout.ts
@@ -1,24 +1,33 @@
-import { defineComponent, h, Ref } from 'vue'
+import { defineComponent, isRef, Ref, Transition } from 'vue'
+import { useRoute } from 'vue-router'
+import { wrapIf } from './utils'
 // @ts-ignore
 import layouts from '#build/layouts'
+
+const defaultLayoutTransition = { name: 'layout', mode: 'out-in' }
 
 export default defineComponent({
   props: {
     name: {
       type: [String, Boolean, Object] as unknown as () => string | false | Ref<string | false>,
-      default: 'default'
+      default: null
     }
   },
   setup (props, context) {
+    const route = useRoute()
+
     return () => {
-      const layout = (props.name && typeof props.name === 'object' ? props.name.value : props.name) ?? 'default'
-      if (!layouts[layout]) {
-        if (process.dev && layout && layout !== 'default') {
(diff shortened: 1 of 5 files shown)


packages/nuxt3/src/pages/runtime/layout.ts
```diff
@@ -1,24 +1,33 @@
-import { defineComponent, h, Ref } from 'vue'
```
pi0, line 1: Shall we move the layout component to core app/runtime with same PR?

danielroe: Would you be okay with me doing it in a separate PR, because we'll also need to extract the useRoute interop?

pi0 (approved):

PR #3070, opened 2022-02-05
feat(nitro): add support for lambda v2 payload format

### LINKED ISSUE #2266: nitro/lambda:  ctx.req.headers should normalize for compat
### Environment

```
NITRO_PRESET=lambda
nuxt packages: v3.0.0-27286006.a08b435 
```

### Reproduction

TBD

### Describe the bug

When deploying to AWS Lambda using Nitro's lambda preset, the case of the request header name is mixed case.

According to [the documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html#:~:text=You%20should%20get%20a%20successful%20response%20with%20a%20payload%20similar%20to%20the%20following%3A), Lambda events using the API Gateway have a mix of lower case kebab case and upper case. However, as per the [node.js documentation](https://nodejs.org/api/http.html#http), `context.req.headers` is defined to be lowercase kebab case only. Currently, nitro does not have this canonicalization in place, so modules that assume headers are lowercase, such as `@nuxtjs/axios`, cannot be used properly.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #2535: Lambda - Layouts ignored - page elements merge
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0-27338323.1e98259`
- Bundler:          `Vite`
- User Config:      `components`, `buildModules`, `publicRuntimeConfig`, `css`, `build`
- Runtime Modules:  `-`
- Build Modules:    `@tailvue/nuxt@0.1.52`

### Reproduction

Repository 

https://github.com/fumeapp/nitro

Source of index page

https://github.com/fumeapp/nitro/blob/production/pages/index.vue

Source of test page with custom layout

https://github.com/fumeapp/nitro/blob/production/pages/test.vue

### Describe the bug

If you visit https://nitro.fume.farm/ you can see `pages/index.vue` is served with our default layout just fine

Then visit https://nitro.fume.farm/test and we should be in `pages/test.vue` but we are seeing our default layout - and some of the elements get merged - page is asking for `layouts/custom.vue`

Install the setup locally with `yarn dev` or `yarn build; yarn start` and the behavior is just fine otherwise.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2535, resolves #2266

### ❓ Type of change

### 📚 Description

AWS Lambda defaults new projects/deploys to v2: see [docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html). This adds support for the new version.

In addition, we normalise Array request and response headers to/from the format AWS is expecting.

### 📝 Checklist

- [x] update tests


acidjazz:
@pi0 👀 👀 👀 👀

packages/nitro/src/runtime/entries/lambda.ts
```diff
@@ -1,21 +1,35 @@
...
+  let url: string
+  let method: string
+
+  if ('resource' in event) {
+    // APIGatewayProxyEvent
+    url = withQuery(event.path, event.queryStringParameters)
+    method = event.httpMethod
+  } else {
+    // APIGatewayProxyEventV2
+    url = withQuery(event.rawPath, event.queryStringParameters)
```
pi0, line 16: Is it possible to simplify with a fallback approach? This can also reduce the chance of incompatibility with nonstandard implementations without `resource` in event.

```js
    url = withQuery(event.path || event.rawPath, event.queryStringParameters)
    method = event.httpMethod || event.requestContext.http.method
```

packages/nitro/src/runtime/entries/lambda.ts
```diff
@@ -1,21 +1,35 @@
...
+  }
+
   const r = await localCall({
     event,
-    url: withQuery(event.path, event.queryStringParameters),
+    url,
     context,
-    headers: event.headers,
-    method: event.httpMethod,
+    headers: Object.fromEntries(Object.entries(event.headers).map(([key, value]) => [key.toLowerCase(), value])),
```
pi0, line 24: We can refactor to `normalizeHeaders`

packages/nitro/src/runtime/entries/lambda.ts
danielroe: Mostly we'd lose the benefit of the `aws-lambda` types if we want to do that. But we can check `'path' in event` instead.

pi0: `url` and `method` seem directly typed as strings above. Do you mean typing for `event` ?

danielroe: Exactly. By type-narrowing as we have done we keep type defs. Otherwise we lose the benefit of them.

packages/nitro/src/runtime/entries/lambda.ts
danielroe: I've extracted the functions. I considered whether to move to unenv but I think these are specific to AWS headers.

packages/nitro/src/runtime/entries/lambda.ts
pi0: I see but seems we are not relying on `event` type either externally based on inference (nor is possible with transpiled runtime) also for the rest of logic...

danielroe: That's because the properties we're currently using are the same in both payload types. Using the type gives us that information, but if in future we add another property then it will warn us. For example, that was what warned me about the outgoing headers issue. If we type it as any we lose that useful info.

I'm not sure what the issue is that you have with type narrowing as I've done: `if ('path' in event')`. Would you explain more?

pi0: There is no point in doing that

danielroe: It preserves the typing. We lose the benefit of different event types by changing the union (`|`) to intersection (`&`). Although it might not matter now, we run the risk of bugs further down the line, for example if someone makes an edit to this file later.

PR #3096, opened 2022-02-07
feat(nuxt3,bridge): add automatic schema augmentation declaration

danielroe:
### ❓ Type of change

### 📚 Description

This adds a new `schema.d.ts` declaration file for augmenting the schema. With this change, modules do not need to declare their own type augments.

### 📝 Checklist


packages/bridge/src/app.ts
```diff
@@ -38,6 +39,12 @@ export function setupAppBridge (_options: any) {
     references.push({ path: resolve(nuxt.options.buildDir, 'types/components.d.ts') })
   })
 
+  // Augment schema with module types
+  addTemplate({
+    ...schemaTemplate,
+    src: ''
```
pi0, line 45: (is this typing issue? Would be nice to be fixed at some point :D)

packages/nuxt3/src/core/nuxt.ts
```diff
@@ -82,9 +84,9 @@ async function initNuxt (nuxt: Nuxt) {
 
   for (const m of modulesToInstall) {
     if (Array.isArray(m)) {
-      await installModule(m[0], m[1], nuxt)
+      await installModule(m[0], m[1])
     } else {
-      await installModule(m, {}, nuxt)
+      await installModule(m, {})
```
pi0, line 89: Refactor: If `installModule` could directly return `{ meta }`, we can create an array of installed modules meta usable for hooks and remove the dependency between `templates` and kit utils for normalizing config modules! (this also allows to register types for modules dynamically installed by another module)

pi0 (changes requested):

packages/nuxt3/src/core/templates.ts
```diff
@@ -117,3 +118,25 @@ export { }
...
 }
+
+const adHocModules = ['router', 'pages', 'auto-imports', 'meta', 'components']
+export const schemaTemplate = {
+  filename: 'types/schema.d.ts',
+  getContents: ({ nuxt }: TemplateContext) => {
+    const moduleInfo = nuxt.options._installedModules.map(m => ({
+      ...m.meta || {},
+      name: m.entryPath || m.meta?.name
+    })).filter(m => m.configKey && m.name && !adHocModules.includes(m.name))
```
pi0, line 129: - Shouldn't `meta.name` prioritize over `entryPath` for name?
- Why don't do normalization in `addModule` itself? `isAdhoc` flag could also be inferred in kit from name (templates should ideally have less logic)
-

packages/bridge/src/app.ts
```diff
@@ -38,6 +39,24 @@ export function setupAppBridge (_options: any) {
...
   })
 
+  // Add module metadata to nuxt.options._installedModules
+  nuxt.options._installedModules = nuxt.options._installedModules || []
+  nuxt.options._installedModules.push(...await Promise.all([
+    ...nuxt.options.buildModules,
+    ...nuxt.options.modules,
+    ...nuxt.options._modules
+  ].map(async (m) => {
+    const meta = await normalizeModule(m).then(([m]) => m.getMeta?.())
```
pi0, line 49: Maybe we can use `moduleContainer.requiredModules` to natively access installed nuxt 2 modules instead?

packages/nuxt3/src/core/templates.ts
danielroe: Example:

```js

// https://v3.nuxtjs.org/docs/directory-structure/nuxt.config
  modules: ['~/modules/test'],
})
```

And:
```js
import { defineNuxtModule } from '@nuxt/kit'

const defaults = {
  foo: true,
}

  meta: {
    name: 'bingle',
    configKey: 'bingleBob',
  },
  defaults,
})
```

As is, we want to augment the module entry path rather than the name of the module, to support local modules (and there may be other reasons too where a module's name doesn't equal a requirable/importable entity).

packages/bridge/src/app.ts
danielroe: Hm. That would be nice. But isn't this code running before that is populated?

packages/nuxt3/src/core/templates.ts
pi0: `importName` seems better but shall we only rely on `m.entryPath` then? `meta.name` is not necessarily a valid import path either without considering typescript resolving same as `resolveModule`. We could expose resolved `entryPath` from [here](https://github.com/nuxt/framework/blob/3598d50b8ceb7d362b5b77aadacacb25d17e8c6a/packages/kit/src/module/install.ts#L20)

packages/bridge/src/app.ts
pi0: Using `modules:done` seems a good idea! It would be nice if we avoid exposing extra `normalizeModule` from kit and have standalone Nuxt2 implementation if we want to support it. (module builder also populate interface therefore is not a big loss for bridge users)

packages/nuxt3/src/core/templates.ts
danielroe: TS is able to follow the aliasing as we do configure it to do so in `tsconfig.json`. But may be worth exposing resolved path for other purposes 👍.

As far as fallback to `meta.name`, this is mostly useful for directly installed modules (e.g. this kind of thing: https://github.com/nuxt/framework/blob/baa0853afa230ca9c4bdabb7db2b98275f59fcc8/packages/bridge/src/setup.ts#L15). If `meta.name` isn't a valid import path, TS will cope - so there seems to be no downside.

pi0 (approved):

PR #3111, opened 2022-02-07
feat(kit): `createResolver` utility

pi0:
### ❓ Type of change

### 📚 Description

A new `createResolver` utility, makes life easier for module authors resolving relative paths.

Usage:

```js
const resolver = createResolver(import.meta.url)
addPlugin(resolver.resolve('runtime/plugin'))
console.log(await resolver.resolvePath('some-lib'))
```

This PR adds some fixes and improvements to `resovlePath` and `resolveAlias` to work when nuxt is not ready yet

### 📝 Checklist

### DIFF
--- a/packages/kit/src/resolve.ts
+++ b/packages/kit/src/resolve.ts
@@ -117,7 +117,7 @@ export interface Resolver {
  */
 export function createResolver (base: string | URL): Resolver {
   if (!base) {
-    base = useNuxt().options.rootDir
+    throw new Error('`base` argument is missing for createResolver(base)!')
   }
 
   base = base.toString()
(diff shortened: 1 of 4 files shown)


packages/kit/src/resolve.ts
```diff
@@ -105,6 +107,30 @@ export function resolveAlias (path: string, alias?: Record<string, string>): str
   return path
 }
 
+export interface Resolver {
+  resolve(...path): string
+  resolvePath(path: string, opts?: ResolvePathOptions): Promise<string>
```
rchl, line 112: To be fair, it's not quite clear by looking at it what's really the difference between those two. Both seem to resolve path but is async and one not. :)

And while I can see the difference while looking at the code, I wonder if the naming should be more specific to make this more clear. Though unfortunately I can't think of any suggestions...

pi0: Fully agree (and unfortunately i couldn't find any better names :D) Idea for this naming was to keep `resolve` same as `path.resolve` and `resolvePath` same as global `resolvePath` but I guess we could deprecate and find better name for `resolvePath`...

PR #3134, opened 2022-02-08
fix(bridge): respect `--fail-on-error` in generation

### LINKED ISSUE #1968: Nuxi generate: exits with success in case of an error
### Environment

- Node Version:     `v17.0.1`
- Nuxt Version:     `2.16.0-27273615.ab1c6cb4`
- Bundler:          `Webpack`
- User Config:      `target`, `bridge`, `vite`, `alias`, `nitro`, `ssr`, `head`, `css`, `plugins`, `components`, `tailwindcss`, `buildModules`, `modules`, `serverMiddleware`, `watch`, `privateRuntimeConfig`, `build`, `storybook`   
- Runtime Modules:  `-`
- Build Modules:    `@nuxtjs/tailwindcss@4.2.1`, `portal-vue/nuxt`, `@nuxt/postcss8@1.1.3`, `@pinia/nuxt@0.1.0`, `nuxt-route-meta@2.3.2`, `@nuxt/bridge@3.0.0-27283410.c53b6a5`

### Reproduction

https://github.com/JabRef/JabRefOnline/pull/480, and in particular the "build" step in https://github.com/JabRef/JabRefOnline/runs/4218058544?check_suite_focus=true 

### Describe the bug

If there is an error during the nitro server build (in my case  #1967), then `nuxi generate` just keeps going and returns the exit code 0 (success). Thus the CI workflow passes although it should fail in the presence of an error. 

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #1968

### ❓ Type of change

### 📚 Description

We weren't passing on any errors when generating pages with Nitro.

This PR (minimally) returns error text, which allows nuxt 2 to respect the `--fail-on-error` flag.

We could _additionally_ always fail on error, which feels safer, and would just be a matter of adding two lines of code here: https://github.com/nuxt/framework/blob/00dea14b4287276caf36c4ca27abe7c6c7ada4a0/packages/bridge/src/nitro.ts#L195-L198

Thoughts welcome.

(**Note**: I also swapped joinURL as the URL being fetched had a double slash.)

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/nitro.ts
+++ b/packages/bridge/src/nitro.ts
@@ -216,15 +216,13 @@ function createNuxt2DevServer (nitroContext: NitroContext) {
     if (!listener) {
       throw new Error('There is no server listener to call `server.renderRoute()`')
     }
-    const r = await fetch(joinURL(listener.url, route), {
+    const res = await fetch(joinURL(listener.url, route), {
       headers: { 'nuxt-render-context': stringifyQuery(renderContext) }
     })
 
-    const html = await r.text()
+    const html = await res.text()
 
-    if (r.status === 500) {
-      return { html, error: r.statusText }
-    }
+    if (!res.ok) { return { html, error: res.statusText } }
 
     return { html }
   }
--- a/packages/bridge/src/nitro.ts
+++ b/packages/bridge/src/nitro.ts
@@ -1,7 +1,7 @@
 import { promises as fsp } from 'fs'
 import fetch from 'node-fetch'
 import { addPluginTemplate, useNuxt } from '@nuxt/kit'
-import { stringifyQuery } from 'ufo'
+import { joinURL, stringifyQuery } from 'ufo'
 import { resolve } from 'pathe'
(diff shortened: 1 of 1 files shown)


packages/bridge/src/nitro.ts
```diff
@@ -216,9 +216,15 @@ function createNuxt2DevServer (nitroContext: NitroContext) {
...
     }
-    const html = await fetch(listener.url + route, {
+    const r = await fetch(joinURL(listener.url, route), {
       headers: { 'nuxt-render-context': stringifyQuery(renderContext) }
-    }).then(r => r.text())
+    })
+
+    const html = await r.text()
+
+    if (r.status === 500) {
```
pi0, line 225: Shall we use `res.ok` instead? There are other codes when generation fails.

danielroe:
@pi0 Shall I also enable `--fail-on-error` by default?

pi0:
Let's keep it same as Nuxt2. We might directly infer from `process.argv` for behavior. (also this generate solution goes away soon so probably not worth spending time on it)

PR #3136, opened 2022-02-08
fix(nitro): handle incoming cookies on lambda payload format 2.0

danielroe:
### ❓ Type of change

### 📚 Description

[AWS Payload 2.0 passes cookies in an array on event](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html) rather than within headers. This PR adds them to the headers object.

In addition, I've attempted a type refactor. I've left the runtime code intact, but by annotating with `as` it should help future readers of the code see how the properties related to the payload version, and it means we also will get a compile error if we access a property that exists in one format but not the other unless we specifically annotate that access.


acidjazz:
I have run this code on my v3 fume+lambda project and confirmed it works, cookies are being set and retrieved via SSR

image

pi0 (approved):

PR #3143, opened 2022-02-09
docs(schema): clarify aliases and use more obvious example path

### LINKED ISSUE #3138: Inconsistent alias documentation
### Environment

Nuxt project info:                                                                                                                                                                         16:09:29

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-27405808.7825e2a`
- Bundler:          `Webpack`
- User Config:      `srcDir`, `vite`, `buildModules`, `css`, `alias`, `build`, `meta`, `publicRuntimeConfig`
- Runtime Modules:  `-`
- Build Modules:    `nuxt-windicss@2.2.4`, `@pinia/nuxt@0.1.8`

👉 Report an issue: https://github.com/nuxt/framework/issues/new                                                                                                                           16:09:29

Done in 2.19s.

### Reproduction

https://codesandbox.io/s/cranky-shirley-hu6zb?file=/tsconfig.json:99-202

1. Try with the lines commented out in tsconfig.json. The import doesnt work.
2. Now uncomment the lines in tsconfig.json, and change import statement to #tsAlias in app.vue
Again the import doesnt work

### Describe the bug

[ A note in `nuxt.config.ts`](https://v3.nuxtjs.org/docs/directory-structure/nuxt.config#alias) suggests that aliases should be added to `tsconfig.json`

> If you are using TypeScript and want to use the alias you define within your TypeScript files, you will need to add the aliases to your paths object within tsconfig.json.

Additionally, [ A note in `tsconfig.json`](https://v3.nuxtjs.org/docs/directory-structure/tsconfig) suggests that aliases will override the autogenerated aliases, and as such should instead be added to `nuxt.config.ts`

> note that if you need to customize your paths, this will completely override the auto-generated path aliases. Instead, it is suggested that you add any path aliases you need to the alias property within your nuxt.config, where they will get picked up and added to the autogenerated tsconfig.

Trying either unfortunately doesn't work, and nuxt throws a module not found error.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #3138

### ❓ Type of change

### 📚 Description

`/project` has led to one or two issues with people thinking it's a shortcut to the project path, as also has `<rootDir>` in the nuxtjs.org docs. I think `/path/to/your/app` (or something similar) might help.

### 📝 Checklist

### DIFF
--- a/packages/schema/build.config.ts
+++ b/packages/schema/build.config.ts
@@ -9,7 +9,7 @@ export default defineBuildConfig({
       name: 'config',
       builder: 'untyped',
       defaults: {
-        rootDir: '/path/to/your/app/'
+        rootDir: '/<rootDir>/'
       }
     },
     'src/index'
--- a/packages/schema/build.config.ts
+++ b/packages/schema/build.config.ts
@@ -9,7 +9,7 @@ export default defineBuildConfig({
       name: 'config',
       builder: 'untyped',
       defaults: {
-        rootDir: '/project/'
+        rootDir: '/path/to/your/app/'
       }
     },
     'src/index'
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -506,6 +506,8 @@ export default {
    * @note Within a webpack context (image sources, CSS - but not JavaScript) you _must_ access
    * your alias by prefixing it with `~`.
    *
+   * @note These aliases will be automatically added to the generated `.nuxt/tsconfig.json` .
+   *
    * @example
    * ```js
    * import { resolve } from 'pathe'
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -506,9 +506,6 @@ export default {
    * @note Within a webpack context (image sources, CSS - but not JavaScript) you _must_ access
    * your alias by prefixing it with `~`.
    *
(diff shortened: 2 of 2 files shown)


packages/schema/build.config.ts
```diff
@@ -9,7 +9,7 @@ export default defineBuildConfig({
       name: 'config',
       builder: 'untyped',
       defaults: {
-        rootDir: '/project/'
+        rootDir: '/path/to/your/app/'
```
pi0, line 12: What about something like this and instead properly document dirs definition (`<rootDir>` is documented)

```suggestion
        rootDir: '/<rootDir>/'
```

packages/schema/src/config/_common.ts
```diff
@@ -506,8 +506,7 @@ export default {
    * @note Within a webpack context (image sources, CSS - but not JavaScript) you _must_ access
    * your alias by prefixing it with `~`.
    *
-   * @note If you are using TypeScript and want to use the alias you define within
-   * your TypeScript files, you will need to add the aliases to your `paths` object within `tsconfig.json` .
+   * @note These aliases will be automatically added to the generated `.nuxt/tsconfig.json` .
```
pi0, line 509: Maybe add a short version of typescript documentation and warn to not override paths and prefer this option instead?

Issue #1036, opened 2021-10-14
HMR (with Vite) randomly skip updates and stop working


mkhennoussi:
### Environment

Nuxt CLI v3.0.0-27236898.4a34c2a                                      
RootDir: G:\PROJECTS\Testing\NUXT\nuxt3-app
Nuxt project info:
```
------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v16.8.0`
- Nuxt Version:     `3-3.0.0-27236898.4a34c2a`
- Package Manager:  `Yarn`
- Bundler:          `Webpack`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------
```

### Describe the bug

HMR with Vite randomly skip updates, I mean some (random I guess) updates on the code don't show up on the browser.

### Reproduction

Just use the basic generated app from the CLI and add/remove some code in the template part of `app.vue`.
Repeat that until you notice that some updates on the code don't show up on the browser.
It is more noticeable if you work fast with `tailwindcss/windicss`.

productdevbook:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.13.0`
- Nuxt Version:     `3.0.0-27255771.f78ec93`
- Package Manager:  `Yarn`
- Bundler:          `Vite`
- User Config:      `buildModules`, `build`
- Runtime Modules:  `-`
- Build Modules:    `@pinia/nuxt@0.1.0`
------------------------------


### Describe the bug


https://user-images.githubusercontent.com/38668796/139138186-7cd201f5-ab03-4a38-93bc-46ebe6cdc2d8.mp4



### Reproduction

https://github.com/productdevbook/nuxt3-app

### Additional context

This seems to me to be in vite.  because when I translate css, it fixes the project. There seems to be a problem with compiling Sass. but we're not entirely sure

### Logs

```shell
✔ Vite server built in 111ms
(node:15424) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
```

productdevbook:
Component used new problem

<img width="1434" alt="CleanShot 2021-11-03 at 06 37 50@2x" src="https://user-images.githubusercontent.com/38668796/140006946-ebdc291a-bad9-4661-ae27-10cf1c7015fc.png">

productdevbook:
css some problem
[...]

PR #3169, opened 2022-02-10
fix(vite): assign different hmr port for each instance

### LINKED ISSUE #3006: vite HMR Broken for <template>
### Environment

- Operating System: `Darwin`
- Node Version:     `v17.3.1`
- Nuxt Version:     `3.0.0-27389502.88da719`
- Bundler:          `Vite`
- User Config:      `vite`, `ssr`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Not needed 

### Describe the bug

Changes to CSS are picked up immediately and it was the same to the rest but ever since I upgraded Nuxt3 yesterday (not sure which version I had before) template changes aren't picked up by and HMR is broken for them

### Additional context

_No response_

### Logs

_No response_

antfu:
### 🔗 Linked issue

Fix #3006, Fix partially #1036

### ❓ Type of change

### 📚 Description

*Thanks a lot for @danielroe's help!*

The root cause of HMR breakage is because in middleware mode of Vite, the HMR port is fixed to 24678. Which means when you have multiple Vite/Nuxt instances running, it causes the later server to connect to the wrong Websocket instance, making the HMR fail to trigger. This PR makes the port dynamic with the currently available ports.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -6,6 +6,7 @@ import viteJsxPlugin from '@vitejs/plugin-vue-jsx'
 import type { Connect } from 'vite'
 
 import { joinURL } from 'ufo'
+import { getPort } from 'get-port-please'
 import { cacheDirPlugin } from './plugins/cache-dir'
 import { analyzePlugin } from './plugins/analyze'
 import { wpfs } from './utils/wpfs'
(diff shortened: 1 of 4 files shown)


packages/vite/src/client.ts
```diff
@@ -15,6 +16,13 @@ import { devStyleSSRPlugin } from './plugins/dev-ssr-css'
 import { DynamicBasePlugin, RelativeAssetPlugin } from './plugins/dynamic-base'
 
 export async function buildClient (ctx: ViteBuildContext) {
+  // TODO: After nitropack refactor, try if we can resuse the same server port as Nuxt
+  const hmrPortDefault = 24678 // Vite's default HMR port
+  const hmrPort = await getPort({
+    port: hmrPortDefault,
+    ports: Array.from({ length: 20 }, (_, i) => hmrPortDefault + 1 + i)
```
pi0, line 23: Any reason not just use a random port?

pi0 (approved):

packages/vite/src/client.ts
antfu: No, I am just not sure about how random works. Could change it to random if you want to

pi0: OS chooses one when port is 0

PR #3180, opened 2022-02-10
fix(nuxt3): re-enable tree-shaking `definePageMeta`

### LINKED ISSUE #3175: Improve warning about `definePageMeta()` outside `setup`
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0-27408580.aece351`
- Bundler:          `Vite`
- User Config:      `target`, `head`, `srcDir`, `css`, `plugins`, `components`, `buildModules`, `build`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

```vue
<script lang="ts">
definePageMeta({
  layout: 'hero'
})

 // ...
})
</script>
```

### Describe the bug

In a recent update, a new warn has been added when using `definePageMeta()` outside the `setup()` method. But it's a bit misleading, as it insinuates that it's *only* usable inside `<script setup>`. But it is indeed possible inside
```ts
  setup () {
    definePageMeta({
      layout: false
    })
  },
  // ....
})
```

### Additional context

Output of `npm run dev`
```
✔ Nitro built in 129 ms
definePageMeta() is a compiler-hint helper that is only usable inside <script setup> of a single file component. Its arguments should be compiled away and passing it at runtime has no effect.
✔ Vite server built in 416ms                      
```

Causing lines:
https://github.com/nuxt/framework/blob/68a227af037cc7c91b00787cadbd2653676e00ad/packages/nuxt3/src/pages/runtime/composables.ts#L27-L29

danielroe:
### 🔗 Linked issue

resolves #3175

### ❓ Type of change

### 📚 Description

This follows up on #3150 - I missed the fact that when we early return, we've potentially also changed code. This is also the cause of the new warning people are receiving about using `definePageMeta` outside of setup.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/pages/macros.ts
+++ b/packages/nuxt3/src/pages/macros.ts
@@ -13,7 +13,7 @@ export const TransformMacroPlugin = createUnplugin((options: TransformMacroPlugi
     enforce: 'post',
     transformInclude (id) {
       const { search, pathname } = parseURL(id)
-      return pathname.endsWith('.vue') || parseQuery(search).macro
+      return pathname.endsWith('.vue') || !!parseQuery(search).macro
     },
     transform (code, id) {
       const originalCode = code
--- a/packages/nuxt3/src/pages/macros.ts
+++ b/packages/nuxt3/src/pages/macros.ts
@@ -16,6 +16,7 @@ export const TransformMacroPlugin = createUnplugin((options: TransformMacroPlugi
       return pathname.endsWith('.vue') || parseQuery(search).macro
     },
     transform (code, id) {
+      const originalCode = code
       const { search } = parseURL(id)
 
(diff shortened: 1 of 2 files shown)


packages/nuxt3/src/pages/macros.ts
```diff
@@ -27,7 +28,9 @@ export const TransformMacroPlugin = createUnplugin((options: TransformMacroPlugi
         }
       }
 
-      if (!parseQuery(search).macro) { return }
+      if (!parseQuery(search).macro) {
+        return originalCode === code ? undefined : code
```
pi0, line 32: Side note: Doesn't rollup expect `null` instead of `undefined`? (however I think both behave similar)

pi0 (approved):

PR #3199, opened 2022-02-11
fix: clean up some type issues and add type test suite

danielroe:
### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

This PR adds a **type test suite** which we can augment as we add new features or fix bugs.

I've also noticed a couple of regressions, which I've addressed (see commit history), notably:

* `nitro.d.ts` wasn't being generated in `prepare` - and when it was generated, the paths were wrong
* `#imports` was untyped

Hopefully having a type testing suite will prevent future regressions.

**Note** that if you are testing locally you'll need to build `@nuxt/schema` first as `NuxtConfig` isn't being properly augmented within the workspace. (I will see what I can do to improve this for ease of testing in development.)

### Next steps

* Once this is merged, we can look at adding a similar test suite for bridge, which will involve adding types for the legacy composition api plugin, etc.

### DIFF
--- a/packages/nuxt3/src/app/nuxt.ts
+++ b/packages/nuxt3/src/app/nuxt.ts
@@ -193,7 +193,7 @@ export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp, 
 /**
  * Returns the current Nuxt instance.
  */
-export function useNuxtApp (): NuxtApp {
+export function useNuxtApp () {
   const vm = getCurrentInstance()
 
   if (!vm) {
@@ -203,7 +203,7 @@ export function useNuxtApp (): NuxtApp {
     return currentNuxtAppInstance
   }
 
-  return vm.appContext.app.$nuxt
+  return vm.appContext.app.$nuxt as NuxtApp
 }
 
 export function useRuntimeConfig (): RuntimeConfig {
--- a/packages/nuxt3/src/auto-imports/module.ts
+++ b/packages/nuxt3/src/auto-imports/module.ts
@@ -90,7 +90,7 @@ export default defineNuxtModule<AutoImportsOptions>({
     // Add generated types to `nuxt.d.ts`
     nuxt.hook('prepare:types', ({ references }) => {
       references.push({ path: resolve(nuxt.options.buildDir, 'types/auto-imports.d.ts') })
-      references.push({ path: resolve(nuxt.options.buildDir, 'types/imports.d.ts') })
+      references.push({ path: resolve(nuxt.options.buildDir, 'imports.d.ts') })
     })
 
     // Watch composables/ directory
@@ -121,7 +121,7 @@ function addDeclarationTemplates (ctx: AutoImportContext) {
   }
 
   addTemplate({
-    filename: 'types/imports.d.ts',
+    filename: 'imports.d.ts',
     getContents: () => toExports(ctx.autoImports)
   })
 
--- a/packages/nuxt3/src/pages/runtime/router.ts
+++ b/packages/nuxt3/src/pages/runtime/router.ts
@@ -6,6 +6,7 @@ import {
   RouterLink,
   NavigationGuard
 } from 'vue-router'
+import { createError } from 'h3'
 import NuxtPage from './page'
 import NuxtLayout from './layout'
 import { callWithNuxt, defineNuxtPlugin, useRuntimeConfig, NuxtApp } from '#app'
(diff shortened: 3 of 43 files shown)


packages/nuxt3/src/app/nuxt.ts
```diff
@@ -193,7 +193,7 @@ export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp,
 /**
  * Returns the current Nuxt instance.
  */
-export function useNuxtApp (): NuxtApp {
```
pi0, line 196: Why? Does build-step still generate return type of `useNuxtApp()` ?

packages/nuxt3/src/pages/runtime/router.ts
```diff
@@ -92,7 +92,7 @@ export default defineNuxtPlugin((nuxtApp) => {
         console.warn(`Unknown middleware: ${entry}. Valid options are ${Object.keys(namedMiddleware).join(', ')}.`)
       }
 
-      const result = await callWithNuxt(nuxtApp, middleware, [to, from])
+      const result = await callWithNuxt(nuxtApp as NuxtApp, middleware, [to, from])
```
pi0, line 95: Can't we explicitly type it at L95? 

```
export default defineNuxtPlugin((nuxtApp: NuxtApp) => {
```

packages/vite/src/dev-bundler.ts
```diff
@@ -79,7 +79,7 @@ async function transformRequest (opts: TransformOptions, id: string) {
       ? withoutVersionQuery
       : pathToFileURL(withoutVersionQuery)
     return {
-      code: `(global, exports, importMeta, ssrImport, ssrDynamicImport, ssrExportAll) => ${genDynamicImport(path, { wrapper: false })}.then(r => { exports.default = r.default; ssrExportAll(r) }).catch(e => { console.error(e); throw new Error(${JSON.stringify(`[vite dev] Error loading external "${id}".`)}) })`,
+      code: `(global, exports, importMeta, ssrImport, ssrDynamicImport, ssrExportAll) => ${genDynamicImport(path.toString(), { wrapper: false })}.then(r => { exports.default = r.default; ssrExportAll(r) }).catch(e => { console.error(e); throw new Error(${JSON.stringify(`[vite dev] Error loading external "${id}".`)}) })`,
```
pi0, line 82: Code style: Shall we add `.href` or `.toStrring()` to L80 and always have path as string?

packages/nuxt3/src/app/nuxt.ts
danielroe: Yes. The issue (IIRC) was that `$router` was present in the `dist` but not `src` versions of `NuxtApp` (because of augmentation). This isn't reproducible when running the test suite against the built module - it's not an error in the code, but reducing a false positive when running locally. The built declarations are still the same as they were previously.

packages/nuxt3/src/pages/runtime/router.ts
danielroe: No, because then `NuxtApp` will reference itself recursively.

packages/vite/src/dev-bundler.ts
danielroe: much nicer.

packages/nuxt3/src/app/nuxt.ts
pi0: This is a testing pipeline issue. If types can be tested properly against the built module, we should do it in CI instead of directly testing `src` types but testing `dist` types.

danielroe: I might be misunderstanding, but we are doing it in CI against dist types.

packages/nuxt3/src/auto-imports/module.ts
```diff
@@ -90,7 +90,7 @@ export default defineNuxtModule<AutoImportsOptions>({
     // Add generated types to `nuxt.d.ts`
     nuxt.hook('prepare:types', ({ references }) => {
       references.push({ path: resolve(nuxt.options.buildDir, 'types/auto-imports.d.ts') })
-      references.push({ path: resolve(nuxt.options.buildDir, 'types/imports.d.ts') })
+      references.push({ path: resolve(nuxt.options.buildDir, 'imports.d.ts') })
```
pi0, line 93: Can't we use `types/` dir for imports?

danielroe: Alas, no. It needs to match the location where imports.mjs is (virtually) generated. Alternatively, we could modify the generated tsconfig.json to _say_ that imports.mjs is generated within `types/imports.mjs`?

PR #3211, opened 2022-02-12
feat(nuxt3): auto generate runtime config type declarations

### LINKED ISSUE #1785: infer types for runtimeConfig
Follow-up on https://github.com/nuxt/framework/pull/1783#discussion_r744606478

rrmesquita:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/1785

### ❓ Type of change

### 📚 Description

This PR introduces the generation of type declarations for both `publicRuntimeConfig` and `privateRuntimeConfig`, with the latter having JSDoc in its values explaining that they're only accessible on the server-side.

Since runtime configs are strongly recommended in Nuxt 3, we could make use of a better development experience.

Resolves #1785

### 📝 Checklist


packages/nuxt3/src/core/templates.ts
```diff
@@ -128,6 +130,25 @@ export const schemaTemplate = {
...
+          ),
+          'PublicRuntimeConfig'
+        )
+      ),
+      normalizeUntypedOutput(
+        generateTypes(
+          deleteSchemaDefaults(
+            withDescription(
+              resolveSchema(nuxt.options.privateRuntimeConfig),
+              'This value is only accessible from server-side.'
```
danielroe, line 146: For nested objects this should only be added on the final property. E.g., in the following case we get the comment on hover on the entire `sanity` object and not on `sanity.token`, which is actually the secret piece.

```ts

  privateRuntimeConfig: {
    sanity: {
      token: 'foo'
    }
  },
  publicRuntimeConfig: {
    sanity: {
      projectId: 42
    }
  }
})
```

danielroe (commented):
This is a really nice enhancement - thanks for your work on it! ❤️ 

Just one minor comment.

rrmesquita:
I've improved the tests, added a test for what you've asked, and implemented it. Hope you like it! <3

packages/nuxt3/src/core/template-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  } else {
+    schema.description = description
+  }
+
+  return schema
+}
+
+export const normalizeUntypedOutput = (str: string) => {
+  return '  ' + str.replace(/^export\s/, '').replace(/\n/g, '\n  ')
+}
```
danielroe, line 28: Might be nice to expose more 'raw' untyped output as an enhancement upstream.

What do you think @pi0?

pi0: Added options for better control in dts gen

PR #3212, opened 2022-02-13
fix(nuxt3): prevent removing and re-adding tags before mount

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/pull/2419

### ❓ Type of change

### 📚 Description

This PR follows up on https://github.com/nuxt/framework/issues/2337. The issue it solves occurs in the initial render cycle, if there are multiple `useMeta` tags for the same tag type, e.g.:

```js
useMeta({ script: /* ... */ })
useMeta({ script: /* ... */ })
```

In this case, immediately triggering `updateDOM` after the first call causes a mismatch because we haven't yet registered the second script tag yet, and so it's removed. It's then added back, triggering the rerun.

By deferring updateDOM to mounted on the initial render, we give Vue a chance to process all these tags.

As a side effect, we are also more performant because we only trigger a single update, rather than a different one for every different `useMeta` call and meta tag we insert.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/meta/runtime/lib/vueuse-head.plugin.ts
+++ b/packages/nuxt3/src/meta/runtime/lib/vueuse-head.plugin.ts
@@ -1,5 +1,5 @@
 import { createHead, renderHeadToString } from '@vueuse/head'
-import { ref, watchEffect, onBeforeUnmount, getCurrentInstance } from 'vue'
+import { ref, watchEffect, onBeforeUnmount, getCurrentInstance, nextTick } from 'vue'
 import type { MetaObject } from '..'
 import { defineNuxtPlugin } from '#app'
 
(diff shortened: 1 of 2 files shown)


packages/nuxt3/src/meta/runtime/lib/vueuse-head.plugin.ts
```diff
@@ -7,23 +7,19 @@ export default defineNuxtPlugin((nuxtApp) => {
   const head = createHead()
 
   nuxtApp.vueApp.use(head)
+  nuxtApp.hook('app:mounted', () => watchEffect(() => head.updateDOM()))
```
pi0, line 10:
```suggestion
  nuxtApp.hookOnce('app:mounted', () => { watchEffect(() => head.updateDOM()) })
```

pi0 (approved):

Issue #2479, opened 2021-12-21
pass on cookies from server-side API calls on SSR response


antoineanorra:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.13.1`
- Nuxt Version:     `3.0.0-27333874.279bfdc`
- Package Manager:  `yarn@1.22.17`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

Currently no reproduction link is up to date with the latest version of Nuxt3 so I can't help you

### Describe the bug

Hello Nuxt 3 !

I have a problem, I try to make a session manager with redis (like express-session) but here I will try to make it in h3/nitro version, I don't know if it is possible ?

`setCookie` of h3 does not send the cookie to the client during the server side rendering.
However, it does send it if a request is executed directly on the api path.

server/api/hello.js
```js
import { setCookie } from 'h3'

export default (req, res) => {
	setCookie(res, "sessionId", "new session id");
	return 'you are now connected ! (in reality not)';
}
```

hello.vue
```js
const { data } = await useFetch("/api/hello", {
	headers: useRequestHeaders(),
});
```

Browser that executes the request to the API in SSR :
![image](https://user-images.githubusercontent.com/48593504/146915926-214442d7-6053-4205-bbfd-9f09513951c6.png)

Directly by making a request to the API with Postman :
![image](https://user-images.githubusercontent.com/48593504/146916952-6b737fba-5114-4e29-a1b2-d75d09ed9aa3.png)

Thanks !

Atinux:
I created a small sandbox, maybe it could help @pi0 @danielroe 

https://stackblitz.com/edit/nuxt-starter-zi9z7u?file=pages%2Findex.vue

danielroe:
In a web browser, any of many requests can set cookies for the browser. However, making requests on the server results in different behaviour. If you want to allow the headers set on the response to `useFetch` to be added to the response from SSR, you'll have to handle that yourself. (You could even write your own composable to do this.)

If we proxy headers (even just cookie headers) from _server_ API requests to the SSR response by default, that risks leaking sensitive information.

antoineanorra:
`If we proxy headers (even just cookie headers) from server API requests to the SSR response by default, that risks leaking sensitive information.`

@danielroe Is it possible to have a boolean configuration to allow this?

hale01:
[...]

PR #3218, opened 2022-02-14
docs: add note about proxying cookies back to client

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/2479

### ❓ Type of change

### 📚 Description

This adds a note that it's possible to proxy cookies from local API calls back to the client on the initial render.

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/1.usage/1.data-fetching.md
+++ b/docs/content/3.docs/1.usage/1.data-fetching.md
@@ -119,10 +119,9 @@ This composable behaves identically to `useFetch` with the `lazy: true` option s
 
 ## Isomorphic fetch
 
-When we call `fetch` in the browser, user headers like `cookie` will be directly sent to the API.
-But during server-side-rendering, since the `fetch` request is originating from the server, it doesn't include the user's browser cookies.
+When we call `fetch` in the browser, user headers like `cookie` will be directly sent to the API. But during server-side-rendering, since the `fetch` request takes place 'internally' within the server, it doesn't include the user's browser cookies, nor does it pass on cookies from the fetch response.
 
-We can use [`useRequestHeaders`](/docs/usage/ssr) to access and proxy cookies to the API from server-side.
+We can use [`useRequestHeaders`](/docs/usage/ssr) to access and proxy cookies to the API from server-side. (If you want to pass on/proxy cookies in the other direction, from an internal request back to the client, you will need to handle this yourself by accessing the raw response object, for example with `$fetch.raw`.)
 
 **Example:**
 


docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -119,10 +119,9 @@ This composable behaves identically to `useFetch` with the `lazy: true` option s
 
 ## Isomorphic fetch
 
-When we call `fetch` in the browser, user headers like `cookie` will be directly sent to the API.
-But during server-side-rendering, since the `fetch` request is originating from the server, it doesn't include the user's browser cookies.
+When we call `fetch` in the browser, user headers like `cookie` will be directly sent to the API. But during server-side-rendering, since the `fetch` request takes place 'internally' within the server, it doesn't include the user's browser cookies, nor does it pass on cookies from the fetch response.
 
-We can use [`useRequestHeaders`](/docs/usage/ssr) to access and proxy cookies to the API from server-side.
+We can use [`useRequestHeaders`](/docs/usage/ssr) to access and proxy cookies to the API from server-side. (If you want to pass on/proxy cookies in the other direction, from an internal request back to the client, you will need to handle this yourself by accessing the raw response object, for example with `$fetch.raw`.)
```
pi0, line 124: This is a little bit unclear IMO `useFetch` is already wrapping `fetch` and this use case is not isomorphic fetch. What about adding new section titles like "Bypass API headers to client" with an example?

danielroe: We can certainly add an example, though it might be a bit verbose, if you saw [my linked codesandbox](https://github.com/nuxt/framework/issues/2479#issuecomment-1038447712).

I think this isn't a section specifically about `useFetch`, right? And it's about a difference between server/client behaviour (e.g. on client-side, cookies are set on `/api/test` fetch, but on initial render, the fetch doesn't set cookies. So (my thought was) to write isomorphic code, you may need to handle this yourself.

But let me know if you still think better to extract into a different section 👍

antoineanorra: IMO, the best place is to put it in the Cookie section https://v3.nuxtjs.org/docs/usage/cookies#handling-cookies-in-api-routes

After "[Handling cookies in API routes](https://v3.nuxtjs.org/docs/usage/cookies#handling-cookies-in-api-routes)"
New section : Propagate the cookie in SSR context

pi0: /cc @danielroe Do you mind updating PR?

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -193,6 +194,34 @@ Here is a list of common headers that are NOT to be proxied:
...
+```ts [composables/fetch.ts]
+export const fetchWithCookie = async (url: string, cookieName: string) => {
+  const cookie = useCookie(cookieName)
+  const response = await $fetch.raw(url)
+  if (process.server) {
+    const cookies = Object.fromEntries(
+      response.headers.get('set-cookie')?.split(',').map((a) => a.split('='))
+    )
+    if (cookieName in cookies) {
+      cookie.value = cookies[cookieName]
```
pi0, line 210: (I like this approach but seems too much complexity for an example... Let's improve it later with a built-in utility for parse and pass)

BTW We could directly invoke `useCookie(cookieName).value = cookies[cookieName` here to improve both readability and perf by not making ref when not needed

Issue #3198, opened 2022-02-11
Test utils support progress


pi0:
- [x] Initial version (#2952)
- [x] Silent verbose build logs (#3259)
- [x] Rewrite tests against new fixture (#3308)
- [x] Limited bridge testing
- [x] Mark playwright optional (resolving from user root)
- [x] `nuxt test` CLI (#3307)
- [x] Test nuxt in development mode (#3755)
- [x] Working browser testing
- [x] Documentation for setup and usage

Chore: 
- [x] Fix vitest issue with jiti esm stub (https://github.com/unjs/unbuild/issues/35)
- [ ] Merge with `test/` utils for build hashing and dedup (#582)

yann-yinn:
Hi there. Will @nuxt/test-utils be able to handle **units testing / component testing** ? It is unclear to me when reading the issues. I can only see some exemples testing the whole page, but not a single component.

pi0:
Let's go for the next steps!

Build hashing: #582

Module starter: https://github.com/nuxt/starter/issues/12

(`nuxi dev --test`: TODO)

pi0:
> Will @nuxt/test-utils be able to handle units testing / component testing

@yann-yinn Honestly it is tricky to answer this today. Unit testing of SSR components needs a full running nuxt environment. Hopefully, we can achieve this with new nitropack possibilities and a custom entry point to introduce selective server rendering and enable unit testing of components within the tests.

PR #3259, opened 2022-02-16
feat: shared logger and silent test logs

antfu:
### 🔗 Linked issue

#3198 

### ❓ Type of change

### 📚 Description

Introduce `logger` and `useLogger` exports from `@nuxt/kit`:

```js
import { logger, useLogger } from '@nuxt/kit'
```

However consola is globally registered, this allows more semantic usage also removed the need for module authors to add another consola dependency when they want to create a namespace logger

This PR also introduces the `logLevel` option to `@nuxt/test-utils` defaulting to `1` which will be auto silent to the warning in the test environment. This option updates level of kit logger.

### 📝 Checklist


packages/schema/src/config/_common.ts
```diff
@@ -739,4 +739,11 @@ export default {
   publicRuntimeConfig: {
     $resolve: (val: Record<string, any> = {}, get) => ({ ...val, app: defu(val.app, get('app')) })
   },
+
+  /**
+   * Log level of Nuxt logs.
+   *
+   * @version 3
+   */
+  logLevel: isTest ? 1 : 3
```
antfu, line 748: LogLevel in consola is an enum (https://github.com/unjs/consola/blob/abb016f568e8942a3c3222258757bfff14304f7c/src/logLevels.js) what do you think is the best way for users to support it Nuxt? Literals with an internal map?

pi0: Using numbers should be fine for now. I intend to make a ts rewrite sometime soon and improving level typijng.

pi0: Another idea: We could set the log level to `1` by default in consola itself when in test environment :D

pi0: Just moved `logLevel` as an option of test-utils. In the next rewrite of consola, we can directly support named levels : https://github.com/unjs/consola/issues/130

PR #3280, opened 2022-02-17
chore: reverse target/from in eslint rule

danielroe:
### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

I think our internal lint rule was enforcing the opposite of what we meant to though I may have misunderstood the intent. But we are, I think, violating it here:

https://github.com/nuxt/framework/blob/aece3518b541663c1731160e42e26e6ac7c79ce8/packages/nuxt3/src/core/nuxt.ts#L5-L8

This PR is an initial stab meant to highlight the issue - we either need to:

1. amend lint error description to reflect what it's meant to do
2. rethink import approach in core/nuxt.ts
3. remove lint rules

cc: @clarkdo

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,9 @@
     "NodeJS": true,
     "$fetch": true
   },
-  "plugins": ["jsdoc"],
+  "plugins": [
+    "jsdoc"
+  ],
   "extends": [
     "plugin:jsdoc/recommended",
     "@nuxtjs/eslint-config-typescript",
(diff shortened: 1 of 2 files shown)


clarkdo (commented):
This change looks good.

Regaording the lint error after the change, @pi0 should we exclude them from config or refacor the code to forbid importing other pacakges in core/nuxt.ts ?

danielroe:
As adding those modules there is temporary, I've disabled the rule for those specific imports.

https://github.com/nuxt/framework/blob/61c429189767fb601301a64e56176c1a38396f21/packages/nuxt3/src/core/nuxt.ts#L106

clarkdo (approved):

Issue #3272, opened 2022-02-16
navigateTo: router is undefined in a client context


Teranode:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.13.2`
- Nuxt Version:     `3.0.0-27417251.289d54e`
- Package Manager:  `npm@8.4.1`
- Bundler:          `Vite`
- User Config:      `srcDir`, `css`, `buildModules`, `googleFonts`, `axios`, `auth`, `publicRuntimeConfig`, `vite`, `build`, `typescript`
- Runtime Modules:  `-`
- Build Modules:    `nuxt-windicss@2.2.5`, `@nuxtjs/auth@file:project/modules/@nuxtjs/auth`, `@nuxtjs/axios@file:project/modules/@nuxtjs/axios`, `@nuxtjs/pinia@file:project/modules/@nuxtjs/pinia`, `@nuxtjs/google-fonts@file:project/modules/@nuxtjs/google-fonts`

### Reproduction

a codesanbox link has been provided, just click the button and open your devtools.
https://codesandbox.io/s/hardcore-pateu-z5jcif

### Describe the bug

_**navigateTo function**_

https://github.com/nuxt/framework/blob/77aeaa3288f90bdeb1d34c373e9a3a05a8b31970/packages/nuxt3/src/pages/runtime/composables.ts#L74

`window.$nuxt.router` is supposed to be `window.$nuxt.config.globalProperties.$router` this will then allow redirection in a client context.

**_redirection with layouts_**

Regardless of this however, if you try to redirect when a page uses a different layout, regardless of this fix there is an error with layouts, so this function can only be used when all pages utilize the same layout. I've also tested this with the router composable and that same error occurs in regards to layouts.  So that's just an entirely different issue altogether.

### Additional context

I suppose this also has to do with this: https://github.com/nuxt/framework/issues/1750

danielroe:
@Teranode Thank you for this - and good spot! Any chance of creating another issue (with repro) about the layout error you mention?

PR #3281, opened 2022-02-17
fix(nuxt3,bridge): correctly reference router from client-side helper

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/3272

### ❓ Type of change

### 📚 Description

This adds correct references to the client-side nuxt router on Bridge & Nuxt 3.

### 📝 Checklist


packages/nuxt3/src/pages/runtime/composables.ts
```diff
@@ -71,7 +71,7 @@ export const navigateTo = (to: RouteLocationRaw) => {
   if (isProcessingMiddleware()) {
     return to
   }
-  const router: Router = process.server ? useRouter() : (window as any).$nuxt.router
+  const router: Router = process.server ? useRouter() : (window as any).$nuxt.config?.globalProperties.$router
```
pi0, line 74: Shall we instead improve legacy app plugin to allow accessing `$nuxt.$router` same as nuxt 2?

packages/bridge/src/runtime/composables.ts
```diff
@@ -185,7 +185,7 @@ export const navigateTo = (to: Route) => {
   if (isProcessingMiddleware()) {
     return to
   }
-  const router: VueRouter = process.server ? useRouter() : (window as any).$nuxt.router
+  const router: VueRouter = process.server ? useRouter() : (window as any).$nuxt.context?.app.router
```
pi0, line 188: (following other comment) then this could be `$nuxt.$router`

packages/nuxt3/src/pages/runtime/composables.ts
danielroe: Hm. I think there was a good reason that globally injected properties were moved off the instance, but I can't recall why 🤔

pi0: No reason not to IMO. They can be anyway accessed from vue instance...

packages/nuxt3/src/app/compat/legacy-app.ts
```diff
@@ -205,6 +205,10 @@ export const legacyPlugin = (nuxtApp: NuxtApp) => {
       // @ts-ignore
       // TODO: https://github.com/nuxt/framework/issues/244
       legacyApp.constructor = legacyApp
+      if ('$router' in nuxtApp) {
```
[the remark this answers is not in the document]
danielroe: I think it's likely because we didn't want to add `$root` and `constructor` to the vueApp.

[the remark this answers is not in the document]
pi0: What do you think if we use a proxy instead? (sorry for bothering on this I know is not related to fix but trying to fix root cause of issue)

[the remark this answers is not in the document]
danielroe: Not a bother! I like the idea. Checking it out.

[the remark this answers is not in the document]
danielroe: Having checked, `nuxtApp.vueApp.$router` is not defined, so whether or not we destructure or use proxy, it's only accessible via `$nuxt.config.globalProperties.$router`.

packages/nuxt3/src/app/compat/legacy-app.ts
```diff
@@ -199,14 +200,18 @@ export const legacyPlugin = (nuxtApp: NuxtApp) => {
...
-
-      // @ts-ignore
-      // TODO: https://github.com/nuxt/framework/issues/244
-      legacyApp.constructor = legacyApp
-
-      window[`$${nuxtApp.globalName}`] = legacyApp
+      window[`$${nuxtApp.globalName}`] = new Proxy(nuxtApp.vueApp as LegacyApp, {
+        get (legacyApp, p: keyof LegacyApp) {
+          if (p === '$router') {
+            return nuxtApp.$router
```
pi0, line 206: Can we instead fall back to `nuxtApp` if  `legacyApp` does not have the prop?

packages/nuxt3/src/app/compat/legacy-app.ts
```diff
@@ -199,14 +200,18 @@ export const legacyPlugin = (nuxtApp: NuxtApp) => {
...
+          }
+          // TODO: https://github.com/nuxt/framework/issues/244
+          if (['$root', 'constructor'].includes(p)) {
+            return legacyApp
```
pi0, line 210: We should then return Proxy instance instead of `legacyApp` to allow nested usage

packages/nuxt3/src/app/compat/legacy-app.ts
danielroe: What makes you think that legacyApp will have the prop?

pi0: It doesn't! After [this line](https://github.com/nuxt/framework/pull/3281/files#diff-f48ebf08f0fc31f5c6f93777f19b730cd25ad6b8b2588eecc58d14b79893cd7dR212) we could fallback to `nuxtApp[p]`

danielroe: Ah, I see.

PR #3288, opened 2022-02-17
fix(nuxt3): generate imports for components with named exports correctly

danielroe:
### ❓ Type of change

### 📚 Description

Previously we would always generate default imports for components, even though `addComponent` supports adding named imports. This PR adds support for that with `genImport`.

The change:

```diff
-  import __nuxt_component_23 from '/my/file/default.ts'
+  import { default as __nuxt_component_23 } from '/my/file/default.ts'
-  import __nuxt_component_24 from '/my/file/named.ts'
+  import { namedExport as __nuxt_component_24 } from '/my/file/named.ts'
```

#### Notes

* I've added an example in `with-components` but it may be better in a unit test - let me know if so and I can extract...

### 📝 Checklist


examples/with-components/nuxt.config.ts
```diff
@@ -13,5 +14,14 @@ export default defineNuxtConfig({
         prefix: 'nuxt'
       }
     ]
+  },
+  hooks: {
```
pi0, line 18: It is probably not a generic use case to add in example. Let's have it in `test/fixture/basic` which can be used for actually testing behavior too.

packages/nuxt3/src/components/loader.ts
```diff
@@ -37,7 +37,7 @@ function transform (content: string, components: Component[]) {
     if (component) {
       const identifier = map.get(component) || `__nuxt_component_${num++}`
       map.set(component, identifier)
-      imports += genImport(component.filePath, identifier)
+      imports += genImport(component.filePath, [{ name: component.export, as: identifier }])
```
pi0, line 40: Note: I guess we could improve `genImport` to avoid generating `default as x` when there is only one default import...

Issue #3216, opened 2022-02-13
If webpack analyze is true, server building crashes


Tragio:
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.18.1`
- Nuxt Version:     `2.16.0-27358576.777a4b7f`
- Package Manager:  `yarn@1.22.17`
- Bundler:          `Webpack`
- User Config:      `ssr`, `head`, `env`, `loading`, `css`, `plugins`, `components`, `modules`, `buildModules`, `alias`, `build`, `bridge`
- Runtime Modules:  `-`
- Build Modules:    `@nuxtjs/eslint-module@3.0.2`, `@nuxtjs/stylelint-module@4.1.0`, `@nuxtjs/html-validator@0.6.0`, `@nuxtjs/style-resources@1.2.1`, `nuxt-purgecss@1.0.0`, `@nuxt/bridge@3.0.0-27409509.e727bb4`

### Reproduction

https://github.com/Tragio/oruga-tree-shaking-bug

Using webpack bundle analyzer with `analyze: true,` on `nuxt.config.js`.

Doing yarn build will crash when building server but if you comment `analyze: true,` on `nuxt.config.js` it will succeed.

### Describe the bug

```
 FATAL  Cannot read property 'replace' of undefined                                                                                                                                            11:01:32

  at getRollupConfig (node_modules/@nuxt/nitro/dist/index.mjs:952:47)
  at build (node_modules/@nuxt/nitro/dist/index.mjs:1055:31)
  at async node_modules/@nuxt/bridge/dist/chunks/module.mjs:242:7
  at async Nuxt.callHook (node_modules/hable/dist/hable.js:1:959)
  at async Builder.build (node_modules/@nuxt/builder-edge/dist/builder.js:264:5)
  at async Object.invoke (node_modules/nuxi/dist/chunks/build.mjs:33:5)
  at async _main (node_modules/nuxi/dist/chunks/index.mjs:384:7)
```

danielroe:
Ideally you should run `nuxi analyze` rather than set `analyze: true` manually. Does that work around this issue for you?

Tragio:
@danielroe yes, using `nuxi analyze` everything still works, but it is nice that @clarkdo is fixing it to avoid people migrating to nuxt bridge facing that. 😄

danielroe:
This seems to be resolved. Let me know if not and I'll reopen it.

PR #3292, opened 2022-02-17
fix(config): add analyze plugin options in bridge mode

clarkdo:
### 🔗 Linked issue
(https://github.com/nuxt/framework/issues/3216)

### ❓ Type of change

### 📚 Description

When running `nuxi analyze` on bridge mode, we'll see error `Cannot read property 'replace' of undefined` which is because the analyze options is `true` instead of default options object.

This pr is fixing this issue by adding default analyze options in bridge mode

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/build.ts
+++ b/packages/schema/src/config/build.ts
@@ -29,7 +29,7 @@ export default {
   analyze: {
     $resolve: (val, get) => {
       if(val !== true) {
-        return val
+        return val ?? false
       }
       const rootDir = get('rootDir')
       return {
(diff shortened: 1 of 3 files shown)


packages/schema/src/config/build.ts
```diff
@@ -25,7 +26,19 @@ export default {
    * ```
    * @type {boolean | typeof import('webpack-bundle-analyzer').BundleAnalyzerPlugin.Options | typeof import('rollup-plugin-visualizer').PluginVisualizerOptions}
    */
-  analyze: false,
+  analyze: {
```
clarkdo, line 29: I'm not sure if this still makes sense. Let me know and I can remove if not.

pi0 (approved):

PR #3307, opened 2022-02-18
feat(nuxi): init `nuxi test` support

antfu:
### ❓ Type of change

### 📝 Checklist


packages/nuxi/src/commands/test.ts
```diff
@@ -0,0 +1,31 @@
...
+    process.env.NODE_ENV = process.env.NODE_ENV || 'production'
+    const rootDir = process.cwd()
+
+    consola.warn('`nuxt test` is currently expiremental')
+
+    const { loadNuxtConfig } = await loadKit(rootDir)
+    const config = await loadNuxtConfig({ rootDir })
+
+    if (config.vite === false) {
+      consola.error('`nuxi test` is currently only supported in Vite mode')
```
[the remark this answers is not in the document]
antfu: Oh yeah, I thought we will reuse some part of Vite. If it's about the runner, we could sure have vitest work along with Webpack. So that said, should be recommand Vitest by default? (and make it swapable with Jest with some config?)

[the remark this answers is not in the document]
pi0: Yeah, let's use vitest as default runner 👍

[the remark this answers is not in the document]
Atinux: Vitest as default runner 💯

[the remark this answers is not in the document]
ricardogobbosouza: Vitest should be the default but we need jest support too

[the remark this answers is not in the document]
pi0: @ricardogobbosouza Indeed and there an option for it but for now we are focusing on one stable test runner.

antfu:
Updates:

- Vitest is now the default runner
- Vitest now expose the programatic APIs so we don't need to spawn a sub process
- Moved the logics to `@nuxt/test-utils` - Thu `nuxi test` will deps on `@nuxt/test-utils` and should prompt auto install if it's not installed

Todo (in future PRs):

- [ ] Package existance check and auto install after `nypm` is ready
- [ ] Support config entires for switch different test runners

packages/nuxi/src/commands/test.ts
```diff
@@ -0,0 +1,27 @@
...
+import { defineNuxtCommand } from './index'
+
+export default defineNuxtCommand({
+  meta: {
+    name: 'test',
+    usage: 'npx nuxi test',
+    description: 'Run tests'
+  },
+  async invoke (argv) {
+    process.env.NODE_ENV = process.env.NODE_ENV || 'production'
```
pi0, line 12: Shouldn't be `test` for default?

packages/nuxi/src/commands/test.ts
```diff
@@ -0,0 +1,27 @@
...
+    const rootDir = process.cwd()
+
+    consola.warn('`nuxt test` is expiremental')
```
pi0, line 15: (what's the point of mentioning this btw? Command is basically starting vitest) 

```suggestion
    consola.warn('`nuxi test` command is expiremental')
```

packages/test-utils/src/runners/vitest.ts
```diff
@@ -0,0 +1,22 @@
...
+  const succeeded = await startVitest(
+    argv._ || [],
+    {
+      root: options.rootDir,
+      ...argv
+    },
+    {
+      // TODO: forward more Vite options
+      resolve: {
+        alias: options.alias
```
pi0, line 14: These are nuxt build options for aliases not for test runners. What use case would you imagine for this?

packages/nuxi/src/commands/test.ts
Atinux: Also it is "experimental"

packages/nuxi/src/commands/test.ts
```diff
@@ -0,0 +1,32 @@
...
+    const { loadNuxtConfig } = await loadKit(rootDir)
+    const config = await loadNuxtConfig({ rootDir })
+
+    // TODO: support Jest and other runners
+    consola.info('Starting Vitest...')
+
+    // TODO: prompt auto install if test-utils is not installed
+    let start: typeof import('@nuxt/test-utils/vitest')['default'] | undefined
+    try {
+      start = await import('@nuxt/test-utils/vitest').then(m => m.default)
```
pi0, line 24: (we could use named `startVitest` export to avoid this)
```suggestion
      start = await import('@nuxt/test-utils/vitest').then(m => m.default || m)
```

pi0:
Updated implementation with a simpler `runTests` from `@nuxt/utils` to be used by CLI and also an example.

Two remaining issues:
- It seems rootDir resolution is buggy when moving to `examples/with-tests` and run `yarn nuxi test`. It works fine with `yarn nuxi test examples/with-test`
- Vitest tries to evaluate `tsconfig` extends which doesn't work before setup is done! We need to disable it same as doing for vite.

packages/test-utils/src/run.ts
```diff
@@ -0,0 +1,35 @@
...
+    [] /* argv */,
+    // Vitest options
+    {
+      root: opts.rootDir,
+      run: true
+    },
+    // Vite options
+    {
+      esbuild: {
+        tsconfigRaw: '{}'
```
pi0, line 27: This options seems still not respected:

image

pi0 (approved):
Lets land it and iterate to improve

PR #3323, opened 2022-02-18
feat(nuxt3): expose `/app` export paths

### LINKED ISSUE #3309: expose app and app types as `nuxt3/app`
We auto-import default composables from `#app`, but if users do need to import manually then this requires using a nuxt-generated `tsconfig.json`.

By exposing these utilities via a subpath import (as well as `#app`) we can improve DX and IDE auto-complete before `tsconfig.json` is generated, or for module authors.

For Bridge, we will expose `nuxt/app`, and when `nuxt3` moves to `nuxt` then they will both be importing from `nuxt/app`...

danielroe:
### 🔗 Linked issue

resolves  #3309

### ❓ Type of change

- [ ] 🧹 Chore

### 📚 Description

This PR exposes `nuxt3/app` (for #app), `nuxt3/pages` and `nuxt3/meta` (for `#meta`) as the runtime entrypoints for those respective parts of nuxt3.

This means that module or library authors, for example, can benefit from nuxt3 types without needing to use a generated `.nuxt/tsconfig.json`.

Because TypeScript isn't quite ready for types for export paths (track in https://github.com/microsoft/TypeScript/issues/33079), we add declaration files at the project root for IDE auto-complete. Both webpack & vite properly respect subpath exports, however, and resolve any imports we need.

Because this is additive, users can continue importing from '#app' - and indeed I would recommend pending any updates to the user-facing docs until nuxt3/bridge are in sync (e.g. both importing from `nuxt/app` or equivalent), perhaps when we move to RC.

⚠️ **Note**: Adopting package subpath exports is inherently risky because it breaks any other possible imports/access to the package, so we should take special care to ensure that ... ⚠️

### 📝 Checklist


packages/nuxt3/package.json
```diff
@@ -10,7 +10,18 @@
     "nuxi": "./bin/nuxt.mjs",
     "nuxt": "./bin/nuxt.mjs"
   },
+  "exports": {
+    ".": "./dist/index.mjs",
+    "./app": "./dist/app/index.mjs",
+    "./package.json": "./package.json"
```
pi0, line 16: We are not exporting `./package.json` subpath for any other package. Any reason is required?

packages/nuxt3/src/core/nuxt.ts
```diff
@@ -104,7 +104,7 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
   const options = await loadNuxtConfig(opts)
 
   // Temporary until finding better placement for each
-  options.appDir = options.alias['#app'] = resolve(distDir, 'app')
+  options.appDir = options.alias['nuxt3/app'] /* for framework playground */ = options.alias['#app'] = resolve(distDir, 'app')
```
pi0, line 107: Is it needed if we internally use `#app` ?

```suggestion
  options.appDir = options.alias['#app'] = resolve(distDir, 'app')
```

danielroe: Nope!

packages/nuxt3/package.json
danielroe: I thought we might require for nuxi upgrade, but I didn't actually test it.

https://github.com/nuxt/framework/blob/88da7197ba4774820a6e530be42fc1f48bf6f6c3/packages/nuxi/src/commands/upgrade.ts#L9-L16

PR #3330, opened 2022-02-19
docs: auto imports don't work in `server/api` 

vencho-mdp:
### 🔗 Linked issue

[composable auto import improvements](https://github.com/nuxt/framework/issues/1273)

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.concepts/4.auto-imports.md
+++ b/docs/content/2.concepts/4.auto-imports.md
@@ -1,6 +1,6 @@
 # Auto imports
 
-Nuxt auto-imports helper functions, composables and Vue APIs to use across your application without explicitly importing them. Based on the directory structure, every Nuxt application can also use auto-imports for its own components, composables and plugins.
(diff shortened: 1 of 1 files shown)


docs/content/2.concepts/4.auto-imports.md
```diff
@@ -56,3 +56,15 @@ Every Nuxt auto-import is exposed by the `#imports` alias that can be used to ma
   const double = computed(() => count.value * 2)
 </script>
 ```
+
+## Imports in `server/api`
```
danielroe, line 60: How about adding this note earlier in the document (just after the line `These functions can be used in components, composables or plugins.`)? https://v3.nuxtjs.org/concepts/auto-imports#nuxt-auto-imports

I'd avoid importing composables in the server API code, personally, so I'm reluctant to use this as an example in the docs.

pi0 (approved):
LGTM. Thanks!

vencho-mdp:
@danielroe Yes! I completely agree with your idea. However, I also moved that section (as you can see in the PR) a little bit above where it was, because if we leave this under the `Nuxt Auto Imports` header, developers might think that this only applies to those imports.

danielroe:
A good solution!

PR #3331, opened 2022-02-19
docs: add docker deployment guide

misaon:
### ❓ Type of change

### 📚 Description
Add Docker deployment document.

### 📝 Checklist


docs/content/3.docs/3.deployment/7.docker.md
```diff
@@ -0,0 +1,129 @@
...
+
+::code-group
+
+```bash [Docker compose]
+docker-compose build
+```
+
+```bash [Docker compose v2]
+docker compose build
+```
```
rrmesquita, line 125: Capitalize here as well.

docs/content/3.docs/3.deployment/7.docker.md
```diff
@@ -0,0 +1,129 @@
...
+
+::code-group
+
+```bash [Docker compose]
+docker-compose exec nuxt sh
+```
+
+```bash [Docker compose v2]
+docker compose exec nuxt sh
+```
```
rrmesquita, line 94: Capitalize here as well.

docs/content/3.docs/3.deployment/7.docker.md
```diff
@@ -0,0 +1,129 @@
...
+
+::code-group
+
+```bash [Docker compose]
+docker-compose up -d
+```
+
+```bash [Docker compose v2]
+docker compose up -d
+```
```
rrmesquita, line 80: How about capitalizing "compose"? e.g. "Docker Compose" and "Docker Compose V2".

docs/content/3.docs/3.deployment/7.docker.md
```diff
@@ -0,0 +1,129 @@
...
+
+RUN yarn install && yarn run build
+
+###
+
+FROM node:17.5-alpine
+
+WORKDIR /app
+
+COPY ./package.json /app/package.json
```
pi0, line 40: Why do we need package.json in second layer?

docs/content/3.docs/3.deployment/7.docker.md
```diff
@@ -0,0 +1,129 @@
...
+COPY --from=nuxtBuild /nuxtBuild/.output /app/.output
```
pi0, line 41: Maybe directly `.output` to `/app`? 

```suggestion
COPY --from=nuxtBuild /nuxtBuild/.output /app
```

docs/content/3.docs/3.deployment/7.docker.md
```diff
@@ -0,0 +1,129 @@
...
+
+RUN ( \
+    echo '#!/bin/sh'; \
+    echo 'if [[ "$NODE_ENV" == "development" ]]; then yarn run dev; else node /app/.output/server/index.mjs; fi' \
```
pi0, line 45: Why not simply use CMD with `[node, /app/server/index.mjs]` ?

docs/content/3.docs/3.deployment/7.docker.md
```diff
@@ -0,0 +1,129 @@
...
+docker compose exec nuxt sh
+```
+
+::
+
+**Nuxt will be available at [http://localhost:3000](http://localhost:3000)**.
+
+### Build into production
+
+To build into production you need to modify the environment `NODE_ENV` and remove `volumes` property, example:
```
pi0, line 102: I'm thinking we could have docker-compose it's whole section and use `docker build` `docker start` for initial docs. Not all deployments are based on compose while guide for building image is an essential part.

docs/content/3.docs/3.deployment/7.docker.md
misaon: You're right, I'll delete it

docs/content/3.docs/3.deployment/7.docker.md
misaon: Because there is support for local development controlled by the variable "NODE_ENV", it is possible to run the image in dev mode

pi0: Ah, I see! Well for development we could directly override the command in `docker-compose.yml` (or last arg of `docker` in case of directly using docker CLI)

docs/content/3.docs/3.deployment/7.docker.md
misaon: It is safer to copy only the builded folder, it can copy to the production eg ".yarn", error logs, etc.

docs/content/3.docs/3.deployment/7.docker.md
misaon: Good idea!

docs/content/3.docs/3.deployment/7.docker.md
pi0: We could surely document this but i'm afraid this is not needed for actual nuxt deployments. `.output` directory is fully standalone with `node_modules` and even niro build meta in a json file.

pi0:
Hi @misaon. Would you like me to help with improvement suggestions?

misaon:
> Hi @misaon. Would you like me to help with improvement suggestions?

hello, I'll take a look at it over the weekend and put it in its final form.  So do we divide docker and docker compose into two sections?

pi0:
Hi @misaon. Again really appreciate your efforts on this 💚 

I'm working on an official docker preset for nitro (https://github.com/unjs/nitro/issues/54). Let's come back to the documentation once had it. Also for development, we probably need a different kind of documentation.

Issue #2748, opened 2022-01-16
impossible to set options on root `~/components` directory


danielroe:
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.17.3`
- Nuxt Version:     `3.0.0`
- Package Manager:  `yarn@3.1.1`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://codesandbox.io/s/nice-robinson-5kz9r?file=/nuxt.config.ts

### Describe the bug

The components module merges any options _in addition to_ the default `~/components` dir which  means any customisation to the options are lost (only for the root dir).

alexcroox:
Appreciate you looking into and raising this issue ❤️

Jlaird:
I just pulled the latest of master and a lot of my components broke. It looks like ```pathPrefix: false``` Is not respected at all now?

This is what I had previously
``` 
components: [
    { path: '~/components', pathPrefix: false },
    { path: '~/components/general', pathPrefix: false },
    { path: '~/components/admin', pathPrefix: false },
    { path: '~/components/newsletters', pathPrefix: false },
    { path: '~/components/web', pathPrefix: false },
  ],

Tahul:
We've also encountered issues with `components` key with @nuxt/content.

`pathPrefix` or `prefix` key does not seem to have any effect anymore with latest versions of Nuxt.

Jlaird:
@[danielroe](https://github.com/danielroe) It sounds like quite a few people can't upgrade to the latest version without fixing this bug. The issue is that `pathPrefix` is never added to `componetDirs` when calling `scanComponents()`.
https://github.com/nuxt/framework/blob/87eb7d0d392e2fb1ed36cdb8f4c96c69720c2276/packages/nuxt3/src/components/module.ts#L95

The description should be updated by removing ` (only for the root dir)`. It doesn't pass any options at all. This seems to have broken more recently. Also recently it seems that the prefix is not added when the file name already had the prefix on it. This might not be a bug but seems confusing. 

Example:
```'WebContentImageGrid': typeof import("../../components/web/WebContentImageGrid.vue")['default'],```

PR #3364, opened 2022-02-22
fix(nuxt3): respect dirs as objects

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/2748

### ❓ Type of change

### 📚 Description

Recently component dirs passed as objects were no longer respected (https://github.com/nuxt/framework/pull/3108). This PR adds a test for object dir syntax.

In addition, because we always added `~/components` to the dirs list, customisations of this root directory were lost. By dropping it from the module options, we _still_ get this added by default (by schema) when the user does not provide input, but allows the user to override or configure this.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/components/module.ts
+++ b/packages/nuxt3/src/components/module.ts
@@ -18,7 +18,7 @@ export default defineNuxtModule<ComponentsOptions>({
     configKey: 'components'
   },
   defaults: {
-    dirs: ['~/components']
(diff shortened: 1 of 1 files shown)


Atinux (approved):

Jlaird:
@danielroe Confirmed it's fixed, thanks!

PR #2223, opened 2021-11-29
fix(nuxt3): change entry to use asynchronous import

phoenix-ru:
### 🔗 Linked issue

#1444

[Discussion](https://github.com/nuxt/framework/discussions/870)

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Async boundary in the client is crucial for supporting Webpack 5 Module Federation. With this PR and an [example repository](https://github.com/phoenix-ru/nuxt3-module-federation-example), Module Federation will be officially supported by Nuxt 3!

Q: Is it a breaking change?
A: The contract for the entry does not change. Webpack 5 build does function as intended. **Vite build also works.**

I am also intending on publishing a Nuxt module for an easy configuration in the projects.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly. Documentation does not need to be updated, this is an internal change which doesn't face users.

PR #3375, opened 2022-02-23
fix(vite,webpack): disable async entrypoint by default

danielroe:
### ❓ Type of change

### 📚 Description

Previously, when running vite, the following error would be printed in dev mode:

```
Uncaught (in promise) TypeError: error loading dynamically imported module
```

This was a side effect of https://github.com/nuxt/framework/pull/2223.

This PR disables async entry by default, though it can be re-enabled with a flag. **Note**: you may well find that you need to disable async entry in development.

This PR also swaps entry <-> bootstrap names (as I think it makes more sense this way). Despite the massive git diff, the only real change is to the webpack/vite configs.

### 👉 Migration

If you were previously relying on the async entrypoint behaviour (for example, with module federation) then you can reenable previous behaviour by adding `experimental.asyncEntry` to your nuxt.config:

```diff
  import { defineNuxtConfig } from 'nuxt3'
  
  export default defineNuxtConfig({
+   experimental: {
+     asyncEntry: true // or process.env.NODE_ENV !== 'development'
+   }
  })
```

### 📝 Checklist


pi0:
I believe we could revert async entry change back to normal for all targets (dev/prod x vite/webpack) and optionally support async method behind a flag.

packages/schema/src/config/build.ts
```diff
@@ -617,5 +617,11 @@ export default {
...
    * @version 2
    */
-  followSymlinks: false
+
+  followSymlinks: false,
+  /**
+   * Set to true to generate an async entry (for module federation support).
+   * @version 3
+   */
+  asyncEntry: false
```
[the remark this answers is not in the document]
danielroe: I like experimental namespace 👍

packages/vite/src/vite.ts
```diff
@@ -50,7 +50,7 @@ export async function bundle (nuxt: Nuxt) {
             // will be filled in client/server configs
             '#build/plugins': '',
             '#build': nuxt.options.buildDir,
-            '/entry.mjs': resolve(nuxt.options.appDir, 'entry'),
+            '/entry.mjs': resolve(nuxt.options.appDir, nuxt.options.dev || !nuxt.options.experimental.asyncEntry ? 'entry' : 'entry.async'),
```
pi0, line 53: Any reason to add . dev check still? Considering async entry is experimental we could let people trying async entry issues

packages/webpack/src/presets/base.ts
```diff
@@ -23,7 +23,7 @@ function baseConfig (ctx: WebpackConfigContext) {
 
   ctx.config = {
     name: ctx.name,
-    entry: { app: [resolve(options.appDir, 'entry')] },
+    entry: { app: [options.experimental.asyncEntry ? resolve(options.appDir, 'entry.async') : resolve(options.appDir, 'entry')] },
```
pi0, line 26: Refactor: we could shorten logic by adding asyncEntry chrck inside template literal:

```suggestion
    entry: { app: [resolve(options.appDir,  options.experimental.asyncEntry? 'entry.async' : 'entry') ] },
```

packages/vite/src/vite.ts
danielroe:
```suggestion
            '/entry.mjs': resolve(nuxt.options.appDir, nuxt.options.experimental.asyncEntry ? 'entry.async' : 'entry'),
```

pi0 (approved):

packages/vite/src/vite.ts
danielroe: The only point to make is that we know this will be buggy.

pi0:
(i guess broke linter 🙈)

PR #3392, opened 2022-02-25
test: add more tests

antfu:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "play": "yarn run nuxi dev playground",
     "release": "yarn && yarn lint && FORCE_COLOR=1 lerna publish -m \"chore: release\" && yarn stub",
     "stub": "lerna run prepack -- --stub",
-    "test:fixtures": "JITI_ESM_RESOLVE=1 vitest",
+    "test:fixtures": "JITI_ESM_RESOLVE=1 vitest test",
     "test:fixtures:webpack": "TEST_WITH_WEBPACK=1 yarn test:fixtures",
     "test:types": "yarn run nuxi prepare test/fixtures/basic && cd test/fixtures/basic && npx vue-tsc --noEmit",
     "test:unit": "JITI_ESM_RESOLVE=1 yarn vitest packages",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "play": "yarn run nuxi dev playground",
     "release": "yarn && yarn lint && FORCE_COLOR=1 lerna publish -m \"chore: release\" && yarn stub",
     "stub": "lerna run prepack -- --stub",
-    "test:fixtures": "JITI_ESM_RESOLVE=1 vitest test",
+    "test:fixtures": "JITI_ESM_RESOLVE=1 vitest",
```
pi0, line 22: This causes all unit tests run as well:

image

packages/test-utils/src/server.ts
```diff
@@ -24,7 +24,7 @@ export async function listen () {
   const port = await getPort({ host, random: true })
 
   ctx.url = 'http://localhost:' + port
-  execa('node', [
+  ctx.listenProcess = execa('node', [
```
pi0, line 27: (small suggestion: maybe call it `serverProcess`?)

test/fixtures/basic/pages/index.vue
```diff
@@ -4,11 +4,16 @@
       <Title>Basic fixture</Title>
     </Head>
     <h1>Hello Nuxt 3!</h1>
-    <div>Config: {{ $config.testConfig }}</div>
+    <div>RuntimeConfig: {{ config.testConfig }}</div>
```
pi0, line 7: We could start splitting features tests into their own route as the tests grow, it might be easier to add more tests to each

package.json
antfu: Oh that's a mistake. Was planned to introduce a `--dir` arg to Vitest for this instead of path matching.

antfu:
Maybe we could do that in future PRs?

package.json
pi0: Can we use a temporary path matching pattern for now?

PR #3396, opened 2022-02-25
fix(nuxt3): share scanned components with loader

### LINKED ISSUE #3393: Auto import of components broken
### Environment

- Operating System: `Linux`
- Node Version:     `v16.14.0`
- Nuxt Version:     `3.0.0-27429889.e45a780`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/jiblett1000/test

### Describe the bug

With the latest update on npm as of this writing, auto importing of components seems to be completely broken. I've linked a reproduction with a fresh nuxt3 install with a single component in the components directory. Running yarn dev results in this error:

``
[Vue warn]: Failed to resolve component: Test
If this is a native custom element, make sure to exclude it from component resolution via compilerOptions.isCustomElement.
``

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #3393

### ❓ Type of change

### 📚 Description

Regression introduced in https://github.com/nuxt/framework/pull/3366. I also fix a type error requiring empty `src: ''` when adding plugin templates, which I spotted at the same time.

### 📝 Checklist


packages/nuxt3/src/components/module.ts
```diff
@@ -135,7 +133,7 @@ export default defineNuxtModule<ComponentsOptions>({
       }
     })
 
-    const loaderOptions = { getComponents: () => components }
+    const loaderOptions = { getComponents: () => options.components }
```
pi0, line 136: Refactor: We should avoid using options as a mutable context as much as possible.

packages/kit/src/plugin.ts
```diff
@@ -68,13 +68,13 @@ export function addPlugin (_plugin: NuxtPlugin | string, opts: AddPluginOptions
 /**
  * Adds a template and registers as a nuxt plugin.
  */
-export function addPluginTemplate (plugin: NuxtPluginTemplate | string, opts: AddPluginOptions = {}): NuxtPluginTemplate {
+export function addPluginTemplate (plugin: Partial<NuxtPluginTemplate> | string, opts: AddPluginOptions = {}): NuxtPluginTemplate {
```
pi0, line 71: What about instead remove `src` from input type? (rather than making whole interface partial)

packages/nuxt3/src/components/module.ts
danielroe: would you like me to rewrite the module in this pr?

pi0: No, we could do it later :)

danielroe:
@pi0 I believe this is safe to merge; the test failure is resolved by upgrading `get-port-please`

Atinux (approved):

PR #3422, opened 2022-02-28
fix(nuxt3): resolve rootDir & srcDir of layers

kevinmarrec:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Resolve `rootDir` & `srcDir` for Nuxt extended layers.

This is an enhancement for `extends` feature : https://github.com/nuxt/framework/issues/3222

Issue #3222, opened 2022-02-14
Extends support


pi0:
Feature:
- [x] Initial support (#3008)
- [x] `components/` dir (#3108)
- [ ] `composables/` dir
- [ ] `plugins/` dir
- [ ] `pages/` dir
- [ ] `middleware/` dir
- [ ] `server/` dir
- [ ] Merge `hook` config key

Documentation:
- [ ] Basic usage 
- [ ] Theming
- [ ] Multi-app

Upstream: 
- [nitropack] hooks for extending dirs
- [c12] install cloned dependencies

Tahul:
Hey there 👋

We've made a review of `extends` feature with @kevinmarrec so I can document it!

I would like to brainstorm few ideas with you, before beginning any changes.

Aside from these points, we've tried the extends feature together with @kevinmarrec and it went really well!

I will start creating a PR to document this with appropriate examples as you mentioned in the first message. 🙂

### Extends configuration

With the current `extends` implementation, once you pass a path to be extended, Nuxt will try to grab every available feature from this extended directory.

It is really great to have such behavior as it makes `extends` a real plug & play feature, but having some flexibility over how we do extend our project would be really great.

What I suggest is to have a similar way to configure your extends as components does:

```ts
defineNuxtConfig({
   extends: [
		'./base', // Would be fully extended
		{
			path: './ui',
			middleware: false, // Would skip middleware extending
			pages: false, // Would skip pages extending
			// All of the rest are set to `true` by default
		}
   ]
})
```

That would maybe include a PR to `c12` package in order to support this.

### Extends pages prefixing

One cool feature would be to be able to rewrite the base path of pages from an extend target.

Let's say you extend your project from a `blog` theme that has default pages exposed.

The blog theme might expose index pages and such from the root of its `pages` directory, so that would be overwritten by your own website.

What you might want then is to have these blog pages nested under a `/blog` prefix in your project.

We could use the previous configuration object to specify this:

```ts
defineNuxtConfig({
   extends: [
		'./base', // Pages would be merged with project-level pages
		{
			path: './blog',
			pagesPrefix: '/blog' // Pages would be nested under `/blog` route prefix
		}
   ]
})
```

### Low to high instead of high to low priority management

One thing we noticed during the review of `extends` is that currently the priority is handled in a `high to low` manner:

```ts
defineNuxtConfig({
   extends: [
		'./ui', // This will override `./base`
		'./base'
   ]
})
```

[...]

PR #3423, opened 2022-02-28
feat!(nuxt3): extends support for composables/ directory

kevinmarrec:
### ❓ Type of change

### 📚 Description

Following up #3222 (`extends` support progress) this PR allows extending `composables/` directory for extended config layers

### 🔨 Misc

Closes #3422 which is already implemented within this PR


packages/kit/src/loader/config.ts
```diff
@@ -37,6 +38,18 @@ export async function loadNuxtConfig (opts: LoadNuxtConfigOptions): Promise<Nuxt
 
   nuxtConfig._nuxtConfigFile = configFile
   nuxtConfig._nuxtConfigFiles = [configFile]
+
+  // Resolve `rootDir` & `srcDir` of layers
+  for (const layer of layers) {
+    layer.config.rootDir = layer.cwd
+    layer.config.srcDir = layer.config.srcDir
+      ? resolve(layer.cwd, resolveAlias(layer.config.srcDir, {
```
pi0, line 46: `srcDir` is intended to be either relative path to rootDir or an absolute path. Why we would need this?

packages/nuxt3/src/auto-imports/module.ts
```diff
@@ -34,8 +34,13 @@ export default defineNuxtModule<AutoImportsOptions>({
     // composables/ dirs
     let composablesDirs = [
       join(nuxt.options.srcDir, 'composables'),
-      ...options.dirs
+      ...options.dirs,
+      ...nuxt.options._extends
+        .map(layer => ['composables', ...layer.config.autoImports?.dirs ?? []]
```
pi0, line 39: For better code readability, can we please simply push to `composablesDirs` array? (with a  a `for of` loop maybe so we could add additional normalization logic if needed)

packages/kit/src/loader/config.ts
```diff
@@ -37,6 +38,18 @@ export async function loadNuxtConfig (opts: LoadNuxtConfigOptions): Promise<Nuxt
 
   nuxtConfig._nuxtConfigFile = configFile
   nuxtConfig._nuxtConfigFiles = [configFile]
+
+  // Resolve `rootDir` & `srcDir` of layers
+  for (const layer of layers) {
+    layer.config.rootDir = layer.cwd
```
pi0, line 44: We might keep overriden `rootDir` value by `layer.config` if it exists:

```suggestion
    layer.config.rootDir = resolve(layer.cwd, layer.config.rootDir)
```

kevinmarrec:
@pi0 Updated the PR from your comments, we'll still need to figure out why `bridge` test fails, the output doesn't help, maybe you have a clue :).

Issue #2998, opened 2022-01-30
.nuxtignore filter support for Nuxt3


danielroe:
I think we should add back nuxtignore filter support for Nuxt3 (and plugin imports). This way we can ignore unwanted paths from plugins.

_Originally posted by @pi0 in https://github.com/nuxt/framework/issues/2163#issuecomment-981612377_

Issue #240, opened 2021-06-18
Make watcher's `ignore` configurable


antfu:
https://github.com/nuxt/framework/blob/62b19c63f33bb8d6f208c934dc1cb16390992459/packages/nuxt3/src/builder.ts#L32-L36

Related #239

PR #3424, opened 2022-02-28
feat: `.nuxtignore` support and `isIgnored` kit utility

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/240, resolves https://github.com/nuxt/framework/issues/2998, resolves https://github.com/nuxt/framework/issues/2163

### ❓ Type of change

### 📚 Description

This PR adds support for `ignore`, `ignorePrefix`, and `ignoreOptions`. These options are also respected when scanning for components, pages, layouts, plugins, middleware, composables and other instances where `resolveFiles` is called.

### 📝 Checklist


packages/schema/src/types/components.ts
```diff
@@ -33,6 +33,10 @@ export interface ScanDir {
    * Ignore patterns that will be run against specified path.
    */
   ignore?: string[]
+  /**
+   * Function to filter files - for more control than the `ignore` pattern.
+   */
+  ignoreFile?: (pathname: string) => boolean
```
pi0, line 39: What would be the usecase of this new option? Using non-hooks can be tricky if one module depends on it and user or another module override.

danielroe: Solely to pass the nuxtignore filter down to the scanning function.

packages/kit/src/ignore.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export interface NuxtIgnore {
+  shouldIgnoreFile: (pathname: string) => boolean
+}
+
+/**
+ * Return a filter function to filter an array of paths
+ */
+export function useIgnore (nuxt = useNuxt()): NuxtIgnore {
+  nuxt._ignore = nuxt._ignore || createIgnore()
```
pi0, line 14: Since we are caching context, why not simply export `shouldIgnoreFile` utility from kit?

packages/schema/src/types/components.ts
pi0: I see but ignore utility from kit is shared. scanner could directly consume it. `ignore[]` option per dir was to workaround removing unwanted component files without access to modify `nuxtignore`.

packages/kit/src/ignore.ts
```diff
@@ -0,0 +1,33 @@
...
+import ignore, { Ignore } from 'ignore'
+import { join, relative } from 'pathe'
+import { useNuxt } from './context'
+
+/**
+ * Return a filter function to filter an array of paths
+ */
+export function isIgnored (pathname: string): boolean {
+  const nuxt = useNuxt()
+  nuxt._ignore = nuxt._ignore || createIgnore()
```
pi0, line 11: Look amazing but only last thing do we still need separation of `createIgnore` logic? :D

packages/kit/src/ignore.ts
```diff
@@ -0,0 +1,27 @@
...
+import { join, relative } from 'pathe'
+import { useNuxt } from './context'
+
+/**
+ * Return a filter function to filter an array of paths
+ */
+export function isIgnored (pathname: string): boolean {
+  const nuxt = useNuxt()
+
+  if (!nuxt._ignore) {
```
pi0, line 12: Maybe we could improve it later, `useNuxt()` silently returns `null` when context is not available. This causes this line fail with strange error like cannot find prop `_ignore` of undefined. We could maybe make an error by default for `useNuxt` or allow an optional argument to throw with a custom message

PR #3428, opened 2022-02-28
fix: sourcemap support for `nuxt:vite-relative-asset`

antfu:
### ❓ Type of change

### 📝 Checklist


packages/vite/src/plugins/dynamic-base.ts
```diff
@@ -42,51 +43,83 @@ export const DynamicBasePlugin = createUnplugin(function (options: DynamicBasePl
...
     },
     enforce: 'post',
-    transform (original, id) {
-      let code = original
+    transform (code, id) {
+      const s = new MagicString(code)
       if (options.globalPublicPath && id.includes('entry.ts')) {
-        code = 'import { joinURL } from "ufo";' +
-          `${options.globalPublicPath} = joinURL(NUXT_BASE, NUXT_CONFIG.app.buildAssetsDir);` + code
+        s.prepend(`import { joinURL } from "ufo";${options.globalPublicPath} = joinURL(NUXT_BASE, NUXT_CONFIG.app.buildAssetsDir);`)
```
danielroe, line 49: We could probably add newlines now that we are generating a source map. (Previously it was to prevent line numbers from changing.)

packages/vite/src/plugins/dynamic-base.ts
```diff
@@ -42,51 +43,83 @@ export const DynamicBasePlugin = createUnplugin(function (options: DynamicBasePl
...
       }
 
       const assetId = code.match(VITE_ASSET_RE)
       if (assetId) {
-        code = 'import { joinURL } from "ufo";' +
-          `export default joinURL(NUXT_BASE, NUXT_CONFIG.app.buildAssetsDir, "${assetId[1]}".replace("/__NUXT_BASE__", ""));`
+        s.overwrite(0, code.length, `import { joinURL } from "ufo";export default joinURL(NUXT_BASE, NUXT_CONFIG.app.buildAssetsDir, "${assetId[1]}".replace("/__NUXT_BASE__", ""));`)
       }
 
       if (code.includes('NUXT_BASE') && !code.includes('const NUXT_BASE =')) {
```
danielroe, line 57: If we've added `NUXT_BASE` in the previous few lines, then we want this to trigger too.

packages/vite/src/plugins/dynamic-base.ts
```diff
@@ -42,51 +43,83 @@ export const DynamicBasePlugin = createUnplugin(function (options: DynamicBasePl
...
 
-      if (code === original) { return }
-      return code
+      if (code === s.toString()) {
+        return
+      }
+      return {
+        code: s.toString(),
+        map: s.generateMap()
+      }
```
danielroe, line 98: should we make it configurable whether or not we generate maps?

packages/vite/src/plugins/dynamic-base.ts
pi0: Doesn't it increase sourcemap size?

packages/vite/src/plugins/dynamic-base.ts
antfu: I know how to read the config from Vite but not sure about Webpack nor Nuxt. Do we have a config for it?

danielroe: I would suggest a global `build.sourcemap` but @pi0 may have another opinion - and I think he might recommend against adding more options to `build`.... ideas welcome for another location.

pi0: We could add a top-level `sourceMap` option with nested options for quality control of sourceMap and enable/disable for different areas. But let's do it in next steps.

PR #3462, opened 2022-03-01
feat(nuxt3): extends support for plugins/ directory

kevinmarrec:
### ❓ Type of change

### 📚 Description

Following up #3222 (`extends` support progress) this PR allows extending `plugins/` directory for extended config layers

### TO DO
- [ ] Merge https://github.com/nuxt/framework/pull/3423 beforehand so we get the `srcDir` thing for layers
- [x] Extends support for `plugins` directory
- [ ] Extends support for `plugins` config key in layers => Need to resolve aliases


pi0:
For supporting aliases in layer, we might introduce new named layer aliases. Changing resolver behavior and aliases might be tricky to be handled properly and sometimes we actualyl do like to allow overriding something from user dir (~/plugins/foo being overriden)

```js
  name: 'myTheme'
  plugins: [
    '~myTheme/plugins/foo'
  ]
})
````

Issue #3161, opened 2022-02-10
Migrating to nitropack


pi0:
Following up initial RFC #1690, nitropack project removes complexity from Nuxt core for server handling and allows adding well tested features to the core.

Starting with a soft migration phase, integration is behind a flag for Nuxt 3 and `@nuxtjs/nitro` will be maintained for a while until stabilizing nitropack support. 

How to opt-in (pending #3160):

- Method1: Using env `EXPERIMENT_NITROPACK=1

```
EXPERIMENT_NITROPACK=1 yarn nuxi build`
```

- Method2: Set `experimentNitropack: true` in `nuxt.config`

```js [nuxt.config.js]

export default defineNuxtConfig({ ​
​experimentNitropack: true
})
```

Roadmap:

- [ ] Enable behind a flag (#3160)
- [ ] Support bridge behind a flag
- [ ] Public [unjs/nitropack](https://github.com/unjs/nitropack) repo
- [ ] Migrate [open issues with nitro label](https://github.com/nuxt/framework/issues?q=is%3Aopen+is%3Aissue+label%3Anitro)
- [ ] Sync last nitro related PRs to upstream
- [ ] Transfer nitro preset tests to nitropack 
- [ ] Integrate route caching rules (#560)

Rigo-m:
Some checks needs to be updated @pi0 ;) waitin for unjs/nitropack to be published ❤️

PR #2795, opened 2022-01-19
feat(vite): experimental `vite-node` support

antfu:
Close #645

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Changes

- Now we no longer need the dev bundler and no need to bundle `server.mjs` anymore
- And we also don't need the watcher since everything is fetched on-demand (via HTTP requests)
- Client manifests are generated from `server.mjs` which is from the requested module tree, and also on-demand
- Now we could have stacktraces for server errors with correct sourcemap :)

### TODO

- [ ] Do a measurement and compare with a Nuxt 3 codebase to see the improvements in numbers
- [ ] Migrate for bridge (in another PR)

PR #3464, opened 2022-03-01
feat(nitro, nuxt3): allow handling otherwise unhandled runtime errors

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/559

### ❓ Type of change

### 📚 Description

This PR implements or documents the following:

- [x] handling errors during the Vue rendering lifecycle (SSR + SPA)
- [x] handling server and client startup errors (SSR + SPA)
- [x] Allow hooking to catch error events (useful for error reporters)
- [x] (Isomorphic) Rendering a production HTML page for production errors and allow customization
- [x] Showing errors in console or proxy to the dev tools (similar to nuxt2 window SSR logs)
- [x] Allow gracefully handle component level errors yet use error reporting framework

##### Iterative improvements

- [x] improve error serialisation, perhaps sharing logic with existing nitro error json serialiser
- [ ] improve error page from `@nuxt/design` to make a more useful debugging HTML page for developers: https://github.com/nuxt/ui/issues/66
- [ ] improve console logging of errors: https://github.com/nuxt/framework/issues/3592
   
##### Nitro-side tasks still to complete

- [x] Errors during API or nitro server lifecycle
- [x] Rendering JSON friendly server errors
- [x] Allow gracefully handle API errors (?)

I've ideas for this - basically skipping deferred most of the nitro side until https://github.com/nuxt/framework/issues/3161, but I'll port the changes here in a follow-up PR to https://github.com/unjs/nitropack when we're happy with it.

##### Other issues

- [ ] Handling source maps properly: https://github.com/nuxt/framework/pull/2795

### 📝 Checklist


packages/nitro/src/runtime/app/render.ts
```diff
@@ -86,15 +96,17 @@ export async function renderMiddleware (req, res: ServerResponse) {
...
-  const rendered = await renderToString(ssrContext)
+  let rendered = await renderToString(ssrContext)
 
   // Handle errors
-  if (ssrContext.error) {
-    throw ssrContext.error
+  if (ssrContext.errors.length) {
+    // Render error page instead
+    rendered = await renderToString(ssrContext)
   }
```
Aareksio, line 110: Am I missing something or it essentially calls isomorphic function `renderToString` with the same arguments again?

Also without a `throw` I do not think we're able to pass `statusCode` and `statusMessage` to h3.

misaon:
Awesome 👍🏼

packages/nitro/src/runtime/app/render.ts
pi0: Indeed. I believe throwing error and doing a full rerender makes more sense..

Aareksio: After closer inspection, it seems I introduced unnecessary confusion:
- `renderToString` is not isomorphic, it modifies `ssrContext`
- there is a nuxt 2 code below in L138-L139 setting `res.statusCode`

Current code looks like a good solution.

---

Since `renderMiddleware` is already in control of `res` object, I suggest editing the L188-L189 for nuxt 3:

```diff
- const error = ssrContext.nuxt && ssrContext.nuxt.error
+ const error = payload.error || payload.errors[0]
res.statusCode = error ? error.statusCode : 200
```

---

It may be necessary to introduce and document `error.statusMessage` handling, since `h3`, unlike `express`, does not provide default status messages for HTTP codes.

pi0:
Checklist from today's call with @danielroe:

- [ ]  Use single error in context (not array of them)
- [ ]  Remove logging of errors (next steps)
- [ ] Rename internal `<nuxt-error>` to `<nuxt-error-page>`
- [ ] (try) use `useState('_error')` instead of global `error` for hydration purpose
- [ ] Use `nitro/error.ts` and `$fetch` to generate error page instead of `nitro/render.ts` (handles API errors as well)
- [ ]  Error handling on client side: only trigger full screen error on 404 or during navigation or when manually triggered (`throwError()`, `showError()`)

Next steps:
  - [ ]  `<NuxtErrorBoundary>` which handles errors and renders them:
        - [ ]  #default - main content
        - [ ]  #error - shows error (and receives it via scope)
        - [ ]  disables the error on navigation
- [ ]  New utility to parse and normalize stack traces, node entry + browser entry
- [ ]  Show SSR logs in client console (including errors and warns) via consola v3 (WIP: https://github.com/unjs/consola/pull/135)

packages/nitro/src/runtime/app/render.ts
```diff
@@ -70,7 +70,9 @@ function renderToString (ssrContext) {
 }
 
 export async function renderMiddleware (req, res: ServerResponse) {
-  let url = req.url
+  // Whether we're rendering an error page
+  const isErrorHandler = req.method === 'GET' && req.body
```
pi0, line 74: body with getting seems a little bit unsemantic. Can't we use a specific header instead?

packages/nitro/src/runtime/app/render.ts
```diff
@@ -86,17 +88,31 @@ export async function renderMiddleware (req, res: ServerResponse) {
...
+  if (error && !isErrorHandler) {
+    // trigger a re-render by onError handler
+    throw error
   }
 
+  // TODO: nuxt3 should not reuse `nuxt` property for different purpose!
+  const payload = ssrContext.payload /* nuxt 3 */ || ssrContext.nuxt /* nuxt 2 */
+
+  // Pass error to the client
+  payload.error = error
```
pi0, line 114: We could pick error object from SSR Context directly in app and pass to payload if needed. Not involving nitro render logic for payload modification.

packages/nitro/src/runtime/server/error.ts
```diff
@@ -56,8 +58,12 @@ export function handleError (error, req: IncomingMessage, res: ServerResponse) {
   }
 
   // HTML response
-  const errorTemplate = is404 ? error404 : (isDev ? errorDev : error500)
-  const html = errorTemplate(errorObject)
+  const fallbackTemplate = is404 ? error404 : (isDev ? errorDev : error500)
+  const { body: html } = await localCall({
+    url: '/_error',
+    headers: req.headers,
+    body: { url: req.url, error: errorObject }
```
pi0, line 65: Can we please use `$fetch` API instead to render the error page? Caching localCall directly would be tricky also helps to use more valid semantics (no GET with Body!)

packages/nuxt3/src/app/components/nuxt-error-page.vue
```diff
@@ -0,0 +1,19 @@
...
+import { error500, errorDev, error404 } from '@nuxt/design'
+
+const props = defineProps({
+  error: Object
+})
+
+const err = props.error
+const handler = err.statusCode === 404 ? error404 : process.dev ? errorDev : error500
+const html = handler({
+  description: err.message,
```
pi0, line 16: Just checking do we still have stack trace rendered?

packages/nuxt3/src/app/components/nuxt-root.vue
```diff
@@ -1,22 +1,34 @@
...
+// vue:setup hook
+const results = nuxtApp.hooks.callHookWith(hooks => hooks.map(hook => hook()), 'vue:setup')
+if (process.dev && results && results.some(i => i && 'then' in i)) {
+  console.error('[nuxt] Error in `vue:setup`. Callbacks must be synchronous.')
 }
+
+// error handling
+const error = useError()
+onErrorCaptured((err, target) => {
+  const results = nuxtApp.hooks.callHookWith(hooks => hooks.map(hook => hook(err, target)), 'vue:error')
```
pi0, line 26: If result of hook handler is async promise (which is not uncommon for error reporting) we could have better behavior to capture and at least show a console error about it while with sync call, we just ignore this.

packages/nuxt3/src/app/components/nuxt-root.vue
```diff
@@ -1,22 +1,34 @@
...
+  if (process.dev && results && results.some(i => i && 'then' in i)) {
+    console.error('[nuxt] Error in `vue:error`. Callbacks must be synchronous.')
+  }
+  if (process.server) {
+    callWithNuxt(nuxtApp, throwError, [err])
```
pi0, line 31: Just checking: Isn't vue-app instance available at this stage to use for throwError composable?

packages/nuxt3/src/app/composables/error.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export const throwError = (_err: string | Error) => {
+  const nuxtApp = useNuxtApp()
+  const error = useError()
+  const err = typeof _err === 'string' ? new Error(_err) : _err
+  if (process.server) {
+    nuxtApp.ssrContext.error = nuxtApp.ssrContext.error || err
+  } else {
+    nuxtApp.callHook('app:error', err, nuxtApp)
```
pi0, line 15: It might be helpful for server side plugins hooking into `app:error` any reason limit to client?

packages/nuxt3/src/app/composables/error.ts
```diff
@@ -0,0 +1,30 @@
...
+    error.value = error.value || err
+  }
+
+  return err
+}
+
+export const clearError = async (redirect?: string) => {
+  const nuxtApp = useNuxtApp()
+  const error = useError()
+  nuxtApp.callHook('app:error:handled')
```
pi0, line 25: Composable name is not in sync with hook. Maybe call it `handlError`/`resolveError` or rename hook to `app:error:cleared` with direct option also being passed?

packages/nuxt3/src/app/composables/error.ts
```diff
@@ -0,0 +1,30 @@
...
+    nuxtApp.ssrContext.error = nuxtApp.ssrContext.error || err
+  } else {
+    nuxtApp.callHook('app:error', err, nuxtApp)
+    error.value = error.value || err
+  }
+
+  return err
+}
+
+export const clearError = async (redirect?: string) => {
```
pi0, line 22: I would use an object interface `({ redirect? })` for possible other actions in the future

packages/nuxt3/src/app/entry.ts
```diff
@@ -18,12 +18,20 @@ if (process.server) {
...
     const nuxt = createNuxtApp({ vueApp, ssrContext })
-
-    await applyPlugins(nuxt, plugins)
-
-    await nuxt.hooks.callHook('app:created', vueApp)
-
-    return vueApp
+    if (ssrContext.error) {
+      // Skip user code
+      return vueApp
```
pi0, line 23: This approach has some benefits but also can make error page broken if depends on anything global injected by plugins even something like i18n... Should we at least document this?

packages/nuxt3/src/app/entry.ts
```diff
@@ -43,19 +51,28 @@ if (process.client) {
...
     })
 
-    vueApp.mount('#__nuxt')
-
-    await nuxt.hooks.callHook('app:mounted', vueApp)
-    await nextTick()
+    try {
+      await applyPlugins(nuxt, plugins)
+    } catch (err) {
+      nuxt.hooks.hookOnce('app:error:handled', () => applyPlugins(nuxt, plugins))
```
pi0, line 61: Revaluation of plugins can cause unwanted behavior. Why we need to do this?

packages/nuxt3/src/app/entry.ts
```diff
@@ -43,19 +51,28 @@ if (process.client) {
...
+      await nuxt.callHook('app:error', err, nuxt)
```
pi0, line 62: Hooks are already registered with same `nuxt` context. We don't need to pass it again.
```suggestion
      await nuxt.callHook('app:error', err)
```

danielroe: This was so the error handling hook receives nuxt to handle. Without this, it can't access `nuxt`.

packages/nuxt3/src/app/entry.ts
danielroe: We only re-evaluate them if there was an error evaluating them the first time. The issue I'm trying to solve is that once the error is handled there's no way to call the plugins again (for example, this would be required to get `$router` working again, if there was an error the first time through).

You're right, if we got part-way through before an error occurred, we'd have an issue with redefining a getter (for injections). Any merit instead in tracking how far we got and 'resuming' plugin execution?

packages/nuxt3/src/app/entry.ts
pi0: How then hook can be registered then? :D (`nuxt.hook('app:error', err => { /* we do already have access to `nuxt` })

packages/nuxt3/src/app/entry.ts
danielroe: Agreed. But it should already be documented:

https://github.com/nuxt/framework/blob/b032e5d44992748b4a2ec027899fab0554f1e771/docs/content/3.docs/1.usage/8.error-handling.md#L52-L54

packages/nuxt3/src/app/entry.ts
danielroe: 😆 Fair point!

packages/nuxt3/src/app/components/nuxt-root.vue
danielroe: Not sure I understand the question.

packages/nuxt3/src/app/components/nuxt-root.vue
danielroe: Good point. Let's just ignore async calls?

packages/nuxt3/src/app/entry.ts
pi0: Tracking failed plugins and only reevaluate them would be a better idea but we still lose order and are potentially missing hooks.  Is it really necessary to consider supporting this method vs simply resolving error by redirect or reload?

packages/nuxt3/src/app/components/nuxt-root.vue
pi0: useNuxtApp tries vue-instance as well to access `nuxtApp` context. I was wondering if it is a vue-lifecyecle error, vue instance wouldn't be available for composables?

packages/nitro/src/runtime/app/render.ts
danielroe: I chose this because it was unsemantic and wouldn't be impossible to replicate by accident.

Refactoring now to use a header instead, but this is also unsemantic as we're passing a payload within a header, which _also_ isn't technically allowed.

pi0: We could use a special query string preserving get or actually makde a POST `$fetch` call with specified path and body param like `__nuxt_error__`

packages/nuxt3/src/app/composables/error.ts
Aareksio, line 22: What are your thoughts on hooking to router event for easier implementation of common scenarios the same way as in Nuxt 2, such as 404 - Not Found:

```html
<p>Page Not Found</p>
<NuxtLink to="/">Back to Home</NuxtLink>
```

docs/content/3.docs/1.usage/8.error-handling.md
```diff
@@ -0,0 +1,90 @@
...
+## Rendering an error page
+
+When Nuxt encounters a fatal error, whether during the server lifecycle, or when rendering your Vue application (both SSR and SPA), it will either render a JSON response (if requested with `Accept: application/json` header) or an HTML error page.
+
+You can customize this error page by adding `~/error.vue` in the source directory of your application, alongside `app.vue`. This page has a single prop - `error` which contains an error for you to handle.
+
+When you are ready to remove the error page, you can call the `clearError` helper function, which takes an optional path to redirect to (for example, if you want to navigate to a 'safe' page).
+
+::alert{type="warning"}
+Make sure not to use anything dependent on Nuxt plugins, such as `$route` or `useRouter`, as these are not guaranteed to have run until you clear the error.
```
Aareksio, line 53: Isn't this a big regression from Nuxt 2? 

Some examples:
- i18n, @pi0 already mentioned it in other comment
- support chat plugin
- configuration fetched via plugin (eg. A/B tests, navigation items, session data, auth)

I imagine we would have to bake all those required and expected features directly into `~/error.vue`. Forcing us either to commit to maintaining 2 versions of the same feature, or write additional abstraction for sole purpose of handling error page.

From outsider (user) perspective it is hard to see how the benefits outweigh costs, especially as plugins are executed just fine in Nuxt 2.

packages/nuxt3/src/app/components/nuxt-error-page.vue
danielroe: Yes. It looks like this:
```js
const error = {
  statusCode: 500,
  statusMessage: 'Internal Server Error',
  message: 'error in setup',
  description: '\n' +
    '    <h1>error in setup</h1>\n' +
    '    <pre><span class="stack">at setup (file://./.nuxt/dist/server/server.mjs:9420:13)</span>\n' +
    '<span class="stack">at _sfc_main.setup (file://./.nuxt/dist/server/server.mjs:9555:23)</span>\n' +
    '<span class="stack internal">at callWithErrorHandling (/project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:157:22)</span>\n' +
    '<span class="stack internal">at setupStatefulComponent (/project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:6977:29)</span>\n' +
    '<span class="stack internal">at setupComponent (/project/node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:6933:11)</span>\n' +
    '<span class="stack internal">at renderComponentVNode (/project/node_modules/@vue/server-renderer/dist/server-renderer.cjs.js:198:17)</span>\n' +
    '<span class="stack internal">at Module.ssrRenderComponent (/project/node_modules/@vue/server-renderer/dist/server-renderer.cjs.js:629:12)</span>\n' +
    '<span class="stack">at default (file://./.nuxt/dist/server/server.mjs:9274:37)</span>\n' +
    '<span class="stack internal">at Module.<anonymous> (/project/node_modules/@vue/server-renderer/dist/server-renderer.cjs.js:958:13)</span>\n' +
    '<span class="stack">at Generator.next (<anonymous>)</span></pre>\n' +
    '    '
}
```

docs/content/3.docs/1.usage/8.error-handling.md
danielroe: The plugins will run _unless_ the error is thrown by your plugin, in which case we gracefully recover...

packages/nuxt3/src/app/entry.ts
danielroe: Implemented resumable plugins. Let me know if you would prefer we remove entirely, as I do think it's likely an edge case.

packages/nuxt3/src/app/composables/error.ts
danielroe: Nice idea. Implemented.

packages/nuxt3/src/app/components/nuxt-root.vue
danielroe: Interestingly, the vue instance is available via `getCurrentInstance()`.

docs/content/3.docs/1.usage/8.error-handling.md
Aareksio: I see it's been modified in 410e4fd886e847e88d0e78ba9b763e9e18eef16f. 

I am a little worried with recommended `~/plugins` auto registration, the order of plugins is not explicitly determined. This may lead to functionality failing in unexpected manner, eg. the same code may behave differently, depending on file names (assuming plugins are sorted by file name):

```
- plugins/
  |- i18n.js
  |- throw-error.js
```

vs

```
- plugins/
  |- throw-error.js
  |- universal-i18n.js <-- the same code as i18n.js
```

As a developer, I'd like to ensure the basic functionality set is available, even if some plugin down the line fails. This is already possible after beforementioned commit. I think it may be worth documenting this close to `plugins`. Taking `i18n` as an example, I can see Vue crashing irrecoverably when encountering `{{ $t('...') }}` in template of layout component.

packages/nuxt3/src/app/components/nuxt-error-page.vue
pi0: Okay then should we preserve message + description? :D

packages/nitro/src/runtime/server/error.ts
```diff
@@ -56,8 +58,17 @@ export function handleError (error, req: IncomingMessage, res: ServerResponse) {
   }
 
   // HTML response
-  const errorTemplate = is404 ? error404 : (isDev ? errorDev : error500)
-  const html = errorTemplate(errorObject)
+  const url = withQuery('/__internal/__error', {
```
pi0, line 61: We could use something like `/_nitro/error` "internal" is unclear if seen in stack traces and another _ for error seems unnecessary...

packages/nitro/src/runtime/server/error.ts
```diff
@@ -56,8 +58,17 @@ export function handleError (error, req: IncomingMessage, res: ServerResponse) {
...
+    url: req.url,
+    error: JSON.stringify(errorObject)
```
pi0, line 63: What about putting URL also in error object to simplify passing it around

packages/nitro/src/runtime/server/error.ts
```diff
@@ -56,8 +58,17 @@ export function handleError (error, req: IncomingMessage, res: ServerResponse) {
...
+  })
+  const html = await $fetch(url, {
+    headers: req.headers as any
```
pi0, line 66: This can make it harder to implement caching support for error pages (error page is not rendered purely based on errorObject but also varies per req). If not necessary i would not pass headers.

packages/nitro/src/runtime/server/error.ts
```diff
@@ -56,8 +58,17 @@ export function handleError (error, req: IncomingMessage, res: ServerResponse) {
...
+  }).catch(async () => {
+    const fallbackTemplate = await import('@nuxt/design').then(r => is404 ? r.error404 : (isDev ? r.errorDev : r.error500))
```
pi0, line 68: I guess this would avoid tree-shaking needed imports by rollup. Same as vue version, we could import fallbacks directly from top.

packages/nuxt3/src/app/components/nuxt-error-page.vue
pi0: We are nott using `description` field anymore...

packages/nuxt3/src/app/components/nuxt-root.vue
```diff
@@ -1,22 +1,31 @@
...
+// vue:setup hook
+const results = nuxtApp.hooks.callHookWith(hooks => hooks.map(hook => hook()), 'vue:setup')
+if (process.dev && results && results.some(i => i && 'then' in i)) {
+  console.error('[nuxt] Error in `vue:setup`. Callbacks must be synchronous.')
 }
+
+// error handling
+const error = useError()
+onErrorCaptured((err, target, info) => {
+  nuxtApp.hooks.callHook('vue:error', err, target, info).catch(e => console.error('[nuxt] Error in `vue:error` hook', e))
```
pi0, line 26: Refactor: Maybe use a better name for `e` like `hookError`? :D 
```suggestion
  nuxtApp.hooks.callHook('vue:error', err, target, info).catch(hookError => console.error('[nuxt] Error in `vue:error` hook', hookError))
```

packages/nuxt3/src/app/entry.ts
```diff
@@ -43,19 +47,28 @@ if (process.client) {
...
     })
 
-    vueApp.mount('#__nuxt')
-
-    await nuxt.hooks.callHook('app:mounted', vueApp)
-    await nextTick()
+    try {
+      await applyPlugins(nuxt, plugins)
+    } catch (err) {
+      nuxt.hooks.hookOnce('app:error:cleared', () => applyPlugins(nuxt, plugins.slice(err.index || 0)))
```
pi0, line 57: I'm afraid this might introduce more edge cases. Plugins are not awaited anymore and their order is changed. A simple reload/redirect might be much safer to error resolution.

packages/nuxt3/src/pages/runtime/router.ts
```diff
@@ -110,6 +116,15 @@ export default defineNuxtPlugin((nuxtApp) => {
   })
 
   nuxtApp.hook('app:created', async () => {
+    router.afterEach((to) => {
+      if (to.matched.length === 0) {
+        callWithNuxt(nuxtApp, throwError, [createError({
```
pi0, line 121: Does getVueInstance also works here? If yes, useNuxt for throwError can directly work witthout callWithNuxt i guess

packages/nuxt3/src/app/entry.ts
danielroe: Plugins are still run in succession and the order of the plugins doesn't change.

Or am I missing something?

packages/nitro/src/runtime/server/error.ts
danielroe: In some cases users may wish to render a different error page if user is logged in, etc.

pi0: I see but probably would be better to encourage using client side logic in such cases. Caching is a bigger perf loose

packages/nuxt3/src/app/entry.ts
pi0: [lifecycle] Order is completely changed

danielroe: Yes, that's right. I can remove this but in that case there is no way to recover from a plugin error other than a hard reload, which may not be successful.

Would it be worth considering a reboot option which would recreate the client side app entirely?

packages/nitro/src/runtime/server/error.ts
danielroe: Presumably we could support the same sort of rules for the error page as the rest of the app so it's up to the user?

pi0: SWR and caching rules would require same to not rely on headers that vary by client.

packages/nuxt3/src/app/entry.ts
pi0: It can make still some issues if scripts have global side-effects. What do you think are the downsides of full reload? It saves simplicity and prevents all edge-cases that can happen when partially reinitializing app...

danielroe: Largely because I believe a full reload might also replicate the original plugin error. Adding a partial solution like this allows for client side error handling, which is more flexible as an error page or app can use error boundaries or error hooks + reactivity.

pi0: Indeed original error might not be solved by reloading but we are talking about plugin error here. They are kinda global dependencies and their boundary is the whole app. Skipping them is not always the best idea either. 

I can see the pattern here we could improve for plugins Defenition itself to have error handling boundary. If you are fine lets postpone it to a later improvement

danielroe: Excellent idea to move to plugin definition 👍

packages/nuxt3/src/app/components/nuxt-error-page.vue
danielroe: That is only because it takes up too much space at the moment (see my comment in the PR description). This is pending a future improvement to the default `@nuxt/design` template that would have space to render the description.

packages/nitro/src/runtime/server/error.ts
danielroe: Issue is that it's more difficult to detect as the `/_nitro` prefix gets stripped by h3, and it's more likely that a user might have a `/error` route than a `/__error` route. Suggestions welcome.

packages/nuxt3/src/app/components/nuxt-error-page.vue
pi0: Shall we find a quick solution for design? Merging like this will have negative DX

packages/nitro/src/runtime/server/error.ts
pi0: Sure then preserving _ makes sense 😊

packages/nuxt3/src/app/components/nuxt-error-page.vue
danielroe: I've updated this PR to duplicate the current experience, so now it's not a regression in terms of DX. And will look at `@nuxt/design`, but I think we can defer to a subsequent PR.

Note that I've added a normalising layer to default error component as a temp workaround until we add separate isomorphic error handling package.

examples/with-errors/middleware/error.global.ts
```diff
@@ -0,0 +1,5 @@
+export default defineNuxtRouteMiddleware((to) => {
+  if ('middleware' in to.query) {
+    throwError('error in middleware')
```
pi0, line 3:
```suggestion
    return throwError('error in middleware')
```

pi0 (approved):

pi0:
Amazing job @danielroe. We should iterate over error handling implementation but this is a really good improvement!

PR #3480, opened 2022-03-02
feat!(nuxt3): allow modules to manage autoImports on `generateApp`

Diizzayy:
### ❓ Type of change

### 📚 Description

I've modified the `autoImport` module to update the autoImport context when the `builder:generateApp` hook is called. 

This change allows modules that provide their own autoImports to manage them ( add / remove as needed ) when the `builder:generateApp` hook is triggered.

```ts
// current drawback: only called on initial run in current nuxt3 versions
nuxt.hook('autoImports:extend', (autoimports) => {
  console.log('autoImports:extend')

  // manage autoImports
})

nuxt.hook('builder:watch', async (_event, path) => {
  if (!path.endsWith('.special')) return

  // run logic needed when important files are modified
  await nuxt.callHook('builder:generateApp')
})
```

Current nuxt3 implementation requires the dev server to be restarted for any changes to autoImports to take effect, because the `autoImports:extend` hook is never recalled after it's initial load.

This PR allows for much better DX and is particularly useful when a module needs to watch certain files within a project then generate code based on those files to be used in the app. 

The video below shows a working example from a [Nuxt 3 GraphQL module](https://github.com/Diizzayy/nuxt-gql) that I'm currently developing ( this is the last feature needed for launch )

> Amazing DX

https://user-images.githubusercontent.com/19627670/156457824-7c3b6e97-c0b5-4d56-a722-45d2d0ad9365.mov 

There may be a better way of achieving this outcome? @danielroe @pi0 

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/auto-imports/composables.ts
+++ b/packages/nuxt3/src/auto-imports/composables.ts
@@ -6,39 +6,45 @@ import { AutoImport } from '@nuxt/schema'
 import { resolveFiles } from '@nuxt/kit'
 import { filterInPlace } from './utils'
 
-export async function scanForComposables (dir: string, autoImports: AutoImport[]) {
-  if (!existsSync(dir)) { return }
+export async function scanForComposables (dir: string | string[], autoImports: AutoImport[]) {
+  const performScan = async (entry: string) => {
+    const files = await resolveFiles(entry, [
+      '*.{ts,js,mjs,cjs,mts,cts}',
+      '*/index.{ts,js,mjs,cjs,mts,cts}'
+    ])
 
-  const files = await resolveFiles(dir, [
-    '*.{ts,js,mjs,cjs,mts,cts}',
-    '*/index.{ts,js,mjs,cjs,mts,cts}'
-  ])
+    await Promise.all(
+      files.map(async (path) => {
+        // Remove original entries from the same import (for build watcher)
+        filterInPlace(autoImports, i => i.from !== path)
 
-  await Promise.all(
-    files.map(async (path) => {
-      // Remove original entries from the same import (for build watcher)
-      filterInPlace(autoImports, i => i.from !== path)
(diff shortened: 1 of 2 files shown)


Atinux:
I let the Framework team review, but I certainly love the DX improvement this adds!

Thank you @Diizzayy 💚

packages/nuxt3/src/auto-imports/composables.ts
```diff
@@ -6,39 +6,45 @@ import { AutoImport } from '@nuxt/schema'
 import { resolveFiles } from '@nuxt/kit'
 import { filterInPlace } from './utils'
 
-export async function scanForComposables (dir: string, autoImports: AutoImport[]) {
-  if (!existsSync(dir)) { return }
+export async function scanForComposables (dir: string | string[], autoImports: AutoImport[]) {
```
pi0, line 9: /cc @antfu We can refactor this helper to unimport too I belive :)

packages/nuxt3/src/auto-imports/composables.ts
```diff
@@ -6,39 +6,45 @@ import { AutoImport } from '@nuxt/schema'
...
-          autoImports.push({ name: exp.name, as: exp.name, from: path })
         }
-      }
-    })
-  )
+      })
+    )
+  }
+
+  for (const entry of Array.isArray(dir) ? dir : [dir]) {
```
pi0, line 45: Any reason why not doing `performScan` inside a `for of` loop?

pi0:
Thanks for PR and well explained changes @Diizzayy 💚 Using `builder:generateApp` hook instead looks good to me.

packages/nuxt3/src/auto-imports/composables.ts
Diizzayy: @pi0 none that I can think of, this appears to be suffice. At least until the refactor to `unimport`

PR #3541, opened 2022-03-07
fix(nitro): handle decoding static filenames

### LINKED ISSUE #3530: Filenames containing space in public folder cannot be downloaded in production
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0-27439932.78761c2`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### app.vue
 ```
<template>
  <div>
    <a href="/test 1.txt">Download</a>
  </div>
</template>
```

Create a simple `test 1.txt` file in the /public folder.

### Describe the bug

It works fine in development environment (`yarn dev`) but it will show 404 error after generating the project.
`yarn build`
`yarn start`

I'm not sure if it's an issue with nuxt 3 or other frameworks, but I tried a quick test with nuxt 2.0 and it seemed to work fine.

### Additional context

_No response_

### Logs

_No response_

### PR DESCRIPTION
**Pull request by danielroe**
### 🔗 Linked issue

resolves #3530

### ❓ Type of change

### 📚 Description

Incoming requests that are URI-encoded aren't getting decoded. This PR adds them to a list of alternative options to resolve, as we can't assume that all incoming requests will be URI encoded.

### 📝 Checklist

### DIFF
--- a/packages/nitro/src/runtime/server/static.ts
+++ b/packages/nitro/src/runtime/server/static.ts
@@ -17,10 +17,11 @@ export default async function serveStatic (req, res) {
   let id = withLeadingSlash(withoutTrailingSlash(parseURL(req.url).pathname))
   let asset = getAsset(id)
 
-  // Try index.html
-  if (!asset) {
-    const _id = id + '/index.html'
-    const _asset = getAsset(_id)
+  const alternatives = [decodeURIComponent(id), id + '/index.html', decodeURIComponent(id) + '/index.html']
+
+  while (!asset && alternatives.length) {
+    const _id = alternatives.shift()
+    const _asset = getAsset(id)
     if (_asset) {
       asset = _asset
       id = _id
--- a/packages/nitro/src/runtime/server/static.ts
+++ b/packages/nitro/src/runtime/server/static.ts
@@ -19,7 +19,7 @@ export default async function serveStatic (req, res) {
   let asset: string
 
   for (const _id of [id, decodedId, id + '/index.html', decodedId + '/index.html']) {
-    const _asset = getAsset(id)
+    const _asset = getAsset(_id)
     if (_asset) {
       asset = _asset
       id = _id
--- a/packages/nitro/src/runtime/server/static.ts
+++ b/packages/nitro/src/runtime/server/static.ts
@@ -15,16 +15,15 @@ export default async function serveStatic (req, res) {
   }
 
   let id = withLeadingSlash(withoutTrailingSlash(parseURL(req.url).pathname))
-  let asset = getAsset(id)
+  const decodedId = decodeURIComponent(id)
+  let asset: string
 
-  const alternatives = [decodeURIComponent(id), id + '/index.html', decodeURIComponent(id) + '/index.html']
-
-  while (!asset && alternatives.length) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/nitro/src/runtime/server/static.ts [hunk 1]
```diff
@@ -17,10 +17,11 @@ export default async function serveStatic (req, res) {
   let id = withLeadingSlash(withoutTrailingSlash(parseURL(req.url).pathname))
   let asset = getAsset(id)
 
-  // Try index.html
-  if (!asset) {
-    const _id = id + '/index.html'
-    const _asset = getAsset(_id)
+  const alternatives = [decodeURIComponent(id), id + '/index.html', decodeURIComponent(id) + '/index.html']
```
pi0, line 20: We could decode `id` once.

packages/nitro/src/runtime/server/static.ts
```diff
@@ -17,10 +17,11 @@ export default async function serveStatic (req, res) {
...
+
+  while (!asset && alternatives.length) {
+    const _id = alternatives.shift()
```
pi0, line 23: Why not simply iterate over try versions instead of mutating array? (it might have little memory footprint)

packages/nitro/src/runtime/server/static.ts
danielroe: I guess we want to avoid decoding something the user doesn't want to be decoded. I think it's safe to assume fs filename is canonical.

pi0: Do you have an example of unwanted behavior? Ideally real filenames shouldn't be encoded so we can assume similar encoding behavior when trying to access them.

danielroe: Counterpoint: Do you have an example of a situation where static files are encoded on disk but user wants them decoded? That seems like an odd case to be optimising this for. It seems more likely if a user has a file named `test%20.txt` that they have chosen this filename intentionally.

pi0: Previously we had issue with Nuxt2 payload files being stored on disk as url encoded. I don't want similar issues to pop again

pi0: Yeah which should be improved... :)

danielroe: Well, a decision for another day in that case. There is always https://github.com/unjs/h3/issues/31.

I'm still wary of this change as I see more negatives that positives, but I've implemented it.

packages/nitro/src/runtime/server/static.ts
```diff
@@ -15,15 +15,15 @@ export default async function serveStatic (req, res) {
...
 
   let id = withLeadingSlash(withoutTrailingSlash(parseURL(req.url).pathname))
-  let asset = getAsset(id)
+  const decodedId = decodeURIComponent(id)
+  let asset: string
 
-  // Try index.html
-  if (!asset) {
-    const _id = id + '/index.html'
+  for (const _id of [id, decodedId, id + '/index.html', decodedId + '/index.html']) {
```
pi0, line 21: If we assume ids are stored as decoded, can we drop non decoded checks?

PR #3542, opened 2022-03-07
fix(kit): allow `src` for `addPluginTemplate`

danielroe:
### ❓ Type of change

### 📚 Description

Previously, we actually errored if `src` was provided, e.g. https://github.com/nuxt/framework/blob/e674d0f60d52e26122ce54f6c82a5723f8fc8e23/packages/bridge/src/capi.ts#L26. This PR makes it optional.

### DIFF
--- a/packages/schema/src/types/nuxt.ts
+++ b/packages/schema/src/types/nuxt.ts
@@ -59,5 +59,5 @@ export interface NuxtApp {
   templates: NuxtTemplate[]
 }
 
-type _TemplatePlugin = Partial<NuxtPlugin> & NuxtTemplate
+type _TemplatePlugin = Omit<NuxtPlugin, 'src'> & NuxtTemplate
 export interface NuxtPluginTemplate extends _TemplatePlugin { }
--- a/packages/schema/src/types/nuxt.ts
+++ b/packages/schema/src/types/nuxt.ts
@@ -59,5 +59,5 @@ export interface NuxtApp {
   templates: NuxtTemplate[]
 }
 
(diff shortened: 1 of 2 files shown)


packages/schema/src/types/nuxt.ts
```diff
@@ -59,5 +59,5 @@ export interface NuxtApp {
   templates: NuxtTemplate[]
 }
 
-type _TemplatePlugin = NuxtPlugin & NuxtTemplate
+type _TemplatePlugin = Partial<NuxtPlugin> & NuxtTemplate
```
pi0, line 62: Does it produce proper jsdocs? `src` definition from Plugin is different from template

danielroe: Previously:

```ts
/**
 * Adds a template and registers as a nuxt plugin.
 */
declare function addPluginTemplate(plugin: Omit<NuxtPluginTemplate, 'src'> | string, opts?: AddPluginOptions): NuxtPluginTemplate;
```

Now:

```ts
/**
 * Adds a template and registers as a nuxt plugin.
 */
declare function addPluginTemplate(plugin: NuxtPluginTemplate | string, opts?: AddPluginOptions): NuxtPlugin;
```

danielroe: CleanShot 2022-03-08 at 16 30 36

pi0, line 62:
```suggestion
type _TemplatePlugin = Omit<NuxtPlugin, 'src'> & NuxtTemplate
```

Issue #3015, opened 2022-01-31
provide universal route handling utilities


danielroe:
Modules should be able to rely on certain route handling utilities, whether or not users are using `~/pages` in their app. This includes:

* supporting middleware (programmatically) without `~/pages`
* supporting layouts (via `<NuxtLayout>` explicitly) without `~/pages`
* using route utilities like `navigateTo()`
* using `<NuxtLink>` within `app.vue`

This also improves user experience. Although universal route handling utilities would not support all the features of `vue-router`, tutorials and experience could be shared, and it improves the possibilities of minimal apps based simply on `app.vue`.

PR #3544, opened 2022-03-07
feat(nuxt3): add `<NuxtLink>` component

lihbr:
🚧 WIP, just opening a draft PR for now so my work is public. Fork is opened to maintainers of https://github.com/nuxt/framework

### 🔗 Linked issue

#2086

### ❓ Type of change

- [x] **📖 Documentation (updates to the documentation or readme)**
- [x] **✨ New feature (a non-breaking change that adds functionality)**

### 📚 Description

*TODO:*
- [x] Test
	> next thing on my todo
- [ ] Slot API
	> can be delayed, not sure which way we want to go for this one 🤔
- [x] Documentation PR
	> will work on it once the PR has stabilize
- [ ] Prefetching support
	> can be delayed, will need your help on it though

### 📝 Checklist

- [x] I have updated the documentation accordingly: #3594 (merged into this one)


packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,296 @@
...
+const firstNonUndefined = <T>(...args: T[]): T => {
+  for (const arg of args) {
+    if (arg !== undefined) {
+      return arg
+    }
+  }
+}
+
+const warnPropConflict = (main: string, sub: string): void => {
+  console.warn(`[nuxt-link] \`${main}\` and \`${sub}\` cannot be used together. \`${sub}\` will be ignored.`)
```
Atinux, line 31:
```suggestion
  console.warn(`[NuxtLink] \`${main}\` and \`${sub}\` cannot be used together. \`${sub}\` will be ignored.`)
```

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,296 @@
...
+
+      // Slot API
+      custom: {
+        type: Boolean as PropType<boolean>,
+        default: undefined,
+        required: false
+      }
+    },
+    setup (props, { slots }) {
+      const router = useNuxtApp()?.$router as Router | undefined
```
Atinux, line 170: Cannot we have a `useRouter()` that returns `null` is not activated? Could be the next steps since #3015 got merged?

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,296 @@
...
+          throw new Error('[nuxt-link] Missing `to` prop.')
+        }
+      })
+
+      // Resolving link type
+      const type = computed<'internal' | 'external'>(() => {
+        if (!router) {
+          // If Vue Router is not enabled then all links are considered external
+          if (typeof to.value === 'object') {
+            throw new TypeError('[nuxt-link] Vue Router is not enabled, therefore route location object cannot be handled by `<NuxtLink />`.')
```
Atinux, line 191:
```suggestion
            throw new TypeError('[NuxtLink] Vue Router is not enabled, therefore route location object cannot be handled by `<NuxtLink />`.')
```

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,296 @@
...
+          }
+
+          if (props.external !== undefined || props.internal !== undefined) {
+            console.warn('[nuxt-link] `external` and `internal` props have no effect when Vue Router is not enabled.')
```
Atinux, line 195:
```suggestion
            console.warn('[NuxtLink] `external` and `internal` props have no effect when Vue Router is not enabled.')
```

Atinux: Also, I think it's possible to use NuxtLink with #3015, I let @danielroe confim

danielroe: It is indeed. Here's the implementation:

https://github.com/nuxt/framework/blob/2b3dbed594122607720be349475a296fa410b3b6/packages/nuxt3/src/app/plugins/router.ts#L211-L215

It would be nice to have a consistent implementation between vue-router/universal routing though obviously not all features/props would make sense without vue-router.

packages/nuxt3/src/app/composables/defineNuxtLink.ts
danielroe: It should do:

https://github.com/nuxt/framework/blob/dbab979a2ed28b8f3d02044079f6b9039114d5ff/packages/nuxt3/src/app/composables/router.ts#L4-L6

packages/nuxt3/src/app/composables/defineNuxtLink.ts
Atinux: We could warn for those when vue-router is not activated

lihbr: The way I see it is that I'd expect anyone using Nuxt 3 to link using the `NuxtLink` component independently if the app is vue-router/universal.

For universal users `NuxtLink` would be a drop-in replacement for the anchor tag (as discussed in the RFC: https://github.com/nuxt/framework/discussions/2086#discussioncomment-1683819) the same way `NuxtImg` is a drop-in replacement for the image tag. Doesn't prevent anyone from using an anchor tag but leaves the opportunity to enhance the regular anchor tag now and in the future.

In this PR the implementation throws when using `NuxtLink` in a universal context with a route location object (because it requires Vue Router to be resolved) and treat any other link as external: https://github.com/lihbr/framework/blob/feat/add-nuxt-link/packages/nuxt3/src/app/composables/defineNuxtLink.ts#L188-L198

Of course happy to keep discussing it! Will work on refining the code & test today ☺️

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
+import { defineComponent, h, resolveComponent, PropType, computed } from 'vue'
+import { RouteLocationRaw, Router } from 'vue-router'
+
+import { useRouter } from '#app'
+
+/**
+ * Determines if a URL is internal or external. This does not detect all relative URLs as internal, such as `about` or `./about`. This function assumes relative URLs start with a "/"`.
+ */
+const isInternalURL = (url: string): boolean => {
```
pi0, line 9: Does it make sense to call it something like `isAppURL` or the negative logic of `isExternalURL`? (second helps to use early return in logic)

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+      const to = computed<string | RouteLocationRaw>(() => {
+        if (props.to) {
+          if (props.href) {
+            warnPropConflict('to', 'href')
+          }
+          return props.to
+        } else if (props.href) {
+          return props.href
+        } else {
+          throw new Error('[NuxtLink] Missing `to` prop.')
```
pi0, line 182:
```suggestion
          throw new Error('[NuxtLink] Either `to` or `href` props are required.')
```

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+        }
+      })
+
+      // Resolving link type
+      const type = computed<'internal' | 'external'>(() => {
+        if (!router) {
+          // If Vue Router is not enabled then all links are considered external
+          if (typeof to.value === 'object') {
+            throw new TypeError('[NuxtLink] Vue Router is not enabled, therefore route location object cannot be handled by `<NuxtLink />`.')
```
pi0, line 191:
```suggestion
            throw new TypeError('[NuxtLink] Route location object cannot be resolved by `<NuxtLink />` when vue-router is disabled (no pages).')
```

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+ */
+const isInternalURL = (url: string): boolean => {
+  /**
+   * @see Regex101 expression: {@link https://regex101.com/r/1y7iod/1}
+   */
+  const isInternal = /^\/(?!\/)/.test(url)
+  /**
+   * @see Regex101 expression: {@link https://regex101.com/r/RnUseS/1}
+   */
+  const isSpecialLink = !isInternal && !/^https?:\/\//i.test(url)
```
pi0, line 17: We might use `hasProtocol` from `ufo` and merge

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+        required: false
+      },
+
+      // Edge cases handling
+      external: {
+        type: Boolean as PropType<boolean>,
+        default: undefined,
+        required: false
+      },
+      internal: {
```
pi0, line 156: Having additional `internal` prop is little bit confusing IMO. Since it is also an edge case, can't we just allow `:external="false"`?

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+
+const firstNonUndefined = <T>(...args: T[]): T => {
+  for (const arg of args) {
+    if (arg !== undefined) {
+      return arg
+    }
+  }
+}
+
+const warnPropConflict = (main: string, sub: string): void => {
```
pi0, line 30: Refactor. This utility can accept the object as the first argument in order to reduce repeated logic of checking prop existence like `checkPropConflicts(obj, key1, key2)`

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+      // Resolving `to` value from `to` and `href` props
+      const to = computed<string | RouteLocationRaw>(() => {
+        if (props.to) {
+          if (props.href) {
+            warnPropConflict('to', 'href')
+          }
+          return props.to
+        } else if (props.href) {
+          return props.href
+        } else {
```
pi0, line 181: Refactor: With `checkPropConflicts` refactor, we could remove nested if/else logic

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+        required: false
+      },
+      href: {
+        type: [String, Object] as PropType<string | RouteLocationRaw>,
+        default: undefined,
+        required: false
+      },
+
+      // Attributes
+      blank: {
```
pi0, line 93: There are more valid values for the target that we are not providing a shortcut for. Since `target` prop already exists, Let's keep it simple and semantic for usage  and remove `blank`

lihbr (commented):
Alright, here's a first MVP of the `<NuxtLink />` component. It includes the base component and tests for it as described in the RFC (#2086)

As mentioned in the PR first post here's are the points that are missing to me now:

1. Slot API: I think we need to discuss/agree on the desired API. It can also be postponed if needed and added later as a new feature.
2. Documentation PR: I'll work on it when we're confident about this PR.
3. Prefetching support: I'm not aware of the current status of Nuxt regarding this feature. If available I'll sync with @danielroe for some guidance, else we can wait for it or delay this feature.

ps. I'm not able to build `bridge` on Windows because symlinks like this one: https://github.com/nuxt/framework/blob/main/packages/bridge/src/runtime/asyncData.ts aren't supported there (or I'm doing something wrong). Because of that I'm not sure what the compiler really expect in the issue on the CI here: https://github.com/nuxt/framework/runs/5465287878?check_suite_focus=true

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+          return props.to
+        } else if (props.href) {
+          return props.href
+        } else {
+          throw new Error('[NuxtLink] Missing `to` prop.')
+        }
+      })
+
+      // Resolving link type
+      const type = computed<'internal' | 'external'>(() => {
```
pi0, line 187: Any reason not to simplify as `isExternal`? (value is binary)

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+        if (type.value === 'internal') {
+          // Internal props
+          return {
+            to: to.value,
+            activeClass: props.activeClass || options.activeClass,
+            exactActiveClass: props.exactActiveClass || options.exactActiveClass,
+            replace: props.replace,
+            ariaCurrentValue: props.ariaCurrentValue
+          }
+        } else {
```
pi0, line 241: Refactor: else is not required since we do have return statement

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+   * @see Regex101 expression: {@link https://regex101.com/r/1y7iod/1}
+   */
+  const isInternal = /^\/(?!\/)/.test(url)
+  /**
+   * @see Regex101 expression: {@link https://regex101.com/r/RnUseS/1}
+   */
+  const isSpecialLink = !isInternal && !/^https?:\/\//i.test(url)
+
+  return isInternal && !isSpecialLink
+}
```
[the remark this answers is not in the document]
pi0: We should respect `app.baseURL` from runtime config for base handling. We can use `withoutBase` from ufo before passing path to this function from component to remove base

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: Agree, makes sense since it's for edge cases!

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,295 @@
...
+          }
+
+          if (props.external !== undefined || props.internal !== undefined) {
+            console.warn('[NuxtLink] `external` and `internal` props have no effect when Vue Router is not enabled.')
+          }
+
+          return 'external'
+        } else if (props.external !== undefined) {
+          // Else if `external` prop is used
+          if (props.internal !== undefined) {
```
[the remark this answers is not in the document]
pi0: Making the difference between `null` and `undefined` could be also a confusing API IMO for users. Can't we introduce `no-rel` for example for that case? Also external/internal could be simplified into one prop (`external`) and converted into boolean when is provided (not undefined)

packages/nuxt3/src/app/composables/defineNuxtLink.ts
danielroe: There should still be a route location object and a router even if `vue-router` is disabled.

lihbr: So it could be resolved? 🤔

packages/nuxt3/src/app/composables/defineNuxtLink.ts
pi0: We refer to internal vue instance, as to app :)  and if I'm currect, we want to make a difference between links could be handled by app instance vue-router and those not to use plain a

packages/nuxt3/src/app/composables/defineNuxtLink.ts
pi0: Resolve is not implemented i guess but Daniel is right. Router itself exists and is mocked even without vue-router being enabled

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: Agree, this sugar is not really helpful after all 🤔

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
pi0: Maybe we could be nicer and fall back to `#` as the default value? When programmatically assigning values, it might be possible that API for any reason lacks link for one of the links and it throws an unnecessary Errro

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: Done 🎉

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: Refactored to `isExternal` ☺️

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: Done 🎉

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: The function has been refactored, just checking if the link starts with a single `/` now (two slashes would be an external URL)

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: Error updated ☺️

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: Updated ☺️

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
lihbr: Updated ☺️

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: OK, so what would be a consistent way of checking if Vue Router is enabled? https://github.com/lihbr/framework/blob/feat/add-nuxt-link/packages/nuxt3/src/app/composables/defineNuxtLink.ts#L177-L178

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
lihbr: I don't have a strong opinion on it but I tend to prefer it failing fast so that developers can handle the missing case quickly (checking if the link is available before rendering) rather than thinking links are working fine and discovering months after that they never did on production.

Maybe `#` + a warning? Up to you ☺️

pi0:
Thanks for the updates @lihbr 🔥 Adding replies here as cannot reply to outdated comment anymore:

> I don't have a strong opinion on it but I tend to prefer it failing fast so that developers can handle the missing case quickly (checking if the link is available before rendering) rather than thinking links are working fine and discovering months after that they never did on production. Maybe # + a warning?

Both vue-router and HTML spec do not require `href` attribute for `a` tags: (https://html.spec.whatwg.org) and there is no downside associated with it. Sometimes links really don't exist for few items in a list!

> If the a eleaceholder for where a link might otherwise have been placed, if it had been relevant.

> Authors may also create an A element that specifies no anchors, i.e., that doesn't specify href, name, or id. Values for these attributes may be set at a later time through scripts.


> Actually, I think it makes more sense to keep it as we still want people to be able to use defineNuxtLink({ externalRelAttribute: null }) or defineNuxtLink({ externalRelAttribute: "" }) to disable the framework's default rel attribute. But let me know!

Makese sense. Then at least can we please ignore both null and undefined equally as falsy values and allow one way (empty string) to disable default value with `defineNuxtLink({ externalRelAttribute: "" })` ?

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,276 @@
...
+      })
+      // TODO: Update when prefetching is implemented
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const prefetchedClass = computed<string | null>(() => null)
+
+      return () => {
+        // TODO: Handle `custom` prop
+        return h(
+          isExternal.value ? 'a' : resolveComponent('RouterLink'),
+          propsOrAttributes.value,
```
pi0, line 269: Refactor idea: We could actually create a conditional return part that splits `propsOrAttributes` logic. It seems we are directly serving it here only and a computed would be extra considering it's purpose

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,276 @@
...
+import { useRouter } from '#app'
+
+/**
+ * Determines if a URL is internal or external. This does not detect all relative URLs as internal, such as `about` or `./about`. This function assumes relative URLs start with a "/"`.
+ */
+const isExternalURL = (url: string): boolean => {
+  /**
+   * @see Regex101 expression: {@link https://regex101.com/r/1y7iod/1}
+   */
+  return !/^\/(?!\/)/.test(url)
```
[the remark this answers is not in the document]
lihbr: Oh, OK, reading through UFO source, does `hasProtocol(str, true)` actually returns `true` for `"//nuxtjs.org"`? If so I'm more than happy to get rid of that function and just use `hasProtocol(str, true)`

[Does `//nuxtjs.org`](https://github.com/unjs/ufo/blob/main/src/utils.ts#L10-L12)

pi0:
> OK indeed wasn't thinking about that. Should it default to "#" then? (rendering href="#") or null? (rendering no href attribute)

null seems good idea!

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
pi0:
```suggestion
  // TODO: Use `ufo.hasProtocol` when issue fixed https://github.com/unjs/ufo/issues/45
  return !/^\/(?!\/)/.test(url)
```

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,276 @@
...
+        } else if (props.href) {
+          return props.href
+        }
+
+        throw new Error('[NuxtLink] Either `to` or `href` props are required.')
+      })
+
+      // Resolving link type
+      const isExternal = computed<boolean>(() => {
+        if (!router) {
```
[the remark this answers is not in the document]
lihbr: Yes I noticed that here: https://github.com/nuxt/framework/pull/3544#discussion_r821914449

What would be a consistent way of checking if app using `~/pages`? 🤔 Should it be a special key on the mocked router?

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: I thought with the "memoization" of computed there was actually a benefit of doing it this way, but happy to get rid of that computed and move it to the render function. Let me know!

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,276 @@
...
+        )
+      })
+      // TODO: Update when prefetching is implemented
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const prefetchedClass = computed<string | null>(() => null)
+
+      return () => {
+        // TODO: Handle `custom` prop
+        return h(
+          isExternal.value ? 'a' : resolveComponent('RouterLink'),
```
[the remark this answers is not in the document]
lihbr: Doesn't Vue already take care of that internally? I'd expect it to do so 🤔

Happy to refactor if not~

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,276 @@
...
+      const router = useRouter() as Router | undefined
+
+      // Resolving `to` value from `to` and `href` props
+      const to = computed<string | RouteLocationRaw>(() => {
+        checkPropConflicts(props, 'to', 'href')
+
+        if (props.to) {
+          return props.to
+        } else if (props.href) {
+          return props.href
```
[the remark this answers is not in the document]
lihbr: By defaulting to `null` this works great yeah!

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,276 @@
...
+
+const checkPropConflicts = (props: NuxtLinkProps, main: string, sub: string): void => {
+  if (props[main] !== undefined && props[sub] !== undefined) {
+    console.warn(`[NuxtLink] \`${main}\` and \`${sub}\` cannot be used together. \`${sub}\` will be ignored.`)
+  }
+}
+
+export function defineNuxtLink (options: DefineNuxtLinkOptions = {}) {
+  return defineComponent({
+    name: 'NuxtLink',
```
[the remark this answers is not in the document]
lihbr: Added! 🎉

Do we want to update warnings/errors to use that name as prefix also (or keep `NuxtLink`)?

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
lihbr: Great catch 😉

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
lihbr: Solved: https://github.com/nuxt/framework/pull/3544#discussion_r821951203

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
pi0: Or we could simply assume always have it! Main reason of universal router utils is to make implementation easier. It can handle route changing even when not having `vue-router` and might either implement or throw an error for resolving link objects

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
pi0: Hmm wouldn't be a bad idea ;)

packages/nuxt3/src/app/composables/defineNuxtLink.ts
pi0: Generally yes but I think this computed is simply assigning prop values not doing any extra logic. There shouldn't be performance gain with separation I guess

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
pi0: Yeah... Mostly saving one tick to execute function that we expect to return same value

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
lihbr: Updated in a768dd4

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: Refactored in 9ef445a 🎉

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
lihbr: Alright, cached it on the component instance here: 19dd9b5

Let me know if that's not what you had in mind!

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
lihbr: OK, so we assume a router is always available and let it fail gracefully in a cases not handled by the mocked one?

pi0:
This looks like an already nice starter! What do you think is pending draft @lihbr ?

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
danielroe: Not yet, but you could add it in this PR.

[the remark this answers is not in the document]
lihbr: Awesome! I think I managed to do it properly in d8f2155 ☺️

lihbr:
> This looks like an already nice starter! What do you think is pending draft @lihbr ?

Hey there, indeed! I just synced with Daniel to fix the two last critical issues (bridge build failing + overwrite issue)

With that in mind, the `<NuxtLink />` component and its documentation is quite ready IMO.

Slot API and prefetching feature can be postponed until available within the framework ☺️

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,248 @@
+import { defineComponent, h, resolveComponent, PropType, computed, ConcreteComponent, DefineComponent } from 'vue'
```
danielroe, line 1: totally minor, but can we call this file `nuxt-link.ts`?

packages/nuxt3/test/app--defineNuxtLink.test.ts
```diff
@@ -0,0 +1,199 @@
+import { expect, describe, it, vi } from 'vitest'
```
danielroe, line 1: also tiny, but could we rename this file `define-nuxt-link.test.ts`?

danielroe (commented):
This is really nice! ❤️ A few tiny stylistic changes. I think merging the examples together would be helpful too.

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,248 @@
...
+  // TODO: Use `ufo.hasProtocol` when issue fixed https://github.com/unjs/ufo/issues/45
+  return !/^\/(?!\/)/.test(url)
+}
+
+const firstNonUndefined = <T>(...args: T[]): T => {
+  for (const arg of args) {
+    if (arg !== undefined) {
+      return arg
+    }
+  }
```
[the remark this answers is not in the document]
lihbr: I have to admit, I just let copilot write this one for me haha

packages/nuxt3/test/app--defineNuxtLink.test.ts
lihbr: Done! 🎉

packages/nuxt3/src/app/composables/defineNuxtLink.ts
lihbr: If that makes more sense, sure! Just updated ☺️

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,248 @@
...
+      }
+    },
+    setup (props, { slots }) {
+      const router = useRouter() as Router | undefined
+
+      // Resolving `to` value from `to` and `href` props
+      const to = computed<string | RouteLocationRaw>(() => {
+        checkPropConflicts(props, 'to', 'href')
+
+        return props.to || props.href || '' // Defaults to empty string (won't render any `href` attribute)
```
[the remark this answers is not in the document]
lihbr: Updated ☺️

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
lihbr: Done 🎉

packages/nuxt3/src/app/composables/defineNuxtLink.ts
Atinux: Agree

packages/nuxt3/src/app/composables/defineNuxtLink.ts
```diff
@@ -0,0 +1,248 @@
...
+  // Edge cases handling
+  external?: boolean;
+
+  // Slot API
+  custom?: boolean;
+};
+
+export function defineNuxtLink (options: DefineNuxtLinkOptions) {
+  const checkPropConflicts = (props: NuxtLinkProps, main: string, sub: string): void => {
+    if (props[main] !== undefined && props[sub] !== undefined) {
```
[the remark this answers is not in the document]
lihbr: Sure! Any idea how I could impact properly `process.dev` in the test? Else I can drop the tests for the warn 🤷‍♀️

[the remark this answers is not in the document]
Atinux: @antfu is working on a way to have the `dev` mode for tests, but so far you could comment them I guess.

examples/nuxt-link-overwrite/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "private": true,
+  "scripts": {
+    "build": "nuxi build",
+    "dev": "nuxi dev",
+    "start": "nuxi preview"
+  },
+  "devDependencies": {
+    "nuxt3": "latest"
+  }
+}
```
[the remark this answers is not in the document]
lihbr: Updated 🎉

packages/nuxt3/src/app/composables/defineNuxtLink.ts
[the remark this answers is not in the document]
pi0: This change leads to using a falsy value like 0 or false if set to either of `to` or `href`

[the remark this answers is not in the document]
danielroe: The check we do when `to.value` is consumed handles the case of `<a>`:

https://github.com/nuxt/framework/pull/3544/files#diff-a4d8558f8a76a612d0d973d8f09416e6e702c4373d0e81ab78c8313ba71f1acbR213

I'm not sure what `<RouterLink>` would do with a falsy to. We could add a `|| ''` to the end of this line if we want to ensure an empty string is output.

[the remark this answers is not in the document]
lihbr: Yeah, I would assume TypeScript would have caught it but let's bullet proof it ☺️

danielroe (approved):

packages/nuxt3/src/app/composables/nuxt-link.ts
```diff
@@ -0,0 +1,242 @@
...
+      }
+    },
+    setup (props, { slots }) {
+      const router = useRouter() as Router | undefined
+
+      // Resolving `to` value from `to` and `href` props
+      const to = computed<string | RouteLocationRaw>(() => {
+        checkPropConflicts(props, 'to', 'href')
+
+        return (props.to ?? props.href) || '' // Defaults to empty string (won't render any `href` attribute)
```
pi0, line 170: Using `??` When `to` prop is set a falsy value like `0` it will be picked and I guess unnecessary transform for incompatible envs...

packages/nuxt3/src/app/composables/nuxt-link.ts
```diff
@@ -0,0 +1,242 @@
...
+      })
+
+      // Resolving link type
+      const isExternal = computed<boolean>(() => {
+        if (props.external !== undefined) {
+          // Else if `external` prop is used
+          return props.external
+        } else if (props.target) {
```
pi0, line 178: _self needs to be excluded: 

```suggestion
        } else if (props.target && props.target !== '_self') {
```

packages/nuxt3/src/app/composables/nuxt-link.ts
pi0:
```suggestion
        return props.to || props.href || '' // Defaults to empty string (won't render any `href` attribute)
```

packages/nuxt3/src/app/composables/nuxt-link.ts
pi0, line 178: Code style: We don't need if/else when early return

packages/nuxt3/src/app/composables/nuxt-link.ts
```diff
@@ -0,0 +1,242 @@
...
+          return false
+        }
+
+        // Else check if `to` is an external URL
+        return isExternalURL(to.value)
+      })
+
+      // TODO: Just resolved for now, requires prefetching implementation
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const shouldPrefetch = computed<boolean>(() => {
```
pi0, line 192: Let's drop it for now and design prefeth support properly before adding props

packages/nuxt3/test/define-nuxt-link.test.ts
```diff
@@ -0,0 +1,201 @@
...
+  const component = defineNuxtLink({ componentName: 'NuxtLink', ...defineNuxtLinkOptions })
+
+  const [type, _props, slots] = (component.setup as unknown as (props: NuxtLinkProps, context: { slots: Record<string, () => unknown> }) =>
+    () => [string, Record<string, unknown>, unknown])(props, { slots: { default: () => null } })()
+
+  return { type, props: _props, slots }
+}
+
+describe('define-nuxt-link:to', () => {
+  it('renders link with `to` prop', () => {
```
pi0, line 39: It would be nice if add e2e tests with basic fixture as well to test actual component's rendering behavior...

packages/nuxt3/src/app/composables/nuxt-link.ts
danielroe: Hm. I'd say that a falsy value is still a value, and we do say that `to` overrides `href`.

pi0:
Issue for prefetch support: https://github.com/nuxt/framework/issues/3649

packages/nuxt3/src/app/composables/nuxt-link.ts
lihbr: Good catch!

PR #3564, opened 2022-03-09
docs(server): add example for early spiking middleware

manniL:
### Linked issue

#3561 

### ❓ Type of change

### 📚 Description

This PR adds an example that showcases the usage of `next` inside a server middleware.

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/2.directory-structure/13.server.md
+++ b/docs/content/3.docs/2.directory-structure/13.server.md
@@ -89,7 +89,7 @@ export default async (req: IncomingMessage, res: ServerResponse) => {
 }
 ```
 
-To pass the request deeper into the application, you can simply `return` inside the function: 
+To pass the request deeper into the application, you can simply `return` inside the function:
 
 ```js
 export default async (req, res) => {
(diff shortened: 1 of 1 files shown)


pi0:
With h3, also not calling next() without end() or returning would have same effect. Any reason to advice using legacy callback syntax?

manniL:
> With h3, also not calling next() without end() or returning would have same effect. Any reason to advice using legacy callback syntax?

Not at all, just the lack of knowledge 😄 
So just returning (or calling `end()`) would be better?

pi0:
Yeah you could simply use `return;` :D I'm planning to also move from `(req, res)` to `(event)` syntax progressively so would be nice if we avoid advising calbacks like next when possible

manniL:
> Yeah you could simply use `return;` :D I'm planning to also move from `(req, res)` to `(event)` syntax progressively so would be nice if we avoid advising calbacks like next when possible

Good to know! Will update the docs then accordingly.

docs/content/3.docs/2.directory-structure/13.server.md
```diff
@@ -89,4 +89,17 @@ export default async (req: IncomingMessage, res: ServerResponse) => {
 }
 ```
 
+To pass the request deeper into the application, you can simply `return` inside the function: 
+
+```js
+export default async (req, res, next) => {
```
pi0, line 95:
```suggestion
export default async (req, res) => {
```

pi0:
Thanks!

PR #3577, opened 2022-03-09
docs(usage): add references of nuxi commands and options

clemcode:
### ❓ Type of change

### 📚 Description

This PR adds a page in the Usage section of the docs to list Nuxi commands and options.

### DIFF
--- a/docs/content/1.getting-started/5.commands.md
+++ b/docs/content/1.getting-started/5.commands.md
@@ -66,28 +66,8 @@ npm run build
 
 Nuxt will create a [`.output`](/docs/directory-structure/output) directory with all your application, server and dependencies ready to be deployed. Check out the [deployment](/docs/deployment) section to learn where and how you can deploy a Nuxt application using Nitro.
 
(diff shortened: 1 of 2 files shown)


clemcode:
With the current sitemap, I think it makes sense to find infos about the `dev` and `build` commands in the Get Started section, as well as a glossary in Usage, but there sure is some duplication between the two pages. Maybe we can:
- Keep a quick explanation of the two main commands in Get Started
- Replace the last sections (Preview, Upgrade, Typecheck) with a link to the CLI glossary

danielroe:
Agreed 👍

Any ideas for tidying up the Docs > Usage? It doesn't really seem right to have CLI commands next to 'data fetching' (which describes composables for use within your Nuxt app). That's not an issue with your work on this PR but it seems slightly disjointed... Do you have a better idea?

pi0 (approved):
Really nice addition! Would be nice if we could mention about environment variable associated with some args but can be done later..

PR #3604, opened 2022-03-11
docs(cloudflare): update build command

Tazer:
### ❓ Type of change

- [X] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

So the problem is if you just copy & paste the documentation from https://v3.nuxtjs.org/docs/deployment/cloudflare/
You will be getting into a problem on github actions with 

```
$ /github/workspace/node_modules/.bin/nuxt build
[log] Nuxt CLI v3.0.0-27449808.ac40c97
Error:  Unexpected token '.'
``` 

But when checking the example repo:  
https://github.com/nuxt/nitro-demo/blob/main/wrangler.toml#L13
It was updated. it that make sense cause we are already building the application before in the gitub actions.

### 📝 Checklist
- [ ] I have linked an issue or discussion. ( DId this PR directly from the docs page ) 

### DIFF
--- a/docs/content/3.docs/3.deployment/2.cloudflare.md
+++ b/docs/content/3.docs/3.deployment/2.cloudflare.md
@@ -32,7 +32,7 @@ bucket = ".output/public"
 entry-point = ".output"
 
 [build]
-command = "NITRO_PRESET=cloudflare yarn nuxt build"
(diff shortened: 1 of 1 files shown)


pi0:
Thanks!

PR #3610, opened 2022-03-11
feat(nuxt3)!: enable using `<NuxtLayout>` without pages integration

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/nuxt.js/issues/12829

### ❓ Type of change

### 📚 Description

This PR extracts the Nuxt layouts integration out of the pages module, meaning `<NuxtLayout>` can be used anywhere in a Nuxt app even if the `~/pages` integration isn't.

It also adds support for `extends` with layouts.

BREAKING CHANGE:
 * we remove the `pages:layouts:extend` hook

**Possible improvements**
- [ ] type the layout prop within `<NuxtLayout>` - this would require a new utility interface exported from #app that we could hook into here and in `PageMeta`

### 👉 Migration

* If you are using `pages:layouts:extend` hook, then you can switch to using `app:resolve`, where you will be able to access `app.layouts`.

### 📝 Checklist


packages/nuxt3/src/core/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { resolveFiles, useNuxt } from '@nuxt/kit'
+import { basename, extname, resolve } from 'pathe'
+import { kebabCase } from 'scule'
+
+export async function resolveLayouts () {
```
pi0, line 5: Rest of resolution logic is in `app.ts`. Would probably makese more sense to keep it there rather than introducing another utils

docs/content/3.docs/2.directory-structure/6.layouts.md
```diff
@@ -30,9 +35,28 @@ In your layout files, you'll need to use `<slot />` to define where the page con
 </template>
 ```
 
-Given the example above, you can use a custom layout like this:
+Here's how you might use that layout in `app.vue`:
 
-```vue
+```vue{}[app.vue]
+<template>
+  <NuxtLayout :name="$route.path === '/' ? 'custom' : false">
```
pi0, line 42: Maybe split into simple layout usage and a dynamic example for more advanced cases using a computed?

packages/nuxt3/src/app/composables/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { h } from 'vue'
```
pi0, line 1: wrapIf is not really a composable but component util. Maybe move to `app/components/utils`?

packages/nuxt3/src/app/composables/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+    return () => slots.default()
+  }
+}
+
+/**
+ * Internal utility
+ *
+ * @private
+ */
+export const wrapIf = (component: Component, props: any, slots: any) => {
```
pi0, line 15: Let's add an `_` to ensure it won't be used by end-users.

docs/content/3.docs/2.directory-structure/6.layouts.md
danielroe: I'll simplify this. Maybe we can keep the example at the end of the file for the computed.

packages/nuxt3/src/core/app.ts
```diff
@@ -70,6 +71,14 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
     app.errorComponent = (await findPath(['~/error'])) || resolve(nuxt.options.appDir, 'components/nuxt-error-page.vue')
   }
 
+  // Resolve layouts
+  app.layouts = []
+  for (const config of [...nuxt.options._extends.map(layer => layer.config), nuxt.options]) {
+    app.layouts.push(...[
+      ...await resolveFiles(config.srcDir, `${config.dir.layouts}/*{${config.extensions.join(',')}}`)
+    ].map(file => ({ name: getNameFromPath(file), file })))
+  }
```
pi0, line 80: We need to dedup/merge Layouts with same name

packages/schema/src/types/nuxt.ts
```diff
@@ -57,6 +57,7 @@ export interface NuxtApp {
   dir: string
   extensions: string[]
   plugins: NuxtPlugin[]
+  layouts: NuxtLayout[]
```
pi0, line 60: Wouldn't be better if we use <string, Layout> record?

danielroe: Good point. And that resolves the dedupe issue too 👍

pi0 (approved):

pi0 (approved):

PR #3622, opened 2022-03-12
fix(nuxt3): extends components with wrongly prefix copy the prefix

Atinux:
### ❓ Type of change

### 📚 Description

Having a simple example using `extends` with a `ui` directory having a `prefix` for its components dir.

```js
ui/
  components/
    Button.vue
  nuxt.config.ts // using { path: './components', prefix: 'U' }
components/
  Hello.vue
app.vue
nuxt.config.ts // using extends: './ui'
```

When looking at the `/.nuxt/types/components.d.ts` it shows:

```ts
// Generated by components discovery
declare module 'vue' {
  export interface GlobalComponents {
    'UHello': typeof import("../../components/Hello.vue")['default'],
    'UButton': typeof import("../../ui/components/Button.vue")['default'],
    'NuxtWelcome': typeof import("../../../../node_modules/@nuxt/ui-templates/dist/templates/welcome.vue")['default'],
    'ClientOnly': typeof import("../../../../packages/nuxt3/src/app/components/client-only")['default']
  }
}
export {}
```

#### The problem?

We should have `Button` instead of `UButton`, the `prefix: U` should be applied only to the `~/ui/component/` dir.

```ts
declare module 'vue' {
  export interface GlobalComponents {
    'Hello': typeof import("../../components/Hello.vue")['default'],
    'UButton': typeof import("../../ui/components/Button.vue")['default'],
    // ...
  }
}
export {}
```

#### Current reproduction

Creation a pull request directly to have an easy way to reproduce the bug:
```bash
yarn install
yarn example extends-components-bug
```

Open http://localhost:4000/_vfs/ and click to `/.nuxt/types/components.d.ts`

I also added a logs in [packages/nuxt3/src/components/module.ts](https://github.com/nuxt/framework/blob/a1769f4b10b875099c916d02934c2c93a6742e76/packages/nuxt3/src/components/module.ts#L62-L64) to debug:

```
componentOptions {                                                                                                                            
  dirs: [
    {
      path: './components',
      prefix: 'U'
    }
  ]
}
layer [                                                                                                                                        
  {
    path: './components',
    prefix: 'U'
  }
]
allDirs [                                                                                                                                     
  {
    path: '/Users/atinux/Projects/nuxt/framework/examples/extends-components-bug/components',
    prefix: 'U'
  },
  {
    path: '/Users/atinux/Projects/nuxt/framework/examples/extends-components-bug/ui/components',
    prefix: 'U'
  }
]
```

Seems that `componentsOptions` is wrongly merged.

### 📝 Checklist

### DIFF
--- a/examples/extends-components-bug/app.vue
+++ b/examples/extends-components-bug/app.vue
@@ -1,0 +1,7 @@
+<template>
+  <div>
+    <!-- Should be <UHello /> -->
+    <UHello />
+    <UButton>Hello</UButton>
+  </div>
+</template>
--- a/examples/extends-components-bug/components/Hello.vue
+++ b/examples/extends-components-bug/components/Hello.vue
@@ -1,0 +1,3 @@
+<template>
+  <div>Hello world</div>
+</template>
--- a/examples/extends-components-bug/nuxt.config.ts
+++ b/examples/extends-components-bug/nuxt.config.ts
@@ -1,0 +1,5 @@
+import { defineNuxtConfig } from 'nuxt3'
+
+export default defineNuxtConfig({
+  extends: './ui'
+})
--- a/examples/extends-components-bug/package.json
+++ b/examples/extends-components-bug/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "example-extends-components-bug",
+  "private": true,
+  "scripts": {
+    "build": "nuxi build",
+    "dev": "nuxi dev",
+    "start": "nuxi preview"
+  },
+  "devDependencies": {
+    "nuxt3": "latest"
+  }
+}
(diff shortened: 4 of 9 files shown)


danielroe:
This issue is  caused by the fact that we are merging the `components` options when we load the nuxt config, before we apply the default values.

https://github.com/unjs/c12/blob/1a6f7d368b643bcebfa38d160c3c31dd7339ae65/src/loader.ts#L94-L108

But the dirs at this point aren't resolved, so there is nothing to tell Nuxt that the incoming dir really belongs to a layer, and effectively we've overwritten the default dir configuration.

pi0:
This issue should be fixed after #3717. `ui/` dir added to the `examples/config-extends` to reproduce behavior.

PR #3656, opened 2022-03-14
fix(bridge,nitro,nuxi,nuxt3,vite): ensure debounced/async handlers run in order

### LINKED ISSUE #3654: duplicate plugins inserted in `app:plugins`
Create a `~/components/CompA.vue` and then rename it to `CompB.vue`.

You'll get the following error:

```
 WARN  [SSR] Error transforming virtual:/playground/.nuxt/plugins/server.mjs: Parse failure: Identifier 'nitroClient_25904e8b' has already been declared (9:7)                                                  15:09:20
Contents of line 9: import nitroClient_25904e8b from "/@fs/packages/nitro/src/runtime/app/nitro.client.mjs";

  Contents of line 9: import nitroClient_25904e8b from "/@fspackages/nitro/src/runtime/app/nitro.client.mjs";
  at ssrTransform (node_modules/vite/dist/node/chunks/dep-9c153816.js:52135:15)
  at doTransform (node_modules/vite/dist/node/chunks/dep-9c153816.js:53052:48)
  at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

The reason this exists is because multiple `generateApp()` hooks are running in parallel, both mutating `app.plugins` at the same time. I have a PR incoming.

danielroe:
### 🔗 Linked issue

resolves #3654

### ❓ Type of change

### 📚 Description

Across the nuxt codebase, this PR uses `debounce` + `debounce.promise` to prevent initialising a new handler before the old one has finished running (which should prevent double access to files, for example, or multiple callbacks pushing the same plugin to `app.plugins`, as in the issue linked above).

At the same time, we drop the timeout from (for example) 300 to 1. This leads to a significant increase in DX as it feels much faster.

This PR also:

* fixes a bug with nuxi dev where any path starting with `.nuxt` was not triggering a reload (such as `.nuxtignore`)
* adds `.env` and `.nuxtrc` files to trigger a full nuxt reload

### 📝 Checklist

### DIFF
--- a/packages/bridge/src/vite/server.ts
+++ b/packages/bridge/src/vite/server.ts
@@ -118,7 +118,7 @@ export async function buildServer (ctx: ViteBuildContext) {
     consola.info(`Server built in ${time}ms`)
     await onBuild()
   }
-  const doBuild = pDebounce(pDebounce.promise(_doBuild), 1)
+  const doBuild = pDebounce(pDebounce.promise(_doBuild), 300)
 
   // Initial build
   await _doBuild()
--- a/packages/bridge/src/vite/server.ts
+++ b/packages/bridge/src/vite/server.ts
@@ -118,7 +118,7 @@ export async function buildServer (ctx: ViteBuildContext) {
     consola.info(`Server built in ${time}ms`)
     await onBuild()
   }
-  const doBuild = pDebounce(_doBuild, 300)
+  const doBuild = pDebounce(pDebounce.promise(_doBuild), 1)
 
   // Initial build
   await _doBuild()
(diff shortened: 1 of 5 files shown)


pi0:
`pdebounce.promise` waits for the **previous** promise. When saving file multiple times, the first call takes precedence which is usually an unwanted behavior to build not the latest code. I think we shouldn't set debounce value to a low number like 1ms. It will loose it's purpose of debouncing effect.

danielroe:
The reason I've combined debounce.promise + debounce is that it will wait for the previous promise to resolve, and _then_ run the function again, ensuring the latest changes are reflected.

pi0:
Awaiting on the promise and combining are fine. Changing debounce delay from `200ms` to `1ms` is almost the same as not having a debnounce effect but a serial await which uses first change not last as `.promise` avoid calling function if previously running.

danielroe:
I think I might be misunderstanding you. The current implementation will use both first and last change, as I've explained.

pi0 (approved):

pi0:
As a regression of this PR, the last build might still be skipped. We are working on a better solutioon.

PR #3664, opened 2022-03-15
docs(routing): improvements

Atinux:
### ❓ Type of change

### 📚 Description

- Renamed `Routing` in **Usage** to `NuxtLink`
- Improved the content of the `pages/` directory in **Directory Structure**

### DIFF
--- a/docs/content/3.docs/1.usage/9.nuxt-link.md
+++ b/docs/content/3.docs/1.usage/9.nuxt-link.md
@@ -1,10 +1,8 @@
-# Routing
-
-## `<NuxtLink>`
+# NuxtLink
 
 Nuxt provides `<NuxtLink>` component to handle any kind of links within your application.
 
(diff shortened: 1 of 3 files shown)


danielroe (approved):
This is a really nice improvement.

docs/content/3.docs/1.usage/9.nuxt-link.md
```diff
@@ -1,10 +1,8 @@
-# Routing
-
-## `<NuxtLink>`
+# NuxtLink
```
danielroe, line 1: I think the idea was that we might cover more on this page than just `<NuxtLink>`

Atinux: I think we shall keep it simple since it will be moved to API Reference later on :)

Atinux: And the `pages/` file should become the Routing guide.

pi0 (approved):

PR #3671, opened 2022-03-15
feat(nuxt3): add `<NuxtErrorBoundary>` component for fine-grained error handling

### LINKED ISSUE #3590: `<NuxtErrorBoundary>` to handle errors and render them
follow-up from https://github.com/nuxt/framework/pull/3464

- #default - main content
- #error - shows error (and receives it via scope)
- disables the error on navigation

danielroe:
### 🔗 Linked issue

closes #3590

### ❓ Type of change

### 📚 Description

This PR adds `<NuxtErrorBoundary>` component to simplify the creation of error interfaces. Due to the limitations of SSR, this component only renders errors on client-side. On SSR, a full page `error.vue` is rendered instead.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/components/nuxt-error-boundary.ts
+++ b/packages/nuxt3/src/app/components/nuxt-error-boundary.ts
@@ -1,0 +1,22 @@
+import { defineComponent, ref, onErrorCaptured, onBeforeUnmount } from 'vue'
+import { useNuxtApp, useRouter } from '#app'
+
+export default defineComponent({
+  setup (_props, { slots }) {
+    const error = ref(null)
+    const nuxtApp = useNuxtApp()
+
+    onErrorCaptured((err) => {
+      if (process.client && !nuxtApp.isHydrating) {
+        error.value = err
+        return false
+      }
+    })
+
+    const clearError = () => { error.value = null }
+    const unregister = useRouter().afterEach(clearError)
+    onBeforeUnmount(unregister)
+
+    return () => error.value ? slots.error?.({ error, clearError }) : slots.default?.()
+  }
+})
--- a/docs/content/3.docs/1.usage/8.error-handling.md
+++ b/docs/content/3.docs/1.usage/8.error-handling.md
@@ -88,3 +88,32 @@ You can call this function at any point on client-side, or (on server side) dire
 * `function clearError (redirect?: string): Promise<void>`
 
 This function will clear the currently handled Nuxt error. It also takes an optional path to redirect to (for example, if you want to navigate to a 'safe' page).
+
+## Rendering errors within your app
+
+Nuxt also provides a `<NuxtErrorBoundary>` component that allows you to handle client-side errors within your app, without replacing your entire site with an error page.
+
+This component is responsible for handling errors that occur within its default slot. On client-side, it will prevent the error from bubbling up to the top level, and will render the `#error` slot instead.
+
+The `#error` slot will receive `error` and `clearError` props. (`clearError` is a method that will trigger re-rendering the default slot; you'll need to ensure that the error is fully resolved first or the error slot will just be rendered a second time.)
+
+::alert{type="info"}
+If you navigate to another route, the error will be cleared automatically.
+::
+
+### Example
+
+```vue [pages/index.vue]
+<template>
+  <!-- some content --> 
+  <NuxtErrorBoundary>
+    <!-- You use the default slot to render your content --> 
+    <template #error="{ error, clearError }">
+      You can display the error locally here.
+      <button @click="clearError">
+        This will clear the error.
+      </button>
+    </template>
+  </NuxtErrorBoundary>
+</template>
+```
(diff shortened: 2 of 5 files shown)


packages/nuxt3/src/app/components/nuxt-error-boundary.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default defineComponent({
+  setup (_props, { slots }) {
+    const error = ref(null)
+    const nuxtApp = useNuxtApp()
+
+    onErrorCaptured((err) => {
+      if (process.client && !nuxtApp.isHydrating) {
+        error.value = err
+        return false
```
pi0, line 12: Would be nice to also emit an event to the component like `@error` for fine gained error access. (maybe also `event.cancel()` support?)

packages/nuxt3/src/app/components/nuxt-error-boundary.ts
```diff
@@ -0,0 +1,22 @@
...
+      }
+    })
+
+    const clearError = () => { error.value = null }
+    const unregister = useRouter().afterEach(clearError)
```
pi0, line 17: Does it always makes sense to do this? If component is part of layout (outside of NuxtPage) or using single `app.vue` with custom routing logic, we cannot be sure that component unmounts by changing the page

packages/nuxt3/src/app/components/nuxt-error-boundary.ts
```diff
@@ -0,0 +1,22 @@
...
+    const nuxtApp = useNuxtApp()
+
+    onErrorCaptured((err) => {
+      if (process.client && !nuxtApp.isHydrating) {
+        error.value = err
+        return false
+      }
+    })
+
+    const clearError = () => { error.value = null }
```
pi0, line 16: Since the error is reactive, why not simply let users use `error.value = null` and create a wrapper?

packages/nuxt3/src/app/components/nuxt-error-boundary.ts
danielroe: I'd suggest that if they want that level of fine-grained access they should just use `onErrorHandled` themselves.

packages/nuxt3/src/app/components/nuxt-error-boundary.ts
danielroe: Yes, we can remove this 👍

packages/nuxt3/src/app/components/nuxt-error-boundary.ts
danielroe: Could you explain more?

pi0: Alternative usage:

```vue [pages/index.vue]
<template>
  <!-- some content --> 
  <NuxtErrorBoundary>
    <!-- You use the default slot to render your content --> 
    <template #error="{ error }">
      You can display the error locally here.
      <button @click="error = null">
        This will clear the error.
      </button>
    </template>
  </NuxtErrorBoundary>
</template>
```

One better approach would be supporting `$emit('clear')` to avoid modifying the parent component (it was mentioned somewhere in vue docs but this pattern i'm refering to: https://medium.com/quick-code/vue-communication-patterns-an-intro-to-props-down-and-events-up-pattern-d53340d2c94)

danielroe: I don't believe emitting an event within a slot via `$emit` would bubble up to the component... I also feel that `clearError` is better because we control what happens within the error boundary component, keeping the logic in the same place. For example, we could later add a hook to this. Or implement a different mechanism rather than just clearing the error.

pi0: We couldalso watch `error` ref for changes if needed in the future.

pi0: Indeed. I guess we could start with simplest API and try ideas later. introducing a partial `clearError` that can have breaking changes is more dangrous than not having it and introducing later if needed :)

danielroe: As you wish.

pi0: Shall we also remove implementation?

danielroe: heh - somehow failed to commit 🙈

pi0 (approved):

sultaniman:
@pi0 @danielroe is there a way to use this in broader set of use cases for example after rendering to handle errors on the page, errors when using `$fetch`?

```vue
<template>
  <button @click="doErr">Do Err</button>
  <!-- ... -->
  <NuxtErrorBoundary>
    <template #error="{ error }">
      Error: {{ error }}
    </template>
  </NuxtErrorBoundary>
</template>

<script setup>
const doErr = () => {
  throw new Error("error msg");
}
</script>
```

PR #3673, opened 2022-03-15
feat(nuxt3): extends support for server/ directory

kevinmarrec:
### ❓ Type of change

### 📚 Description

Following up #3222 (`extends` support progress) this PR allows extending `server/` directory for extended config layers

### TO DO
- [x] Extends support for `server` directory
- [ ] Improve code to have less redundant logic ?
- [ ] Extends support for `serverMiddleware` config key in layers => maybe later ? /cc @pi0

### DIFF
--- a/examples/config-extends/base/server/api/base.ts
+++ b/examples/config-extends/base/server/api/base.ts
@@ -1,0 +1,1 @@
+export default () => 'base'
--- a/examples/config-extends/server/api/hello.ts
+++ b/examples/config-extends/server/api/hello.ts
@@ -1,0 +1,1 @@
+export default () => 'hello'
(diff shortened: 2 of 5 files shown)


pi0 (approved):

pi0 (approved):

yassilah:
Hi! I love the  new `extends` feature of Nuxt 3 and I'm really amazed by the possibilities it offers. However, when it comes to extending the `server` directory, I would also expect it to be possible for an "upper" layer to override a "lower" layer with default handlers. Right now, because of the order of the `scanDirs`, it isn't possible; though I found a quick workaround by enforcing it manually. Maybe this should be the default behavior?

```ts
useNuxt().hook('nitro:init', nitro => {
    nitro.hooks.hook('rollup:before', nitro => {
        nitro.options.scanDirs.reverse()
    })
})
```

PR #3679, opened 2022-03-15
fix: use `perfect-debounce` to handle trailing run of promise

danielroe:
### ❓ Type of change

### 📚 Description

This fixes the remaining issue with https://github.com/nuxt/framework/pull/3656 with https://github.com/unjs/perfect-debounce.


packages/nitro/src/server/dev.ts
```diff
@@ -128,7 +128,7 @@ export function createDevServer (nitroContext: NitroContext) {
   let watcher: FSWatcher
   function watch () {
     if (watcher) { return }
-    const dReload = debounce(debounce.promise(() => reload().catch(console.warn)), 200, { before: true })
+    const dReload = debounce(() => reload().catch(console.warn), 200, { leading: true })
```
pi0, line 131: As discussed, enabling leading for server has a negative effect for server

```suggestion
    const dReload = debounce(() => reload().catch(console.warn), 100)
```

packages/nuxt3/src/core/builder.ts
```diff
@@ -48,7 +48,7 @@ function watch (nuxt: Nuxt) {
     ]
   })
 
-  const watchHook = debounce(debounce.promise((event: 'add' | 'addDir' | 'change' | 'unlink' | 'unlinkDir', path: string) => nuxt.callHook('builder:watch', event, path)), 1)
+  const watchHook = debounce((event: 'add' | 'addDir' | 'change' | 'unlink' | 'unlinkDir', path: string) => nuxt.callHook('builder:watch', event, path), 25, { leading: true })
```
pi0, line 51: This is also not a frequent operation. Adding debounce timeout help to avoid rebuild on bulk fs actions.

```suggestion
  const watchHook = debounce((event: 'add' | 'addDir' | 'change' | 'unlink' | 'unlinkDir', path: string) => nuxt.callHook('builder:watch', event, path), 100)
```

packages/nuxi/src/commands/dev.ts
```diff
@@ -69,7 +69,7 @@ export default defineNuxtCommand({
 
     // Watch for config changes
     // TODO: Watcher service, modules, and requireTree
-    const dLoad = debounce(debounce.promise(load), 250)
+    const dLoad = debounce(load, 250)
```
pi0, line 72: Maybe use a default 100ms?
```suggestion
    const dLoad = debounce(load, 100)
```

packages/bridge/src/vite/server.ts
```diff
@@ -118,7 +118,7 @@ export async function buildServer (ctx: ViteBuildContext) {
     consola.info(`Server built in ${time}ms`)
     await onBuild()
   }
-  const doBuild = pDebounce(pDebounce.promise(_doBuild), 300)
+  const doBuild = debounce(_doBuild, 300)
```
pi0, line 121:
```suggestion
  const doBuild = debounce(_doBuild, 100)
```

danielroe:
Would you be okay with moving 100 -> 25 ms default?

PR #3682, opened 2022-03-15
fix(pkg): add implicit dependencies

danielroe:
### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

`vue-router`, `unimport` and `p-debounce` (soon to be removed) were being inlined


packages/bridge/package.json
```diff
@@ -73,6 +73,9 @@
     "vue": "^2",
     "vue-router": "^3"
   },
+  "peerDependencies": {
```
pi0, line 76: unimport is actually a dependency of bridge. We are importing nuxt 3 sources that depend on it.

pi0:
Related upstream improvement: https://github.com/unjs/unbuild/issues/9

We should really fail CI to avoid this issue happening again :D

packages/bridge/package.json
danielroe: yes, you're right. but unfortunately because we are nohoisting bridge deps, adding this dependencies fails build with a ts error. will try a ts-ignore workaround.

danielroe: oh, I see you're on it.

PR #3684, opened 2022-03-15
feat(router): add proper server side redirection to navigateTo

qnp:
### 🔗 Linked issue

See this discussion : https://github.com/nuxt/framework/discussions/3576

### ❓ Type of change

### 📚 Description

- Simplify how `router` is retrieved
- Handle a proper server-side redirection when using `navigateTo` such as right after an `asyncData` in a page. Can be seen as old Nuxt2’s `redirect`.
- Add an optional `replace` boolean option, so that the router navigation uses `replace` instead of `push` client-side (proper redirection, no new entry in history).
- Add tests

See the above mentioned discussion for more context.

### 📝 Checklist

As it is a non-documented usage of `navigateTo` and outside a router middleware, I'm not sure where to document it.


packages/bridge/src/runtime/composables.ts
```diff
@@ -172,12 +172,26 @@ const isProcessingMiddleware = () => {
...
+
+export const navigateTo = (to: Route, options: NavigateToOptions = {}) => {
   if (isProcessingMiddleware()) {
     return to
   }
-  const router: VueRouter = process.server ? useRouter() : (window as any).$nuxt.$router
-  return router.push(to)
+  const router = useRouter()
+  const { ssrContext } = useNuxtApp()
+  if (process.server && ssrContext) {
```
danielroe, line 185:
```suggestion
  if (process.server && useNuxtApp().ssrContext) {
```

packages/bridge/src/runtime/composables.ts
```diff
@@ -172,12 +172,26 @@ const isProcessingMiddleware = () => {
...
+    // Server-side redirection using h3 res from ssrContext
+    const { res } = ssrContext
+    const { route } = router.resolve(to)
+    res.writeHead(302, { Location: route.path })
```
danielroe, line 189:
```suggestion
    res.writeHead(302, { Location: route.fullPath })
```

packages/nuxt3/src/app/composables/router.ts
```diff
@@ -45,12 +45,26 @@ const isProcessingMiddleware = () => {
...
+
+export const navigateTo = (to: RouteLocationRaw, options: NavigateToOptions = {}) => {
   if (isProcessingMiddleware()) {
     return to
   }
-  const router: Router = process.server ? useRouter() : (window as any).$nuxt.$router
-  return router.push(to)
+  const router = useRouter()
+  const { ssrContext } = useNuxtApp()
+  if (process.server && ssrContext) {
```
danielroe, line 58:
```suggestion
  if (process.server && useNuxtApp().ssrContext) {
```

packages/nuxt3/src/app/composables/router.ts
```diff
@@ -45,12 +45,26 @@ const isProcessingMiddleware = () => {
...
+    // Server-side redirection using h3 res from ssrContext
+    const { res } = ssrContext
+    const { path } = router.resolve(to)
+    res.writeHead(302, { Location: path })
```
danielroe, line 62:
```suggestion
    const { fullPath } = router.resolve(to)
    res.writeHead(302, { Location: fullPath })
```

packages/bridge/src/runtime/composables.ts
```diff
@@ -172,12 +172,25 @@ const isProcessingMiddleware = () => {
...
     return to
   }
-  const router: VueRouter = process.server ? useRouter() : (window as any).$nuxt.$router
-  return router.push(to)
+  const router = useRouter()
+  if (process.server && useNuxtApp().ssrContext) {
+    // Server-side redirection using h3 res from ssrContext
+    const { res } = useNuxtApp().ssrContext
+    const { route } = router.resolve(to)
+    res.writeHead(302, { Location: route.fullPath })
```
pi0, line 188: There is a possibility that `navigateTo` is being used in server-side during the rendering process which in this case, headers are already sent and this can lead to an error. We might at least handle such case with a warning

packages/nuxt3/src/app/composables/router.ts
```diff
@@ -45,12 +45,25 @@ const isProcessingMiddleware = () => {
...
     return to
   }
-  const router: Router = process.server ? useRouter() : (window as any).$nuxt.$router
-  return router.push(to)
+  const router = useRouter()
+  if (process.server && useNuxtApp().ssrContext) {
+    // Server-side redirection using h3 res from ssrContext
+    const { res } = useNuxtApp().ssrContext
+    const { fullPath } = router.resolve(to)
+    res.writeHead(302, { Location: fullPath })
```
pi0, line 61: There is a possibility that `navigateTo` is being used in server-side during the rendering process which in this case, headers are already sent and this can lead to an error. We might at least handle such case with a warning

packages/bridge/src/runtime/composables.ts
```diff
@@ -172,12 +172,25 @@ const isProcessingMiddleware = () => {
   return false
 }
 
-export const navigateTo = (to: Route) => {
+export interface NavigateToOptions {
+  replace?: boolean
+}
+
+export const navigateTo = (to: Route, options: NavigateToOptions = {}) => {
```
pi0, line 179: @danielroe Can we somehow reuse logic from nuxt3 code to avoid duplication?

packages/bridge/src/runtime/composables.ts
```diff
@@ -172,12 +172,25 @@ const isProcessingMiddleware = () => {
...
-  const router: VueRouter = process.server ? useRouter() : (window as any).$nuxt.$router
-  return router.push(to)
+  const router = useRouter()
+  if (process.server && useNuxtApp().ssrContext) {
+    // Server-side redirection using h3 res from ssrContext
+    const { res } = useNuxtApp().ssrContext
+    const { route } = router.resolve(to)
+    res.writeHead(302, { Location: route.fullPath })
+    return res.end()
+  } else {
```
misaon, line 190: useless else?

packages/nuxt3/src/app/composables/router.ts
```diff
@@ -45,12 +45,25 @@ const isProcessingMiddleware = () => {
...
+    return res.end()
+  } else {
```
misaon, line 63: useless else?

packages/bridge/src/runtime/composables.ts
danielroe: The issue is the return/argument typing. I think it might be possible by adding a wrapper function with types that calls an untyped utility function, but I think that's possibly worse than duplicated code, given that there are actually differences in the logic.

packages/bridge/src/runtime/composables.ts
danielroe: yes 👍 let's remove them

packages/bridge/src/runtime/composables.ts
pi0: Only difference i could spot is `router.resolve` returning fullPath vs route in v2. But sure let's keep them separate nothing important...

pi0: Only difference i could spot is `router.resolve` returning fullPath vs route in v2. But sure let's keep them separate nothing important...

packages/nuxt3/src/app/composables/router.ts
pi0: Changing to use `sendRedirect` from `h3`. We can improve behavior later there. (https://github.com/unjs/h3/blob/11ef8e3c7be4513eb1720186027aa71b5e623a8d/src/utils/response.ts#L24)

pi0: Changing to use `sendRedirect` from `h3`. We can improve behavior later there. (https://github.com/unjs/h3/blob/11ef8e3c7be4513eb1720186027aa71b5e623a8d/src/utils/response.ts#L24)

pi0:
Thanks @[qnp](https://github.com/qnp) 💯

PR #3697, opened 2022-03-16
fix(macros): ignore nullish id for transform

antfu:
### ❓ Type of change

### 📝 Checklist


packages/nuxt3/src/pages/macros.ts
```diff
@@ -13,6 +13,7 @@ export const TransformMacroPlugin = createUnplugin((options: TransformMacroPlugi
     name: 'nuxt:pages-macros-transform',
     enforce: 'post',
     transformInclude (id) {
+      if (!id) { return }
```
pi0, line 16: Should we also exclude '\x00' for virtual assets?

```suggestion
      if (!id || id.startsWith('\x00')) { return }
```

PR #3707, opened 2022-03-16
ci: reduce ci running

danielroe:
### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

Not sure if there is a reason for the current behaviour, but thought it would save some CI minutes.

This skips `build-release` job rather than just a step in the job, and doesn't run full test suite for docs changes (and vice versa).


pi0:
Initially it was to test `yarn build` behavior but now we have a ched step for it, makes sense to reduce 👍🏼

PR #3721, opened 2022-03-16
refactor: add some more explicit types

danielroe:
### 📚 Description

This adds some explicit return types to `navigateTo` and improves `callWithNuxt` experience within codebase.


packages/bridge/src/runtime/composables.ts
```diff
@@ -177,7 +177,7 @@ export interface NavigateToOptions {
   replace?: boolean
 }
 
-export const navigateTo = (to: Route, options: NavigateToOptions = {}) => {
+export const navigateTo = (to: Route, options: NavigateToOptions = {}): Promise<Route | unknown> | Route => {
```
pi0, line 180: the latest release of h3 is returning `Promise<void>` now. Should we maybe drop `unknown`?

danielroe: Yes, we should. Shall I upgrade in this pr for ease?

pi0: dear renovate seems sleeping 💤  Lets upgrade h3 too :)

Issue #3675, opened 2022-03-15
windows bridge fails when running dev


caikan:
### Environment

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v16.13.1`
- Nuxt Version:     `2.16.0-27358576.777a4b7f`
- Package Manager:  `npm@8.1.2`
- Builder:          `webpack`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://github.com/caikan/bug-nuxt-framework/tree/repro-dev-startup-fatal

### Describe the bug

It seems to happen only on Windows platform, so I can't repro it in codesandbox.

### Logs

```shell
FATAL  Cannot read properties of null (reading 'close')

  at Builder.unwatch (node_modules\@nuxt\builder-edge\dist\builder.js:666:30)
  at Builder.close (node_modules\@nuxt\builder-edge\dist\builder.js:674:10)
  at node_modules\@nuxt\builder-edge\dist\builder.js:195:42
  at node_modules\hable\dist\hable.js:1:990
  at node_modules\hable\dist\hable.js:1:208
  at processTicksAndRejections (node:internal/process/task_queues:96:5)
  at async Nuxt.callHook (node_modules\hable\dist\hable.js:1:959)
  at async Nuxt.close (node_modules\@nuxt\core-edge\dist\core.js:363:5)
  at async load (/project/node_modules/nuxi/dist/chunks/dev.mjs:6727:11)
  at async /project/node_modules/nuxi/dist/chunks/dev.mjs:6687:11

i Preparing project for development
i Initial build may take a while
i Discovered Components: .nuxt/components/readme.md
√ Builder initialized
√ Nuxt files generated

 ERROR  Failed to compile with 1 errors


 ERROR  in ./.nuxt/client.js

Module build failed (from ./node_modules/unplugin/dist/webpack/loaders/transform.js):
Error: ENOENT: no such file or directory, open '\project\.nuxt\client.js'

 @ multi ./node_modules/@nuxt/bridge/dist/runtime/capi.plugin.mjs ./node_modules/eventsource-polyfill/dist/browserify-eventsource.js (webpack)-hot-middleware/client.js?reload=true&timeout=30000&ansiColors=&overlayStyles=&path=%2F__webpack_hmr%2Fclient&name=client ./.nuxt/client.js
```

vis97c:
Same happened to me after reinstalling the dependencies. Also on windows.

[...]

PR #3737, opened 2022-03-17
feat: support reactivity transform

antfu:
### 🔗 Linked issue

#2591

### ❓ Type of change

### 📚 Description

Enable by `experimental.reactivityTransform: true`

### 📝 Checklist


examples/with-reactivity-transform/nuxt.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineNuxtConfig } from 'nuxt3'
+
+export default defineNuxtConfig({
+  modules: [
+    '@nuxt/ui'
+  ],
+  experimental: {
+    reactivityTransform: true
+  }
+  // vite: false
```
antfu, line 10: If switch to webpack, I got this error, which I assume might not be related to this PR?

image

danielroe: We've changed how you switch to webpack. It's now `builder: 'webpack'` rather than setting `vite: false`.

pi0 (approved):
LGTM. Shall we also add a basic test for this flag?

antfu:
> LGTM. Shall we also add a basic test for this flag?

Since it's kinda global - do you want it to be enabled for the basic fixture or create a new fixture for it?

pi0:
It would be great to test with same fixture if possible to save CI time 😊

pi0:
Tests are still failing since `@vitejs/plugin/vue` is using `transformWithEsbuild` without respecting global esbuild options to ignore `extends` field...

baixiaoyu2997:
@antfu  nuxt bridge error,  because not set `experimental` config. i think  should handle this code 
```
nuxt.options.experimental.reactivityTransform
```

antfu:
@baixiaoyu2997 Bridge does not support reactivityTransform currently, it's a Nuxt 3 only feature

baixiaoyu2997:
> @baixiaoyu2997 Bridge does not support reactivityTransform currently, it's a Nuxt 3 only feature

我没有开启experimental这个配置，但是报错了：
```
Cannot read properties of undefined (reading 'reactivityTransform')      
```
from this code: 
https://github.com/nuxt/framework/blob/f69126e8f4adb71bbed55990c97c6e2f9bdd7dec/packages/nuxi/src/utils/prepare.ts#L71

如果需要复现 我可以提供在线代码

senar9:
same problem: https://github.com/nuxt/framework/issues/3675#issue-1169961542

pi0:
Issues with bridge should be solved by #3763

misaon:
@pi0 @antfu $useState / $useCookie will be? ❤️

Atinux:
Is it normal that the documentation has not been generated for the `experimental` option on https://v3.nuxtjs.org/docs/directory-structure/nuxt.config ?

xSuffix:
Does reactivity transform work with nitro? When importing $ref from 'vue/macros' my IDE seems to find the correct file (.../node_modules/.pnpm/vue@3.2.37/node_modules/vue/macros.d.ts) but Rollup is giving me an error, saying it could not load .../node_modules/.pnpm/vue@3.2.37/node_modules/vue/dist/vue.cjs.js/macros.

I tried explicitly installing Vue 3, as suggested in https://github.com/nuxt/framework/discussions/2028, but still got the error.

Reproduction:
https://stackblitz.com/edit/nuxt-starter-tcndrb?file=server/api/data.get.ts

PR #3742, opened 2022-03-17
feat(test-utils)!: add support for running with env variables

danielroe:
### 🔗 Linked issue

related to https://github.com/nuxt/framework/issues/3437

### ❓ Type of change

### 📚 Description

This adds support for passing environment variables and restarting the server without rebuilding a fixture, which will be handy for making assertions about the effect of runtime configuration on a nuxt server.

### 👉 Migration

`listen` has been renamed to `startServer`. It will now also terminate any existing server process when run/rerun.

### 📝 Checklist


antfu (approved):

packages/test-utils/src/server.ts
```diff
@@ -13,13 +17,27 @@ export async function listen () {
     resolve(ctx.nuxt.options.nitro.output.dir, 'server/index.mjs')
   ], {
     env: {
+      ...ctx.options.env || {},
+      ...options.env || {},
```
pi0, line 21: Can we instead proxy current process.env to the running server? What makes us need to explicitly override them for server process?

packages/test-utils/src/server.ts
```diff
@@ -13,13 +17,27 @@ export async function listen () {
...
       PORT: String(port),
       NODE_ENV: 'test'
     }
   })
   await waitForPort(port, { retries: 8 })
 }
 
+export function terminateServerProcess () {
```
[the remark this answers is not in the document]
pi0: Also, i would suggest simplifying its naming to `stopServer`. Internally we might not always use a process for running server.

packages/test-utils/src/server.ts
```diff
@@ -13,13 +17,27 @@ export async function listen () {
...
+  const ctx = useTestContext()
+  if (ctx.serverProcess) {
+    ctx.serverProcess.kill()
+  }
+}
+
+export async function restartServer (options: ListenOptions = {}) {
```
pi0, line 36: We can add an assertion to `listen` (maybe also rename to `startServer`?) that stops the existing server before starting a new one. Having `restartServer` as a shortcut is fine but i feel we could merge functionality to main start fn.

danielroe: I think if server exists we can always stop 👍

packages/test-utils/src/server.ts
danielroe: I think we might have unexpected consequences - like differing behaviour based on the environment where the test is run.

pi0: On the other hand, it makes it harder to test behavior directly using an env or dot-env and `nuxi test` as we need to manually pass such options. Additionally, we are not going to always use an isolation (#3753 for instance depends on nitro worker which currently at least passed all env). Can we please for now pass entire env and discuss about isolation later?

packages/test-utils/src/server.ts
```diff
@@ -4,22 +4,35 @@ import { getRandomPort, waitForPort } from 'get-port-please'
...
+      ...ctx.options.env || {},
+      ...options.env || {},
       PORT: String(port),
       NODE_ENV: 'test'
     }
   })
   await waitForPort(port, { retries: 8 })
 }
 
+export function stopServer (ctx = useTestContext()) {
```
pi0, line 30: Future compat: Would be better to return a promise

packages/test-utils/src/server.ts
danielroe: Sure. So if we need to modify env before restarting server we can just modify `process.env` directly?

packages/test-utils/src/server.ts
```diff
@@ -4,22 +4,30 @@ import { getRandomPort, waitForPort } from 'get-port-please'
...
     env: {
+      ...process.env,
       PORT: String(port),
       NODE_ENV: 'test'
     }
   })
   await waitForPort(port, { retries: 8 })
 }
 
+export async function stopServer (ctx = useTestContext()) {
```
pi0, line 25: Refactor for consitancy: 
```suggestion
export async function stopServer () {
ctx = useTestContext()
```

packages/test-utils/src/server.ts
```diff
@@ -4,22 +4,30 @@ import { getRandomPort, waitForPort } from 'get-port-please'
 import { fetch as _fetch, $fetch as _$fetch, FetchOptions } from 'ohmyfetch'
 import { useTestContext } from './context'
 
-export async function listen () {
+export async function startServer () {
   const ctx = useTestContext()
   const port = await getRandomPort()
   ctx.url = 'http://localhost:' + port
+  stopServer()
```
pi0, line 11:
```suggestion
  await stopServer()
```

pi0: PS: We could stop server before allocating a new random port...

packages/test-utils/src/server.ts
danielroe: this is because it's likely common to run in an `afterAll` or `afterEach` hook where the context has been unset

packages/test-utils/src/server.ts
```diff
@@ -4,22 +4,31 @@ import { getRandomPort, waitForPort } from 'get-port-please'
...
+      ...process.env,
       PORT: String(port),
       NODE_ENV: 'test'
     }
   })
   await waitForPort(port, { retries: 8 })
 }
 
+export async function stopServer () {
+ctx = useTestContext()
```
pi0, line 26:
```suggestion
  ctx = useTestContext()
```

pi0 (approved):

packages/test-utils/src/server.ts
danielroe: this code will fail as is...

PR #3750, opened 2022-03-17
test: preserve server instance in watch mode

antfu:
### ❓ Type of change

### 📚 Description

Currently, when you made changes to the test files, it will trigger the test-utils to re-build the entire app and wait for the server to start (which could take minutes)

This PR bumps the version of `vitest`. With the new exposed API we are able to preserve the server instance across runs and reuse it.

/cc @danielroe 

### 📝 Checklist


packages/test-utils/src/setup/vitest.ts
```diff
@@ -1,9 +1,21 @@
...
+
+  // When running in watch mode, `runOnce` will preserve the state across runs
+  // so we can reuse the server for much faster test reruns
+  const hooks = await vitest.runOnce(create)
+
+  if (vitest.isFirstRun()) {
+    vitest.beforeAll(hooks.setup, 60000)
+  }
+  if (!vitest.isWatchMode()) {
+    vitest.afterAll(hooks.afterAll)
```
pi0, line 14: This seems to make hooks bound to vitest behavior support. What if we normally always maps beforeAll/afterAll but add a new option to test utils like `preserveContext: true` to change setup behavior based on this?

packages/test-utils/src/setup/vitest.ts
```diff
@@ -1,9 +1,21 @@
 import type { TestHooks } from '../types'
 
-export default async function setupVitest (hooks: TestHooks) {
+export default async function setupVitest (create: (() => TestHooks)) {
   const vitest = await import('vitest')
-  vitest.beforeAll(hooks.setup, 60000)
+
+  // When running in watch mode, `runOnce` will preserve the state across runs
+  // so we can reuse the server for much faster test reruns
+  const hooks = await vitest.runOnce(create)
```
pi0, line 8: Can setup be smart enough to detect if context already exists and avoid creating new hooks?

test/bridge.test.ts
```diff
@@ -1,12 +1,12 @@
 import { fileURLToPath } from 'url'
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, runOnce } from 'vitest'
 import { setup, $fetch } from '@nuxt/test-utils'
 
 describe('fixtures:bridge', async () => {
-  await setup({
+  await runOnce(() => setup({
```
pi0, line 6: Would be nice also here if setup could be smart enough to reuse existing context instead of manually bounding it with vitest.

packages/test-utils/src/setup/vitest.ts
antfu: Yes, there is an entry-file based on a unique key generated under the hood, when the hmr is triggered and it will retrieve the result generated from the previous setup.

test/bridge.test.ts
antfu: Oh this is no longer needed, I will remove it

pi0:
I'm concerned the current method is too specific to vite and that we could do alternative methods to avoid extra rebuild with hashing. Let's redo it later once had `nuxi dev --watch`.

PR #3757, opened 2022-03-17
fix(nuxt3): use shared module for dynamic paths

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/3437

### ❓ Type of change

### 📚 Description

This started as adding tests for dynamic paths + relative assets. This uncovered a regression where `__webpack_public_path` wasn't being set because it wasn't top-of-entry.

This PR refactors the public path to import from two virtual templates:

1. a `client-config` file that mocks `#config` for isomorphic usage (and correspondingly we remove the `#config` import protection rule as imports from `#config` now work in vite)
2. a `dynamic-paths` file that mirrors https://github.com/nuxt/framework/blob/main/packages/nitro/src/runtime/app/paths.ts so helper functions are located in a central place rather than being embedded throughout the app

### 📝 Checklist


packages/nuxt3/src/core/templates.ts
```diff
@@ -168,3 +168,36 @@ export const layoutTemplate: NuxtTemplate = {
...
+      nuxt.options.dev
+        ? `const appConfig = ${JSON.stringify(nuxt.options.app)}`
+        : 'const appConfig = config.app',
+
+      'export const NUXT_BASE_URL = () => appConfig.baseURL',
+      'export const NUXT_BUILD_ASSETS_DIR = () => appConfig.buildAssetsDir',
+
+      'export const NUXT_BUILD_ASSETS_URL = (...path) => joinURL(NUXT_PUBLIC_ASSETS_URL(), NUXT_BUILD_ASSETS_DIR(), ...path)',
+
+      'export const NUXT_PUBLIC_ASSETS_URL = (...path) => {',
```
pi0, line 197: Can we please have consistent naming with nitro paths (https://github.com/nuxt/framework/blob/main/packages/nitro/src/runtime/app/paths.ts) for function names?

packages/vite/src/client.ts
```diff
@@ -23,7 +23,8 @@ export async function buildClient (ctx: ViteBuildContext) {
     },
     resolve: {
       alias: {
-        '#build/plugins': resolve(ctx.nuxt.options.buildDir, 'plugins/client')
+        '#build/plugins': resolve(ctx.nuxt.options.buildDir, 'plugins/client'),
+        '#config': resolve(ctx.nuxt.options.buildDir, 'client-config.mjs')
```
pi0, line 27: We can make it private to be like `#_config` and keep for internal usage while advocate composable version of config to be used by users for future compatibility of per-request variable config.

packages/nuxt3/src/core/templates.ts
danielroe: The reason is that we want to ensure these don't conflict with other locally scoped variables. I can prefix with `__` perhaps?

packages/vite/src/client.ts
danielroe: If we do that then we will also have to create a new alias to point `#_config` to the nitro config path (or add a wrapper template - `config.client.mjs` and `config.server.mjs`?).

pi0: Should is fine for server and nitro usage. My main concern is app where composables are possible. We might find better way to also drop #config with nitropack.

packages/nuxt3/src/core/templates.ts
pi0: This is a new module right? How they might conflict?

danielroe: because we are importing them by name into other files:

https://github.com/nuxt/framework/blob/a590c401d003bb01dee0feb84f0e989780ca4e3a/packages/vite/src/plugins/dynamic-base.ts#L63-L65

pi0: Since treeshaking is not relavant, can we import * as _paths and use _paths.baseURL()?

packages/vite/src/client.ts
pi0: Same for client build (webpack and vite) #client usage shouldn't be possible. As we discussed, it is a temporary solution for base url handling

danielroe: Agreed, it's a temporary solution. But I'm not sure what change you are asking for here.

packages/nuxt3/src/core/templates.ts
danielroe: Sure 👍

packages/vite/src/client.ts
pi0: Not providing #config alias which could be used by users (and we drop in the future) but an alias that indicates to be internal :)

danielroe: Ah. My point is if we add this alias here in the client config, we will also have to add a similar `#_config` alias in the server config that we somehow resolve to point to the nitro `#config`. At the moment we are handling it by leaving it as an external and letting Nitro deal with it.

pi0: Having #_config alias for server looks good for now 😊

packages/nuxt3/src/core/plugins/import-protection.ts
```diff
@@ -18,7 +18,7 @@ export const vueAppPatterns = (nuxt: Nuxt) => [
   [/(^|node_modules\/)@vue\/composition-api/],
   ...nuxt.options.modules.filter(m => typeof m === 'string').map((m: string) =>
     [new RegExp(`^${escapeRE(m)}$`), 'Importing directly from module entrypoints is not allowed.']),
-  ...['#static-assets', '#static', '#config', '#assets', '#storage', '#server-middleware']
+  ...['#static-assets', '#static', '#assets', '#storage', '#server-middleware']
```
pi0, line 21: I think we can revert it back:
```suggestion
  ...['#static-assets', '#static', '#config', '#assets', '#storage', '#server-middleware']
```

packages/nuxt3/src/core/templates.ts
```diff
@@ -168,3 +168,40 @@ export const layoutTemplate: NuxtTemplate = {
     ].join('\n')
   }
 }
+
+export const clientConfigTemplate: NuxtTemplate = {
+  filename: 'config.client.mjs',
+  getContents: () => 'export default __NUXT__.config'
```
pi0, line 174:
```suggestion
  getContents: () => 'export default window.__NUXT__.config'
```

packages/nuxt3/src/core/templates.ts
```diff
@@ -168,3 +168,40 @@ export const layoutTemplate: NuxtTemplate = {
...
+}
+
+export const serverConfigTemplate: NuxtTemplate = {
+  filename: 'config.server.mjs',
+  getContents: () => [
+    'import config from \'#config\'',
```
pi0, line 180:
```suggestion
    'import config from \'#_config\'',
```

packages/nuxt3/src/core/templates.ts
```diff
@@ -168,3 +168,40 @@ export const layoutTemplate: NuxtTemplate = {
...
+    'export default config'
+  ].join('\n')
+}
+
+export const publicPathTemplate: NuxtTemplate = {
+  filename: 'dynamic-paths.mjs',
```
pi0, line 186: Any reason not using `paths.mjs` (same as nitro) instead of `dynamic-paths.mjs`?

packages/webpack/src/presets/base.ts
```diff
@@ -118,6 +118,11 @@ function baseAlias (ctx: WebpackConfigContext) {
     ...options.alias,
     ...ctx.alias
   }
+  if (ctx.isClient) {
+    ctx.alias['#_config'] = resolve(ctx.nuxt.options.buildDir, 'config.client.mjs')
```
pi0, line 122: Refactor: only difference is in `client.` / `server.`. We could join if branches.

packages/nuxt3/src/app/entry.ts
```diff
@@ -1,3 +1,4 @@
+import '#build/dynamic-paths.mjs'
```
pi0, line 1: We are not having side-effects with `dynamic-paths`. Why is this required?

packages/nuxt3/src/core/plugins/import-protection.ts
danielroe: No, it will warn because of the virtual template we create internally.

packages/nuxt3/src/core/templates.ts
danielroe: No, that would be fine 👍

packages/nuxt3/src/app/entry.ts
danielroe: It has a side effect in webpack builder

packages/nuxt3/src/core/plugins/import-protection.ts
pi0: Can we use `#_config` within virtual template then? (i'm fine even if it requires an additional alias with nitro but not allow `#config` import in client. it is addictive and potentially breaking change in the future)

packages/nuxt3/src/app/entry.ts
pi0: Can you please elaborate more I can understand? The generated template seems to consume config only in order to export utilities. (even if necessary, at least adding an exaplanation comment)

danielroe: It is injected by the plugin.

pi0 (approved):

PR #3763, opened 2022-03-18
fix(bridge): handle missing `experimental` key on bridge

danielroe:
### 🔗 Linked issue

follow up on https://github.com/nuxt/framework/pull/3737#issuecomment-1072171838

### ❓ Type of change

### 📚 Description

Schema hasn't processed options so `experimental` doesn't exist.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/utils/prepare.ts
+++ b/packages/nuxi/src/utils/prepare.ts
@@ -68,7 +68,7 @@ export const writeTypes = async (nuxt: Nuxt) => {
     .filter(f => typeof f === 'string')
     .map(id => ({ types: getNearestPackage(id, modulePaths)?.name || id }))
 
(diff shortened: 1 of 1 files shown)


baixiaoyu2997:
thanks, danielroe, i'm not sure this issue  is solved， i also found few code like this
https://github.com/nuxt/framework/blob/f69126e8f4adb71bbed55990c97c6e2f9bdd7dec/packages/webpack/src/presets/vue.ts#L18
https://github.com/nuxt/framework/blob/f69126e8f4adb71bbed55990c97c6e2f9bdd7dec/packages/vite/src/vite.ts#L69

danielroe:
@baixiaoyu2997 That code is only used in Nuxt 3, not Bridge.

PR #3776, opened 2022-03-18
fix(kit): add typings to kit server middleware

Rigo-m:
### ❓ Type of change

### 📚 Description

Add types to `addServerMiddleware`.
I've discovered that I can place an object with `path` and `handler` inside `addServerMiddleware` by taking a look at tailwindcss nuxt module. I think this should be auto-documented via types.

I need help to better type this, I'm not so ts-savvy yet.

### 📝 Checklist


packages/kit/src/server.ts
```diff
@@ -1,6 +1,12 @@
 import { useNuxt } from './context'
+import { IncomingMessage, ServerResponse } from 'http'
+
+export interface ServerMiddleware {
+  path: string,
+  handler: (req: IncomingMessage, res: ServerResponse) => void
```
pi0, line 6: Handler can be also be a string (path to middleware). Reference types: https://github.com/nuxt/framework/blob/e2e021cae68b1e62f0d7d8ecfb9ca6b4889ff9d5/packages/nitro/src/server/middleware.ts#L9

Rigo-m: I have improved the handler type. Although, I wonder, why not importing directly the ServerMiddleware interface from the nitro package?

pi0: Thanks. Nitro package will be dropped soon but indeed later we could depend on the types from nitropack. Do you mind to also use middleware type from h3?

Rigo-m: Check the latest commit

PR #3783, opened 2022-03-18
feat(nuxt3): extends support for pages/ & middleware/ directories

kevinmarrec:
### ❓ Type of change

### 📚 Description

Following up #3222 (`extends` support progress) this PR allows extending `pages/` & `middleware/` directories for extended config layers.

- [x] Extends support for `pages` directory
- [x] Extends support for `middleware` directory
- [x] Handle overrides

### ℹ️ Override behavior

```ts
export default {
  extends: ['a', 'b']
}
```

If they match same `name` , pages and middlewares defined in `a` can be overriden by `b`, and finally by the project itself.


packages/nuxt3/src/pages/module.ts
```diff
@@ -99,6 +103,7 @@ export default defineNuxtModule({
       filename: 'middleware.mjs',
       async getContents () {
         const middleware = await resolveMiddleware()
+        await nuxt.callHook('pages:middleware:extend', middleware)
```
kevinmarrec, line 106: Info: I extracted the hook call from `resolveMiddleware` utility function to here.

packages/nuxt3/src/pages/module.ts
```diff
@@ -68,7 +72,7 @@ export default defineNuxtModule({
     addTemplate({
       filename: 'routes.mjs',
       async getContents () {
-        const pages = await resolvePagesRoutes(nuxt)
+        const pages = await resolvePagesRoutes()
```
kevinmarrec, line 75: I refactored `resolvePagesRoutes` to use `useNuxt`

packages/nuxt3/src/pages/utils.ts
```diff
@@ -24,14 +24,29 @@ interface SegmentToken {
...
-  const files = await resolveFiles(pagesDir, `**/*{${nuxt.options.extensions.join(',')}}`)
+export async function resolvePagesRoutes (): Promise<NuxtPage[]> {
+  const nuxt = useNuxt()
 
-  // Sort to make sure parent are listed first
-  files.sort()
+  // Page layers priority (Low to High):
+  // Extended Layer (1) < Extended Layer (2) < ... < Extended Layer (N-1) < Extended Layer (N) < Local layer
+  // Therefore, we make the local layer last
+  const pageLayers = [...nuxt.options._layers.slice(1), nuxt.options._layers[0]]
```
pi0, line 33: It is intentionally sorted in _layers from high to low

kevinmarrec: Yes, the thing I wanted to point here IS that project (local) layer needs to be last and not first

kevinmarrec: Unless you meant I juste need to reverse the array ?

danielroe:
What do you think about adding some tests for this feature?

kevinmarrec:
@danielroe Yes, I'm planning to add tests for the `extends` features, I'm gonna write tests for the ones that already have been merged first :)

packages/nuxt3/src/pages/utils.ts
pi0: local layer is the first :)

pi0:
Shall we add extends to the same basic fixture? i guess it won't harm or add side-effect to the other tests but saves CI build time.

packages/nuxt3/src/pages/utils.ts
kevinmarrec: Alright so :

```js
{
  extends: [
    './ui',
    './base'
  ]
}
```

means `ui` overrides `base` ?

I thought the last item of array would override the items before :)

pi0: Actually it is the other way and why layouts sorted like it 🙈 I'm fine to later on change this convention from low to top but needs changes to several places including nitropack and c12.

wsaca:
I'm not sure if this change is related, but routes are not working correctly, now it only resolves the first folder inside "pages".

kevinmarrec:
@wsaca Can you provide a reproduction repository cause I can't reproduce it. Thanks !

wsaca:
Hi @kevinmarrec, I created a simple example, try to navigate to /apis, /apps or /dbs. Only the first folder is resolved.
- https://codesandbox.io/s/lively-darkness-cd2x8c

kevinmarrec:
Thanks @wsaca , @danielroe created an issue with your reproduction and I'm gonna check this asap

PR #3795, opened 2022-03-20
fix(nitro): list global middleware before API middleware

### LINKED ISSUE #3792: Creating properties in middleware no longer is available to endpoints
### Environment

- Nuxt Version:     `3.0.0-27460489.53fbca7`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/TheColaber/nuxt-3-bug

### Describe the bug

When running `yarn dev` and opening up your localhost, properties created by the middleware on the `req` object aren't available in the API endpoints as they used to. After causing the server to rebuild by saving a file in the server directory, the property created on `req` becomes available. Check out my repro for a better understanding.

### Additional context

This bug started in this commit: https://github.com/nuxt/framework/commit/227ceed
The source of the problem seems to be the order of `nitroContext.scannedMiddleware` set by `_build` and `_watch`. Before this commit, the order looked something like this:
```js
[
  {
    route: '/',
    handle: 'C:/.../nuxt-3/server/middleware/index.ts'
  },
  {
    route: '/api/data',
    handle: 'C:/.../nuxt-3/server/api/data.ts',
    lazy: true
  }
]
```
but after this commit, it looks like this:
```js
[
  {
    route: '/api/data',
    handle: 'C:/.../nuxt-3/server/api/data.ts',
    lazy: true
  },
  {
    route: '/',
    handle: 'C:/.../nuxt-3/server/middleware/index.ts'
  }
]
```
I think this might be the cause of the problem.

### Logs

_No response_

TheColaber:
### 🔗 Linked issue

#3792

### ❓ Type of change

### 📚 Description

Resolves #3792 
https://github.com/nuxt/framework/commit/227ceed was sorting the `scannedMiddleware` by using `localeCompare` on the routes which would cause all the middleware to be put at the bottom of the array when they should be at the top. The `filesToMiddleware` function (used in the `scanMiddleware` function) already sorts the routes individually. This PR just removes the code that resorts the code that was causing problems between the middleware and the API endpoints, see #3792.

### 📝 Checklist

I'm not entirely sure about this second checkbox and weather this needs documentation or not.


pi0:
Thanks!

Issue #3672, opened 2022-03-15
Dynamically creating <NuxtLink> with <Component> no longer works


ataylor32:
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.17.6`
- Nuxt Version:     `3.0.0-27454426.4cefce4`
- Package Manager:  `npm@8.5.4`
- Bundler:          `Vite`
- User Config:      `meta`, `publicRuntimeConfig`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-92nabt?file=app.vue

### Describe the bug

Code like this used to work, but no longer does:

```html
<Component :is="1 === 1 ? 'NuxtLink' : 'a'" to="/example">
  This should be a link to the "Example" page
</Component>
```

### Additional context

Two others with the same problem:

https://discord.com/channels/473401852243869706/897487139888062506/953235583403196476

https://discord.com/channels/473401852243869706/897487139888062506/953281246472056882

manniL:
With https://github.com/nuxt/framework/pull/3544, this might not be necessary anymore 😋 

Nevertheless, creating `NuxtLink` via `<component is="">` should still be possible.

ja1984:
@manniL In this example it happens to be a NuxtLink, but I think <component :is=""> is broken, I'm having issues getting it working locally, and I also took the liberty to update @ataylor32 s example https://stackblitz.com/edit/github-92nabt-973keb?file=app.vue and as you can see it does not render the test component, it tries to render a <test> component.

snarcraft:
I think it comes from this MR https://github.com/nuxt/framework/pull/3544
A workaround would be to use an instance of NuxtLink from defineNuxtLink
```
const nl = defineNuxtLink({})
```

przyb:
While the new `NuxtLink` is much more robust and you won't need to dynamically exchange it with an `a` like in the OP's example, we still need to be able to do that when we need to do a "sometime-a-link". My use case is a list of projects on a portfolio website. Some of them will link locally to a case study, others to external websites. But for some older projects, I only want to list them, so I need to replace a `NuxtLink` with a simple `div`. The workaround with `defineNuxtLink({})` worked well but I feel it would be nice to be able to import `NuxtLink` or at least have this better documented :)

jerryjappinen:
Also ran into this. I very often have components that might or might not render a link, just like @przyb mentioned there. I might have a button component that does something like this:

```js
  computed: {
[...]

PR #3814, opened 2022-03-21
feat(nuxt3): support auto-loading lazy and custom-resolved components

danielroe:
### 🔗 Linked issue

for example, https://github.com/nuxt/framework/issues/3672 - though not marking as resolving it as we may also wish to globally register `NuxtLink`

### ❓ Type of change

### 📚 Description

This PR:

* adds components and autoImports keys to `nuxt.config` documentation
* supports auto-loading `Lazy`-prefixed components
* adds type-hinting for `Lazy`-prefixed components
* supports custom resolving components with `resolveComponent` (to allow users to do more elaborate dynamic components without needing to set `global: true`, which can have a negative performance effect). See https://github.com/nuxt/framework/issues/3672 for example.

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/2.directory-structure/4.components.md
+++ b/docs/content/3.docs/2.directory-structure/4.components.md
@@ -47,6 +47,39 @@ If you have a component in nested directories such as:
 For clarity, we recommend that the component's file name matches its name. (So, in the example above, you could rename `Button.vue` to be `BaseFooButton.vue`.)
 ::
 
+## Dynamic components
+
+If you want to use the Vue `<component :is="someComputedComponent">` syntax, then you will need to use the `resolveComponent` helper provided by Vue.
+
+For example:
+
+```vue
+<template>
+  <component :is="clickable ? MyButton : 'div'" />
+</template>
+
+<script setup>
+const MyButton = resolveComponent('MyButton')
+</script>
+```
+
+Alternatively, though not recommended, you can register all your components globally, which will create async chunks for all your components and make them available throughout your application.
+
+```diff
+  import { defineNuxtConfig } from 'nuxt3'
+  
+  export default defineNuxtConfig({
+    components: {
++     global: true,
++     dirs: ['~/components']
+    },
+  })
+```
+
+::alert{type=info}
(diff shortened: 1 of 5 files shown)


pi0:
Instead of re-intorducing `lazy-` prefixed components, I believe we could expose a new global utility like `getAsyncComponent()` that basically wraps `resolveComponent` but also hints loader to use lazy imported version. In the future, we can use it for smarter prefetch of async components.

danielroe:
@pi0 So instead of:

```vue
<template>
  <LazyTheFooter v-if="someCondition" />
</template>
```

we would do:

```vue
<template>
  <component is="LazyTheFooter" v-if="someCondition" />
</template>

<script setup>
  const LazyTheFooter = resolveAsyncComponent('TheFooter')
</script>
```

Initially I think this is more verbose without need. We can use the `<LazyTheFooter>` to extract the same info for smarter prefetch. Where necessary to do something more custom, this PR already makes it possible to do `const LazyTheFooter = resolveComponent('LazyTheFooter')`, with the same effect as the `resolveAsyncComponent` call above.

pi0:
Please see https://github.com/nuxt/components/pull/212 for some context.

There are mainly two use cases for lazy loading components:

- 1) When we intentionally want a component to be lazy-loaded because of it's size, etc.
- 2) When we cannot predict their name to detect at built-time and _have to_ lazy load them.

I assume your example is the first case. When a file name ends with `.async` we assign a flag to it and regardless of the loader and global usage they are async. No need for `Lazy` prefix usage and all smart prefetching already possible.

Second is I guess what you mentioned as:

> supports custom resolving components with resolveComponent (to allow users to do more elaborate dynamic components without needing to set global: true, which can have a negative performance effect).

```vue
<template>
  <component :is="clickable ? MyButton : 'div'" />
</template>
<script setup>
const MyButton = resolveComponent('MyButton')
</script>
```

This would be:

```vue
<template>
  <component :is="clickable ? MyButton : 'div'" />
</template>
<script setup>
const MyButton = getAsyncComponent('MyButton')
</script>
```

Issue #3798, opened 2022-03-20
Components won't load when `srcDir` is set


cerino-ligutom:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.13.0`
- Nuxt Version:     `3.0.0-27460489.53fbca7`
- Package Manager:  `npm@8.3.0`
- Builder:          `vite`
- User Config:      `srcDir`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

1. Create a new nuxt3 app
2. Set `srcDir` in `nuxt.config.ts` to `src/`
3. Create a `src` folder on the root directory
4. Move `app.vue` to `src` folder
5. Create a `components` folder inside `src` folder.
6. Create a Vue SFC file inside `components` folder`.
7. Use this new component in `app.vue`
8. Start the project and go to `localhost:3000`, there should be an error on the CLI.

### Describe the bug

Setting `srcDir` on `nuxt.config.ts` seems to break the component resolution. This used to work fine but after reinstalling my dependencies, it broke.

![image](https://user-images.githubusercontent.com/6721822/159161232-6185e8ac-2563-4766-8d26-86769904b7de.png)

### Additional context

Could it be related to https://github.com/nuxt/framework/pull/3782 ?

cerino-ligutom:
Will make do with the workaround on https://github.com/nuxt/framework/issues/3746#issuecomment-1072959742 for the meantime.

PR #3815, opened 2022-03-21
fix(nuxt3): respect `srcDir` and normalize string in component dirs

### LINKED ISSUE #3746: Autoimport components dir
### Environment

- Operating System: `Linux`
- Node Version:     `v17.6.0`
- Nuxt Version:     `3.0.0-27458584.91fd16a`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `privateRuntimeConfig`, `app`, `components`, `nitro`, `srcDir`, `build`, `buildModules`, `modern`, `tailwindcss`, `meta`
- Runtime Modules:  `-`
- Build Modules:    `@nuxtjs/tailwindcss@5.0.2`

### Reproduction

--

### Describe the bug

when i trying to start my project with new version of nuxt 3 i have error:
- `[Vue warn]: Failed to resolve component: NavLink
If this is a native custom element, make sure to exclude it from component resolution via compilerOptions.isCustomElement.
Invalid value used as weak map key`

my `nuxt.config.ts`:
```ts
import { defineNuxtConfig } from 'nuxt3';

	privateRuntimeConfig: {
		secretJWT: process.env.SECRET_JWT,
	},
	app: {
		buildAssetsDir: 'assets/',
	},
	components: {
		global: true,
		dirs: ['~/components'],
	},
	nitro: {
		output: {
			dir: 'build',
		},
		externals: {
			inline: ['uuid'],
		},
	},
	srcDir: 'src',
	build: {
		splitChunks: {
			pages: true,
			layouts: true,
		},
	},
	buildModules: ['@nuxtjs/tailwindcss'],
	modern: true,
	tailwindcss: {
		configPath: '~~/tailwind.config.js',
		viewer: false,
	},
	meta: {
		title: 'Santrek Admin',
		link: [
			{
				rel: 'icon',
				href: '/favicon.svg',
				type: 'image/svg+xml',
			},
		],
	},
});

```

### Additional context

-

### Logs

```shell
-
```

danielroe:
### 🔗 Linked issue

resolves #3746, resolves https://github.com/nuxt/framework/issues/3798

### ❓ Type of change

### 📚 Description

This fixes two regressions.

1. we handle the possibility of strings in `components.dirs`
2. we resolve `~` to the `srcDir` rather than `cwd` of the layer (is there a reason we're not just using the `alias` config of the layer?)

### 📝 Checklist


packages/nuxt3/src/components/module.ts
```diff
@@ -24,39 +24,38 @@ export default defineNuxtModule<ComponentsOptions>({
...
       }
-      const dirs = (dir.dirs || [dir]).filter(_dir => _dir.path)
+      const dirs = (dir.dirs || [dir]).map(dir => typeof dir === 'string' ? { path: dir } : dir).filter(_dir => _dir.path)
       return dirs.map(_dir => ({
         ..._dir,
-        path: resolve(cwd, resolveAlias(_dir.path, {
-          ...nuxt.options.alias,
-          '~': cwd
-        }))
+        path: resolveComponentsDir(_dir.path)
```
pi0, line 50: Aliases to srcDir (`~` and `@`) should have the same value across layers. We could keep using `cwd` to allow `path: './components'` possible in layers and `options.srcDir` for resolving aliases, regardless of which layer we are working on. Later, this could be better by introducing named layers (`~theme/components`)

danielroe: I think you might need to explain the usage a little more then.

You mean `~` should refer to the parent alias? In that case it should either refer to the parent project's _source_ directory, or the layer's _source_ directory - not the `cwd` of the layer. I also think it should have the same value as `@`, and (personally) I think it is more intuitive to resolve it with regard to the _layer's_ source directory, but happy to defer on that as long as it's consistent. 

**Note**: at the moment, the parent directory is broken, even if there are no layers.

pi0: Can we please only fix top-level isue in this PR and preserve aliases? 

Would be happy to discuss later and explain trade-offs of alias handling for layers.

danielroe: Very happy to remove `@`, by the way, or make any other changes. I may have missed something that you're aware of, but I think this is a minimal PR that doesn't otherwise change the implementation other than resolving the issues linked.

Issue #3624, opened 2022-03-12
&.something scss syntax breaks styles


toniengelhardt:
### Environment

Latest Nuxt 3 and windicss, see reproduction...

### Reproduction

https://stackblitz.com/edit/nuxt3-windi-css-bug

### Describe the bug

The following `scss` syntax broke somewhere between version `3.0.0-27424151.dbab979` and the latest version `3.0.0-27451165.114cbe3`

```scss
.repo-list-item {
  @apply p-4 bg-white dark:bg-gray-800 md:(shadow-lg rounded-lg);
  &.status-0 {
    .indicator {
      @apply border-gray-300 dark:border-gray-600;
    }
  }
}
```

What happens is that once `&.something...` is added the styles for `.repo-list-item` under `@apply p-4 ...` are no longer added. Note that this is windicss.

### Additional context

The version where it still works had a bug and required `components: { global: true }`, maybe the fix for that created the problem, but just a wild guess...

<hr>

**Update** 

The following also doesn't work:

```scss
.repo-list-item {
  @apply p-4 bg-white dark:bg-gray-800 md:(shadow-lg rounded-lg);
  .indicator {
    @apply border-gray-300 dark:border-gray-600;
  }
}
```

It seems like nested `@apply`s break the ones on the top-level selector.

danielroe:
Would you raise this on `nuxt-windicss`? 🙏 

By the way, the reproduction doesn't seem to be behaving as you're suggesting - both components display the same as each other on both of versions of nuxt you mention (though on the recent version not even the grey background is showing - even without nesting).

toniengelhardt:
@danielroe so sorry, I tested something in the reproduction and forgot to fork 🙈

Now you can see the difference again. What I found is that actually not the `&` is responsible, but actually any nested selector that has an `@apply` breaks the `@apply` on the top-level selector. 

In my app the problem seems to be coming from Nuxt, not nuxt-windicss, bc it works with `3.0.0-27424151.dbab979` and doesn't work with `3.0.0-27451165.114cbe3`, every other package version and the code being the same.

I'll try to fully reproduce this in StackBlitz 🙏🏽

toniengelhardt:
@danielroe I managed to reproduce.

`3.0.0-27451165.114cbe3` broken:
https://stackblitz.com/edit/nuxt3-windi-css-bug

`3.0.0-27424151.dbab979` working: 
https://stackblitz.com/edit/nuxt3-windi-css-bug-part-2

[...]

PR #3827, opened 2022-03-22
fix(vite): sort postcss plugins to add autoprefixer & cssnano last

danielroe:
### 🔗 Linked issue

related: #3624

### ❓ Type of change

### 📚 Description

We have some code to add `cssnano` & `autoprefixer` last in webpack. This adds a similar effect for vite:

https://github.com/nuxt/framework/blob/73ba30fb69b91a25ee595daa447f3d6988b4dc8c/packages/webpack/src/utils/postcss.ts#L8-L26

### 📝 Checklist

### DIFF
--- a/packages/vite/src/css.ts
+++ b/packages/vite/src/css.ts
@@ -4,28 +4,25 @@ import type { ViteOptions } from './vite'
 import { distDir } from './dirs'
 
 export function resolveCSSOptions (nuxt: Nuxt): ViteOptions['css'] {
-  const css: ViteOptions['css'] = {
+  const css: ViteOptions['css'] & { postcss: Exclude<ViteOptions['css']['postcss'], string> } = {
     postcss: {
       plugins: []
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/vite/src/css.ts
```diff
@@ -4,28 +4,25 @@ import type { ViteOptions } from './vite'
...
-    const opts = plugins[name]
-    if (!opts) {
-      continue
-    }
-    const plugin = requireModule(name, {
-      paths: [
-        ...nuxt.options.modulesDir,
-        distDir
-      ]
+  const lastPlugins = ['autoprefixer', 'cssnano']
```
pi0, line 13: Do we need same fix for webpack too? (or normalize within schema normalizers)

danielroe: webpack already sorts (see PR description) but I can refactor to make the implementation the same.

I think it makes sense to sort at this point, rather than in schema, as modules may add their own plugins to the object and we want this to be done last.

PR #3841, opened 2022-03-22
feat(nuxi): scaffold files with `nuxi add`

Baroshem:
### 🔗 Linked issue

#3496 

### ❓ Type of change

### 📚 Description

I have created a MVP of a new CLI method to add new Nuxt 3 app elements like endpoints, plugins, etc from terminal (by using nuxi CLI). Please keep in mind that this is not a finished functionality. It is a Proof of Concept that I can develop when this functionality will be approved by core maintainers.

This is actually the first time I am contributing to this project so if I have done something wrong, please do not hesitate to correct me :)

How does it work?

When we type:

```bash
nuxi add endpoint test
```

it will automatically create a new file called `test.ts` inside `server/api` directory of our Nuxt 3 project with following content:

```ts
export default (req, res) => 'Hello test'
```

I planned to develop this for all elements of Nuxt like middlewares, components, pages, layouts, composables, etc and providing schematics so that Nuxt users could very easily generate new things. The next step would also be to provide options like i.e. for endpoints option to select a server framework (http/h3 and depending from the result provide different typings for req/res as explained in the docs -> https://v3.nuxtjs.org/docs/directory-structure/server#examples

Let me know what you think about it and as always, I am open for any feedback :)

### 📝 Checklist


pi0:
Thanks for PR @Baroshem. Scaffolding support for nuxi CLI was a really missing feature. 

In the future, I plan to introduce a new unjs package for advanced scafolding support but it is a nice idea to start with a POC. Will add few comments on changes to make POC ready to land 🚀

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,43 @@
...
+import { loadKit } from '../utils/kit'
+
+export default defineNuxtCommand({
+  meta: {
+    name: 'add',
+    usage: 'npx nuxi add endpoint test',
+    description: ''
+  },
+  async invoke(args) {
+    const rootDir = resolve('.')
```
pi0, line 13: const rootDir = resolve(args._[0] || '.')

Baroshem:
@misaon 

I am using Nest on a daily basis and really liked the experience of the CLI. Maybe we could add something similar to Nuxt ecosystem as I am working with it everyday :)

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    switch (schematic) {
+      case 'endpoint':
+        const endpointPath = `${rootDirPath}/server/api`
+        const newEndpointPath = resolve(endpointPath, `${name.toLowerCase()}.ts`)
+
+        if(!(existsSync(endpointPath) && readdirSync(endpointPath).length)) {
+          await fsp.mkdir(endpointPath, { recursive: true })
+        }
+        await fsp.writeFile(newEndpointPath, `export default (req, res) => 'Hello ${name}'`)
```
pi0, line 30: It would be nice to add a local utility function to the end of this file for `writeTemplate` and refaxtor logic.

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,43 @@
...
+    const rootDir = resolve('.')
+    const schematic = args._[0]
+    const name = args._[1];
+
+    const { loadNuxt } = await loadKit(rootDir)
+    const nuxt = await loadNuxt({ rootDir, config: { _export: true } })
+
+    const rootDirPath = nuxt.options.rootDir
+
+    switch (schematic) {
```
pi0, line 22: Instead of a switch case, we can have an object mapping from scaffold name to a function that returns `{ path, template }` like:

```js
const templates = {
  api: () => {  return { path, template  },
  plugin: () => { ... }
}
```
```

**Comment by pi0** [comment]
Certainly! From directory structure, i would like to have this templates to start with:

- component
- composable
- layout
- middlware
- page
- plugin
- api
- server-middleware / serverMiddleware (alias)
- app
- config
- nuxtignore
- tsconfig

File: packages/nuxi/src/commands/add.ts [hunk 4]
```diff
@@ -0,0 +1,43 @@
+import { defineNuxtCommand } from "./index";
+import { existsSync, promises as fsp, readdirSync } from 'fs'
+import { resolve } from 'pathe'
+import { loadKit } from '../utils/kit'
+
+export default defineNuxtCommand({
+  meta: {
+    name: 'add',
+    usage: 'npx nuxi add endpoint test',
```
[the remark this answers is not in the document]
Baroshem: In this case `test` is the name of the endpoint that is supposed to be created, not a test file (like unit test). I could add description for all supported elements. It works like this:

`nuxi add <SCHEMATIC> <NAME> <?OPTIONS>`

alvarosabu (commented):
Amazing work Jakub, looks promising. Maybe schematics for composables and components? Is an idea... maybe those 2 are to simple to scaffold without the need of having a cli cmd. Just food for thought 💚

misaon:
> @misaon
> 
> I am using Nest on a daily basis and really liked the experience of the CLI. Maybe we could add something similar to Nuxt ecosystem as I am working with it everyday :)

The combination of Nuxt 3 and NestJS is a tremendously powerful thing for efficient application development.  Maybe in the future we will see even greater integration of these two environments.  Thanks for your work!

Intevel:
Good Work 💚

packages/nuxi/src/commands/add.ts
Baroshem: I have noticed that in other commands this works quite well because there are no arguments provided by the user however, what do you think to have this rootDir argument as the last argument provided? User will have to provide a SCHEMATIC like composable and NAME like test and in your proposed solution he would also have to remember to add root dir before every command. Maybe we could use the last argument provided and by default use '.'?

Something like this:

`nuxi add composable useSomething [rootDir]` <- if rootDir is not provided use '.' instead

instead of 

`nuxi add [rootDir] composable useSomething`

Atinux: I agree that the rootDir shall be at the end and optional.

Atinux: We shall also read the `srcDir` from nuxt.config to make sure to create the files in the correct directories.

pi0: There is a lighter loadNuxtConfig available from kit for srcDir support.

Also fair point about root dir position. We can also use a named argument like --root-dir for now

jpsc:
This is a neat idea. There are a lot of good examples on how to model this looking at rails and ember for example.
There is a benefit on Nuxt 3: since most of the things are automatically imported the logic to add these templates is as simple as creating the files, we will not need to update existing files.

I would just like if these templates don't get complex enough that it "makes sense" to use a remote template. This should work without one being online 👍

Baroshem:
@pi0 @danielroe @Atinux 

Do you maybe have some recommendations for template generating library to use here? Using string interpolation will work but is really difficult to maintain and read.

pi0:
> Do you maybe have some recommendations for template generating library to use here? Using string interpolation will work but is really difficult to maintain and read.

As mentioned earlier, we plan to create one suitable for scaffolding (which was actually reason didn't add this CLI feature so far). Using string templates should be fine for now. We use them in other places and i guess there are not much dynamic features in initial templates we need.

Atinux:
Could we maybe imagine the possibility to have API routes (ex: `/api/_nuxt/nuxi`) to run the commands from an API call?

Baroshem:
@Atinux I am afraid it is out of the scope of my knowledge. I am just starting to learn nuxt framework

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+    const schematic = args._[0]
+    const name = args._[1]
+
+    const { loadNuxt } = await loadKit(rootDir)
+    const nuxt = await loadNuxt({ rootDir, config: { _export: true } })
+
+    const rootDirPath = nuxt.options.rootDir
+
+    const templates = {
+      api: {
```
pi0, line 23: Templates should be a **function** and simply return a full path to the file we expect to be generated (relative to srcDir). Example: `{ 
path: `server/api/${name}.ts` ,contents: 'export default...'`.

Write function can resolve path to srcDir then.

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+        path: `${rootDirPath}/server/api`,
+        template: `export default (req, res) => 'Hello ${name}'`,
```
pi0, line 25: It would be nice to include `defineHandle` from h3 for typing support and use object syntax to allow easily extending API:

```js
import { defineHandle } from 'h3'

export default defineHandle((req, res) => {
  return 'example response from name'
})
```

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+        fileName: name,
+        fileExtension: 'ts'
+      },
+      component: {
+        path: `${rootDirPath}/components`,
+        template: `<template>
+  <div>${name}</div>
+</template>
+
+<script setup></script>`,
```
pi0, line 41:
```suggestion
<script lang="ts" setup></script>`,
```

pi0, line 41: I guess vue3 convention is that script is above template. Also shall we add an empty style scoped block?

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+        fileName: name,
+        fileExtension: 'vue'
+      },
+      composable: {
+        path: `${rootDirPath}/composables`,
+        template: `export const ${name} = () => {
+  return useState(${name}, () => 'bar')
```
pi0, line 48: We could simply return a ref or nothing.

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+}`,
+        fileName: !name.includes('use') ? `use${name.charAt(0).toUpperCase() + name.slice(1)}` : name,
```
pi0, line 50: We can use `scule.upperFirst` (https://github.com/unjs/scule#upperfirststr)

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+        fileName: name,
+        fileExtension: 'vue'
+      },
+      page: {
+        path: name.includes('/') ? `${rootDirPath}/pages/${name.split('/')[0]}` : `${rootDirPath}/pages`,
+        template: `<template>
+  <div>${name}</div>
+</template>
+
+<script setup></script>`,
```
pi0, line 78: (same change suggestions as for component template)

Atinux:
Could be in another PR then :blush:

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+        fileName: name.includes('/') ? name.split('/')[1] : name,
+        fileExtension: 'vue'
+      }
+    }
+
+    const schema = templates[schematic]
+
+    if (!schema) { throw new Error('Schematic was not found. Pleasy try one of the schematics described in the docs') }
```
pi0, line 86: Can we please simply call these templates rather than schematic?

packages/nuxi/src/commands/add.ts
```diff
@@ -0,0 +1,102 @@
+import { existsSync, promises as fsp, readdirSync } from 'fs'
+import { resolve } from 'pathe'
+import { loadKit } from '../utils/kit'
+import { defineNuxtCommand } from './index'
+
+export default defineNuxtCommand({
+  meta: {
+    name: 'add',
+    usage: 'npx nuxi add api|plugin|component|composable|middleware|layout|page <name> [rootDir]',
```
pi0, line 9: Tip: We can now use Object.keys(templates).join('|') to generate list of available templates.

pi0: (for ordering, i would refactor templates to src/utils/templates.ts)

packages/nuxi/src/commands/add.ts
Baroshem: I wanted to implement it later with the support for options. So that later we could have `nuxi add api test --server=h3` that would result in creating the template you proposed but for the sake of simplicity I wanted to create the most basic api endpoint possible so that a user could run the default command and immediately see the result. These options could also include http as explained in the docs here -> https://v3.nuxtjs.org/docs/directory-structure/server#examples

Let me know what you think about it.

packages/nuxi/src/commands/add.ts
Baroshem: I have been thinking about it but was not sure what lang I should add for the style. For sure it should be scoped so maybe for now just leave a style without any lang. What do you think?

packages/nuxi/src/commands/add.ts
Baroshem: Do you mean something like this?

```
  return useState(${name}, () => 'bar') || null
```

packages/nuxi/src/commands/add.ts
Baroshem: Sounds good, will do that!

packages/nuxi/src/commands/add.ts
Baroshem: Not sure about that as we need file name, extension and the path later in the writeTemplate function. I am creating an empty repository with a path, then creating a file inside of this directory I need the name of the file and the extension (as some files generated are .ts and some are .vue)

packages/nuxi/src/commands/add.ts
pi0: Seems good! Suggesting ts by default for better type hints

packages/nuxi/src/commands/add.ts
pi0: I mean for the template, we can keep it almost empty. 

```js
export const useFoo = () => {
  return ref()
}
```

packages/nuxi/src/commands/add.ts
pi0: You can use dirname(path) to get parent dir. And extension is part of file name already.

packages/nuxi/src/commands/add.ts
pi0: h3 is anyway Nuxt 3 server. With nitropack, this will be simpler since with auto imports we can drop import from h3, yet wrapper is still needed.

packages/nuxi/src/commands/add.ts
Baroshem: I dont think that it is a good idea because in the docs the first example that pops up uses useState to define even a simple composable. I would recommend leaving it as it is to keep it up to date with the docs

pi0: Docs is currently a mix of toturial and API guide. It is expected one someone using this templates, is already familiar with usage. Besides that useState is more nuxt specific and needs docs to be read for it while a composables normally returns a ref. With o without universal state.

Rigo-m:
> Could we maybe imagine the possibility to have API routes (ex: `/api/_nuxt/nuxi`) to run the commands from an API call?

@Atinux Can you elaborate on this/create an RFC? I'm up for taking a chance on it.

pi0:
> Can you elaborate on this/create an RFC? I'm up for taking a chance on it.

We plan to implement CLI API service as a part of the dev tools project which I believe Sebastien's idea originated from. I will make sure to ping you with implementation issue if like to help 💚

packages/nuxi/src/utils/templates.ts
```diff
@@ -0,0 +1,55 @@
...
+    path: `/server/api/${name}.ts`,
+    template: `export default (req, res) => 'Hello ${name}'`
+  }),
+  plugin: (name: string) => ({
+    path: `/plugins/${name}.ts`,
+    template: 'export default defineNuxtPlugin(nuxtApp => {})'
+  }),
+  component: (name: string) => ({
+    path: `/components/${name}.vue`,
+    template: `<script lang="ts" setup></script>
```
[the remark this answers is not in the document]
pi0: Using template literal is easier to maintain. We can indeed do as @Rigo-m suggested or trim final strings.

Baroshem:
Guys, anything else I could do here to help making this ready to merge? :)

pi0 (approved):

PR #3850, opened 2022-03-23
fix(bridge): replace nuxt 3 auto-imports

### LINKED ISSUE #3727: Warnings import from #app and vue
### Environment

- Operating System: `Linux`
- Node Version:     `v16.4.1`
- Nuxt Version:     `2.16.0-27358576.777a4b7f`
- Builder:          `webpack`
- User Config:      `app`, `bridge`, `alias`, `ssr`, `target`, `publicRuntimeConfig`, `head`, `loading`, `css`, `plugins`, `components`, `buildModules`, `modules`, `apollo`, `i18n`, `typescript`, `router`, `vuetify`, `build`
- Runtime Modules:  `@nuxtjs/color-mode@3.0.2`, `@nuxtjs/apollo@4.0.1-rc.5`, `cookie-universal-nuxt@2.1.5`
- Build Modules:    `@nuxt/bridge@3.0.0-27457855.7c0b914`, `@pinia/nuxt@0.1.8`, `@nuxtjs/vuetify@1.12.3`, `@nuxtjs/i18n@7.2.0`

### Reproduction

https://github.com/devind-team/devind-template 

Steps:
- cd client
- yarn
- yarn run dev

### Describe the bug

I gotting warnings when run in dev and prod mode.

warn in in ./.nuxt/imports.mjs:
- "export 'clearError' was not found in '#app'  
- "export 'defineNuxtLink' was not found in '#app'
- "export 'effect' was not found in 'vue' 
- "export 'isProxy' was not found in 'vue'
- "export 'isShallow' was not found in 'vue' 
- "export 'onRenderTracked' was not found in 'vue' 
- "export 'onRenderTriggered' was not found in 'vue'
- "export 'throwError' was not found in '#app'
- "export 'useCssVars' was not found in 'vue'
- "export 'useError' was not found in '#app' 
- "export 'useTransitionState' was not found in 'vue'
- "export 'withCtx' was not found in 'vue' 
- "export 'withDirectives' was not found in 'vue' 
- "export 'withKeys' was not found in 'vue'
- "export 'withMemo' was not found in 'vue'
- "export 'withModifiers' was not found in 'vue'
- "export 'withScopeId' was not found in 'vue'

### Additional context

image

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #3727

### ❓ Type of change

### 📚 Description

With the move to unimports, we had a few regressions. Namely presets were being merged rather than replaced. (So the vue3 presets were added onto the vue2 ones in the bridge implementation.) Likely we want to entirely replace these presets, and we have a convenient hook for this purpose.

### 📝 Checklist


packages/bridge/src/auto-imports.ts
```diff
@@ -9,21 +8,23 @@ const CapiHelpers = new Set(Object.keys(CompositionApi))
 
 export function setupAutoImports () {
   const nuxt = useNuxt()
+  nuxt.hook('autoImports:sources', (presets) => {
+    presets.length = 0
```
pi0, line 12: Doesn't this override user options and other module preset extends?

packages/nuxt3/src/auto-imports/module.ts
```diff
@@ -36,7 +36,7 @@ export default defineNuxtModule<Partial<AutoImportsOptions>>({
 
     // Create a context to share state between module internals
     const ctx = createUnimport({
-      presets: defaultPresets,
+      presets: options.presets,
```
antfu, line 39: Nice catch! 😄

packages/bridge/src/auto-imports.ts
danielroe: Good point!

antfu: Can't we just provide it via the options?

danielroe: defu adds to the default array rather than replacing it.

antfu: Or maybe we cna handle the default value manually? Like

```ts
if (!options.presets.length) {
  options.presets.push(...defaultPresets)
}
```

I think users should also have this ability to override the default behaviour via options.

pi0: Before unimport migration, we had a possibility to set `enabled: false` to each preset or auto import. And bridge could iterate over presets and disable nuxt 3 built-in presets instead of removing all.

antfu: Ok I will add it to unimport now

danielroe: I've pushed a change that mutates the array in place, just to see it.

However, I don't like mutating this kind of thing though as it's the same _object_ as the `defaultPresets`. (Even setting `enabled: false` feels wrong.)

Can I suggest instead that we change the merge behaviour for the auto-imports module in this PR and use `schema.$resolve` instead? @pi0

danielroe: Oops, didn't see your message @antfu. I can update using a disabled flag, although note my comment above.

danielroe: pending on https://github.com/unjs/unimport/pull/14

pi0 (approved):

Issue #2817, opened 2022-01-19
different default scrollBehavior that resets page scroll on navigation


fanckush:
### Environment

```
- Operating System: `Darwin`
- Node Version:     `v17.3.1`
- Nuxt Version:     `3.0.0-27369360.33ebb01`
- Package Manager:  `yarn@1.22.17`
- Bundler:          `Vite`
- User Config:      `vite`, `ssr`
- Runtime Modules:  `-`
- Build Modules:    `-`
```

### Reproduction

No reproduction needed imo

### Describe the bug

When navigating using a `<NuxtLink to="/othePage" />` the page scroll is not resetting.
In other words, the page does not `scrollToTop` on route change

K-Honkawa:
It may be solved by installing a plugin for router.

https://github.com/nuxt/framework/issues/1707#issuecomment-961983939

fanckush:
@K-Honkawa thanks! i'll try that but the default behaviour should be to reset the scroll so i'll keep this open as a bug

K-Honkawa:
I don't think this is a Nuxt3 issue as NuxtLink is RouterLink in `vue-router` and is the default behavior of vue-rouer.

>  https://github.com/nuxt/framework/blob/c80870ad8c4a1c1b044d7a5ae27e41e997f8a75f/packages/nuxt3/src/pages/runtime/router.ts#L20

fanckush:
That could be the case but it is possible that NuxtLink extends RouterLink and overrides it's behaviour

daver987:
Any news on this issue, or is there a quick fix to get pages to scroll to top on page change. :)

danielroe:
@daver987 See above: https://github.com/nuxt/framework/issues/2817#issuecomment-1017155226

jamaluddinrumi:
and in case you wanna add a bit delay, you can check it out

https://github.com/nuxt/framework/discussions/1661#discussioncomment-2159828

danielroe:
This is now configurable with https://github.com/nuxt/framework/pull/3485. See https://v3.nuxtjs.org/docs/directory-structure/pages#router-options for documentation.

fanckush:
@danielroe awesome! this should replace the workaround mentioned above https://github.com/nuxt/framework/issues/2817#issuecomment-1017155226
but if my understanding is correct it doesn't fix the default behavior which is originally described in this issue. If that's the case then I won't close this just yet

Bodokh:
[...]

PR #3851, opened 2022-03-23
feat(nuxt): default router scroll behavior

joel-wenzel:
### 🔗 Linked issue

[Discussion #3338](https://github.com/nuxt/framework/discussions/3338)

### ❓ Type of change

### 📚 Description

Added new hook for `page:transition:finish` emitted from NuxtPage. Hook fires after the vue transition `afterLeave` event.

Also added default scroll behavior, which can still be overridden by using `app/router.options`

### 📝 Checklist


packages/nuxt3/src/pages/routing/default-router.options.ts
```diff
@@ -0,0 +1,86 @@
...
+
+if (process.client) {
+  if ('scrollRestoration' in window.history) {
+    setScrollRestoration('manual')
+
+    // reset scrollRestoration to auto when leaving page, allowing page reload
+    // and back-navigation from other pages to use the browser to restore the
+    // scrolling position.
+    window.addEventListener('beforeunload', () => {
+      setScrollRestoration('auto')
```
posva, line 19: I don't think you need this whole block as the router already handles on load scroll

packages/nuxt3/src/pages/routing/default-router.options.ts
```diff
@@ -0,0 +1,86 @@
...
+        // coords will be used if no selector is provided,
+        // or if the selector didn't match any element.
+        if (to.hash) {
+          let hash = to.hash
+          // CSS.escape() is not supported with IE and Edge.
+          if (
+            typeof window.CSS !== 'undefined' &&
+            typeof window.CSS.escape !== 'undefined'
+          ) {
+            hash = '#' + window.CSS.escape(hash.substr(1))
```
posva, line 61: I think you can go much simpler 🙂 
The router already tries to handle most scenarios: https://github.com/vuejs/router/blob/main/src/scrollBehavior.ts#L77

packages/nuxt3/src/pages/routing/default-router.options.ts
```diff
@@ -0,0 +1,86 @@
...
+          }
+          try {
+            const elem = document.querySelector(hash)
+
+            // vue-router does not incorporate scroll-margin-top on its own.
+            if (elem) {
+              const offset = parseFloat(getComputedStyle(elem).scrollMarginTop)
+              position = {
+                selector: hash,
```
posva, line 70: `selector` was deprecated in favor of `el`

packages/nuxt3/src/pages/runtime/page.ts
```diff
@@ -16,14 +16,36 @@ export default defineComponent({
...
+
     return () => {
       return h(RouterView, {}, {
         default: (routeProps: RouterViewSlotProps) => routeProps.Component &&
-            _wrapIf(Transition, routeProps.route.meta.pageTransition ?? defaultPageTransition,
-              wrapInKeepAlive(routeProps.route.meta.keepalive, h(Suspense, {
-                onPending: () => nuxtApp.callHook('page:start', routeProps.Component),
-                onResolve: () => nuxtApp.callHook('page:finish', routeProps.Component)
-              }, { default: () => h(routeProps.Component, { key: generateRouteKey(props.pageKey, routeProps) } as {}) }))).default()
+          _wrapIf(Transition, getTransitionProps(routeProps),
```
posva, line 42: Here you are always wrapping with Transition but this is not always desired

packages/nuxt3/src/pages/routing/default-router.options.ts
joel-wenzel: @posva Cleaned that up. The only thing that seemed like it wasn't handled was `scrollMarginTop` so I left that in.

packages/nuxt3/src/pages/runtime/page.ts
joel-wenzel: @posva Going through the _wrapIf it should only render the Transition component if the transition props are truthy unless I am missing something

posva: I will check again, I read the diff a bit too quickly

pi0:
@joel-wenzel sorry, your PR got stalled. Can you please consider rebasing it?

@posva Can you please confirm if this changes are looking good to you or any change needed?

fanckush:
Merging this PR would fix the issue #2817 btw

packages/nuxt3/src/pages/routing/default-router.options.ts
```diff
@@ -0,0 +1,47 @@
...
+import { nextTick } from 'vue'
+import { useNuxtApp } from '#app'
+
+// https://router.vuejs.org/api/#routeroptions
+export default <RouterOptions>{
+  scrollBehavior (to, from, savedPosition) {
+    const nuxtApp = useNuxtApp()
+    // If the returned position is falsy or an empty object, will retain current scroll position
+    let position
+    const isRouteChanged = to !== from
```
vis97c, line 11: Please support for cases were the only difference is the hash (anchor links), in those cases the scroll should be handled by the browser and not reseted every time.

I was going to add it as an issue but since there is a pending MR i think is best to have it here.

vis97c (changes requested):

joel-wenzel:
> @joel-wenzel sorry, your PR got stalled. Can you please consider rebasing it?

@pi0 No problem. Rebased. Should be all good now

packages/nuxt3/src/pages/routing/default-router.options.ts
joel-wenzel: @vis97c Can you provide a reproduction of the use case you are trying to solve/avoid or elaborate a little further? 
I pushed a change to handle anchor tag navigation when on the same page and it works I as I would expect. 

Returning nothing from scroll behavior for those scenarios doesn't delegate anything to the browser. I could be missing something though

vis97c: Yeah sure. Not sure if it needs reproduction but using anchor links on a page triggers a router update.

```html
<a href="#about">about</a>
<div id="about"></div>
```

When clicking the link the page should be scrolling to whatever position "#about" div I located but the current approach would reset the scroll to 0 preventing that from happening. Let me know if the example is not clear enough so I can implement a reproduction.

joel-wenzel: That scenario is handled here for sure. This is my third test page:
```html
<div>
    <a href="#hash">Hash</a>
    <a href="#hash2">Hash2</a>
    <div style="padding-top: 2000px">
      child child page
    </div>
    <h6 id="hash" style="scroll-margin-top: 50px;">
      Should go here
      <a href="#hash2">Hash2</a>
    </h6>
    <h6 id="hash2" style="scroll-margin-top: 50px; margin-top: 500px">
      Should go here
    </h6>
    <p style="margin-bottom: 2000px">
      Lorem ipsum dolor sit amet consectetur adipisicing elit. Omnis, modi quis dolorum asperiores impedit aspernatur nam! Aut illo explicabo eligendi rem consequuntur quam! Esse cupiditate officiis ullam nihil laudantium porro.
    </p>
  </div>
```

joel-wenzel:
@pi0 I merged instead of rebase. If you would rather I make a new branch/PR let me know

fanckush:
bump

Rigo-m:
> When will it be merged? Current Nuxt 3 page link behaviour is counter-intuitive because page sticks on current position of previous page.
> 
> Update 2022-06-04: When playing around with Content Wind I noticed that scroll behaviour was as expected. So I checked for a custom router configuration et voila:
> 
> ```
> import type { RouterConfig } from '@nuxt/schema'
> 
> // https://router.vuejs.org/api/#routeroptions
> export default <RouterConfig>{
>   scrollBehavior (to, _, savedPosition) {
>     const nuxtApp = useNuxtApp()
> 
>     // If history back
>     if (savedPosition) {
>       // Handle Suspense resolution
>       return new Promise((resolve) => {
>         nuxtApp.hooks.hookOnce('page:finish', () => {
>           setTimeout(() => resolve(savedPosition), 50)
>         })
>       })
>     }
>     // Scroll to heading on click
>     if (to.hash) {
>       setTimeout(() => {
>         const heading = document.querySelector(to.hash) as any
> 
>         return window.scrollTo({
>           top: heading.offsetTop,
>           behavior: 'smooth'
>         })
>       })
>       return
>     }
> 
>     // Scroll to top of window
>     return { top: 0 }
>   }
> }
> ```
> 
> This might also solve your problems with latest plain Nuxt 3 apps.

The problem of this, at least in content-wind, is that the scroll starts when the next page has yet to be mounted

axelthat:
Any updates?

joel-wenzel:
> Any updates?

I am waiting until I hear from the core team before fixing merge conflicts. I have fixed them 3 or 4 times now only to become outdated again

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -21,17 +21,39 @@ export default defineComponent({
...
+            _wrapIf(Transition, getTransitionProps(routeProps),
               wrapInKeepAlive(routeProps.route.meta.keepalive,
                 isNested && nuxtApp.isHydrating
                   // Include route children in parent suspense
                   ? h(routeProps.Component, { key: generateRouteKey(props.pageKey, routeProps) } as {})
                   : h(Suspense, {
                     onPending: () => nuxtApp.callHook('page:start', routeProps.Component),
-                    onResolve: () => nuxtApp.callHook('page:finish', routeProps.Component)
+                    onResolve: () => {
+                      nextTick(() => nuxtApp.callHook('page:finish', routeProps.Component))
```
posva, line 55: won't this trigger multiple times for nested routes? That would be problematic

posva: Ah I see, it's only for root transitions :+1:

joel-wenzel:
@pi0 No worries. Should be all set now if you want to test.

axelthat:
Nuxt team, please look into this 🙏

jakubM817:
> Nuxt team, please look into this 🙏

please too

nandi95:
This new hook might help with this issue: https://github.com/vuejs/core/issues/5844

Gyurmatag:
When this will be merged? :O

rashidpathiyil:
Waiting for a solution.....

axelthat:
Any new updates?

kemalcany:
Hoping that this PR will find itself merged into one of the upcoming rcs soon!

jiblett1000:
> Hoping that this PR will find itself merged into one of the upcoming rcs soon!

You can snag the one from the 'Wind' nuxt content template. Just copy the `/app/router.config.ts`.

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -28,20 +27,42 @@ export default defineComponent({
...
+    function getTransitionProps (routeProps: RouterViewSlotProps) {
+      const metaTransition = routeProps.route.meta.pageTransition
+      const onAfterLeave = () => {
+        nuxtApp.callHook('page:transition:finish', routeProps.Component)
+      }
+
+      if (typeof metaTransition === 'boolean') {
+        return metaTransition && {
+          ...defaultPageTransition,
+          onAfterLeave
```
nandi95, line 39: maybe for bit of safety we could check if `onAfterLeave` was already defined so we're not overwriting it.
If defined, you could wrap that in a function where you call the defined callback and then emit the event

joel-wenzel: @nandi95 good thinking, added it

packages/nuxt/src/pages/routing/default-router.options.ts
```diff
@@ -0,0 +1,60 @@
...
+    const nuxtApp = useNuxtApp()
+    // If the returned position is falsy or an empty object, will retain current scroll position
+    let position
+    const isRouteChanged = to !== from
+
+    // savedPosition is only available for popstate navigations (back button)
+    if (savedPosition) {
+      position = savedPosition
+    } else if (isRouteChanged) {
+      position = { left: 0, top: 0 }
```
Atinux, line 17: Something to note here is that we may not want to `scrollToTop` if we are changing routes inside a child page.

See the reference in Nuxt 2: https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/router.scrollBehavior.js#L25-L32

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -34,22 +35,26 @@ export default defineComponent({
...
+            defaultPageTransition,
+            { onAfterLeave: () => { nuxtApp.callHook('page:transition:finish', routeProps.Component) } }
+          ].filter(Boolean))
+          return _wrapIf(Transition, hasTransition && transitionProps,
             wrapInKeepAlive(props.keepalive ?? routeProps.route.meta.keepalive ?? (defaultKeepaliveConfig as KeepAliveProps), h(Suspense, {
               onPending: () => nuxtApp.callHook('page:start', routeProps.Component),
-              onResolve: () => nuxtApp.callHook('page:finish', routeProps.Component).finally(done)
-            }, { default: () => h(Component, { key, routeProps, pageKey: key, hasTransition: !!transitionProps } as {}) })
+              onResolve: () => {
+                nextTick(() => nuxtApp.callHook('page:finish', routeProps.Component).finally(done))
```
pi0, line 55: Heads up @danielroe we are calling `page:finish` in nextTick by this PR

packages/nuxt/src/pages/routing/default-router.options.ts
pi0: Added a simple fix with `3ec9051b077e0ad8ae2ff8bb03bd5c3c428a65a9`. BTW thanks for reference. I think in next steps we shall also support `scrollToTop` pageMeta.

pi0 (approved):

pi0:
Thank you so much for working on this @joel-wenzel and sorry it took long to land!

I've made a few refactors and fixes on top of your work.

lamdinh95:
Hello @pi0, I've updated to rc.13 today but the scroll behaviour do not work correctly.  When I navigate back, the scroll doesn't go to the save position but it's reset instead. 
The problem cause by this line in file: packages/nuxt/src/pages/runtime/router.options.ts
`const hasTransition = to.meta.pageTransition !== false && from.meta.pageTransition !== false;`
to.meta.pageTransition was undefined so the strict equal not work in here.

danielroe:
This should be resolved in the [edge channel](https://v3.nuxtjs.org/guide/going-further/edge-channel/), or in the next RC.

Let me know if not.

PR #3854, opened 2022-03-23
fix(vite-node): entry path and executing in production

antfu:
### 🔗 Linked issue

https://github.com/nuxt/nuxt3-docs/pull/2

### ❓ Type of change

### 📝 Checklist


packages/vite/src/vite.ts
```diff
@@ -32,6 +32,8 @@ export async function bundle (nuxt: Nuxt) {
     ports: Array.from({ length: 20 }, (_, i) => hmrPortDefault + 1 + i)
   })
 
+  const asyncEntry = nuxt.options.experimental.asyncEntry || nuxt.options.experimental.viteNode
```
pi0, line 35: Should we also use dev check? Async entry is not needed for production build with vite-node

antfu: Maybe it's better to directly mutate `experimental.asyncEntry` if viteNode is true? If so, where do you think it's the best place to do that?

pi0: We can use schema `$resolve` to infer default value of `asyncEntry` to be enabled on `dev && viteNode` condition. (example usage of $resolve: https://github.com/nuxt/framework/blob/main/packages/schema/src/config/_common.ts#L473)

antfu: Possible to access nuxt's config? Like this?

```js
 $resolve: (val, get) => val ?? get('dev') && get('expierments.viteNode') ?? false,
```

pi0: Yes we can :)

packages/vite/src/vite-node.ts
```diff
@@ -55,7 +56,11 @@ function createViteNodeMiddleware (ctx: ViteBuildContext): Connect.NextHandleFun
 }
 
 export async function prepareDevServerEntry (ctx: ViteBuildContext) {
-  const entryPath = resolve(ctx.nuxt.options.appDir, 'entry.async')
+  let entryPath = resolve(ctx.nuxt.options.appDir, 'entry.async.mjs')
+  if (!fse.existsSync(entryPath)) {
+    entryPath = resolve(ctx.nuxt.options.appDir, 'entry.async')
+  }
```
antfu, line 62: This is a workaround. In production build, the `.mjs` is required for node to be able the `.mjs` file, while in dev stub, it uses `.ts` and failed load. Not sure if there is a good solution for that?

pi0 (approved):
LGTM. Only waiting for @danielroe to confirm if there are no concerns with using vite in dev with async entry

danielroe:
I can't seem to test this as it throws an error when I run it on the playground. Previously using an async entry threw an error on the _first_ client-side load. If that's not happening, then I don't have any concerns with using an async entry in dev.

pi0:
Thanks for checking @danielroe. I guess issue with #_config is another topic. Let's have this then if there are no concerns.

PR #3884, opened 2022-03-24
feat: support async plugins and middlewares

### LINKED ISSUE #3438: restore context with ast transformer for middleware & plugins
We could support a AST transformer to restore context similar to `script setup` and use for `defineNuxtRouteMiddleware`, `defineNuxtPlugin`, `async setup() {}`, etc making it possible to use navigate utils in async contexts.

(as a simpler solution, we could move `navigateTo` and `abortNavigation` to `nuxtApp` instance? `useNuxtApp()` can be cached before any await operation)

_Originally posted by @pi0 in https://github.com/nuxt/framework/issues/3105#issuecomment-1031788002_

antfu:
### 🔗 Linked issue

Close #3438

~~Currently, only async plugin is supported. Middleware is on the way. You can merge this for now, or I will update this PR later.~~

### ❓ Type of change

### 📝 Checklist


packages/nuxt3/src/core/plugins/unctx.ts
```diff
@@ -0,0 +1,31 @@
...
+  let app: NuxtApp | undefined
+  let middlewares: NuxtMiddleware[] = []
+  nuxt.hook('app:resolve', (_app) => { app = _app })
+  nuxt.hook('pages:middleware:extend', (_middlewares) => { middlewares = _middlewares })
+
+  return createUnplugin(() => ({
+    name: 'unctx:transfrom',
+    enforce: 'post',
+    transformInclude (id) {
+      return Boolean(app?.plugins.find(i => i.src === id) || middlewares.find(m => m.path === id))
```
antfu, line 19: I guess the failing tests on Windows are caused by module id not matching with the source path here.

I didn't not their shape so not sure how to normalize them. I am setting up my Windows env to test it, which might take some time. Before that any help is greatly welcome!

packages/nuxt3/src/core/plugins/unctx.ts
```diff
@@ -0,0 +1,31 @@
...
+  const transformer = createTransformer({
+    asyncFunctions: ['defineNuxtPlugin', 'defineNuxtRouteMiddleware', 'withAsyncContext', 'callAsync']
+  })
+
+  let app: NuxtApp | undefined
+  let middlewares: NuxtMiddleware[] = []
+  nuxt.hook('app:resolve', (_app) => { app = _app })
+  nuxt.hook('pages:middleware:extend', (_middlewares) => { middlewares = _middlewares })
+
+  return createUnplugin(() => ({
```
pi0, line 15: FYI, i've added a `/plugin` export from unctx for the plugin (https://github.com/unjs/unctx/blob/main/src/plugin.ts) which we could use.

packages/nuxt3/src/core/plugins/unctx.ts
```diff
@@ -0,0 +1,31 @@
...
+import { createTransformer } from 'unctx/transform'
+import { createUnplugin } from 'unplugin'
+
+export const UnctxTransformPlugin = (nuxt: Nuxt) => {
+  const transformer = createTransformer({
+    asyncFunctions: ['defineNuxtPlugin', 'defineNuxtRouteMiddleware', 'withAsyncContext', 'callAsync']
+  })
+
+  let app: NuxtApp | undefined
+  let middlewares: NuxtMiddleware[] = []
```
pi0, line 11: Grammar: The plural form of `middleware` is also `middleware` unless you mean different kinds of middleware. (@danielroe correct me please if wrong)

packages/nuxt3/src/core/plugins/unctx.ts
```diff
@@ -0,0 +1,31 @@
+import { Nuxt, NuxtApp, NuxtMiddleware } from '@nuxt/schema'
+import { createTransformer } from 'unctx/transform'
+import { createUnplugin } from 'unplugin'
+
+export const UnctxTransformPlugin = (nuxt: Nuxt) => {
+  const transformer = createTransformer({
+    asyncFunctions: ['defineNuxtPlugin', 'defineNuxtRouteMiddleware', 'withAsyncContext', 'callAsync']
```
pi0, line 7: unctx, concats and preserves built-ins (https://github.com/unjs/unctx/blob/main/src/transform.ts#L31)

```suggestion
    asyncFunctions: ['defineNuxtPlugin', 'defineNuxtRouteMiddleware']
```

packages/nuxt3/src/core/plugins/unctx.ts
pi0: (after investigating windows) maybe is this something worth to normalize with unplugin when calling transformInclude?

packages/nuxt3/src/app/nuxt.ts
```diff
@@ -171,26 +174,17 @@ export function isLegacyPlugin (plugin: unknown): plugin is LegacyPlugin {
...
 /**
  * Ensures that the setup function passed in has access to the Nuxt instance via `useNuxt`.
  *
  * @param nuxt A Nuxt instance
  * @param setup The function to call
  */
 export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp | _NuxtApp, setup: T, args?: Parameters<T>) {
-  setNuxtAppInstance(nuxt as NuxtApp)
-  const p: ReturnType<T> = args ? setup(...args as Parameters<T>) : setup()
-  if (process.server) {
```
pi0, line 189: Side effect: Since we assume there is once instance of nuxt active per window, it was safe to not unset `nuxt` in client-side while `callAsync` does unset. 

Cons with this change is that we have a more consistant restriction but downside is breaking change and also losing the ability to simply use nuxt composables in global context of client for various purposes.

pi0: In order to revert back, we could use singleton methods of unctx and `callAsync` in server-side. This also helps os saving performance by only doing async transform for server bundle.

packages/nuxt3/src/core/plugins/unctx.ts
antfu: Wondering what's the difference between them? Is it related to `file://`, `c:/` or just slashes?

packages/nuxt3/src/core/plugins/unctx.ts
antfu: Interesting, but I actually hesitate about this behavior of `unctx`, since then you can't disable the transformation of those built-in entries.

pi0: Well, to be fair we shouldn't even include `callAsync`. It was only nice for testing purposes of unctx without using `withAsyncContext` in each test. In real use cases, `callAsync(obj)`, obj is an external dependency and `withAsyncContext` is a kinda special case but agree we could make them opt-in.

packages/nuxt3/src/core/plugins/unctx.ts
antfu: Yeah, I know it. But using it means we need to pass the options to the unplugin like:

```ts
UnctxTransformPlugin().vite({ ...options })

// instead of
UnctxTransformPlugin(nuxt).vite()
```

Meanwhile having the unplugin directly would allow us to have better control of it.

antfu:
This is ready :)

packages/nuxt3/src/core/plugins/unctx.ts
pi0: We can always map nuxt options to plugin options. Are you thinking of any nuxt specific plugin option we might not be able to easily pass?

antfu: Since the options of the unctx are handled by this wrapper, which means if we want to reuse it, we'd need to have:

```ts
const options = { ... }
return {
  vite: () => unctxPlugin.vue(options),
  webpack: () => unctxPlugin.webpack(options),
}
```

I just thought it's much code to reuse anyway, but if you prefer I can change it like the code above.

packages/nuxt3/src/core/nuxt.ts
```diff
@@ -57,14 +58,17 @@ async function initNuxt (nuxt: Nuxt) {
...
   const config = {
     rootDir: nuxt.options.rootDir,
     patterns: vueAppPatterns(nuxt)
   }
   addVitePlugin(ImportProtectionPlugin.vite(config))
   addWebpackPlugin(ImportProtectionPlugin.webpack(config))
 
+  // Add unctx transform
+  addVitePlugin(UnctxTransformPlugin(nuxt).vite())
+  addWebpackPlugin(UnctxTransformPlugin(nuxt).webpack())
```
pi0, line 70: I'm wondering why we need to create `UnctxTransformPlugin` wrapper two times and not creating once (`unctxPlugin(options)`) and pass as `instance.vite()` / `instance.webpack()`?

packages/nuxt3/src/core/plugins/unctx.ts
pi0: I'm wondering if unplugin wrapper options could be a factory function, we could wrap nuxt logic like this:

```js
const plugin = unctxPlugin(() => {
  let app: NuxtApp | undefined
  let middleware: NuxtMiddleware[] = []
  nuxt.hook('app:resolve', (_app) => { app = _app })
  nuxt.hook('pages:middleware:extend', (_middlewares) => { middleware = _middlewares })
  return {
    asyncFunctions: ['defineNuxtPlugin', 'defineNuxtRouteMiddleware'],
    transformInclude (id) {
      return Boolean(app?.plugins.find(i => i.src === id) || middleware.find(m => m.path === id))
    }
  }
})
```

pi0: I'm wondering if unplugin wrapper options could be a factory function, we could wrap nuxt logic like this:

```js
const plugin = unctxPlugin(() => {
  let app: NuxtApp | undefined
  let middleware: NuxtMiddleware[] = []
  nuxt.hook('app:resolve', (_app) => { app = _app })
  nuxt.hook('pages:middleware:extend', (_middlewares) => { middleware = _middlewares })
  return {
    asyncFunctions: ['defineNuxtPlugin', 'defineNuxtRouteMiddleware'],
    transformInclude (id) {
      return Boolean(app?.plugins.find(i => i.src === id) || middleware.find(m => m.path === id))
    }
  }
})
```

pi0 (approved):
Overall looks good 💯 Let's revise unplugin implementation later.

packages/nuxt3/src/core/plugins/unctx.ts
```diff
@@ -0,0 +1,31 @@
+import { Nuxt, NuxtApp, NuxtMiddleware } from '@nuxt/schema'
+import { createTransformer } from 'unctx/transform'
+import { createUnplugin } from 'unplugin'
+
+export const UnctxTransformPlugin = (nuxt: Nuxt) => {
+  const transformer = createTransformer({
+    asyncFunctions: ['defineNuxtPlugin', 'defineNuxtRouteMiddleware']
```
pi0, line 7: Small note: Could we add `withAsyncContext` (from unctx) to auto imports? This change would allow to use async pattern in rest of composables without explicitly requiring us to mark them.

PR #3894, opened 2022-03-25
docs: update `serverMiddleware` to add warning on function and make example esm

### LINKED ISSUE #3891: @nuxt/kit - addServerMiddleware does not seem to run in production.
### Environment

- Operating System: `Darwin`
- Node Version:     `v17.5.0`
- Nuxt Version:     `2.15.8`
- Package Manager:  `npm@8.4.1`
- Builder:          `webpack`
- User Config:      `build`, `modules`, `contentful`
- Runtime Modules:  `nuxt-3-contentful-module@0.1.0`
- Build Modules:    `-`

NOTE: Runtime module is a locally developed module.

### Reproduction

1. Create a local module which uses @nuxt/kit addServerMiddleware,
2. add a simple middleware with a path and handler in the form of a function rather than a module path,
3. run standard dev script, middleware works as expected,
4. run standard build script, middleware does not seem to have been appended to the server middleware option.

### Describe the bug

See reproduction above.

**NOTES:**

- Server middleware passed in automatically seems to work fine, but the object form with a function passed as the handler as [described in docs](https://v3.nuxtjs.org/docs/directory-structure/nuxt.config#servermiddleware) does not, either via the kit helper method or directly in nuxt.config.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #3891

### ❓ Type of change

### 📚 Description

This clarifies that functions will only run in development mode. Also updated syntax of example to ESM (might be worth changing the example too 🤔).

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -344,7 +344,7 @@ export default {
    *   'redirect-ssl',
    *   // Will register file from project server-middleware directory to handle /server-middleware/* requires
    *   { path: '/server-middleware', handler: '~/server-middleware/index.js' },
-   *   // We can create custom instances too - but only in development mode
+   *   // We can create custom instances too, but only in development mode, they are ignored for the production bundle.
    *   { path: '/static2', handler: serveStatic(__dirname + '/static2') }
    * ]
    * ```
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -328,12 +328,14 @@ export default {
    * Server middleware are connect/express/h3-shaped functions that handle server-side requests. They
    * run on the server and before the Vue renderer.
    *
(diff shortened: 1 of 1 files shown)


packages/schema/src/config/_common.ts
```diff
@@ -328,21 +328,23 @@ export default {
...
+   *
    * @example
    * ```js
    * serverMiddleware: [
    *   // Will register redirect-ssl npm package
    *   'redirect-ssl',
    *   // Will register file from project server-middleware directory to handle /server-middleware/* requires
    *   { path: '/server-middleware', handler: '~/server-middleware/index.js' },
-   *   // We can create custom instances too
+   *   // We can create custom instances too - but only in development mode
```
pi0, line 347:
```suggestion
   *   // We can create custom instances too, but only in development mode, they are ignored for the production bundle.
```

PR #3897, opened 2022-03-25
fix(nuxt3): improve types of `uniqueBy` utility

kevinmarrec:
### ❓ Type of change

### 📚 Description

Improves types for the `uniqueBy` utility function, enabling autocompletion on second parameter `key` 😎

image

@pi0 This utility being used at two places in the same package (**nuxt3**), I'll let you refactor it to only define it once somewhere :)


pi0:
Thank you!

PR #3900, opened 2022-03-25
fix(bridge): define mixin outside plugin, to avoid memory leak

danielroe:
### 🔗 Linked issue

follow up on https://github.com/nuxt/framework/pull/3898

### ❓ Type of change

### 📚 Description

this defines the meta mixin outside the plugin so we can add the installed flag in bridge...

### 📝 Checklist


pi0:
Since we are already guarding bridge-compat, is this change necessary?

danielroe:
The issue is that the mixin object is defined inline, which means the `__nuxt_installed` flag is set freshly each time, and it can't deduplicate. By moving it out of function, we get the protection applied.

PR #3905, opened 2022-03-25
fix(schema): add nitropack server types

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/content-next/pull/82#issuecomment-1079091868

### ❓ Type of change

### 📚 Description

If the user does not have `nitropack` installed this falls back to `any` type.

@pi0 If you think we need to defer further, happy to pend this.

### 📝 Checklist


pi0:
Shall we delagate this to later? I would probably use a better name to access server form nuxt as devServer. 

We might use an inline typing if you would like to for nuxt2 only features and depcated flags...

danielroe:
Yes, I'm happy to wait until the move to nitropack. We should likely also provide a type export for the kind of server exported, which I'll pick up there.

PR #3906, opened 2022-03-25
fix(bridge): don't call nuxt app if instance is inaccessible

### LINKED ISSUE #3904: nuxt app instance unavailable error on all pages
Sorry for wrongly opened issue. I cannot edit labels. This issue related with ```bridge``` 🙏 

> @antlionguard Thanks - following up with a PR on that mixin: #3900.

There's something broke after this PR. I got this two error in my all pages. @danielroe 

Firstly i got this:
resim

and after refresh i got this:
resim

_Originally posted by @antlionguard in https://github.com/nuxt/framework/issues/3431#issuecomment-1079097686_

danielroe:
### 🔗 Linked issue

resolves  #3904

### ❓ Type of change

### 📚 Description

We don't need to call `useNuxtApp` (and indeed cannot) if there is no instance.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/meta/runtime/plugin.ts
+++ b/packages/nuxt3/src/meta/runtime/plugin.ts
@@ -14,9 +14,9 @@ const metaMixin = {
   [metaConfig.mixinKey] () {
     const instance = getCurrentInstance()
     const options = instance?.type || /* nuxt2 */ instance?.proxy?.$options
-    const nuxtApp = useNuxtApp()
     if (!options || !('head' in options)) { return }
 
+    const nuxtApp = useNuxtApp()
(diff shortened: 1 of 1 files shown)


packages/nuxt3/src/meta/runtime/plugin.ts
```diff
@@ -14,9 +14,9 @@ const metaMixin = {
   [metaConfig.mixinKey] () {
     const instance = getCurrentInstance()
```
pi0, line 15: Shall we early return when instance is undefined? 

Current change has similar effect but we do unnecessary operations and it is not clear in code with behavior.

PR #3908, opened 2022-03-25
docs: remove unnecessary `return` in example

harmyderoman:
No need to return `abortNavigation` and `navigateTo` in `defineNuxtRouteMiddleware`

### ❓ Type of change

Docs

### 📚 Description

No need to return `abortNavigation` and `navigateTo` in `defineNuxtRouteMiddleware`

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/2.directory-structure/7.middleware.md
+++ b/docs/content/3.docs/2.directory-structure/7.middleware.md
@@ -27,9 +27,9 @@ Route middleware are navigation guards that receive the current route and the ne
 ```js
 export default defineNuxtRouteMiddleware((to, from) => {
   if (to.params.id === '1') {
-    return abortNavigation()
+    abortNavigation()
   }
-  return navigateTo('/')
+  navigateTo('/')
 })
 ```
 


pi0:
Thanks for PR @harmyderoman. Indeed they are not needed but semantically making example readable that calling those functions causes rest of the logic to stop. Besides navigateTo is likely to return a promise which preserving chain is better than discarding.

pi0:
It shouldn't. I'll investigate.

danielroe:
Would you share a reproduction @harmyderoman? 🙏

harmyderoman:
Sorry guys, I did that infinity loop by trying to run `return navigateTo(to.path)`. 😄 
So stupid error.
It's just not so clear from the docs what this function returns and how to use it properly.

pi0:
No worries @harmyderoman and indeed thanks for sharing the cause. I think we might have improved redirectTo to avoid such loops.

harmyderoman:
Cool! So I'll close this PR.
Thank you for your great work!

lixiaofa:
I also did that infinity loop by trying to run `return navigateTo(to.path)` .So stupid mistake.
What is the correct writing method?

> No worries @harmyderoman and indeed thanks for sharing the cause. I think we might have improved redirectTo to avoid such loops.

 I also did that infinity loop by trying to run `return navigateTo(to.path)` .So stupid mistake.
What is the correct writing method？
@pi0 @harmyderoman

harmyderoman:
@lixiaofa if it's not fixed in new version, try it without `return`

danielroe:
You should not navigate to the current path as that will perform an extra navigation. Just return:

```js
if (whitelist.includes(to.path)) {
  return
}
```

You can read more on https://router.vuejs.org/guide/advanced/navigation-guards.html#navigation-guards.

lixiaofa:
> @lixiaofa if it's not fixed in new version, try it without `return`

All permissions will be separated if they do not return. 
Failed to achieve the effect of routing guard.

lixiaofa:
> You should not navigate to the current path as that will perform an extra navigation. Just return:
> 
> ```js
> if (whitelist.includes(to.path)) {
>   return
> }
> ```
> 
> You can read more on https://router.vuejs.org/guide/advanced/navigation-guards.html#navigation-guards.

Thank you very much！

Issue #3385, opened 2022-02-25
Throw "Hydration node mismatch" error when using teleport


msidolphin:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.11.0`
- Nuxt Version:     `3.0.0-27428764.271289c`
- Package Manager:  `yarn@1.22.10`
- Bundler:          `Vite`
- User Config:      `meta`, `css`, `build`, `buildModules`, `components`, `vite`, `vueuse`, `windicss`
- Runtime Modules:  `-`
- Build Modules:    `@pinia/nuxt@0.1.8`, `@nuxtjs/svg@0.4.0`, `@vueuse/nuxt@7.6.2`, `nuxt-windicss@2.2.6`, `unplugin-icons/nuxt`
------------------------------

### Reproduction

https://stackblitz.com/edit/github-dyh5gp?file=app.vue

### Describe the bug

here is mini reproduction
```vue
<template>
  <teleport to="body" disabled>
      <div v-show="visible">Hydration node mismatch</div>
  </teleport>
</template>

<script lang="ts" setup>
import { defineProps } from 'vue'
defineProps({
    visible: {
        type: Boolean,
        default: false,
    }
})
</script>
```

```vue
<template>
  <teleport-example :visible="state.visible"></teleport-example>
  <button @click="state.visible = !state.visible">toggle1</button>
</template>

<script lang="ts" setup>
import TeleportExample from './comp.vue'
const state = reactive({
  visible: false,
});
</script>

```

<img width="864" alt="image" src="https://user-images.githubusercontent.com/26672484/155649509-79151e9f-3026-4e8b-ba0b-8a7b1585e80f.png">

our users have given us feedback on this issue [6233](https://github.com/element-plus/element-plus/issues/6233)

### Logs

```shell
[Vue warn]: Hydration node mismatch:
- Client vnode: button 
- Server rendered DOM: <div style=​"display:​none;​">​Hydration node mismatch​</div>​  
  at <Examples> 
  at <Index onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< undefined > key="/" > 
  at <BaseTransition mode="out-in" appear=false persisted=false  ... > 
  at <Transition name="page" mode="out-in" > 
  at <RouterView > 
  at <NuxtPage> 
  at <App> 
  at <NuxtRoot>
```

danielroe:
Let's follow in #1907

PR #3909, opened 2022-03-25
feat(nitro): support for rendering ssr teleports to body

### LINKED ISSUE #1907: Hydration Error at use teleport
### Environment

* node: LTS
* nuxt: latest

### Reproduction

https://codesandbox.io/s/nuxt3-teleport-bug-6b2vr

### Describe the bug

Hello 😀

Hydration error occurs when using teleport in nuxt3.
I inserted a teleport target in app.vue and it works normally until code movement.
However, since hydration mismatch occurs, the function of the code to be teleported cannot operate normally.

### Additional context

_No response_

### Logs

```shell
[Vue warn]: Hydration children mismatch in <div>: server rendered element contains fewer child nodes than client vdom. 
  at <Example> 
  at <Index onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< 
undefined
> key="/" > 
  at <BaseTransition mode="out-in" appear=false persisted=false  ... > 
  at <Transition name="page" mode="out-in" > 
  at <NuxtLayout key=0 name=undefined > 
  at <RouterView> 
  at <NuxtPage> 
  at <App> 
  at <Root> 

Hydration completed but contains mismatches.
```

danielroe:
### 🔗 Linked issue

resolves #1907, https://github.com/nuxt/framework/issues/3385

### ❓ Type of change

### 📚 Description

This PR adds support for using teleports in Nuxt 3, including an example based on https://vuejs.org/guide/built-ins/teleport.html.

### 📝 Checklist


pi0:
This is a nice initiative, however obviously not best solution using cheerio nor vue3 seems have a proper solution for teleports right now.

It would be nice if we could have a limited parser (maybe #id only selectors) to do a quick replacement? for POC.

Before that, it would be nice having a section in docs for Teleports to mention usage limitations and <ClientOnly> workaround.

Ideally by moving this feature to nitropack, we could make a nitro plugin, and start support teleport behind a flag to avoid overhead when not needed.

danielroe:
Agreed.

Maybe, priority list of:

1. docs with workaround for client-only teleport support
2. implementation with parser supporting `<body>` (likely the most common) and `#id` only
3. (possibly not needed) flag for full cheerio support (or a more advanced custom parser)

For anyone else's reference, teleport SSR support is up to frameworks to implement: see https://github.com/vuejs/core/issues/3869#issuecomment-852115768.

toniengelhardt:
Small side-node regarding the client-only solution: it would be amazing if the `<Teleport>` tag would be implicitly client-only and would not require wrapping into a `<ClientOnly>` tag.

danielroe:
I've adapted this PR for an ultra-minimal support for teleports, using the existing template. I think that this is a reasonable approach to include out-of-the-box, with more advanced use cases to be developed under a flag later on. (Given that teleports are meant to be outside the Vue app, the body target should cover most use cases, with the client-only approach being an escape hatch only...)

docs/content/3.docs/1.usage/10.teleports.md
```diff
@@ -0,0 +1,37 @@
...
+  </Teleport>
+</template>
+```
+
+## Example: client-side teleport
+
+```vue
+<template>
+  <ClientOnly>
+    <Teleport to="#outside-vue">
```
pi0, line 30: Shall we introduce `NuxtTeleport` with the ability to automatically fallback into csr?

packages/nuxt3/src/core/templates.ts
```diff
@@ -80,7 +80,7 @@ export const appViewTemplate = {
   {{ HEAD }}
 </head>
 
-<body {{ BODY_ATTRS }}>
+<body {{ BODY_ATTRS }}>{{ BODY_TELEPORTS }}
```
pi0, line 83: Nice!

docs/content/3.docs/1.usage/10.teleports.md
pi0: Could also have `<NuxtTeleport ssr>` or `<NuxtTeleport body>` :)

danielroe: A quick check seems that the vue compiler optimises teleports differently, meaning we can't just wrap `Teleport` in our own component.

For example, this is how `<Teleport>` renders an empty slot:
```html
<!----><!---->
```

This is how it gets rendered if we use `<NuxtTeleport>`:

```html
<!--[--><!--]--><!---->
```

pi0: Thanks for checking! Well if that's the case, it means we have a serious limitation with vue teleports if want to create a reusable teleport wrapper. Do you think of it as a serious downside or that we could experiment with it somehow to see limitations?

danielroe: I think it's a serious blocker to an all-purpose wrapper. I would suggest we can use `<NuxtTeleport>` only as a shortcut for a client-only teleport. Or maybe call it `<ClientTeleport>` for clarity?

packages/nuxt3/src/core/templates.ts
pi0: Could be also `BODY_PREPEND`

docs/content/3.docs/1.usage/10.teleports.md
pi0: Having docs for <ClientOnly> if fine then!

danielroe: do you mean it's fine as-is, or that we should remove `<NuxtTeleport>` and just add a note to the docs?

pi0: I like the idea of having <NuxtTeleport> also for server since at least can ensure target is body by default and warn or fallback to client-side if is otherwise. Just not sure what is main blocker or needed to be tracked in upstream. Was thinking we could add component in a later step and not block PR neither make an early decision like <ClientTeleport> introduction.

danielroe: I'm happy with that. I'll amend this PR to remove the component, and we can see about creating an issue on vuejs/core.

danielroe: Raised https://github.com/vuejs/core/issues/5653.

PR #3843, opened 2022-03-22
docs: added data fetching migration

danpastori:
### 🔗 Linked issue
#3769 

### ❓ Type of change

- [ x ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This is a start to all of the different options provided by Nuxt 3 for data fetching. I wrote this up to where I was with the migration of our app. I know there is a ton more that we will have to touch on.

@danielroe : The big question I came across when migrating our app from Nuxt 2 to Nuxt 3 was how to design reusable functions that call an API through fetch. For example, if I were to make an API endpoint that accepted a `search` term or a `pagination` term, what would be the best way to abstract that? Interface with the provided `$fetch` object? If you have a good example of this, I'll submit a PR to the docs as well. I noticed a few questions in the discussion forum asking this as well.

I'll keep updating this as I migrate pieces of our app as well. If there's any changes, let me know!

### 📝 Checklist

- [ x ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #3914, opened 2022-03-25
docs: update data fetching for `$fetch` and refresh 

danpastori:
### 🔗 Linked issue
Proposed docs update

### ❓ Type of change

### 📚 Description
While working on migrating our app, I found that directly calling the API wasn't specifically mentioned. I could have missed it or missed a point early on, but I figured that adding this to the official docs would be extremely helpful.

@danielroe added a great example on the branch for migrating data from Nuxt 2 to Nuxt 3 that I added to the PR for the official docs. This way it's easily found for those coming from a "new Nuxt 3 install" perspective and not a migration.
https://github.com/nuxt/framework/blob/docs/migration-guide/docs/content/3.docs/1.migration/8.data-fetching.md
https://github.com/nuxt/framework/pull/3843

I also used the example found in the discussion: https://github.com/nuxt/framework/discussions/2445 to show two different ways to show to the developer.

Hopefully this will help others in the same boat! If there's anything you wish to add, I'd definitely love to learn myself!

### 📝 Checklist


docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -295,3 +295,77 @@ export default defineComponent({
   </header>
 </template>
 ```
+## Changing URL Parameters with `asyncData`
```
pi0, line 298: We could make this section more general as "Refreshing Data". With the explanation of `refrash` callback and `watch` options.

As for the example, In order to properly trigger refrash on url and query changes, `useFetch` can be used:

```js
const { data } = await useFetch(() => `users?page=${page.value}&take=6`, { baseURL: config.API_BASE_URL }
);
```

We might add it as an example to `useFetch` section.

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -295,3 +295,77 @@ export default defineComponent({
...
+function next() {
+  page.value++;
+  refresh();
+}
+</script>
+```
+The key to making this work is to call the `refresh()` method returned from the `useAsyncData()` composable when a query parameter has changed.
+
+## Directly calling an API Endpoint
+There are instances where you may need to directly call the API. Nuxt 3 provides a globally available $fetch method that has the same API as the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch).
```
pi0, line 327:
```suggestion
There are instances where you may need to directly call the API. Next 3 provides globally available `$fetch` method using [unjs/ohmyfetch](https://github.com/unjs/ohmyfetch) as well as `fetch`
with the same API as the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch).
```

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -295,3 +295,77 @@ export default defineComponent({
...
+## Changing URL Parameters with `asyncData`
+Sometimes throughout the course of your user's page visit, you may need to refresh the data loaded from the API. This can happen if the user chooses to paginate, filter results, search, etc. 
+
+You can make use of the `refresh()` method returned from the `useAsyncData()` composable to refresh the data with different query parameters:
+
+```vue
+<script setup>
+const page = ref(1);
+
+const { data:users, loading, refresh, error } = await useAsyncData( 'users', () =>
```
pi0, line 307: As for the example, In order to properly trigger refrash on url and query changes, `useFetch` can be used:

```js
const { data } = await useFetch(() => `users?page=${page.value}&take=6`, { baseURL: config.API_BASE_URL }
);
```

We might add it as an example to `useFetch` section.

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -295,3 +295,77 @@ export default defineComponent({
...
+  refresh();
+}
+</script>
+```
+The key to making this work is to call the `refresh()` method returned from the `useAsyncData()` composable when a query parameter has changed.
+
+## Directly calling an API Endpoint
+There are instances where you may need to directly call the API. Nuxt 3 provides a globally available $fetch method that has the same API as the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch).
+
+It has a number of benefits, including:
```
pi0, line 329:
```suggestion
Using `$fetch` has a number of benefits, including:
```

docs/content/3.docs/1.usage/1.data-fetching.md
danpastori: @pi0 I moved this up under the `useFetch()` section.  Do you have an example of watching parameters? If so, I can inspect and explain it.

docs/content/3.docs/1.usage/1.data-fetching.md
danpastori: I got this updated in thanks!

danpastori:
Thanks for the feedback. I see there's a PR about contributing to the docs: https://github.com/nuxt/framework/pull/3923 I'll make sure to thoroughly read it for any more updates I have!

Let me know if I need to change anything else.

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -168,6 +168,31 @@ watch(posts, (newPosts) => {
 </script>
 ```
 
+## Refreshing Data
```
pi0, line 171:
```suggestion
## Refreshing Data

```

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -295,3 +320,51 @@ export default defineComponent({
   </header>
 </template>
 ```
+
+## Directly calling an API Endpoint
```
pi0, line 324:
```suggestion
## Directly calling an API Endpoint

```

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -328,3 +355,52 @@ export default defineComponent({
...
+
+There are instances where you may need to directly call the API. Next 3 provides globally available `$fetch` method using [unjs/ohmyfetch](https://github.com/unjs/ohmyfetch) as well as `fetch`
+with the same API as the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch).
```
danielroe, line 362:
```suggestion
There are instances where you may need to directly call the API. Nuxt 3 provides a globally available `$fetch` method using [unjs/ohmyfetch](https://github.com/unjs/ohmyfetch) (in addition to `fetch`)
with the same API as the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch).
```

docs/content/3.docs/1.usage/1.data-fetching.md
```diff
@@ -328,3 +355,52 @@ export default defineComponent({
...
+    page.value++;
+    loadUsers(page.value);
+  }
+
+  function previous () {
+    page.value--;
+    loadUsers(page.value);
+  }
+
+  loadUsers(page.value);
```
danielroe, line 402: As there is only one instance of users I would suggest instead something like.

```suggestion
  const page = ref(1);
  const users = ref([]);

  watch(page, async (page) => {
    users.value = await $fetch(
      '/api/v1/users',
      {
        method: 'GET',
        params: {
          page,
          take: 6
        }
      }
    )
  }, { immediate: true })

  function next () {
    page.value++;
  }

  function previous () {
    page.value--;
  }
```

danielroe: But I'd also point out that this isn't yet a reusable composable. Maybe it would be worth giving an example of that?

pi0: In regards of example above as daniel noted, it is not a reusable composable and directly using `useFetch` is better since it can automatically do the watching as well. 

> Also, In Nuxt 2, I made a plugin that wrapped our API endpoints, this.$api.{resource}.show(). This way I had a single source of truth whenever the API interface changed (i.e. /v1 to /v2). Would it be helpful to give an example about wrapping an entire API Resource (GET, POST, PATCH, DELETE endpoints) as a composable? 

This is a really nice pattern as well. I would like to have it in examples/ section at least! For this PR, i would suggest to keep changes minimal and mergable and iterate over with next ideas.

PR #3921, opened 2022-03-27
feat(nuxi): support mode flags for `add` command

Diizzayy:

### 📚 Description

Allow `--client|--server` flags to be used with `nuxi add` to specify that the respective entity should be either client / server only.

This PR also handles the case where a user manually specifies a file extension when using `nuxi add`, With current implementation `nuxi add plugin test.ts` would result in `/plugins/test.ts.ts` being generated.

*Side note*: Nuxt 3 encourages typescript usage for optimal developer experience, hence though different entities allow for varying extensions, `nuxi add` defaults to (and enforces) `.vue` and `.ts`.

### 📝 Checklist


packages/nuxi/src/commands/add.ts
```diff
@@ -33,8 +34,21 @@ export default defineNuxtCommand({
...
+    }
+
+    if (modeProvided) {
+      name = name.replace(modeProvided, '')
+      if (!mode) { mode = modeProvided.replace(/\./, '') }
+    }
+
     // Resolve template
-    const res = templates[template]({ name })
+    const res = templates[template]({ name, mode })
```
pi0, line 51: Templates that support `.client` / `.server` convention can themselves apply file or template changes.

packages/nuxi/src/utils/templates.ts
```diff
@@ -1,7 +1,7 @@
 import { upperFirst } from 'scule'
 
 interface Template {
-  (options: { name: string }): { path: string, contents: string }
+  (options: { name: string, mode?: string }): { path: string, contents: string }
```
pi0, line 4:
```suggestion
  (options: { name: string, mode?: 'client' | 'server' }): { path: string, contents: string }
```

packages/nuxi/src/utils/templates.ts
```diff
@@ -15,27 +15,34 @@ export default defineHandle((req, res) => {
 `
 })
 
-const plugin: Template = ({ name }) => ({
-  path: `plugins/${name}.ts`,
-  contents: `
+const plugin: Template = (opts) => {
+  opts.name = opts.mode ? `${opts.name}.${opts.mode}` : opts.name
```
pi0, line 19: Code style: instead of mutating opts, we can use 

```
path: opts.mode ? `${opts.name}.${opts.mode}` : opts.name
```

packages/nuxi/src/commands/add.ts
```diff
@@ -15,7 +15,8 @@ export default defineNuxtCommand({
     const cwd = resolve(args.cwd || '.')
 
     const template = args._[0]
-    const name = args._[1]
+    let name = args._[1]
+    let mode = (args.client && 'client') || (args.server && 'server')
```
pi0, line 19: For now, i would add a validator to ensure mode is only allowed with `['plugin']` and otherwise throw an error of bad command usage.

pi0: Why error instead of warn: We do need to ensure CLI behavior is solid and predictable. In the future when introducing programmatic API, and if an old version of CLI is picked not support client/server for an entity, it needs to properly fail and indicate operation didn't went as expected.

packages/nuxi/src/utils/templates.ts
```diff
@@ -15,27 +15,34 @@ export default defineHandle((req, res) => {
...
 export default defineNuxtPlugin((nuxtApp) => {})
   `
-})
+  }
+}
 
-const component: Template = ({ name }) => ({
-  path: `components/${name}.vue`,
-  contents: `
+const component: Template = (opts) => {
```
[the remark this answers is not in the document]
Diizzayy: Agreed!

packages/nuxi/src/commands/add.ts
```diff
@@ -29,12 +30,21 @@ export default defineNuxtCommand({
...
+      process.exit(1)
+    }
+
     // Load config in order to respect srcDir
     const kit = await loadKit(cwd)
     const config = await kit.loadNuxtConfig({ cwd })
 
+    const extRegex = /\.[a-z]+/g
+    const extProvided = name.match(extRegex)
+    if (extProvided) { name = name.replaceAll(extRegex, '') }
```
pi0, line 44: This feature causes over-complexity IMO and besides adding strange behavior and edge cases. Not all extensions are to be stripped from the file name. Even client/server when filesystem structure is not supporting it.

Diizzayy: Understood, let's omit both that feature and inferring the mode from the file name

Diizzayy: @pio Just account for following extensions ( `\.(js|mjs|ts|jsx|tsx|vue)+$` ) should be suffice to prevent producing a double `.ts.ts`

pi0: Sure but even if we omit it, it can cause a strange behavior to consider using `add component Foo.ts`, removing the extension, and forcing to be `Foo.vue` unless we ensure to reapply it. For sure possible but too much complexity is added to code. We can plan to improve this behavior later with extension inference.

Diizzayy: Agreed, We can introduce better handling of extensions at a later point. With current implementation `add component Foo.ts` results in `Foo.ts.vue`, so we'll make sure to revisit this

docs/content/3.docs/1.usage/8.cli.md
```diff
@@ -127,7 +129,7 @@ The `add` command generates new elements:
 * **component**: `npx nuxi add component TheHeader`
 * **composable**: `npx nuxi add composable foo`
 * **layout**: `npx nuxi add layout custom`
-* **plugin**: `npx nuxi add plugin analytics`
+* **plugin**: `npx nuxi add plugin analytics` or `npx nuxi add plugin firebase --client` (generates `/plugins/firebase.client.ts`)
```
Atinux, line 132:
```suggestion
* **plugin**: `npx nuxi add plugin analytics` or `npx nuxi add plugin sockets --client` (generates `/plugins/sockets.client.ts`)
```

Atinux: Avoiding to use any brand here :)

pi0 (approved):
Thanks! I've made a small refactor that enables modifiers for components, API, etc per template.

PR #3929, opened 2022-03-28
feat: `refreshNuxtData` function and `app:data:refresh` hook

antfu:
### ❓ Type of change

### 📚 Description

See documentation on https://github.com/nuxt/framework/blob/feat/app-refresh/docs/content/3.docs/1.usage/1.data-fetching.md#usedatarefresh

### 📝 Checklist


Atinux:
Nice work @antfu 

I would personally rename it to `useDataRefresh()` and `app:data:refresh` hook.

Also, it should also work for `useFetch` since it uses `useAsyncData` under the hood (documentation wise)

Atinux (approved):
Good for me 👍 
Waiting for Daniel and Pooya to review 😊

danielroe (commented):
I love it!

A couple of ideas/questions for thought:

1. An API suggestion - do we really need to pass the keys at the `useDataRefresh` stage? Might it be worth (also?) supporting passing them when used?

   ```js
   const refresh = useDataRefresh()
   // refresh everything
   refresh()
   // refresh just one thing
   refresh('count')
   ```

1. Is there any merit in modifying the default `refresh` function returned from `useAsyncData`? Perhaps with a flag it could trigger a full refresh?

1. On client side, which is probably when we'll be using this, we could just have a globally available `refreshData` helper that calls `useNuxtApp()` when used, and could be directly usable within functions rather than requiring a composable to be called in setup - much like `navigateTo` etc...

Atinux:
We could imagine something simpler indeed like `useNuxtApp().refreshData()` (actually I am using `useNuxtApp().callHook('app:data:refresh')`

Adding the keys makes it a bit tricky since we cannot guess the keys when using `useFetch` and the main purpose here is to refresh the whole app data.

I believe that we may not need a composable for something that global.

antfu:
We could actually have `refreshData` globally as 

```ts
refreshData()
refreshData('count')
```

I was just not sure about:

- The auto-import conversion, ideally we don't want auto-import arbitrary functions by default but with some conventions like starting with `use` or `on`. Or do you have a better name in mind?
- The auto-import currently won't work in the template directly, which means users would need to do that alias themselves in order to use it

```html
<script setup>
const refresh = refreshData
<script>

<template>
  <button @click="refresh">Refresh</button>
<template>
```

Having a hanging variable looks a bit weird. So I was thinking, since we need to do the alias anyway maybe a function would be more clear and more explicit. WDYT?

packages/nuxt3/src/app/composables/asyncData.ts
```diff
@@ -166,6 +169,14 @@ export function useLazyAsyncData<
   return useAsyncData(key, handler, { ...options, lazy: true })
 }
 
+export function useDataRefresh (keys?: string | string[]) {
```
pi0, line 172: Instead of a deferred function factory, I would call it `refreshData({ keys })`. It is not returning a reference but simply doing an action also we might want to extend filtering possibilities in the future, therefore, using options object.

BTW I understand the benefit of a "data refresher" pattern but users can actually really easily create such a wrapper by themselves while having a basic building-block for it.

pi0: Perf: We might early return in this function for `process.server`

pi0:
> Adding the keys makes it a bit tricky since we cannot guess the keys when using useFetch (and potentially useAsyncData is we make the key optional too). The main purpose here is to refresh the whole app data.

In future iterations, we might directly return the used key from the result of `useAsyncData` / `useFetch` to solve this limitation :) 

```js
const { data, key } = useFetch(...)

const refrash = () => refreshData(key)
```

Atinux:
I think exposing `refreshData` from `useNuxtApp` will make more sense since it is related to it directly:

```vue
<script setup>
const { refreshData } = useNuxtApp()
</script>

<template>
  <button @click="refreshData">Refresh</button>
<template>
```

At the end it is simply an alias of `useNuxtApp().callHook('app:data:refresh')`

pi0:
> I think exposing refreshData from useNuxtApp will make more sense since it is related to it directly:

We don't need to add overhead to the base of `nuxtApp` for hook shortcuts. As more hooks and their shortcuts adding, it adds uncessarility to the bundle size. 

`refreshData`, similar to other utils, can implicitly access nuxtApp using `useNuxtApp` internally :)

pi0:
Another note: It probably needs Nuxt prefix to be `refrashNuxtData` and not conflicting with user functions 😅 At least also making usage simpler without template auto import support:

```js
```vue
<script setup>
const refreshData = () => refrashNuxtData()
</script>

<template>
  <button @click="refreshData">Refresh</button>
<template>
```
```

**Comment by Atinux** [comment]
Let's go for `refreshNuxtData()` 👍

File: examples/use-async-data/app.vue [hunk 2]
```diff
@@ -1,19 +1,31 @@
 <script setup>
 const ctr = ref(0)
-const { data, refresh, pending } = await useAsyncData('/api/hello', () => $fetch(`/api/hello/${ctr.value}`), { watch: [ctr] })
+const showMountain = ref(false)
+const refresh = refreshNuxtData()
+const { data, pending } = await useAsyncData('/api/hello', () => $fetch(`/api/hello/${ctr.value}`), { watch: [ctr] })
```
pi0, line 5: Okay, new refreshNuxtData is nice but we still have a simple `refresh` exposed to use and demonstrate here :D

pi0: I see the purpose of testing but we are unnecessarily making examples complex. They are not for testing. I would suggest to revert this change to make PR mergeable and later on add browser tests + a better example that has global refrash button somewhere outside of component.

pi0: (it can be it's own data-refrash example too)

examples/use-async-data/app.vue
```diff
@@ -1,19 +1,30 @@
...
-        Refresh Data
-      </NButton>
+    <div class="flex justify-center gap-2">
       <NButton :disabled="pending" @click="ctr++">
         +
       </NButton>
+      <NButton @click="showMountain = !showMountain">
+        {{ showMountain ? 'Hide' : 'Show' }} Mountain
+      </NButton>
+      <NButton :disabled="pending" @click="refreshAll">
```
pi0, line 23: pending state is for counter example...

examples/use-async-data/app.vue
pi0: Updated example to seperate two mountain / counter examples by component and a global refrash in main app.vue: https://github.com/nuxt/framework/pull/3929/commits/d9f58d85a933bed1f129c70cc05544abed42d6eb

pi0: This, however, revealed a new issue we have. the pending state of `useFetch` is not updated when global refreeshNuxtData invoked.

PR #3930, opened 2022-03-28
docs: init testing section

antfu:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/3.docs/4.advanced/5.testing.md
+++ b/docs/content/3.docs/4.advanced/5.testing.md
@@ -1,0 +1,61 @@
+# Testing
+
+::alert{icon=👉}
+Test utils are still in development and the API and behavior may change. Currently it's for module authors to preview but not yet ready for testing production apps.
+::
+
+In Nuxt 3, we have a rewritten version of `@nuxt/test-utils` available as `@nuxt/test-utils-edge`. We support [Vitest](https://github.com/vitest-dev/vitest) and [Jest](https://jestjs.io/) as the test runner.
+
+## Installation
+
+```bash
+yarn add --dev @nuxt/test-utils-edge vitest
+```
+
+## Setup
+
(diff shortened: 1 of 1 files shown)


docs/content/3.docs/4.advanced/5.testing.md
```diff
@@ -0,0 +1,61 @@
...
+})
+```
+
+Behind the scenes, `setup` performs a number of tasks in `beforeAll`, `beforeEach`, `afterEach` and `afterAll` to setup the Nuxt test environment correctly.
+
+### Setup Options
+
+> TODO:
+
+## Testing an app
```
pi0, line 40: The goal of this (advanced) section is testing nuxt modules. We might keep intro simple and instead focus on documenting possible setup options + exposed utilities

antfu:
Guess we could keep it simple for now, or do you think what is missing currently?

pi0:
Thank! Looks good to me! Only a few items could make it perfect to start with:

- In first example, we can advice to use `server: true` or enable it by default 
- A list of utility exports from `@nuxt/test-utils` and a short description for them
- A list of available option names and short description for them

antfu:
Oh haha, I misread your last comment and thought you are saying not listing them 🤣  I will do it!

packages/test-utils/src/context.ts
```diff
@@ -12,8 +12,8 @@ export function createTestContext (options: Partial<TestOptions>): TestContext {
     setupTimeout: 60000,
     dev: !!JSON.parse(process.env.NUXT_TEST_DEV || 'false'),
     logLevel: 1,
-    server: options.browser,
-    build: options.browser || options.server,
+    server: true,
+    build: (options.browser !== false) && (options.server !== false),
```
pi0, line 16: BTW why this change is needed? Simply when either of server (default) or browser options are enabled, we need to build (by default)

```suggestion
    build: (options.browser !== false) || (options.server !== false),
```

PR #3934, opened 2022-03-28
fix(vite-node): bugfixes for production build

antfu:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/entry.async.ts
+++ b/packages/nuxt3/src/app/entry.async.ts
@@ -2,7 +2,7 @@ import { CreateOptions } from '#app'
 
 const entry = process.server
   ? (ctx?: CreateOptions['ssrContext']) => import('#app/entry').then(m => m.default(ctx))
-  : () => import('#app/entry').then(m => m.default)
+  : () => import('#app/entry').then(m => m.default())
 
 if (process.client) {
   entry()
--- a/packages/nuxt3/src/app/entry.async.ts
+++ b/packages/nuxt3/src/app/entry.async.ts
@@ -2,7 +2,7 @@ import { CreateOptions } from '#app'
 
 const entry = process.server
   ? (ctx?: CreateOptions['ssrContext']) => import('#app/entry').then(m => m.default(ctx))
-  : () => import('#app/entry').then(m => m.default())
+  : () => import('#app/entry').then(m => m.default)
 
 if (process.client) {
   entry()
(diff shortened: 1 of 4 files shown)


packages/nuxt3/src/app/nuxt.ts
```diff
@@ -116,6 +116,9 @@ export function createNuxtApp (options: CreateOptions) {
 
   // Expose runtime config
   if (process.server) {
+    options.ssrContext.runtimeConfig = options.ssrContext.runtimeConfig || {}
```
pi0, line 119: Can we directly populate ssrContext with vite-node and avoid this workaround? This is unclear we are making a fix for vite-node only

antfu: I just wasn't sure why we have this error. Do you have a reference where is supposed to be set so I can debug a bit.

packages/nuxt3/src/app/entry.async.ts
```diff
@@ -2,7 +2,7 @@ import { CreateOptions } from '#app'
 
 const entry = process.server
   ? (ctx?: CreateOptions['ssrContext']) => import('#app/entry').then(m => m.default(ctx))
-  : () => import('#app/entry').then(m => m.default)
+  : () => import('#app/entry').then(m => m.default())
```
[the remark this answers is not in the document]
antfu: Oh, I thought it was a typo (and somehow it did block vite-node), as the format does not match with L4.

packages/nuxt3/src/app/nuxt.ts
pi0: This should be normally filled at runtime by nitro: https://github.com/nuxt/framework/blob/9ebea904f4f3a7bd63b2a8ce5ecc5e592abf80c0/packages/nitro/src/runtime/app/render.ts#L90

packages/nuxt3/src/app/entry.async.ts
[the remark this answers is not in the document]
pi0: Yeah, little bit confusing with async entry. With client, we directly evaluate while with server, we let renderer do evaluate

packages/vite/src/runtime/server.mjs
```diff
@@ -44,11 +44,11 @@ async function writeManifest (extraEntries) {
   await fs.writeFile(join(dir, 'client.manifest.mjs'), 'export default ' + JSON.stringify(clientManifest, null, 2), 'utf8')
 }
 
-export default (async () => {
+export default async (ssrContext) => {
   const { default: render } = await runner.executeFile(entry)
-  const result = await render()
+  const result = await render(ssrContext)
```
antfu, line 49: Now I understand - the root cause is this one, while I was debugging from the bottom up 😄

PR #3939, opened 2022-03-28
feat(nuxt3): extends support for `app/router.options`

kevinmarrec:
### ❓ Type of change

### 📚 Description

Following up #3222 (`extends` support progress) this PR allows extending `app/` directory (`app/router.options`) for extended config layers.

- [x] Extends support for `app` directory
- [x] Tests

### ℹ️ Override behavior

It uses `defu` to merge router options if `app/router.options` exists for multiple layers


kevinmarrec:
Need to deal with the `test-types` not passing :)

packages/nuxt3/src/pages/module.ts
```diff
@@ -83,15 +84,22 @@ export default defineNuxtModule({
...
       getContents: async () => {
         // Check for router options
-        const routerOptionsFile = await findPath('~/app/router.options')
-        const configRouterOptions = genObjectFromRawEntries(Object.entries(nuxt.options.router.options)
-          .map(([key, value]) => [key, genString(value as string)]))
+        const routerOptionsFiles = (await Promise.all(nuxt.options._layers.map(
+          async layer => await findPath(resolve(layer.config.srcDir, 'app/router.options'))
+        ))).filter(Boolean)
+
+        const routerOptions = defu({}, ...await Promise.all(routerOptionsFiles.map(async file => await import(file))))
```
pi0, line 91: The `app` runtime files are not to be directly imported in Node.js namespace which is basically why we use `app/` instead of `nuxt.config`.

packages/nuxt3/src/pages/module.ts
```diff
@@ -83,15 +84,22 @@ export default defineNuxtModule({
...
+
+        const genObject = (object: object) => genObjectFromRawEntries(
+          Object.entries(object).map(([key, value]) => [key, genString(value as string)])
+        )
+
         return [
-          routerOptionsFile ? genImport(routerOptionsFile, 'routerOptions') : '',
```
pi0, line 90: We can instead, generate multiple import statements in the generated file and merge them there.

(Side note: in general, not all files `app/` directory can be merged and they are supposed to **override** templates. we are making an exception by doing this kind of merge)

packages/nuxt3/src/pages/module.ts
kevinmarrec: Got it !

packages/nuxt3/src/pages/module.ts
kevinmarrec: Gonna refactor the code so the imports & the merges are done in the generated file (as like before but we multiple imports if it needs so)

packages/nuxt3/src/pages/module.ts
```diff
@@ -83,15 +83,20 @@ export default defineNuxtModule({
...
+          ...routerOptionsFiles.map((file, index) => genImport(file, `routerOptions${index}`)),
           `const configRouterOptions = ${configRouterOptions}`,
           'export default {',
           '...configRouterOptions,',
-          routerOptionsFile ? '...routerOptions' : '',
+          // We need to reverse spreading order to respect layers priority
+          ...routerOptionsFiles.map((_, index) => `...routerOptions${index},`).reverse(),
```
kevinmarrec, line 99: @pi0 Do we care about trailing `,` for last item ?

pi0: It is totally fine but you can use `.join(',')` to avoid trailing comma

kevinmarrec: We can't use `join` cause it would result in losing `\n` join later on, one solution would be to `,\n` join what's inside the `export default {}`

pi0 (approved):
LGTM. Thanks 💚

Issue #3598, opened 2022-03-11
Vite-node support progress


pi0:
Using vite-node integration, we could lazily load and evaluate server chunks just like clients without bundling the whole app each time. Since this is a rather big change, we are progressively rolling it out.

Opting in: Use `experimental.viteNode: true` in `nuxt.config` or `EXPERIMENTAL_VITE_NODE=1` environment variable.

```js
// nuxt.config
experimental: { viteNode: true }
```

- [x] Initial support (https://github.com/nuxt/framework/pull/2795)
- [x] Use nuxt server middleware
- [x] Use envrionemnt variables to provide config instead of template
- [x] Generate files outside of runtime (endpoint to get manifest)

Atinux:
What are the current blockers to have progress on this?

pi0:
Nothing is blocker on nuxt side I suppose. We only need to iterate to move the feature out of the experimental state with more stability.

pi0:
The initial checklist is done. However we need still testing to move it out of the experimental flag.

PR #3967, opened 2022-03-30
fix(vite-node): improve server.mjs

antfu:
### 🔗 Linked issue

#3598

- Use process.env to pass server URLs instead of file replacement. 
- Improved manifest generation. 
- Fix root options for vite node runner

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite-node.ts
+++ b/packages/vite/src/vite-node.ts
@@ -62,18 +62,18 @@ export async function prepareDevServerEntry (ctx: ViteBuildContext) {
     entryPath = resolve(ctx.nuxt.options.appDir, 'entry.async')
   }
 
-  const raw = await fse.readFile(resolve(distDir, 'runtime/server.mjs'), 'utf-8')
   const host = ctx.nuxt.options.server.host || 'localhost'
   const port = ctx.nuxt.options.server.port || '3000'
   const protocol = ctx.nuxt.options.server.https ? 'https' : 'http'
-  const code = raw
-    .replace('__NUXT_SERVER_FETCH_URL__', `${protocol}://${host}:${port}/__nuxt_vite_node__/`)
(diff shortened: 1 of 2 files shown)


packages/vite/src/vite-node.ts
```diff
@@ -62,18 +62,18 @@ export async function prepareDevServerEntry (ctx: ViteBuildContext) {
...
-    resolve(ctx.nuxt.options.buildDir, 'dist/server/server.mjs'),
-    code,
-    'utf-8'
+
+  process.env.NUXT_VITE_SERVER_FETCH = `${protocol}://${host}:${port}/__nuxt_vite_node__/`
+  process.env.NUXT_VITE_SERVER_ENTRY = entryPath
+  process.env.NUXT_VITE_SERVER_BASE = ctx.ssrServer.config.base || '/_nuxt/'
+  process.env.NUXT_VITE_SERVER_ROOT = ctx.nuxt.options.rootDir
+
+  await fse.copyFile(
```
pi0, line 74: Since we are not depending on template anymore, can we avoid copy and directly use `runtime/server.mjs` as entrypoint?

antfu: I'd love to, but how could we do that?

pi0: Hmm actually we can update via nitro options but seems little hacky with legacy support of webpack.

What about generating something minimal like this inside entry:

```js
export { default } from `path/runtime/server.mjs`
```

Suggestion: would be nice to also rename `server.mjs` to `vite-node.mjs` for clarity.

antfu: Oh since it will write `client.manifest.mjs`, maybe it's better to do that after #3969? Otherwise it writes the file to the runtime folder

pi0 (approved):

PR #3968, opened 2022-03-30
fix(vite-node): on-demand manifest generation

antfu:
### 🔗 Linked issue

On top of #3967 

It's not working correctly yet, because of two blockers:

#### 1 - On-demand manifest loading timing
Since with Vite Node, everything is on-demand now. Thus we don't know the module graph until we actually trigger the rendering. While `vue-bundler-renderer` seems to assume the format has been bundled based and requests the manifest and server bundle at the same timing
https://github.com/nuxt/framework/blob/ac40c9746cc516ab117e9edca1c8260a7fbd83dd/packages/nitro/src/runtime/app/render.ts#L18-L24
It causes the FOUC on initial load (and missing entries for newly discovered CSS modules). I see two ways of fixing it:

- A: Make the `clientManifest` option of `createRenderer` possibility to be a function in `vue-bundler-renderer` so we could evaluate after we request the rendering
- B: Allow the server bundle to return a manifest object along with the rendered app
https://github.com/nuxt/framework/blob/949def61791a32ef7d96c0f8702ef1a8a5fa5383/packages/vite/src/runtime/server.mjs#L23-L24

#### 2 - Manifest Cache

Since this change, the manifest is now on-demand and does not change during the development. Causing it reuses the cache and does not re-fetch:

https://github.com/nuxt/framework/blob/ac40c9746cc516ab117e9edca1c8260a7fbd83dd/packages/nitro/src/runtime/app/render.ts#L14

Do you know if there is a way to force cleaning the cache, or did I missed anything?


antfu:
FOUC fixed by #4106

pi0:
Quickly looking it is an amazing improvement!! Two changes would like to add before merging:

- Serialize vite-node options object as one env and access with destr 
- Use get method to pass id

pi0:
On latest commit, using `yarn example vite-node` i get `_fetch is not a function ()` error on second SSR requests:

image

PR #3974, opened 2022-03-30
feat(vite): upgrade to vite 2.9.0

danielroe:
### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

This PR upgrades to vite 2.9.0, which has a raft of new features and improvements.

~~Notably it adds [non-blocking pre bundling of dependencies](https://github.com/vitejs/vite/pull/6758), which I think means we can drop our own pre-warming fo the server.~~

~~**Note**: there is currently an issue with our `vite-node` integration. I suspect @antfu is already on it.~~


pi0:
> Notably it adds https://github.com/vitejs/vite/pull/6758, which I think means we can drop our own pre-warming fo the server.

Are you sure it also triggers build warmup for entry or just doing same prebundling (of client) in background?

pi0:
Noticed this issue on `main` since we are already using `2.9`:

image

antfu:
The error is not related to Vite 2.9, but rather a misalignment that only happens on stub mode (build mode works fine). I'll investigate the issue in another PR but I think it should block the implements of Vite 2.9.

> warm up

The optimization in 2.9 becomes lazier and non-blocking, while I guess our warm-up could still benefit the initial loading time. I think we could keep it.

antfu (approved):

danielroe:
Thanks @antfu! I'll restore the warm up.

antfu (approved):

PR #3983, opened 2022-03-31
fix(vite-node): wrong entry been resolved in stub mode

antfu:
### 🔗 Linked issue

Fixes https://github.com/nuxt/framework/pull/3974#issuecomment-1084166442

`process.server` was `undefined` so it resolved to the client entry and caused loaded app is an function instead of a server-rendered app. 

This only happens on the stub mode of development, but not the production build.
TBH I am not sure why it's happening and if this is the correct fix.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/vite/src/runtime/server.mjs
+++ b/packages/vite/src/runtime/server.mjs
@@ -50,6 +50,8 @@ async function writeManifest () {
 let render
 
 export default async (ssrContext) => {
+  // Workaround for stub mode
+  // https://github.com/nuxt/framework/pull/3983
   process.server = true
   render = render || (await runner.executeFile(entry)).default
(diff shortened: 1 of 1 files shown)


packages/vite/src/runtime/server.mjs
```diff
@@ -50,6 +50,7 @@ async function writeManifest () {
 let render
 
 export default async (ssrContext) => {
+  process.server = true
```
pi0, line 53:
```suggestion
  // Workaround for stub mode
  // https://github.com/nuxt/framework/pull/3983
  process.server = true
```

pi0 (approved):

pi0:
Thanks!

PR #3985, opened 2022-03-31
feat(nuxt)!: use paylod cache for initial data fetching by default

pi0:
### ❓ Type of change

### 📚 Description

Both `useFetch` and `useAsyncData` avoid fetching twice at the same time when a promise is running but skip any value in payload cache for **initial refrash**. A simple way to reproduce behavior:

```vue
<script setup lang="ts">
await useFetch('http://icanhazip.com', { key: 'test' })
await useFetch('http://icanhazip.com', { key: 'test', server: false })
</script>
```

The script above, causes another client-side fetching on initialization, despite the fact that we already have the same key in payload cache. There are other situations where this happens as well, including fetching in multiple components when a race-condition can happen for fast fetches.

This PR changes the default behavior to **use payload cache for initial fetch** when exists. Running the above example, will not cause another client-side request anymore. To avoid this new behavior, a new `{ initialCache: false }` options can be provide (* see notes).

Any subsequent call of `refrash` either automated with watchers or manually triggered by user, will still cause fetching and updating payload cache.

Slightly breaking changes:
- `refresh(_force)` option is dropped. It wasn't making sense to have it and it was limiting future compatibility with refresh options. It should be either `refresh()` or `refresh({})` (there are no public options yet)
- A code relying on the buggy behavior with double fetch might be broken. `{ initialCache: false }` can be used as a workaround.

---

(*) `useFetch`'s `cache` option can be either a [Request Cache strategy](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache) or boolean. Value will be mapped to boolean for internal `useAsyncData ` options. 

In the future, we can extend `cache` option with different new strategies. Ideally consistent with Request Cache API.

**Update**: Using explicit `initialCache` for time being to avoid possibly usage conflict in the future development of cache support. (https://github.com/nuxt/framework/pull/3985/commits/d6436a183fd4f521e662a3f56a2a3569aad6f63a)

### DIFF
--- a/packages/nuxt3/src/app/composables/asyncData.ts
+++ b/packages/nuxt3/src/app/composables/asyncData.ts
@@ -31,7 +31,7 @@ export interface RefreshOptions {
 export interface _AsyncData<DataT> {
   data: Ref<DataT>
   pending: Ref<boolean>
-  refrash: (opts?: RefreshOptions) => Promise<void>
+  refresh: (opts?: RefreshOptions) => Promise<void>
   error?: any
 }
 
@@ -87,7 +87,7 @@ export function useAsyncData<
     error: ref(nuxt.payload._errors[key] ?? null)
   } as AsyncData<DataT>
 
-  asyncData.refrash = (opts = {}) => {
+  asyncData.refresh = (opts = {}) => {
     // Avoid fetching same key more than once at a time
     if (nuxt._asyncDataPromises[key]) {
       return nuxt._asyncDataPromises[key]
@@ -125,7 +125,7 @@ export function useAsyncData<
     return nuxt._asyncDataPromises[key]
   }
 
-  const initialFetch = () => asyncData.refrash({ _initial: true })
+  const initialFetch = () => asyncData.refresh({ _initial: true })
 
   const fetchOnServer = options.server !== false && nuxt.payload.serverRendered
 
--- a/packages/nuxt3/src/app/composables/asyncData.ts
+++ b/packages/nuxt3/src/app/composables/asyncData.ts
@@ -153,7 +153,7 @@ export function useAsyncData<
     }
     const off = nuxt.hook('app:data:refresh', (keys) => {
       if (!keys || keys.includes(key)) {
-        return initialFetch()
+        return asyncData.refresh()
       }
     })
     if (instance) {
--- a/packages/nuxt3/src/app/composables/asyncData.ts
+++ b/packages/nuxt3/src/app/composables/asyncData.ts
@@ -24,14 +24,14 @@ export interface AsyncDataOptions<
   cache?: boolean
 }
 
-export interface RefrashOptions {
+export interface RefreshOptions {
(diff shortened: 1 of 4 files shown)


packages/nuxt3/src/app/composables/asyncData.ts
```diff
@@ -131,17 +143,17 @@ export function useAsyncData<
...
-      instance._nuxtOnBeforeMountCbs.push(asyncData.refresh)
+      instance._nuxtOnBeforeMountCbs.push(initialFetch)
     } else {
       // 4. Navigation (lazy: false) - or plugin usage: await fetch
-      asyncData.refresh()
+      initialFetch()
     }
     if (options.watch) {
-      watch(options.watch, () => asyncData.refresh())
+      watch(options.watch, () => initialFetch())
```
danielroe, line 152: Currently this prevents `watch` option from triggering updates.

This, for example, doesn't work as expected:

https://github.com/nuxt/framework/blob/1a7b570c82311e04812b1bfc913c455d9218d133/examples/composables/use-async-data/components/CounterExample.vue#L1-L5

was there a reason it shouldn't be treated as a subsequent refresh?

```suggestion
      watch(options.watch, () => asyncData.refresh())
```

packages/nuxt3/src/app/composables/asyncData.ts
```diff
@@ -131,17 +143,17 @@ export function useAsyncData<
...
     }
     const off = nuxt.hook('app:data:refresh', (keys) => {
       if (!keys || keys.includes(key)) {
-        return asyncData.refresh()
+        return initialFetch()
```
danielroe, line 156: shouldn't in this case also force refresh the data?

```suggestion
        return asyncData.refresh()
```

toooni:
@pi0 I am not sure if this is working properly. If I do have a page which fetches some data. The data is submitted from the server on the first request and the client by default does not make a "new" request to refresh the data. But as I see it the client also doesn't fetch new data if the same function is called again by f.e. a client route change. Is this intentional? How do I cache the first request (payload from the server) but avoid caching from there on when the same function is called on client side.

StefanCardnell:
@pi0 @danielroe Similar to @toooni. I am using `useAsyncData` to fetch some data from an external API. However, I am noticing that all of my API requests are cached. I am working on an app where users need to see the latest version of the data, never cached. If they leave a page and come back, and that page obtains its initial data via `useAsyncData`, they will get a cached version. Am I expected to place `{ initialCache: false}` on every API - which in the case of my app, is basically all of them (I imagine for a lot of people, also??)

pi0:
I understand your concern about global behavior. It is tricky because sometimes, we also do not expect asyncData result is being changed to keep it in memory and it makes double calls. Hard to judge what fits for everyone. But you are right.

I'm going to work on an improvement soon about shared state. We could make it better to respect cache status specially with `useFetch`.

Issue #3883, opened 2022-03-24
nuxt-bridge  ERROR  only on WINDOWS OS - [worker] Cannot find module


stephenjason89:
### Environment

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v14.17.0`
- Nuxt Version:     `2.16.0-27358576.777a4b7f`
- Package Manager:  `yarn@1.22.17`
- Builder:          `webpack`
- User Config:      `srcDir`, `ssr`, `dev`, `alias`, `image`, `generate`, `head`, `typescript`, `publicRuntimeConfig`, `privateRuntimeConfig`, `css`, `loading`, `loadingIndicator`, `plugins`, `components`, `buildModules`, `modules`,
`apollo`, `proxy`, `axios`, `vuetify`, `build`, `bridge`
- Runtime Modules:  `@nuxt/image@0.6.2`, `@nuxtjs/axios@5.13.6`, `@nuxtjs/apollo@4.0.1-rc.5`, `@nuxtjs/proxy@2.1.0`, `nuxt-social-meta@1.0.0`
- Build Modules:    `nuxt-storm@1.1.3`, `@nuxtjs/router@1.7.0`, `nuxt-compress@5.0.0`, `@nuxtjs/vuetify@1.12.3`, `@nuxt/bridge@3.0.0-27467056.93c4dfe`
------------------------------

### Reproduction

![image](https://user-images.githubusercontent.com/3810177/159923644-4113497d-3f0c-4af8-be93-8007655283a1.png)
![image](https://user-images.githubusercontent.com/3810177/159923719-478f8186-9829-4e39-a26f-6dd12f7b6579.png)
![image](https://user-images.githubusercontent.com/3810177/159923840-ba2e81aa-ac0a-4bff-84a1-a2b095bdcd9a.png)
![image](https://user-images.githubusercontent.com/3810177/159922855-798e3ed2-ff3f-4767-b51e-45ad0b031376.png)

### Describe the bug

I keep getting this error and it goes in an infinite loop of rebuilding the project.
This same project works without problem on my Macbook

### Additional context

The part where it says ERROR  Cannot start nuxt:  Plugin not found: C:\Users\stephenjasonwang\Development\foodat.ph\.nuxt\apollo-module.js

I can clearly see that the file is present in .nuxt\apollo-module.js

This is a working branch on my macbook running macos monterey

danielroe:
[...]

Issue #3147, opened 2022-02-09
`nuxi dev` infinite refresh loop


DamianGlowala:
### Environment

- Operating System: `Windows_NT`
- Node Version:     `v16.13.1`
- Nuxt Version:     `2.16.0-27358576.777a4b7f`
- Package Manager:  `npm@8.1.2`
- Bundler:          `Webpack`
- User Config:      `head`, `srcDir`, `css`, `plugins`, `components`, `buildModules`, `modules`, `moment`, `tailwindcss`, `pwa`, `i18n`, `axios`, `proxy`, `router`, `build`, `alias`, `server`, `bridge`
- Runtime Modules:  `@nuxtjs/axios@5.13.6`, `8n`, `@nuxtjs/pwa@3.3.5`, `@nuxtjs/proxy@2.1.0`, `cookie-universal-nuxt@2.1.5`
- Build Modules:    `@nuxtjs/stylelint-module@4.1.0`, `@nuxtjs/moment@1.6.1`, `@nuxt/postcss8@1.1.3`, `@pinia/nuxt@0.1.8`, `@nuxt/bridge@3.0.0-27406801.569aaeb`

### Reproduction

As much as I cannot publicly share the repository containing the project's source code, I'd be more than happy to provide access through e.g. TeamViewer to one of the Nuxt team members.

### Describe the bug

I have recently attempted to "experience Nuxt 3 features on existing Nuxt 2 projects" by following the instructions:
[https://v3.nuxtjs.org/getting-started/bridge/](https://v3.nuxtjs.org/getting-started/bridge/)

When running `npm run dev` (`nuxi dev`), the project builds successfully, but is stuck in an infinite refresh loop, as shown on the below video:
[https://drive.google.com/file/d/1Ce0Qv67EvyBbucwgFj5pmAy5o5Y73RHD/view?usp=sharing](https://drive.google.com/file/d/1Ce0Qv67EvyBbucwgFj5pmAy5o5Y73RHD/view?usp=sharing)

I tried adding `vite: false` to `nuxt.config.ts`, but with no luck. There are no other issues or resources on the Internet that I would find helpful in resolving mine.

### Additional context

`package.json` file content:

```
{
    "name": "(...)",
    "version": "1.0.0",
    "private": true,
    "scripts": {
        "dev": "nuxi dev",
        "build": "nuxi build",
        "start": "nuxi preview"
    },
    "dependencies": {
        "@chenfengyuan/vue-countdown": "^1.1.5",
[...]

PR #3999, opened 2022-03-31
fix(nuxi): ignore buildDir on windows too

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/3675
resolves https://github.com/nuxt/framework/issues/3883
may resolve some of the issues reported in https://github.com/nuxt/framework/issues/3147, for example https://github.com/nuxt/framework/issues/3147#issuecomment-1060124361

### ❓ Type of change

### 📚 Description

Previously we were checking a string like `C:\\test\\.nuxt\\somefile.js` against `C:\\test\\.nuxt/` which would of course never match, meaning on Windows dev servers were immediately interrupting themselves and restarting when they noticed changes in `.nuxt`.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -1,4 +1,4 @@
-import { resolve, relative } from 'pathe'
+import { resolve, relative, normalize } from 'pathe'
 import chokidar from 'chokidar'
 import { debounce } from 'perfect-debounce'
 import type { Nuxt } from '@nuxt/schema'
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/commands/dev.ts
```diff
@@ -73,7 +73,7 @@ export default defineNuxtCommand({
     const watcher = chokidar.watch([rootDir], { ignoreInitial: true, depth: 1 })
     watcher.on('all', (event, file) => {
       if (!currentNuxt) { return }
-      if (file.startsWith(withTrailingSlash(currentNuxt.options.buildDir))) { return }
+      if (normalize(file).startsWith(withTrailingSlash(normalize(currentNuxt.options.buildDir)))) { return }
```
pi0, line 76: A note: `buildDir` should be already normalized (all options from nuxt)

danielroe: Yes, it's probably just on Bridge that it is an issue.

pi0: Would be nice to try normalizing `srcDir`/`buildDir`/`rootDir` for bridge if that's the case. Potentially avoiding more of issues like this.

pi0: Precisely @Eckhardt-D it wasn't ignoring on windows...

Eckhardt-D: Great stuff. Thanks for the fix!

PR #3915, opened 2022-03-26
feat(nuxi): Entity management shortcuts

Diizzayy:
As outlined in nuxt/nuxt.js#13550 by @pi0 @Atinux. This POC draft is our first step in introducing [`Entity management shortcuts`](#3903) to the `Nuxi CLI`.

`Entity Shortcuts` are meant to provide a simple yet robust way to quickly manage projects via CLI in this first stage. This functionality will later be leveraged by API Routes (ex: `/__nuxi/<entity>/<command>`) and Nuxt Devtools to allow for project management directly from the browser.

Nuxt is heavily dependent on proper filesystem structure, hence this PR aims to bridge to the gap making it easier for both newcomers as well as veterans of Nuxt to minimize the time it takes it scan and manage projects.

Upon arrival of a proper unjs templating util, we'll be able to begin full scale extension of this functionality

Initial implementation is as follows (Though the possibilities are vast):
```ts
nuxi <entity> <command> <name?>

nuxi modules show
nuxi modules add @unocss/nuxt
nuxi composables new useProfile
nuxi pages new "articles/[slug]"
```

the `show` command is meant to display an overview of the filesystem structure for the given entity. We also explored the possibility of using a `--resolve-all` flag for leveraging a nuxt3 instance to get a fully resolved list of all the entities which may be extended by modules (pages, plugins, composables, middlewares, etc...), Though we've decided for now (performance in mind) that the full resolution would be a better fit for Nuxt Devtools.

Any ideas or suggestions are welcome!

Todo:
- [x] Initial Functionality
- [ ] Refactor code @pi0
- [ ] Update Documentation

---

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #4001, opened 2022-03-31
feat(nuxi): manage modules via cli

Intevel:
### 🔗 Linked issue

https://github.com/nuxt/rfcs/issues/31
https://github.com/nuxt/framework/discussions/569
#3915

### ❓ Type of change

### 📚 Description

With the new commands you can now add different Nuxt 3 modules (& bridge) to the project via Nuxi. The given module name is matched with a list that is currently still done manually (maybe in the future an API?), the name is matched to prevent that not Nuxt modules are installed. The module is added to the `.nuxtrc` file which is merged with the normal nuxt config.

carbon

### 📝 Checklist

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -10,7 +10,7 @@
   "devDependencies": {
     "jiti": "^1.16.0",
     "nuxt": "npm:nuxt3@latest",
-    "nuxt-website-theme": "^0.1.0",
+    "nuxt-website-theme": "^0.1.2",
     "pathe": "^1.0.0",
     "scule": "^1.0.0",
     "untyped": "^1.0.0"
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -1035,6 +1035,32 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@nuxt/kit@npm:^3.0.0-rc.14":
+  version: 3.0.0-rc.14
+  resolution: "@nuxt/kit@npm:3.0.0-rc.14"
+  dependencies:
+    "@nuxt/schema": 3.0.0-rc.14
+    c12: ^1.0.1
+    consola: ^2.15.3
+    defu: ^6.1.1
+    globby: ^13.1.2
+    hash-sum: ^2.0.0
+    ignore: ^5.2.0
+    jiti: ^1.16.0
+    knitwork: ^1.0.0
+    lodash.template: ^4.5.0
+    mlly: ^1.0.0
+    pathe: ^1.0.0
+    pkg-types: ^1.0.1
+    scule: ^1.0.0
+    semver: ^7.3.8
+    unctx: ^2.1.0
+    unimport: ^1.0.0
+    untyped: ^1.0.0
+  checksum: 129af9745d168b6c057a7fad9df15920830fc003ce3d00ca8d0535de5322dd955759bf5439a30f04bd7c166a4c10dbc451b94c7cfdbfcb5acfe6e55257272011
+  languageName: node
+  linkType: hard
+
 "@nuxt/postcss8@npm:^1.1.3":
   version: 1.1.3
   resolution: "@nuxt/postcss8@npm:1.1.3"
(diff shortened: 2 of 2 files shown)


Intevel:
@Baroshem

pi0:
Thanks for the PR @Intevel. Some notes:

- Please disable local linter (prettier) and run `yarn lint --fix`
- We can start with `nuxi module add <npm name>` syntax (later integrate with [nuxt modules](https://github.com/nuxt/modules) DB but is not straight forward)
- `lmify` is deprecated. Working on a new package. It is not safe to rely on.

Intevel:
Thank you @pi0. 
I had already talked with @Atinux that we draft this PR for now. I fixed the linting & disabled prettier. Currently have some kind of manual DB inserted with names, similar to slugs. That would be just an idea.

packages/nuxi/src/commands/module.ts
```diff
@@ -0,0 +1,60 @@
...
+      process.exit(1)
+    }
+    if (!name) {
+      consola.error('Name argument is missing!')
+      process.exit(1)
+    }
+
+    // Checking if provided package is a Nuxt Module
+    const module = await findModule(name)
+    if (!module) { consola.error('Please check the module list, and ensure if') }
```
Baroshem, line 32: Is this error message correct? It seems to end without proper description 😉

packages/nuxi/src/utils/modules.ts
```diff
@@ -0,0 +1,82 @@
...
+    packageName: 'nuxt-hue'
+  },
+  {
+    name: 'grapqhl',
+    packageName: 'nuxt-graphql-client'
+  }
+]
+
+export async function findModule (name: string) {
+  const modules = availableModules.filter(
```
Baroshem, line 77: Maybe you can use find method instead?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

packages/nuxi/src/utils/modules.ts
```diff
@@ -0,0 +1,82 @@
...
+    name: 'nuxt-hue',
+    packageName: 'nuxt-hue'
+  },
+  {
+    name: 'grapqhl',
+    packageName: 'nuxt-graphql-client'
+  }
+]
+
+export async function findModule (name: string) {
```
Baroshem, line 76: Maybe better here would be to name this method findModulePackageName and just return the module package name? If I am correct only the package name is being used later on

packages/nuxi/src/commands/module.ts
Intevel: I will fix it, thank you 😅

packages/nuxi/src/utils/modules.ts
Intevel: I think it's too long ^^

Atinux:
Thank you for the PR :)

Agree to start with `nuxi module add <npm name>` since it allow anyone to use new packages that are not officially listed.

Intevel:
Should the `npm name` still be checked against a list to make sure it is a Nuxt module?

Intevel:
> Thanks for the PR @Intevel. Some notes:
> 
> 
> 
> - Please disable local linter (prettier) and run `yarn lint --fix`
> 
> - We can start with `nuxi module add <npm name>` syntax (later integrate with [nuxt modules](https://github.com/nuxt/modules) DB but is not straight forward)
> 
> - `lmify` is deprecated. Working on a new package. It is not safe to rely on. 
> 
> 

@pi0 https://github.com/Intevel/qpind Maybe this instead of `lmify` for this?

pi0:
Nice work on qpind @Intevel 💚 However we already have [antfu/install-pkg](https://github.com/antfu/install-pkg) and other solutions, I would still wait for new unjs package designed for nuxi and nuxt universal integration.

Intevel:
@pi0 Any updates on this?

Baroshem:
Folks, can we start working on this feature again? I think this would be an awesome DX improvement to the Nuxi tool. I can help with implementing this as well :)

Intevel:
Surely, so the package installing can we do with install-pkg but where do we get the data from to check if this is a nuxt module or not?

Baroshem:
Can't we just parse the user input against a regexp and if it includes the `nuxt` keywoard assume that it is a module? @nuxt/@nuxtjs/nuxt-something should all work with that. I think even right now, you will only handle properly the modules that are in the modules array so we can always log to the user (Module not found in the supported modules DB/array, please configure the module by yourself)

Intevel:
@pi0 What do you think? Is a regex in this case the best way?

Intevel:
@Baroshem pushed some new stuff, what are you're thoughts about it? (btw i know its not an regex, wrong commit message)

Baroshem:
@Intevel 

Yes, this is what I meant in my proposition in the Discord channel. Nice work! Now, let's wait for Pio, Daniel, or Sebastien to give their opinion and recommend next steps :)

manniL:
@Intevel Did you close this intentionally?

pi0:
Supporting module management from CLI is definitely one thing we always dreamt of having but has been tricky so far because we lacked the CLI infra and AST modification tool to make it happen.

Would be happy to have you on board in next weeks on this feature @Intevel and sorry was delayed in the roadmap.

Baroshem:
I can help with this as well. Just let me know :)

Intevel:
Nice to hear, thank you.

Issue #3172, opened 2022-02-10
nuxi generate: ReferenceError: System is not defined


yassilah:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.13.0`
- Nuxt Version:     `2.16.0-27358576.777a4b7f`
- Package Manager:  `yarn@1.22.17`
- Bundler:          `Webpack`
- User Config:      `target`, `srcDir`, `dir`, `bridge`, `head`, `components`, `generate`, `build`, `plugins`, `buildModules`, `modules`, `alias`, `vite`, `hooks`, `layouts`, `autoImports`, `database`, `info`, `style`, `fonts`
- Runtime Modules:  `normalizedModule()`
- Build Modules:    `@nuxt/bridge@3.0.0-27408103.b243891`, `normalizedModule()`, `normalizedModule()`, `normalizedModule()`, `normalizedModule()`, `normalizedModule()`, `normalizedModule()`, `normalizedModule()`, `normalizedModule()`, `normalizedModule()`, `@vueuse/nuxt@7.6.1`, `unplugin-icons/nuxt`, `@pinia/nuxt@0.1.8`
------------------------------

### Reproduction

Simply run `nuxi generate` in a nuxt bridge project, then run `nuxi preview`.

### Describe the bug

An error is thrown from the `entry-legacy` file: `Uncaught (in promise) ReferenceError: System is not defined`.

IlyaSemenov:
It's not only `nuxi generate` but also `nuxi build`. There is no SystemJS runtime in the final production build, and all client-side code doesn't load and doesn't start.

Issue #3738, opened 2022-03-17
align bridge vite with nuxt 3 vite


danielroe:
Currently the vite implementation in Bridge diverges somewhat from within Nuxt 3. This is a problem for a couple of different reasons.

1. it increases the maintenance burden and doubles potential bug-surface
2. it is misleading for people migrating to nuxt 3, because the experience and compatibility is subtly different

One example is using `@vitejs/plugin-legacy`. The [context for using it](https://github.com/nuxt/vite/pull/179) is the need for native ES module support in browsers. Otherwise, adopting vite for Nuxt 2 would mean unsupporting browsers that historically Nuxt 2 has supported. See [**browser support table on MDN**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#browser_support).

I propose we align the implementations of vite in both Bridge & Nuxt 3, including (where possible) either removing use of `@vitejs/plugin-legacy` or porting it to the Nuxt 3 vite implementation.

---

**Related issues with `@vitejs/plugin-legacy`**

- [ ] it will be tricky to add support for dynamic base urls - seen whilst tackling https://github.com/nuxt/framework/issues/3570
- [ ] the polyfill is loaded async which means it may not have been executed before entry, leading to a fatal error: https://github.com/nuxt/framework/issues/3172 - solution in separate PR

Jackliu007888:
so how to resolve it

PR #4005, opened 2022-03-31
refactor(bridge): align bridge with vite and inline systemjs polyfill in entry

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/3172

### ❓ Type of change

### 📚 Description

This PR:

1. makes bridge + vite-builder more closely aligned, including porting support for dynamic base paths to bridge vite, including using `untyped` to give us just the `vite` default options
2. inlines systemjs polyfill for now (otherwise it was susceptible to a race condition and possibly not being loaded - note that this is a temporary fix and we will remove legacy support in bridge: https://github.com/nuxt/framework/issues/3738

A lingering issue is that Bridge does not respect static URLs in terms of custom baseURL or cdnURL - this can be deferred to another PR.

### 📝 Checklist


packages/bridge/src/vite/vite.ts
```diff
@@ -18,41 +22,38 @@ async function bundle (nuxt: Nuxt, builder: any) {
     p.src = nuxt.resolver.resolvePath(resolve(nuxt.options.buildDir, p.src))
   }
 
+  // TODO: After nitropack refactor, try if we can resuse the same server port as Nuxt
```
pi0, line 25: This is tricky with ws integrartion. We need to investigate a proxy solution to bypass to vite-port still, this is needed unless vite provides another way for HMR without listening. In both cases comment could be removed.

packages/bridge/src/vite/vite.ts
```diff
@@ -18,41 +22,38 @@ async function bundle (nuxt: Nuxt, builder: any) {
...
+  const hmrPortDefault = 24678 // Vite's default HMR port
+  const hmrPort = await getPort({
+    port: hmrPortDefault,
+    ports: Array.from({ length: 20 }, (_, i) => hmrPortDefault + 1 + i)
+  })
+
+  nuxt.options.vite = applyDefaults(NuxtConfigSchema, { ...nuxt.options, server: {} }).vite
```
pi0, line 32: This can be problematic in the future as we add more to the Nuxt 3 schema for bridge support which is limited anyway. I would suggest to keep defaults and normalization inside bridge.

packages/vite/src/plugins/dynamic-base.ts
```diff
@@ -18,6 +16,15 @@ export const RelativeAssetPlugin = function (): Plugin {
 
       for (const file in bundle) {
         const asset = bundle[file]
+        if (asset.fileName.includes('legacy') && asset.type === 'chunk' && asset.code.includes('innerHTML')) {
```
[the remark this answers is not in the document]
danielroe: We can remove this as soon as I remove the `@vitejs/legacy` plugin when I implement https://github.com/nuxt/framework/issues/3738.

packages/bridge/src/vite/vite.ts
danielroe: This is just a duplicate of the current vite builder implementation + comment, and ideally we can resolve both at the same time with the same solution.

pi0: Not sure how comment landed in vite builder :D Indeed ideally if we came with a solution we can backport it to bridge as well. But let's remove comment. It is misleading and I'm not even sure easily possible to be a TODO item.

packages/bridge/src/vite/vite.ts
danielroe: I will go ahead with that, though I guess if https://github.com/unjs/untyped/issues/27 lands, we could revisit?

pi0: Hmm not exactly. As we discussed earlier, schema versioning and untyped are for **upward compatibility**.

Schema for old types of bridge and Nuxt 2 are only kept as a reference that we don't reuse the same namespace and if did, being careful to not conflict with Nuxt 2. Bridge support has to stop at some point and having less dependency helps for a more gentle approach.

Issue #4009, opened 2022-03-31
Option to add CORS headers to static assets


hubermat:
### Discussed in https://github.com/nuxt/framework/discussions/4003

<div type='discussions-op-text'>

<sup>Originally posted by **hubermat** March 31, 2022</sup>
Hello,

our Nuxt-based project serves, besides other API routes, static files (i.e. files from `./public`). These static files will also be accessed from outside our web application, but browsers apply CORS policies. Up to now I could not find a way to add `access-control-allow-origin` headers to the static files that are served by the Nuxt server.

I have created a pull request, that adds a new config option in nuxt.config.js:

```js
render: {
  static: {
    /**
     * Whether to add headers to allow access from any origin
     *
     * If set to true, all responses of static files from the `public/` directory will contain the following headers:
     * Access-Control-Allow-Origin *
     * Access-Control-Allow-Methods GET
     * Access-Control-Allow-Headers Origin, X-Requested-With, Content-Type, Accept
     * X-Frame-Options ALLOWALL
     */
    allowFromAnyOrigin: false
  },
}
```

If allowFromAnyOrigin is set to true, for development mode, a function setHeaders will be passed to serve-static to set these headers. For production mode, the public assets will be flagged and the headers will be set in the static middleware.
</div>

chengzhuo5:
It`s useful for me:
```ts
import { defineCorsEventHandler } from '@nozomuikuta/h3-cors';
export default defineNuxtConfig({
  builder: 'webpack',
  devServerHandlers: [
    {
      route: '/',
      handler: defineCorsEventHandler({}),
    },
  ],
  // With vite
  vite: {
    server: {
      headers: {
        'Access-Control-Allow-Origin': '*',
      }
    }
  },
  // or webpack
  webpack: {
    devMiddleware: {
      headers: {
        'Access-Control-Allow-Origin': '*',
      },
    },
  },
});
```

pi0:
You can try routeRules to set CORS for `/_nuxt` assets: https://v3.nuxtjs.org/guide/concepts/rendering#route-rules. For your own, create a sub directory inside public dir like `public/static` and set `/static: { cors: true }` rule.

chengzhuo5:
It`s invalid with webpack builder.




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "nuxt/framework"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年10月21日(星期五) 下午4:11
[...]

PR #4010, opened 2022-03-31
feat: add nuxt config option render.static.allowFromAnyOrigin

hubermat:
### 🔗 Linked issue

#4009 

### ❓ Type of change

### 📚 Description

This pull request adds a new config option in nuxt.config.js:

```js
render: {
  static: {
    /**
     * Whether to add headers to allow access from any origin
     *
     * If set to true, all responses of static files from the `public/` directory will contain the following headers:
     * Access-Control-Allow-Origin *
     * Access-Control-Allow-Methods GET
     * Access-Control-Allow-Headers Origin, X-Requested-With, Content-Type, Accept
     * X-Frame-Options ALLOWALL
     */
    allowFromAnyOrigin: false
  },
}
```

If allowFromAnyOrigin is set to true, for development mode, a function setHeaders will be passed to serve-static to set these headers. For production mode, the public assets will be flagged and the headers will be set in the static middleware.

This allows static files from a Nuxt server to be served to clients from another domain, circumventing a CORS policy error.

### 📝 Checklist

- [ ] I have updated the documentation accordingly.  => so far only in schema/src/config/render.ts

### DIFF
--- a/packages/nitro/src/context.ts
+++ b/packages/nitro/src/context.ts
@@ -62,6 +62,7 @@ export interface NitroContext {
   }
   storage: StorageOptions,
   assets: AssetOptions,
+  staticAllowFromAnyOrigin: boolean,
   _nuxt: {
     majorVersion: number
     dev: boolean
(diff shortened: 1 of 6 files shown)


pi0:
Thanks for the pull request @hubermat. I think it makes sense that we add an option for adding CORS headers for dist static assets. However, `renderer` namespace is deprecated for Nuxt 3 and we need to implement this either using a server middleware or new routing rules (comming with nitropack and #560). Let's track issue via #4009.

PR #4019, opened 2022-04-01
fix(nuxt3)!: add default cookie path to `/`

### LINKED ISSUE #4018: useCookie sets path different from root
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-27480156.02d2d1a`
- Builder:          `vite`
- User Config:      `css`, `buildModules`, `vite`, `build`
- Runtime Modules:  `-`

### Reproduction

There's no need.

Captura de pantalla de 2022-04-01 13-03-28

Captura de pantalla de 2022-04-01 13-04-34

### Describe the bug

When useCookie is used in a nested page (for example `/auth/sign-in`) and a value is set... The cookie is saved with the path `/auth`.

Also useCookie cannot be told to use the `/` path.

This can cause issue #3876.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #4018

### ❓ Type of change

### 📚 Description

This defaults cookie path to app's baseURL (though we could instead set it to `/` by default if there's a reason)

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/composables/cookie.ts
+++ b/packages/nuxt3/src/app/composables/cookie.ts
@@ -4,7 +4,7 @@ import { parse, serialize, CookieParseOptions, CookieSerializeOptions } from 'co
 import { appendHeader } from 'h3'
 import type { NuxtApp } from '@nuxt/schema'
 import destr from 'destr'
-import { useNuxtApp, useRuntimeConfig } from '#app'
+import { useNuxtApp } from '#app'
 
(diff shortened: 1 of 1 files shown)


pi0:
I agree about adding default path!  But this is a hard decision since we need to align with setCookie with h3 for API routes. Using / seems most straight forward start to keep consistency. And with that, i guess can directly introduce breaking change in h3 for default value.

Issue #4021, opened 2022-04-01
`useFetch` returns the `.data` property as `any`


uki00a:
### Environment

```
------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.13.2`
- Nuxt Version:     `3.0.0-27479058.e4956c5`
- Package Manager:  `yarn@1.22.5`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------
```

### Reproduction

When running `npx nuxi typecheck` on https://stackblitz.com/edit/github-etxewc, the following error occurs:

```shell
❯ npx nuxi typecheck
Nuxt CLI v3.0.0-27479058.e4956c5                                               20:52:06
app.vue:10:18 - error TS2339: Property 'name' does not exist on type 'Ref<any>'.

10 console.log(data.name); // A type error occurs on this line
                    ~~~~


Found 1 error in app.vue:10


 ERROR  Command failed with exit code 2: vue-tsc --noEmit                      20:52:13

  at makeError (node_modules/nuxi/dist/chunks/index4.mjs:1253:11)
  at handlePromise (node_modules/nuxi/dist/chunks/index4.mjs:2082:26)
  at async Object.invoke (node_modules/nuxi/dist/chunks/typecheck.mjs:52:7)
  at async _main (node_modules/nuxi/dist/cli.mjs:57:7)
```

### Describe the bug

[The documentation](https://github.com/nuxt/framework/blob/323803832a87b7bf2783210e1769f2f7f39a3d5b/docs/content/3.docs/1.usage/1.data-fetching.md#usefetch) say that `useFetch` infers its return type. However, the actual return type seems to be declared as `AsyncData<any>` (https://unpkg.com/browse/nuxt3@3.0.0-27479058.e4956c5/dist/app/composables/fetch.d.ts)

### Additional context

I think that the problem comes from the following code:

https://github.com/nuxt/framework/blob/02d2d1ae35ab94deaf5ca21b2c37826085ff3116/packages/nuxt3/src/app/composables/fetch.ts#L45-L51

Changing this to the following may resolve the problem 🤔

```ts
 const _asyncDataOptions = { 
   ...opts, 
   watch: [ 
     _request, 
     ...(opts.watch || []) 
   ] 
 } 
```

danielroe:
It is a Ref, so it should be:

```diff
- console.log(data.name);
+ console.log(data.value.name);
```

[...]

PR #4058, opened 2022-04-04
fix(nuxt3): use inferred type for `useFetch`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/4021

### ❓ Type of change

### 📚 Description

I'll follow up with a PR to fix the test suite, which should have caught this.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/composables/fetch.ts
+++ b/packages/nuxt3/src/app/composables/fetch.ts
@@ -42,15 +42,17 @@ export function useFetch<
     cache: typeof opts.cache === 'boolean' ? undefined : opts.cache
   }
 
-  const asyncData = useAsyncData(key, () => {
-    return $fetch(_request.value, _fetchOptions) as Promise<_ResT>
-  }, {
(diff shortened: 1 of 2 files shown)


packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -42,17 +42,15 @@ export function useFetch<
     cache: typeof opts.cache === 'boolean' ? undefined : opts.cache
   }
 
-  const _asyncDataOptions: AsyncDataOptions<any> = {
```
pi0, line 45: Is there a way to fix ts issue without inlining options?

danielroe: Yes. It was just simpler and, by relying on inference we reduce the risk of future breakage. I can push a fix that doesn't inline them.

pi0: I see. Actually i did refactor on purpose to extend object later for caching. Avoid inlining + inference would be nice!

PR #4066, opened 2022-04-04
refactor(nuxt3,bridge)!: rename `useMeta` to `useHead`

danielroe:
### 🔗 Linked issue

resolves nuxt/nuxt.js#13651

### ❓ Type of change

### 📚 Description

This renames `useMeta` to `useHead` to improve DX and reduce confusion with `definePageMeta`.

We also:
 * remove support for typed Nuxt 2 `head` in `nuxt.config` as this conflicts with new `nuxt.options.head` typings
 * continue to support `useMeta` in both Bridge & Nuxt, but display a warning the first time it's used
 * add tests for nuxt 3 useHead

### 👉 Migration

Rename `useMeta` to `useHead` anywhere it occurs in your code, and use `app.head` in your `nuxt.config` rather than `meta`.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/_app.ts
+++ b/packages/schema/src/config/_app.ts
@@ -66,9 +66,48 @@ export default {
     */
     cdnURL: {
       $resolve: (val, get) => get('dev') ? null : val || null
-    }
+    },
+    /**
+     * Set default configuration for `<head>` on every page.
+     *
+     * @example
+     * ```js
+     * app: {
+     *   head: {
+     *     meta: [
+     *       // <meta name="viewport" content="width=device-width, initial-scale=1">
+     *       { name: 'viewport', content: 'width=device-width, initial-scale=1' }
+     *     ],
+     *     script: [
+     *       // <script src="https://myawesome-lib.js"></script>
+     *       { src: 'https://awesome-lib.js' }
+     *     ],
+     *     link: [
+     *       // <link rel="stylesheet" href="https://myawesome-lib.css">
+     *       { rel: 'stylesheet', href: 'https://awesome-lib.css' }
+     *     ],
+     *     // please note that this is an area that is likely to change
+     *     style: [
+     *       // <style type="text/css">:root { color: red }</style>
+     *       { children: ':root { color: red }', type: 'text/css' }
+     *     ]
+     *   }
+     * }
+     * ```
+     * @type {typeof import('../src/types/meta').MetaObject}
+     * @version 3
+     */
+    head: {
+      $resolve: (val, get) => {
+        return defu(val, get('meta'), {
+          meta: [],
+          link: [],
+          style: [],
+          script: []
+        })
+      }
+    },
   },
-
   /**
    * The path to a templated HTML file for rendering Nuxt responses.
    * Uses `<srcDir>/app.html` if it exists or the Nuxt default template if not.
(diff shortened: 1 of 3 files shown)


packages/nuxt3/src/app/index.ts
```diff
@@ -8,7 +8,7 @@ export * from './components'
 export type { PageMeta } from '../pages/runtime'
 // eslint-disable-next-line import/no-restricted-paths
 export type { MetaObject } from '../meta/runtime'
-export { useMeta } from '#meta'
+export { useHead, useMeta } from '#meta'
```
pi0, line 11: Shall we also update `#meta` references?

packages/nuxt3/src/meta/runtime/composables.ts
```diff
@@ -11,7 +11,20 @@ import { useNuxtApp } from '#app'
...
   const resolvedMeta = isFunction(meta) ? computed(meta) : meta
-  useNuxtApp()._useMeta(resolvedMeta)
+  useNuxtApp()._useHead(resolvedMeta)
+}
+
+// TODO: remove useMeta support when Nuxt 3 is stable
+const _warned = {}
+const warnOnce = (id: string, message: string) => {
+  if (!_warned[id]) {
+    console.warn('[meta]', message)
```
pi0, line 23: Util should be either reusable or specific to meta :D I would move tag to `useMeta` to make it a generic util and then add todo + `@deprecated` on top of `useMeta`

packages/schema/src/config/_app.ts
```diff
@@ -124,19 +124,40 @@ export default {
...
+   *  style: [
+   *    // <style type="text/css">:root { color: red }</style>
+   *    { children: ':root { color: red }', type: 'text/css' }
+   *  ]
+   * }
+   * ```
+   * @type {typeof import('../src/types/meta').MetaObject}
    * @version 2
    */
   head: {
```
pi0, line 152: This is really bad about conflicting with Nuxt 2 config :( All efforts of adding Nuxt 2 to the schema were to avoid such conflict... Do you think of any new non-conflicting option name? /cc @Atinux

danielroe: Indeed, that was originally why we picked `useMeta` instead of `useHead`. We could entirely deprecate setting meta tags in config and advise setting them in `app.vue` or a plugin?

pi0: Hmm, sure. Also, we can make use of `app/head.ts` for additional head tags. But the main purpose of config, was for allowing modules easily add their head and meta tags without writing (and adding the overhead of) a plugin. They will loose it this way.

danielroe: I see these options:

1. leave it as `meta` in `nuxt.config`, which will be inconsistent with `useHead`, perhaps also accepting `meta` in component options rather than `head`
2. move it somewhere else in the schema
3. normalize `head` schema so `vue-meta` options are normalised -> useHead and vice versa (or pick new schema for options)
4. type it as arrays of `any` and recommend users set it with `useHead` for type support...

pi0: We should normally do something like (2) unless api is cross version compatible (which is not. Features also differe beween vue-meta and vueuse)

I'm thinking to use app namespace (app.head), however have to carefully change it's purpose and only pick some options to runtimeConfig from app.

danielroe: Shall I go ahead and migrate this to `app.head` then?

packages/schema/src/config/_common.ts
```diff
@@ -744,6 +744,15 @@ export default {
...
   publicRuntimeConfig: {
-    $resolve: (val: Record<string, any> = {}, get) => ({ ...val, app: defu(val.app, get('app')) })
+    $resolve: (val: Record<string, any> = {}, get) => ({
+      ...val,
+      app: {
+        baseURL: get('app.baseURL'),
+        buildAssetsDir: get('app.buildAssetsDir'),
+        assetsPath: get('app.assetsPath'),
+        cdnURL: get('app.cdnURL'),
+        ...val.app || {},
```
pi0, line 754: Doesn't it add any custom app.head to runtimeCOnfig?

danielroe: `val` here is `publicRuntimeConfig` so it will only add whatever the user adds in `publicRuntimeConfig.app`.

fiamma06:
After rename **useMeta** to **useHead** - **useMeta** still work correctly with notice, but **useHead** doesn't work
https://v3.nuxtjs.org/examples/composables/use-head - also don't work

danielroe:
Thanks. This is an issue in `vueuse/nuxt`, which is used in these examples: https://github.com/vueuse/vueuse/blob/main/packages/nuxt/ssr-plugin.mjs

PR: https://github.com/vueuse/vueuse/pull/1481

PR #4092, opened 2022-04-05
fix(nuxt3)!: sync route object to currently resolved page

### LINKED ISSUE #4086: suspend route object until route suspense resolves
At the moment the route changes before the route component fully loads ([sandbox](https://stackblitz.com/edit/nuxt3-windicss-bm1znm?file=app.vue)).

We already provide our own `useRoute` composable. We can update our implementation so the `route` object returned continues to show the past route until the route component suspense resolves, so this kind of inconsistency is not present.

danielroe:
### 🔗 Linked issue

resolves #4086, resolves https://github.com/nuxt/framework/discussions/2459

### ❓ Type of change

### 📚 Description

This PR does two things:

1. correctly resolves the route for use within plugins (previously `useRoute()` did not return the correct route until the plugin lifecycle had fully run)
2. creates a 'suspended' route that mirrors the behaviour of the page component - so it does not resolve to the new route until the page component resolves

I would especially welcome testing on this given the nature of the change.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/pages/runtime/router.ts
+++ b/packages/nuxt3/src/pages/runtime/router.ts
@@ -3,7 +3,8 @@ import {
   createRouter,
   createWebHistory,
   createMemoryHistory,
-  NavigationGuard
+  NavigationGuard,
+  RouteLocation
 } from 'vue-router'
 import { createError } from 'h3'
 import NuxtPage from './page'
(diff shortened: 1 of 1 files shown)


packages/nuxt3/src/pages/runtime/router.ts
```diff
@@ -52,10 +54,23 @@ export default defineNuxtPlugin((nuxtApp) => {
     get: () => previousRoute.value
   })
 
+  // Allows suspending the route object until page navigation completes
+  const path = process.server ? nuxtApp.ssrContext.req.url : withoutBase(window.location.href, window.location.origin)
```
pi0, line 58: Do you know a reference how vue-router internally resolve initial path?

pi0: Does it make sense to somehow reuse same implementation to avoid any possible mismatch?

danielroe: Makes sense. At the moment, we don't support baseURLs with `#` in them (i.e. we just get 404s). Is this a use-case we want to support?

Atinux:
Nice work @danielroe 

It would be nice to test indeed and check if it works when using a nested page too 😊

danielroe:
Tested ✅

**Note**: At the moment, a nested page's async setup doesn't block the loading of the parent, which means it doesn't block the updating of the `route` object. That may be a separate issue to track, but it doesn't seem to be an issue with this PR.

misaon:
@danielroe I fighting with following case:

app.vue
```vue
<template>
  <div>
    <NuxtLayout>
      <NuxtPage/>
    </NuxtLayout>
  </div>
</template>
```

pages/index.vue
```vue
<template>
  <div>
    <NuxtLink to="/hi.html">
      Go to Hi page
    </NuxtLink>
  </div>
</template>
```


pages/[slug].html.vue
```vue
<template>
  <div>
    <h1>Hi</h1>
  </div>
</template>

<script lang="ts" setup>
const route = useRoute()

console.log(route.params)
</script>
```

console:
```js
{}
```

How can i get `slug` param value now after your change?

misaon:
Im going back to `3.0.0-27486179.cb74ab6`

Issue #4122, opened 2022-04-06
Can't access current route within page component setup


danielroe:
Example: https://stackblitz.com/edit/github-t3dxlt

see https://github.com/nuxt/framework/pull/4092#issuecomment-1090069023

frbuceta:
Confirmed, it's happening to me too

PR #4124, opened 2022-04-06
fix(nuxt3): move new router behavior to `useActiveRoute`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/4122

### ❓ Type of change

### 📚 Description

Follow-up on https://github.com/nuxt/framework/pull/4092; users may legitimately wish to access current route _within_ the page that is displaying. It doesn't make sense to suspend the route within the page that is loading, even if we do so elsewhere.

This PR provides the current route to the children of NuxtPage, exempting them from the suspension behaviour, meaning effectively `useRoute` only suspends the routes outside pages (e.g. in layouts or app.vue)...

An alternative/future approach would be to allow users to be in full control of this behaviour, perhaps allowing `useRoute` to take an option to enable the suspension behaviour for that route alone. Thoughts welcome.

### 📝 Checklist


misaon:
Thanks @danielroe ! 👍🏼

pi0:
There is too much overhead by injecting this way also makes `useRoute` behavior inconsistent. I propose that we introduce the new behavior of #4092 as the new `useActiveRoute` composable which changes after suspense.

pi0 (approved):

PR #4139, opened 2022-04-06
fix(kit): use `route` instead of `path` in `ServerMiddleware`

farnabaz:
### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

This PR will improve compatibility with Nitropack and internal usage of `addServerMiddleware` .
As Nitropack uses `route` and `handler` for middleware, I assume `path` and `handle` will be deprecated. 

### 📝 Checklist

### DIFF
--- a/packages/kit/src/server.ts
+++ b/packages/kit/src/server.ts
@@ -2,7 +2,7 @@ import type { Middleware } from 'h3'
 import { useNuxt } from './context'
 
 export interface ServerMiddleware {
-  path?: string,
+  route?: string,
   handler: Middleware | string
 }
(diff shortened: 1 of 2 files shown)


Rigo-m:
Can it be considered breaking? https://github.com/nuxt/framework/discussions/2883

pi0:
> Can it be considered breaking?

[resolve handlers](https://github.com/nuxt/framework/blob/main/packages/nuxt3/src/core/nitro.ts#L172) should normally add backward compatibility for old names.

PR #4168, opened 2022-04-07
docs: add note about nuxt full restart feature

stafyniaksacha:
### ❓ Type of change

### 📚 Description

This add a note about how nuxt distinct full restart and partial update and add an example on how to benefit from full restart (to split configuration file)

Based on: https://github.com/nuxt/framework/blob/main/packages/nuxi/src/commands/dev.ts#L87-L89

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/3.directory-structure/18.nuxt.config.md
+++ b/docs/content/2.guide/3.directory-structure/18.nuxt.config.md
@@ -18,3 +18,19 @@ export default defineNuxtConfig({
 
 ::ReadMore{link="/api/configuration/nuxt.config"}
 ::
+
+To ensure your configuration is up to date, nuxt will make a full restart when detecting changes in the nuxt main configuration file, the `.env`, `.nuxtignore` and `.nuxtrc` dot files.
+
+If you want to externalize configuration and benefit from full restart, you can use `*.nuxt.config` suffix:
+
+
+```ts
+import { defineNuxtConfig } from 'nuxt3'
+
+// nuxt will restart when changes detected on meta.nuxt.config
+import { meta } from './meta.nuxt.config'
+
+export default defineNuxtConfig({
+  meta,
+})
+```
--- a/docs/content/2.guide/3.directory-structure/18.nuxt.config.md
+++ b/docs/content/2.guide/3.directory-structure/18.nuxt.config.md
@@ -16,21 +16,7 @@ export default defineNuxtConfig({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


docs/content/2.guide/3.directory-structure/18.nuxt.config.md
```diff
@@ -18,3 +18,19 @@ export default defineNuxtConfig({
 
 ::ReadMore{link="/api/configuration/nuxt.config"}
 ::
+
+To ensure your configuration is up to date, nuxt will make a full restart when detecting changes in the nuxt main configuration file, the `.env`, `.nuxtignore` and `.nuxtrc` dot files.
+
+If you want to externalize configuration and benefit from full restart, you can use `*.nuxt.config` suffix:
```
pi0, line 24: This is mostly a bug-as-a-feature because of [current regex](https://github.com/nuxt/framework/blob/c7ff6d17406f5f3912c24c91f84cd15109e6f1b7/packages/nuxi/src/commands/dev.ts#L87) 😅 I'm not sure we can keep supporting it in the longer feature it was never intended to! Let's do nested support properly and for now, keep using single config file.

stafyniaksacha: To take an example, with unocss nuxt module, we can declare rules in our configuration: https://github.com/unocss/unocss/tree/main/packages/nuxt

Would the good way be that the module listen for a file `unocss.config.js` and then do the reload?

Atinux: I fully agree, having a unocss.config.ts is a great idea, cc @antfu

stafyniaksacha: We may want to update the regex to set `nuxt.config` as prefix instead of suffix, or why not create another setting `watchRestart` ?

pi0: **Off topic:** Supporting extra configuration files is a fair point when integrations provide strong tooling that depends on it. However, I wouldn't recommend it that much. It makes the developer experience less consistent with Nuxt and also impossible to manage that config with upcoming dev tools UI and CLI shortcuts.

pi0:
The main tip is good enough 👍🏼 Once we fixed TODO item to watch dependency tree, we shall update the note about auto restart on dependencies but as I mentioned above, if the nuxt config is getting bigger that needs splitting into multiple files, something is wrong. Nuxt is beautiful when minimal :)

PR #4171, opened 2022-04-07
docs(api): add javascript to code-group

clemcode:
### ❓ Type of change

### 📚 Description

This PR adds a code-group to `useAsyncData` API to support both TypeScript and JavaScript notation.

### DIFF
--- a/docs/content/3.api/1.composables/use-async-data.md
+++ b/docs/content/3.api/1.composables/use-async-data.md
@@ -3,7 +3,13 @@
 ::ReadMore{link="/guide/features/data-fetching"}
 ::
 
-```ts
+::code-group
+
+```js[JavaScript]
+const { data, pending, refresh, error } = useAsyncData(key, handler, options?)
+```
+
+```ts[TypeScript]
 const {
   data: Ref<DataT>,
   pending: Ref<boolean>,
@@ -19,6 +25,8 @@ const {
   }
 )
 ```
+
+::
 
 ## Params
 


Atinux (approved):

pi0:
Thanks for the PR @clemcode but the first section of API reference for composables, is not for usage information but explains function signature with help of a well-known typescript notion for additional information and shall be auto generated in the future with untyped.

This is similar to [vue 3 composable references](https://vuejs.org/api/reactivity-utilities.html#torefs)...

What if we additionally add JS syntax to the usage information?

- Function signature (TS Type)
- Usage example (JS Code)
- Params explanation (List)
- More examples

Atinux:
I see, actually I personally found it hard to understand with the TS syntax, but seems similar to Vue anyway.

pi0:
Certainly and same for me when i need to learn usage. Usage Example right after types can be readable JS code (instead of switch, we provide both API reference and quick usage at one glance)

Atinux:
Pushing an update to be more similar to Vue and adding an example at the end.

Would need feedback to have the correct way to show the types @pi0 @danielroe (I kind of copy / paste from our async data composable), I am not the best for TypeScript.



I think we can take inspiration from https://vuejs.org/api/reactivity-core.html#watcheffect for instance

danielroe (approved):
Yes, this is nice. 👍

docs/content/3.api/1.composables/use-async-data.md
```diff
@@ -1,23 +1,32 @@
 # `useAsyncData`
 
-::ReadMore{link="/guide/features/data-fetching"}
-::
+Within your pages, components and plugins you can use useAsyncData to get access to data that resolves asynchronously.
```
dammy001, line 3:
```suggestion
Within your pages, components, and plugins you can use useAsyncData to get access to data that resolves asynchronously.
```

docs/content/3.api/1.composables/use-async-data.md
```diff
@@ -1,23 +1,32 @@
...
+  pick?: PickKeys
+  watch?: WatchSource[]
+  initialCache?: boolean
+}
+
+type DataT = {
   data: Ref<DataT>,
   pending: Ref<boolean>,
   refresh: () => Promise<void>,
   error: Ref<any>
```
danielroe, line 28:
```suggestion
  data: Ref<DataT>
  pending: Ref<boolean>
  refresh: () => Promise<void>
  error: Ref<any>
```

docs/content/3.api/1.composables/use-async-data.md
```diff
@@ -1,23 +1,32 @@
...
-  options?: {
-    lazy: boolean,
-    server: boolean,
-    watch: WatchSource[]
-  }
-)
+  options?: AsyncDataOptions
+): Promise<DataT>
+
+type AsyncDataOptions = {
```
Atinux, line 14: Should it be an `interface` here?

docs/content/3.api/1.composables/use-async-data.md
```diff
@@ -1,23 +1,32 @@
...
+  server?: boolean
+  lazy?: boolean
+  default?: () => DataT
+  transform?: Transform
```
Atinux, line 18: Can we show it as a function like for `default`?

docs/content/3.api/1.composables/use-async-data.md
pi0: type seems nice also 👍🏼

docs/content/3.api/1.composables/use-async-data.md
```diff
@@ -1,23 +1,32 @@
...
+  pick?: PickKeys
+  watch?: WatchSource[]
```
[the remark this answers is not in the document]
pi0: `WatchSource` is a known vue type. I think is fine line this but we could also use something like `vue.WatchSource` to make more clear.

pi0 (approved):

PR #4210, opened 2022-04-08
fix(nuxt3): improve error types for `useAsyncData` and `useFetch` 

Mini-ghost:
### ❓ Type of change

### 📚 Description

Let `useFetch()` and `useLazyFetch()` return error can defined type by generics, like below:

```ts
const { data, error } = useFetch<TData, TError>('url')
```

### 📝 Checklist


packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -67,5 +69,8 @@ export function useLazyFetch<
   request: Ref<ReqT> | ReqT | (() => ReqT),
   opts: Omit<UseFetchOptions<_ResT, Transform, PickKeys>, 'lazy'> = {}
 ) {
-  return useFetch(request, { ...opts, lazy: true })
+  return useFetch<ResT, ErroT, ReqT, _ResT, Transform, PickKeys>(request, {
```
danielroe, line 72:
```suggestion
  return useFetch<ResT, ErrorT, ReqT, _ResT, Transform, PickKeys>(request, {
```

packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -16,10 +16,11 @@ export interface UseFetchOptions<
   FetchOptions
   {
   key?: string
- }
+}
 
 export function useFetch<
   ResT = void,
+  ErroT = any,
```
danielroe, line 23:
```suggestion
  ErrorT = any,
```

packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -50,15 +51,16 @@ export function useFetch<
     ]
   }
 
-  const asyncData = useAsyncData(key, () => {
-    return $fetch(_request.value, _fetchOptions) as Promise<_ResT>
+  const asyncData = useAsyncData<_ResT, ErroT, Transform, PickKeys>(key, () => {
```
danielroe, line 54:
```suggestion
  const asyncData = useAsyncData<_ResT, ErrorT, Transform, PickKeys>(key, () => {
```

packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -50,15 +51,16 @@ export function useFetch<
...
+    return $fetch(_request.value, _fetchOptions)
   }, _asyncDataOptions)
 
   return asyncData
 }
 
 export function useLazyFetch<
   ResT = void,
+  ErroT = any,
```
danielroe, line 63:
```suggestion
  ErrorT = any,
```

Mini-ghost:
@danielroe I adjusted the generics naming, but I'm not sure why the test the `CI / test-fixtures (windows-latest, 16)` is wrong

packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -50,15 +51,16 @@ export function useFetch<
...
+  const asyncData = useAsyncData<_ResT, ErrorT, Transform, PickKeys>(key, () => {
+    return $fetch(_request.value, _fetchOptions)
   }, _asyncDataOptions)
 
   return asyncData
 }
 
 export function useLazyFetch<
   ResT = void,
+  ErrorT = any,
```
pi0, line 63: Should we default it to `Error`? 

```suggestion
  ErrorT = Error,
```

danielroe: It will be a boolean on client-side, so we could default it to `boolean | Error`:
```suggestion
  ErrorT = boolean | Error,
```

pi0: In this case, maybe we should have something similar to `_ResT` maybe? So that even if user provides a custom error type, we include possibility of boolean value.

danielroe: I think we should probably allow for user to provide a custom error transformation to return a custom error so it's not _always_ boolean on client-side: https://github.com/nuxt/framework/issues/2122

packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -50,15 +77,40 @@ export function useFetch<
...
   ResT = void,
+  ErrorT = Error,
+  ReqT extends string = string,
+  _ResT = ResT extends void ? FetchResult<ReqT> : ResT,
+  Transform extends (res: _ResT) => any = (res: _ResT) => _ResT,
+  PickKeys extends KeyOfRes<Transform> = KeyOfRes<Transform>
+> (
+  request: Ref<ReqT> | ReqT | (() => ReqT),
+  opts?: Omit<UseFetchOptions<_ResT, Transform, PickKeys>, 'lazy' | 'sercer'> & { server?: true }
+): AsyncData<PickFrom<ReturnType<Transform>, PickKeys>, ErrorT | true>;
```
danielroe, line 97: should this be?

```suggestion
): AsyncData<PickFrom<ReturnType<Transform>, PickKeys>, ErrorT | boolean>;
```

packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -16,14 +18,39 @@ export interface UseFetchOptions<
...
+  ErrorT = Error,
   ReqT extends FetchRequest = FetchRequest,
   _ResT = ResT extends void ? FetchResult<ReqT> : ResT,
   Transform extends (res: _ResT) => any = (res: _ResT) => _ResT,
-  PickKeys extends KeyOfRes<Transform> = KeyOfRes<Transform>
+  PickKeys extends KeyOfRes<Transform> = KeyOfRes<Transform>,
+> (
+  request: Ref<ReqT> | ReqT | (() => ReqT),
+  opts?: Omit<UseFetchOptions<_ResT, Transform, PickKeys>, 'server'> & { server?: true }
+): AsyncData<PickFrom<ReturnType<Transform>, PickKeys>, ErrorT | true>;
```
danielroe, line 33:
```suggestion
): AsyncData<PickFrom<ReturnType<Transform>, PickKeys>, ErrorT | boolean>;
```

packages/nuxt3/src/app/composables/fetch.ts
Mini-ghost: But if `useFetch` be invoked successfully on server side, then error will be `null` rather than `false`.

danielroe: but `server: true` async data/fetch calls _also_ run on client-side when navigating between pages.

packages/nuxt3/src/app/composables/fetch.ts
```diff
@@ -50,15 +77,40 @@ export function useFetch<
...
 export function useLazyFetch<
   ResT = void,
+  ErrorT = Error,
+  ReqT extends string = string,
+  _ResT = ResT extends void ? FetchResult<ReqT> : ResT,
+  Transform extends (res: _ResT) => any = (res: _ResT) => _ResT,
+  PickKeys extends KeyOfRes<Transform> = KeyOfRes<Transform>
+> (
+  request: Ref<ReqT> | ReqT | (() => ReqT),
+  opts?: Omit<UseFetchOptions<_ResT, Transform, PickKeys>, 'lazy' | 'sercer'> & { server?: true }
```
danielroe, line 96:
```suggestion
  opts?: Omit<UseFetchOptions<_ResT, Transform, PickKeys>, 'lazy' | 'server'> & { server?: true }
```

packages/nuxt3/src/app/composables/fetch.ts
danielroe: I think it would be possible to remove the `true` from the error type in `server: false` fetches, but I wonder if the additional boilerplate to do so is worth it? Possibly we can defer until https://github.com/nuxt/framework/issues/2122.

pi0 (approved):

pi0:
Thanks @Mini-ghost @danielroe 💚

PR #4215, opened 2022-04-08
fix(nuxt3,bridge): rename error handler url

danielroe:
### ❓ Type of change

### 📚 Description

with nitropack change, we have access to full url - this resolves that issue. We also choose to rename to `__nuxt_error`. (Two underscores are needed to avoid it matching the asset url early return test.)

### 📝 Checklist


danielroe:
the snapshot tests won't pass until https://github.com/unjs/nitro/pull/58 merges... I'll remove it first from this pr, but we will need to add when we upgrade nitropack.

pi0 (approved):

Issue #4244, opened 2022-04-11
Cookie can't be set together with redirect


Lanayx:
### Environment

Nuxt CLI v3.0.0-27491748.3186841                                                                                                    11:06:50  
RootDir: C:\Work\Test                                                                                                                        11:06:56
Nuxt project info:                                                                                                                               11:06:56

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v16.13.1`
- Nuxt Version:     `3.0.0-27491748.3186841`
- Package Manager:  `npm@8.1.2`
- Builder:          `vite`
- User Config:      `css`, `build`, `buildModules`, `vite`, `typescript`
- Runtime Modules:  `-`
- Build Modules:    `@pinia/nuxt@0.1.8`
------------------------------

### Reproduction

When using this code and performing SSR request (when this code is executed on server side)
```
useCookie('myCookie').value = '123';
await navigateTo('/test');
```
cookie "myCookie" is not set

### Describe the bug

Setting cookie together with redirecting doesn't work when executing SSR request

### Additional context

A bit context why is this needed. When using external auth server it redirect to callback url together with the token (for example `http://localhost:3000/loggedIn?accessToken=abcd`). I want to write down this token to the cookie and redirect user further (for example `http://localhost:3000/dashboard`), so I need both set cookie and redirect functionality

### Logs
On simple pages it just silently fails, in more complex case I see the following warning:

```shell
[nitro] [dev] [unhandledRejection] Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
```

danielroe:
This is because we defer writing the server cookie until the HTML is fully rendered (in case it's set more than once) - but if you redirect then we never reach that point... If you want to set a cookie _and_ redirect, you can do this manually:

```js
import { serialize } from 'cookie-es'
import { appendHeader } from 'h3'

export default defineNuxtPlugin(async () => {
  const route = useRoute()
  if (route.path === '/') {
    const event = useRequestEvent()
    appendHeader(event, 'Set-Cookie', serialize('test', '123'))
    await navigateTo('/test')
  }
})
```

Lanayx:
@danielroe Thank you, I'll try that way! But it would be nice if this could be done not only inside plugin, but inside any component as well.

danielroe:
The same code can be run in a component setup function.

pi0:
[...]

PR #4253, opened 2022-04-11
fix(nuxt3): handle prematurely ended `res` object

danielroe:
### 🔗 Linked issue

related: #4244

### ❓ Type of change

### 📚 Description

It's possible before these points for the request to have been ended (such as in a previous `app:rendered` or in a router middleware).

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/composables/cookie.ts
+++ b/packages/nuxt3/src/app/composables/cookie.ts
@@ -65,7 +65,7 @@ function writeClientCookie (name: string, value: any, opts: CookieSerializeOptio
 }
 
 function writeServerCookie (event: CompatibilityEvent, name: string, value: any, opts: CookieSerializeOptions = {}) {
-  if (event) {
+  if (event && !event.res.writableEnded) {
     // TODO: Try to smart join with existing Set-Cookie headers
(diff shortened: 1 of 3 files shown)


packages/nuxt3/src/app/composables/cookie.ts
```diff
@@ -65,7 +65,7 @@ function writeClientCookie (name: string, value: any, opts: CookieSerializeOptio
 }
 
 function writeServerCookie (event: CompatibilityEvent, name: string, value: any, opts: CookieSerializeOptions = {}) {
-  if (event) {
+  if (event && !event.res.writableEnded) {
```
pi0, line 68: This is not fixing any issue but silently ignoring a cookie addition that is expected but is not possible anymore!

danielroe: it's an internal function so I think we could easily rename. I would not expect this to error; I think a warning would be okay.

pi0: It is used by `useCookie`. Silent errors are not a fix for lifecycle issues!

pi0:
DX to be improved in h3.

PR #4254, opened 2022-04-11
feat: introduce simplified `runtimeConfig` option

pi0:
### ❓ Type of change

### 📚 Description

Since nitropack migration, nitro introduced a simplified concept of `runtimeConfig` that is a [server] runtime config with possibility of overriding with `NUXT_` environment vairables and special `public.*` and `app.*` subkeys that are only exposed to the client but can be also updated at runtime. This PR brings the same simplified interface to `nuxt` config with backward compatibility for `publicRuntimeConfig` and `privateRuntimeConfig`.

Migration:

Update:

```js
  privateRuntimeConfig: {
    secretKey: ''
  },
  publicRuntimeConfig: {
    baseURL: ''
  }
})
```

To:

```
  privateRuntimeConfig: {
    secretKey: '',
    public: {
      baseURL: ''
    }
  }
})
```

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -4,6 +4,7 @@ import createRequire from 'create-require'
 import { pascalCase } from 'scule'
 import jiti from 'jiti'
 import defu from 'defu'
+import { RuntimeConfig } from '../types/config'
 
 export default {
   /**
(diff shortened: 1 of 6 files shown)


packages/schema/src/config/_common.ts
```diff
@@ -695,63 +696,56 @@ export default {
   /**
    * Runtime config allows passing dynamic config and environment variables to the Nuxt app context.
    *
-   * The value of this object is accessible from server only using `$config` or `useRuntimeConfig`.
-   * It will override `publicRuntimeConfig` on the server-side.
+   * The value of this object is accessible from server only using `useRuntimeConfig`.
    *
-   * It should hold _private_ environment variables (that should not be exposed on the frontend).
+   * It mainly should holds _private_ configuration which is not exposed on the frontend.
```
danielroe, line 701:
```suggestion
   * It mainly should hold _private_ configuration which is not exposed on the frontend.
```

packages/schema/src/config/_common.ts
```diff
@@ -695,63 +696,56 @@ export default {
...
    * This could include a reference to your API secret tokens.
    *
+   * Anything under the `public` and `app` will be exposed to the frontned as well.
```
danielroe, line 704:
```suggestion
   * Anything under `public` and `app` will be exposed to the frontend as well.
```

packages/schema/src/config/_common.ts
```diff
@@ -695,63 +696,56 @@ export default {
...
+   *
    * Values are automatically replaced by matching env variables at runtime, e.g. setting an environment
-   * variable `API_SECRET=my-api-key` would overwrite the value in the example below.
-   * Note that the env variable has to be named exactly the same as the config key.
+   * variable `API_KEY=my-api-key` would overwrite the value in the example below.
```
danielroe, line 707:
```suggestion
   * variable `API_KEY=my-api-key PUBLIC_BASE_URL=/foo/` would overwrite the two values in the example below.
```

tobiasdiez:
I have to say that I found the distinction of private vs public quite intuitive. What do you think about the following interface?
```ts
  private: {
    secretKey: ''
  },
  public: {
    baseURL: ''
  }
})
```

msonowal:
and hope in server context also accessing is same

Atinux:
Checking locally, to see the DX :blush:

Atinux:
Looks good to me, but I think we have an issue with HMR and `.env`, see the video to understand:




https://user-images.githubusercontent.com/904724/162755073-6ad5192f-fe18-471d-8fac-d77729f79dc1.mp4

pi0:
> I have to say that I found the distinction of private vs public quite intuitive. What do you think about the following interface?

@tobiasdiez I get your POV. This was what we initially thought of for Nuxt 2 config API. But other than DX cons/pros, this is a new technical definition change with nitro that config originates from server. Term `private` makes less sense.

> and hope in server context also accessing is same

@msonowal It will be universally same with new runtimeConfig (as you can see in Daniel's examples)

> Looks good to me, but I think we have an issue with HMR and .env, see the video to understand:

Thanks for checking @Atinux. Unsetting a `.env` seems as edge case to me but definitely something we can improve and fix. This is not an issue with nitro or composable. But nuxt full start and c12 behavior that loads `.env` to shared `process.env` which can be also usable for `nuxt.config` and modules.

PR #4221, opened 2022-04-08
feat(nuxt3): add `titleTemplate` (and viewport/charset shorthand)

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/3522, https://github.com/nuxt/framework/discussions/1768

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This adds `titleTemplate` support to Nuxt 3, as well as enabling `viewport` and `charset` shorthands (which were type-hinted to users but hitherto ineffective).

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #4260, opened 2022-04-11
fix(nuxt3): re-enable computed metadata in useHead

### LINKED ISSUE #4258: Ability to use computed in `useHead`
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-27494471.1091d45`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-onqjwc?file=app.vue

### Describe the bug

Using a computed to set as title is not reactive:

```ts
const firstname = ref('')
const lastname = ref('')
const fullname = computed(() => {
  return `${firstname.value} ${lastname.value}`.trim() || 'Account'
})
useHead({
  title: fullname.value,
})
```

- Open in a new window
- Update the firstname and/or lastname input
- The window title should be updated

Passing the `ref` will result with an `[object object]` as window title.

### Additional context

It only works when using the components: https://stackblitz.com/edit/nuxt-starter-wrxbyb?file=app.vue

If this is a normal behaviour, I believe this should be documented.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #4258

### ❓ Type of change

### 📚 Description

This resolves the regression for computed metadata (which worked before https://github.com/nuxt/framework/pull/4221).

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/head/runtime/lib/vueuse-head.plugin.ts
+++ b/packages/nuxt3/src/head/runtime/lib/vueuse-head.plugin.ts
@@ -17,22 +17,24 @@ export default defineNuxtPlugin((nuxtApp) => {
 
   const titleTemplate = ref<MetaObject['titleTemplate']>()
 
-  nuxtApp._useHead = (meta: MetaObject | ComputedGetter<MetaObject>) => {
-    titleTemplate.value = (unref(meta) as MetaObject).titleTemplate || titleTemplate.value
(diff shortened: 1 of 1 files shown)


Atinux (approved):
I guess it is too early to add a test case for it?

PR #4295, opened 2022-04-12
fix(nuxt3): as backward compatible runtime config for server side

Atinux:
### 🔗 Linked issue

Following https://github.com/nuxt/framework/pull/4254, proxy is not working on SSR

### ❓ Type of change

### DIFF
--- a/packages/nuxt3/src/app/nuxt.ts
+++ b/packages/nuxt3/src/app/nuxt.ts
@@ -119,33 +119,35 @@ export function createNuxtApp (options: CreateOptions) {
   }
 
   // Expose runtime config
+  let runtimeConfig
   if (process.server) {
-    nuxtApp.provide('config', options.ssrContext.runtimeConfig)
+    runtimeConfig = options.ssrContext.runtimeConfig
     // Client's runtime-config
     nuxtApp.payload.config = {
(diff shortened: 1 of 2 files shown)


packages/nuxt3/src/app/nuxt.ts
```diff
@@ -119,33 +119,35 @@ export function createNuxtApp (options: CreateOptions) {
   }
 
   // Expose runtime config
+  let runtimeConfig
```
pi0, line 122: Style: We could use inline branching for server/client.

packages/nuxt3/src/app/nuxt.ts
```diff
@@ -118,35 +118,39 @@ export function createNuxtApp (options: CreateOptions) {
...
+        return target.public
+      }
+      return target[prop] ?? target.public[prop]
+    },
+    set (target, prop, value) {
+      if (prop === 'public' || prop === 'app') {
+        return false // Throws TypeError
+      }
+      target[prop] = value
+      target.public[prop] = value
```
pi0, line 147: Server config is frozen. We need to disallow this (have to check)

Atinux: What about adding `if (process.server) { return false }` ?

PR #4298, opened 2022-04-12
fix(nuxi): don't use glob

danielroe:
### ❓ Type of change

### 📚 Description

If we can't resolve `glob` from `nuxi` entrypoint, it will throw even though we're not using the feature in `clearDir`. this disables it: see https://github.com/isaacs/rimraf#options


pi0:
Lets drop it at all :) #4308

Issue #4290, opened 2022-04-12
Incorrect #nitro alias type in nuxt v3.0.0-27495976.e31c604


TrungRueta:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.0`
- Nuxt Version:     `3.0.0-27495976.e31c604`
- Package Manager:  `npm@8.3.1`
- Builder:          `vite`
- User Config:      `typescript`, `buildModules`, `components`, `windicss`, `autoImports`, `router`, `experimental`, `features`, `globalName`, `build`, `vite`, `loading`, `cli`, `alias`, `privateRuntimeConfig`, `publicRuntimeConfig`, `css`
- Runtime Modules:  `-`
- Build Modules:    `@vueuse/nuxt@8.2.5`, `nuxt-windicss@2.2.10`
------------------------------

### Reproduction

- install new nuxt with v
- open new ts file (server dir) then try import `useRuntimeConfig` from `#nitro`.
- see error about missing type .

<img width="1181" alt="image" src="https://user-images.githubusercontent.com/12226007/162958241-0ec2df9f-1a9c-4f43-b751-51c6536ef5f1.png">

### Describe the bug

Look like bug somewhere block system get `d.ts` file, because as i checked, we had definition file `../index.d.ts` along with `../index.mjs`.

### Additional context

For anyone using this version and get same issue, this is workaround:

- open or create new file in root project named : `shim.d.ts`
- insert snippet to teach js get correct d.ts file

```typescript
// fix nuxt 3 issue missing type hint in [#nitro]
declare module '#nitro' {
  export * from 'nitropack/dist/runtime/index.d'
}
```

pi0:
Hi. You can simply drop `#nitro` import. `useRuntimeConfig` is globally available.

danielroe:
@TrungRueta Would you create a new issue with a reproduction please? :pray

TrungRueta:
@danielroe Here you go, thank for take care issue <3

https://stackblitz.com/edit/github-fjh4sx?file=server%2Fapi%2Fapi.ts

PR #4300, opened 2022-04-12
fix(nuxi): strip extensions from absolute paths (for types)

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/4290

### ❓ Type of change

### 📚 Description

As Nitro path was absolute, it alone was not having extension stripped, which mean TS couldn't resolve

### 📝 Checklist


pi0:
(PR is pending since its main effect on the framework is enabling `#nitro` usage but needs to be converted into `#internals/nitro`. In the meantime, global auto imports are available)

Issue #3649, opened 2022-03-14
`<NuxtLink>` prefetch support


pi0:
Ref: #2086, #3544

Smartly prefetch next pages with `<NuxtLink>`.

Considerations:
- Support payload prefetching for static routes
- Lower priority and disable for slow connections
- Consider using Link Prefetching

References:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ
Nuxt 2 prefetch: https://github.com/nuxt/nuxt.js/blob/dev/packages/vue-app/template/components/nuxt-link.client.js

Atinux:
I would also call a hook from it to let plugins & modules prefetch also content when the link is prefetching:

```ts
callHook(’app:link:prefetch’, link) 
```

danielroe:
To collect some more ideas:

* I think there is also space for doing build-time optimisation for prefetching, perhaps updating the generated manifest so we can use it when rendering initial page HTML (e.g. rather than prefetching all possible dynamic imports, we prioritise pages linked to from the page).
* with `nuxt/scripts` we can provide modules or users a way to add scripts with automatic preconnect, prefetching, etc. as required (maybe providing a hook as @Atinux suggests?)

pi0:
@Atinux Regarding the `callHook(’app:link:prefetch’, link)` idea, I fully agree we need to allow a programmatic way to do this but why not expose `nuxtApp.prefetch(link)` or `prefetchLink(link)` instead? 

Hooks are usually for actually _hooking_ into a lifecycle event. It makes sense if we define a life event that starts prefetching for page something like this:

```js
// Calls on initial and navigations
nuxtApp.hook('app:prefetch', (links: string[], prefetch: ((link) => Promise<void>)) => {
  // links is an array of links to be prefetched
  // prefetch utility is provided to request more prefetches
})
```

pi0:
@danielroe Manifest idea is nice but it probably needs a full prerendering stage to have predicates. Since nitro is already allowing use of `X-NITRO-PREFETCH`, maybe we can enhance when prefetching next pages to respect this header and prefetch all of relevant links of next?

Example:

- User loads `/profiles`
- We observe links to `/user/anna` and `/user/bob` (NuxtLink prefetch)
- For prefetching we use background fetch to both `/user/{anna/bob}`
- Page handler for `/user/:name` adds a prefetch header `X-NITRO-PREFETCH: '/img/users/:name.png'`
[...]

PR #4329, opened 2022-04-13
feat(nuxt): support prefetching `<nuxt-link>`

Mini-ghost:
### 🔗 Linked issue

#3649 #2086

### ❓ Type of change

### 📚 Description

Support `prefetch` prop for `<nuxt-link>`. This implementation is refer to https://github.com/nuxt/framework/discussions/2086#discussioncomment-1683819. Need a polyfill for `requestIdleCallback` ?

Also, I'm not sure how to use Link Prefetching like below, because `path` are unpredictable.

```html 
<link rel="prefetch" as="script" href="path" />
````

The implementation of this PR refers to nuxt2, maybe we can generate a router manifest. and mapping file path when triggering prefetch  ? 

### 📝 Checklist

### CODE REVIEW

**Comment by Mini-ghost** [comment]
@pi0 @Atinux I will add polyfill for `requestIdleCallback`. And I think we have two opportunities to add polyfill

1. Add polyfill at file start and run in client side.

```ts
const requestIdleCallback = process.client
  ? window.requestIdleCallback || function (cb) {
    const start = Date.now()
    const idleDeadline = {
      didTimeout: false,
      timeRemaining () {
        return Math.max(0, 50 - (Date.now() - start))
      }
    }
    return window.setTimeout(function () {
      cb(idleDeadline)
    }, 1)
  }
  : (() => {}) as any as Window['requestIdleCallback']
 ```

2. When component mounted and need prefetch, implement with Singleton.

```ts
let requestIdleCallback: Window['requestIdleCallback'] | null = null
```

```ts
onMounted(() => {
  if (!shouldPrefetch.value) { return }

  // add `requestIdleCallback` polyfill if not support and need prefetch
  if (!requestIdleCallback) {
    requestIdleCallback = window.requestIdleCallback || function (cb) {
      const start = Date.now()
      const idleDeadline = {
        didTimeout: false,
        timeRemaining () {
          return Math.max(0, 50 - (Date.now() - start))
        }
      }
      return window.setTimeout(function () {
        cb(idleDeadline)
      }, 1)
    }
  }
})
```

Which is the preferred approach?

danielroe:
@Mini-ghost The first approach seems better to me 👍

danielroe (approved):
Thank you! Looks good to me but @pi0 may have some more ideas for improvement.

pi0 (approved):

pi0:
Thank you so much for helping on this feature @Mini-ghost <3

I've made few more reactors to the PR to simplify do small fixes and also integrated with new payload prerendering using `link:prefetch` hook to be ready to land.

PR #4332, opened 2022-04-13
docs(api): apply use-async-data page structure to use-fetch

clemcode:
### ❓ Type of change

### 📚 Description
This PR applies the same structure to `useFetch` API reference than to `useAsyncData`:
- Type
- Params
- Return Values
- Example


clemcode:
Not sure about the function and options signature, I based myself on the previous documentation but clearly need a review on that.

pi0 (approved):

PR #2772, opened 2022-01-18
feat(nuxi): call nuxt `listen` hook for dev

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/2499

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [] ✨ New feature (a non-breaking change that adds functionality)
- [] ⚠️ Breaking change (fix or feature that would cause existing functionality to change) 

### 📚 Description

The listen hook is a feature from nuxt v2 that is typed in nuxt v3. but nothing calls it (as far as I can tell).

It's a useful hook to expose the server used for local development, giving module authors functionality to see the server address and hook into the server if needed.

Specifically a module I'm working on needs it, this is the equivalent nuxt 2 code.

```ts
nuxt.hook('listen', (ctx) => {
  setServerContext({
    url: ctx.listeners[0].url,
    server: ctx.listeners[0].server,
    app: ctx.app,
  })
})
```

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly. - **no hook documentation section, needs it's own page**

PR #4346, opened 2022-04-14
fix(nuxi): call listen on each nuxt restart

### LINKED ISSUE #3835: Listen hook not called again on hot reload
### Environment

Nuxt project info:                                                                                                                                                                                       14:32:14

- Operating System: `Darwin`
- Node Version:     `v14.18.2`
- Nuxt Version:     `3.0.0-27404337.c9c0171`
- Bundler:          `Vite`
- User Config:      `privateRuntimeConfig`, `buildModules`, `googleFonts`, `css`, `vite`, `ngrok`
- Runtime Modules:  `-`
- Build Modules:    `~/modules/google-fonts`, `~/modules/ngrok`

👉 Report an issue: https://github.com/nuxt/framework/issues/new                                                                                                                                         14:32:14

✨  Done in 2.68s.

### Reproduction

```ts
import ngrok from "ngrok";
import type { Ngrok } from "ngrok";
import { defineNuxtModule } from "@nuxt/kit";
import consola from "consola";

// https://github.com/bubenshchykov/ngrok
export type ModuleOptions = Partial<Ngrok.Options>;
declare module "@nuxt/schema" {
  interface NuxtConfig {
    [CONFIG_KEY]?: ModuleOptions;
  }
}
const CONFIG_KEY = "ngrok";

  meta: {
    // Usually  npm package name of your module
    name: "ngrok",
    // The key in `nuxt.config` that holds your module options
    configKey: CONFIG_KEY,
    // Compatibility constraints
    compatibility: {
      nuxt: "^3.0.0",
    },
  },
  // Default configuration options for your module
  defaults: {},
  hooks: {},
  async setup(moduleOptions, nuxt) {
    nuxt.hook("listen", async () => {
      consola.success("I am listening...");
    });

    // Disconnect ngrok connection when closing nuxt
    nuxt.hook("close", () => {
      consola.log("I have stopped listening");
    });
  },
});

```

### Describe the bug

In my ngrok module that I'm working on for Nuxt 3, I need to listen to the `listen` hook and then connect to ngrok.

When I save my file, I see that the server s restarted because I get the local and network URLs in my console, but the listen event is not fired. I close the ngrok connection on the `close` hook just as is done in the Nuxt 2 version of the ngrok module.

If you re-save the nuxt config file you'll see the disconnect/close happen and the listen hook never fired again.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #3835

### ❓ Type of change

### 📚 Description

This is a tricky one. Current behaviour is that the `listen` hook only runs once, and it's called by `nuxi`. Even when Nuxt restarts, the same listener persists, so on one level, it makes sense to call it once.

However, modules and other nuxt context re-initialises so it probably makes sense to re-call this hook on each reinitialisation of nuxt.

We might also consider adding functionality to allow users to hook into the Nitro dev server, which might mean this PR is unneeded.

Context: https://github.com/nuxt/framework/pull/2772

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -61,6 +61,7 @@ export default defineNuxtCommand({
         }
         currentNuxt = await loadNuxt({ rootDir, dev: true, ready: false })
         await currentNuxt.ready()
+        await currentNuxt.hooks.callHook('listen', listener.server, listener)
         await Promise.all([
           writeTypes(currentNuxt).catch(console.error),
           buildNuxt(currentNuxt)
@@ -105,9 +106,6 @@ export default defineNuxtCommand({
     })
 
     await load(false)
-    if (currentNuxt) {
-      await currentNuxt.hooks.callHook('listen', listener.server, listener)
-    }
 
     return 'wait'
   }


pi0 (approved):
LGTM. Modules shouldn't really depend or use this hook and have server logic in runtime but at least is a good backward compatibility feature

PR #4372, opened 2022-04-15
test: init hmr testing

antfu:
To make it simple, I used git to manage the version reset instead of copying the fixture. It requires no unstaged changes to the fixture when running the HMR tests which I guess could be acceptable?


test/basic.test.ts
```diff
@@ -370,3 +373,59 @@ describe('dynamic paths', () => {
...
+      expect((await page.$('.sugar-counter').then(r => r.textContent())).trim())
+        .toEqual('Sugar Counter 13 x 2 = 26  Inc')
+
+      // modify file
+      let indexVue = await fsp.readFile(join(fixturePath, 'pages/index.vue'), 'utf8')
+      indexVue = indexVue
+        .replace('<Title>Basic fixture</Title>', '<Title>Basic fixture HMR</Title>')
+        .replace('<h1>Hello Nuxt 3!</h1>', '<h1>Hello Nuxt 3! HMR</h1>')
+      indexVue += '<style scoped>\nh1 { color: red }\n</style>'
+      await fsp.writeFile(join(fixturePath, 'pages/index.vue'), indexVue)
```
pi0, line 405: Instead of directly writing to fixture dir, we can use a temporary directory (gitignored) and `extends` option.

antfu: Would that be equivalent to user operation on changing the file to trigger HMR?

pi0: Yes. All dirs behave with the same logic.

antfu:
Disabled the test on Windows for now so we can move forward.

danielroe (approved):

Issue #929, opened 2021-10-13
Webpack: SCSS Styles doesn't preload in dev


proskurindima:
### Environment

Local:
------------------------------
- Operating System: `Linux`
- Node Version:     `v14.17.1`
- Nuxt Version:     `3-3.0.0-27234503.d5127e9`
- Package Manager:  `Yarn`
- Bundler:          `Webpack`
- User Config:      `modules`, `styleResources`, `css`
- Runtime Modules:  `@nuxtjs/style-resources@1.2.1`
- Build Modules:    `-`
------------------------------

### Describe the bug

When I use SCSS with Webpack, styles doesn't preload.
You can see it in [codeSanbox](https://codesandbox.io/s/ecstatic-field-zvuql?file=/nuxt.config.ts)

### Reproduction

1. Add node-sass, sass and sass-loader packages in your package.json
2. Disable Vite in nuxt.config file. vite=false
3. Add some styles in <style> tag with attr lang="scss". For example:
```
<style lang="scss">
.test {
  font-size: 24px;
  font-family: sans-serif;
}
</style>
```
4. Reload page
5. See, that styles doesn't preload. They've applied after small delay

### Additional context

I've reproduced this error in [codeSanbox](https://codesandbox.io/s/ecstatic-field-zvuql?file=/nuxt.config.ts).

Also, this problem happens with Vite. There is one more [ticket](https://github.com/nuxt/framework/issues/905)

danielroe:
At the moment inlining styles is not supported for webpack + vue-loader: https://github.com/vuejs/vue-style-loader/issues/56.

danielroe:
This is no longer reproducible for me. Let me know if not and I'll happily reopen.

PR #4388, opened 2022-04-15
fix(webpack): set `extractCSS: true` by default

danielroe:
### 🔗 Linked issue

#929

### ❓ Type of change

### 📚 Description

`vue-loader` no longer supports extracting styles with `vue-style-loader` - see https://github.com/vuejs/vue-style-loader/issues/56.

This PR removes the dependency and sets `extractCSS` to true by default (which [is recommended](https://webpack.js.org/loaders/css-loader/#extract) in any case for production by webpack). If we are not able to inline styles then this is likely a better default, though it will be a breaking change when we re-introduce `vue-style-loader` later on.

### 📝 Checklist


danielroe:
Looking at https://github.com/pi0/vue-loader/commit/22894febd6eda43bce033553602b40f96d730182, this is not sufficient to enable us to inline styles, and we're already able to identify CSS files to preload when `extractCSS` is set to true. So I'm not sure what it would add. (Maybe it has been fixed upstream in vue-loader?)

I would suggest we merge this PR, as vite isn't inlining styles either. Then we can look to add a new feature to both vite + webpack later to inline critical CSS later on. wdyt @pi0?

syifeng:
It's not worked when I used vite , I am confused

danielroe:
Vite extracts CSS by default. You can turn off inlining styles by setting `experimental.inlineSSRStyles` to false in your nuxt config.

(In future please create a new issue rather than commenting on an old issue.)

PR #4422, opened 2022-04-19
fix(nuxt3)!: include route children in parent suspense

danielroe:
### ❓ Type of change

### 📚 Description

This wraps the suspense for child routes into it's parent, which means `page:finish` is now accurate (it means all route components have been fully resolved). However, it's a breaking change in that the hook is now only called _once_, with the top-level route component.

### 👉 Migration

The `page:start` and `page:finish` hooks are now only called _once_, with the top-level route component rather than any nested route components.

### 📝 Checklist


pi0:
@danielroe Would you mind updating breaking change tracker?

PR #4427, opened 2022-04-19
fix(nuxt3): initialise router with normalised url

### LINKED ISSUE #4417: Server-side rendering (SSR) of a custom Nuxt error page leads to Vue Router warning: No match found for location with path "/__nuxt_error?..."
### Environment

```
Nuxt CLI v3.0.0-27501460.04a72f8
RootDir: /home/projects/github-hf1mjb
Nuxt project info:

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-27501460.04a72f8`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
```

### Reproduction

https://stackblitz.com/edit/github-hf1mjb?file=pages/index.vue

### Describe the bug

Hello.

I get a Vue Router warning in the Terminal when a custom Nuxt error page is used:

```
[Vue Router warn]: No match found for location with path "/__nuxt_error?url=/test/404&statusCode=404&statusMessage=Route+Not+Found&message=+&description"
```

Best wishes,
Sergey.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #4417

### ❓ Type of change

### 📚 Description

`ssrContext.req.url` displays the full error handler URL rather than the cleaned-up version. This PR changes it to use the path that the error handler is invoked for.

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/pages/runtime/router.ts
+++ b/packages/nuxt3/src/pages/runtime/router.ts
@@ -82,7 +82,7 @@ export default defineNuxtPlugin((nuxtApp) => {
   }
 
   // Allows suspending the route object until page navigation completes
-  const path = process.server ? nuxtApp.ssrContext.req.url : createCurrentLocation(baseURL, window.location)
+  const path = process.server ? nuxtApp.ssrContext.url : createCurrentLocation(baseURL, window.location)
   const _activeRoute = shallowRef(router.resolve(path) as RouteLocation)
   const syncCurrentRoute = () => { _activeRoute.value = router.currentRoute.value }
   nuxtApp.hook('page:finish', syncCurrentRoute)


packages/nuxt3/src/pages/runtime/router.ts
```diff
@@ -82,7 +82,7 @@ export default defineNuxtPlugin((nuxtApp) => {
   }
 
   // Allows suspending the route object until page navigation completes
-  const path = process.server ? nuxtApp.ssrContext.req.url : createCurrentLocation(baseURL, window.location)
+  const path = process.server ? nuxtApp.ssrContext.url : createCurrentLocation(baseURL, window.location)
```
pi0, line 85: This is a good fix per se but also worries me that error URL could be properly encoded that this doesn't happens with URL mismatch.

danielroe: Do you have an example of what that might look like? We are using `withQuery` and `useQuery` which (if I understand rightly) _should_ make this an upstream issue to add tests for in `ufo`.

pi0: > should make this an upstream issue to add tests for in ufo.

ufo utils don't update formating and are based on non-encoded by default. This is kinda normal behavior.

> Do you have an example of what that might look like

URL we are generating for `/_error/*` fetch call can be encoded.

PR #4430, opened 2022-04-19
fix(nuxt3): improve redirect lifecycle

### LINKED ISSUE #4244: Cookie can't be set together with redirect

danielroe:
### 🔗 Linked issue

resolves #4244

### ❓ Type of change

### 📚 Description

This PR improves a number of things around cookies + redirect lifecycle:

1. it always sets a cookie value (not only if the value has changed) - let me know if I should revert this @pi0 but it seems that I would expect `useCookie('test', { default: () => 'value' })` to set a value on the initial SSR load.
2. it adds an additional `app:redirected` hook for setting final headers etc. in case of redirect (in which case there is no `app:rendered` to hook into
3. we use `navigateTo` helper within router plugins so there's a central place for setting the redirect
4. we accept an optional `redirectCode` via `navigateTo` options (defaulting to 301, which is a technical change from h3 default value of 302, but the same as the previous router plugin implementation)

### 📝 Checklist

### DIFF
--- a/packages/nuxt3/src/app/composables/cookie.ts
+++ b/packages/nuxt3/src/app/composables/cookie.ts
@@ -33,7 +33,11 @@ export function useCookie <T=string> (name: string, _opts?: CookieOptions<T>): C
     watch(cookie, () => { writeClientCookie(name, cookie.value, opts as CookieSerializeOptions) })
   } else if (process.server) {
     const nuxtApp = useNuxtApp()
-    const writeFinalCookieValue = () => writeServerCookie(useRequestEvent(nuxtApp), name, cookie.value, opts)
+    const writeFinalCookieValue = () => {
+      if (cookie.value !== cookies[name]) {
+        writeServerCookie(useRequestEvent(nuxtApp), name, cookie.value, opts)
+      }
+    }
     nuxtApp.hooks.hookOnce('app:rendered', writeFinalCookieValue)
     nuxtApp.hooks.hookOnce('app:redirected', writeFinalCookieValue)
   }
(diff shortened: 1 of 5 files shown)


pi0:
>  I would expect useCookie('test', { default: () => 'value' }) to set a value on the initial SSR load.

I expect the same but not adding a `Set-Cookie` header if `test` cookie (in `req.headers`) is already `value`.

pi0 (approved):

PR #4438, opened 2022-04-19
feat!: migrate to nitro 0.3x

pi0:
### ❓ Type of change

### 📚 Description

Breaking change: `#nitro` is no longer available (Replaced with `#internal/nitro`). For useRuntimeConfig, you can use global version using auto import.

### 📝 Checklist


Atinux:
We agree it has to be renamed here @pi0 ? https://github.com/nuxt-community/supabase-module/blob/main/src/runtime/server/api/session.ts#L2

Does the auto-import works for modules exposing runtime server API?

pi0:
No auto import ignores anything in `node_modules`. I've added back `#nitro` for backward compatibility. It shouldn't be a breaking change anymore but please use `#imports` when possible. They are aliases.

Atinux:
Will update to use `#imports` :+1:

PR #4443, opened 2022-04-19
docs: Remove useless -- in getting started docs

gris:
### ❓ Type of change

### 📚 Description

WIth npm the command works perfectly fine without the added `--`. So it is better to simplify it.

### 📝 Checklist

### DIFF
--- a/docs/content/1.getting-started/1.quick-start.md
+++ b/docs/content/1.getting-started/1.quick-start.md
@@ -90,7 +90,7 @@ npm run dev -o
 ```
 
 ```bash [pnpm]
-pnpm run dev -o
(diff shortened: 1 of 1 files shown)


Atinux:
Actually it does not work, the `--` is used by NPM to forward all the other arguments to the command ran.

gris:
Hey, @Atinux . Thank you very much for your reply. It does work fine for me. I'm using npm version `8.5.0`.

gris:
Oh, thank you very much. I didn't noticed that the intended argument was not being forwarded to the command, I've only checked if there was an error. Sorry for that.

PR #4453, opened 2022-04-20
chore: setup IssueUp

antfu:
https://github.com/antfu/issue-up

To enable it, we need to install [the app](https://github.com/apps/issue-up) for both Nuxt and UnJS org


antfu:
Oh, I didn't notice that. Should be fixed now. Thanks

PR #4459, opened 2022-04-20
fix(schema): preserve top-level undefined keys in runtime config

danielroe:
### 🔗 Linked issue

#4458

### ❓ Type of change

### 📚 Description

This is a behaviour of `defu` that we have previously resolved in https://github.com/unjs/nitropack/pull/28 and https://github.com/nuxt/framework/pull/2456, but which has regressed...

It will also need to be paired with a PR to nitro and c12.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -720,14 +720,19 @@ export default {
    * @version 3
    */
   runtimeConfig: {
-    $resolve: (val: RuntimeConfig, get) => defu(val, {
+    $resolve: (val: RuntimeConfig, get) => ({
       ...get('publicRuntimeConfig'),
       ...get('privateRuntimeConfig'),
-      public: get('publicRuntimeConfig'),
+      ...val || {},
+      public: {
+        ...get('publicRuntimeConfig'),
+        ...val?.public || {},
+      },
       app: {
         baseURL: get('app').baseURL,
         buildAssetsDir: get('app').buildAssetsDir,
         cdnURL: get('app').cdnURL,
+        ...val?.app || {},
       }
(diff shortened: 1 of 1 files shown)


packages/schema/src/config/_common.ts
```diff
@@ -720,14 +720,19 @@ export default {
    * @version 3
    */
   runtimeConfig: {
-    $resolve: (val: RuntimeConfig, get) => defu(val, {
```
pi0, line 723: Have you e2e tested this? c12 and nitro use default multiple times.

pi0: Also, we have a chance to map all `NITRO_` and `NUXT_` values into runtime config without explicit definition.

danielroe: Yes - see PR description. It will need PRs to c12 & nitro. Issue is only at _top-level_ IIRC so nested values with undefined are fine.

Tricky thing about automatic mapping is that it doesn't work both ways (ie. given a value in the config object, we can find the env variable name, but not vice versa).

For example, NUXT_BASE_URL maps to all of the following:

```js
{
  base: {
    url: ''
  },
  baseURL: '',
  baseUrl: '',
  base_url: '',
  BaseUrl: '',
  // etc.
}
```

danielroe: We can alternatively _require_ fallback values in runtimeConfig?

pi0: I think we can make a convention work as top-level camelCase. using `undefined` as default is not a solution I would keep forever... But let's merge it in last minutes.

Issue #2278, opened 2021-12-03
Response status code is not 404 but 200 with custom pages/404.vue


levchak0910:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v14.16.0`
- Nuxt Version:     `3.0.0-27265876.3cd4494`
- Package Manager:  `npm@7.17.0`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://stackblitz.com/edit/github-tpmisr

### Describe the bug

When using a `pages/` directory and navigating to non existent page, result is a default nuxt 404 page and **statusCode = 404**. This is expected
After adding a custom `pages/404.vue` page, the content of the page is returned as expected, **but statusCode = 200** which is not _expected*_

### Additional context

If this behavior by design, how _expected*_ behavior could be achieved?

levchak0910:
One more interesting thing
If forcefully set statusCode to 404, custom `pages/404.vue` is ignored and shown default nuxt 404 page

```javascript
// pages/404.vue
setup() {
    if(process.server) {
      const nuxtApp = useNuxtApp()
      const error = new Error()
      error.statusCode = 404
      nuxtApp.ssrContext.error = error
    }
  },
```

danielroe:
Agreed - let's track this in https://github.com/nuxt/framework/discussions/559.

maxdzin:
@danielroe I'm still getting the same 200 rather than 404. It still not yet corrected or it's required some additional things to make it work?

7ute:
@maxdzin @levchak0910 This works for me, effectively setting the right status code (nuxt `3.0.0-27338323.1e98259`)
```diff
setup() {
    if (process.server) {
        const nuxtApp = useNuxtApp();
        const error = new Error();
        error.statusCode = 404;
-       nuxtApp.ssrContext.error = error;
+       nuxtApp.ssrContext.nuxt.error = error;
    }
},
```

maxdzin:
> 

@7ute Great, thanks! I thought that it is supposed to return a `404` error automatically. :)

PR #4460, opened 2022-04-20
fix(nuxt): return 404 if hitting `404.vue`

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/2278

### ❓ Type of change

### 📚 Description

Given that we support a custom `404.vue` as a catchall response, we should set the status code to be 404 as well.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/runtime/router.ts
+++ b/packages/nuxt/src/pages/runtime/router.ts
@@ -161,6 +161,8 @@ export default defineNuxtPlugin((nuxtApp) => {
           statusCode: 404,
           statusMessage: `Page not found: ${to.fullPath}`
         })])
+      } else if (process.server && to.matched[0].name === '404' && nuxtApp.ssrContext) {
+        nuxtApp.ssrContext.res.statusCode = 404
       }
     })
 


packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -161,6 +161,8 @@ export default defineNuxtPlugin((nuxtApp) => {
           statusCode: 404,
           statusMessage: `Page not found: ${to.fullPath}`
         })])
+      } else if (process.server && to.matched[0].name === '404' && nuxtApp.ssrContext) {
```
pi0, line 164: Is this even a documented feature?

danielroe: It's not documented, but it is implemented:

https://github.com/nuxt/framework/blob/4712e994112a97feba447b3edb4afb65e968ed63/packages/nuxt/src/pages/utils.ts#L80-L81

and tested:

https://github.com/nuxt/framework/blob/4712e994112a97feba447b3edb4afb65e968ed63/packages/nuxt/test/pages.test.ts#L8-L19

PR #4473, opened 2022-04-20
docs: add link to hybrid rendering RFC

Atinux:
### ❓ Type of change

### 📚 Description

Add link to our hybrid rendering RFC

### DIFF
--- a/docs/content/2.guide/5.deploy/2.static-hosting.md
+++ b/docs/content/2.guide/5.deploy/2.static-hosting.md
@@ -1,11 +1,7 @@
 # Static Hosting
 
 ::alert{type=warning icon=🚧}
-This feature is still experimental and under development for Nuxt 3.
-::
-
-::alert{type=info icon=👀}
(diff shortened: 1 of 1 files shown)


docs/content/2.guide/5.deploy/2.static-hosting.md
```diff
@@ -4,6 +4,10 @@
 This feature is still experimental and under development for Nuxt 3.
 ::
 
+::alert{type=info icon=👀}
+We are looking for feedback on the [Hybrid Rendering (route caching rules) RFC](https://github.com/nuxt/framework/discussions/560)
```
pi0, line 8: It is not available yet for collecting feedback. Nitro support is super experimental and dummy :D We can do it by next RC with features like swr, static, redirect, proxy, etc.

Also I think this can be more fit into a guide or concept for "Hybrid Rendering"

pi0: BTW i would add a notice about full-static generation with payload extraction.

Atinux: Updated to remove RFC link and add a note about full static.

PR #4476, opened 2022-04-20
feat(kit): accept function in `inlineOptions` and `configOptions`

ricardogobbosouza:
### ❓ Type of change

### 📚 Description

Allow module options to accept functions

### 📝 Checklist


pi0:
Thanks for the PR @ricardogobbosouza. Can you please share an example?

ricardogobbosouza:
I'm working on nuxt3 support in robots module

And this option with function will not work, without this pr

https://github.com/nuxt-community/robots-module#function

pi0:
Hi @ricardogobbosouza. Sorry for checking on this late!

Adding this feature is probably not generic enough. We are not awaiting on function result and the function might need more context utils from the module other than `nuxt` instance itself.

Regarding robots and Nuxt 3: We are moving away from runtime logic like this from build time (and config) to the runtime. This module can use hooks (like `robots:sitemap`) or an [async] callback option like (`robots: { sitemap: async () => [] }`) to keep semi compatibility with Nuxt 2 usage but also a runtime handler to render sitemap on demand in runtime (or prerender for static). Would be more than happy to have a chat in discord about this.

Issue #4217, opened 2022-04-08
Failed to resolve dependencies when starting Nuxt under workspaces


eden:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v17.8.0`
- Nuxt Version:     `3.0.0-27490192.4253d2e`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

1. Create a minimal workspace: `echo '{"private": true, "workspaces": ["app"]}' > package.json`
2. Create the starter nuxt 3 app inside the same directory: `npx nuxi init app`
3. Update `app/package.json` to add a name and version (since yarn 1.x requires this in its workspaces).
    * `sed -i '' 's/"private"/"name":"app","version":"1.0.0","private"/' app/package.json`
4. The cwd should look like this:
```
./app
./app/app.vue
./app/README.md
./app/.gitignore
./app/package.json
./app/nuxt.config.ts
./app/tsconfig.json
./package.json
```
6. Run `yarn` to setup the workspace
7. Test out the workspace's dev: `yarn workspace app dev` -- this works
8. Build: `yarn workspace app build` -- this fails (see logs)

### Describe the bug

Unable to build a nuxt 3 app within a yarn 1.x workspace. Development works fine. Build will also work if I nuke the workspace's `node_modules` and link it directly to the parent directory's `node_modules`.

### Logs

```shell
Nuxt CLI v3.0.0-27490192.4253d2e                                                                              14:58:13
ℹ Client built in 659ms                                                                                       14:58:15
ℹ Vite client warmed up in 661ms                                                                              14:58:15
ℹ Building server...                                                                                          14:58:15
✔ Server built in 241ms                                                                                       14:58:15
✔ Generated public .output/public                                                                       nitro 14:58:15
start Building server...                                                                                nitro 14:58:15

 WARN  '@nuxt/devalue' is imported by ../node_modules/nuxt3/dist/core/runtime/nitro/renderer.mjs, but could not be resolved – treating it as an external dependency


 WARN  '@vue/runtime-dom' is imported by @vue/runtime-dom?commonjs-external, but could not be resolved – treating it as an external dependency


 WARN  '@vue/shared' is imported by @vue/shared?commonjs-external, but could not be resolved – treating it as an external dependency


[...]

PR #4488, opened 2022-04-21
fix(nuxt): resolve `vue/server-renderer` and `vue/compiler-sfc`

antfu:
### 🔗 Linked issue

Partially fix https://github.com/nuxt/framework/issues/4217

It will be fixed by the combination of this PR and https://github.com/unjs/nitro/pull/136

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -3,10 +3,10 @@ import { resolve, join } from 'pathe'
 import { createNitro, createDevServer, build, prepare, copyPublicAssets, writeTypes, scanHandlers, prerender } from 'nitropack'
 import type { NitroEventHandler, NitroDevEventHandler, NitroConfig } from 'nitropack'
 import type { Nuxt } from '@nuxt/schema'
-import { resolvePath } from '@nuxt/kit'
 import defu from 'defu'
 import fsExtra from 'fs-extra'
 import { toEventHandler, dynamicEventHandler } from 'h3'
+import { resolvePath } from 'mlly'
 import { distDir } from '../dirs'
 import { ImportProtectionPlugin } from './plugins/import-protection'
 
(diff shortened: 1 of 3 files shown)


pi0 (approved):
LGTM.

packages/nuxt/src/core/nitro.ts
```diff
@@ -64,8 +64,8 @@ export async function initNitro (nuxt: Nuxt) {
     },
     alias: {
       // TODO: #590
-      'vue/server-renderer': 'vue/server-renderer',
-      'vue/compiler-sfc': 'vue/compiler-sfc',
+      'vue/server-renderer': await resolvePath('vue/server-renderer'),
```
pi0, line 67: Can you please double-check if it resolves to the _esm_ entry?

antfu: Nice catch! Fixed

packages/nuxt/src/core/nitro.ts
```diff
@@ -64,8 +64,8 @@ export async function initNitro (nuxt: Nuxt) {
     },
     alias: {
       // TODO: #590
-      'vue/server-renderer': 'vue/server-renderer',
-      'vue/compiler-sfc': 'vue/compiler-sfc',
+      'vue/server-renderer': await resolvePath('vue/server-renderer', { conditions: ['import'] }),
```
pi0, line 67: I guess mlly resolver should apply import condition by default following module resolution. But we can path `url` (as node modules dir search path to it so that if a newer version is specified by users will be picked, same as `resolvePath` in kit)

pi0:
@antfu Can you please check if we still need this with last version of nitro with node_modules fix?

antfu:
I can confirm it's fixed by upgrading nitropack to v0.3.7. Awesome!

PR #4187, opened 2022-04-07
fix: nitropack fixes and upgrade to 0.2.x

pi0:
Fixes regarding nitropack:

- avoid overriding nitro's output dir (resolves #4179)
- provide nuxt aliases (resolves #4173)
- update nitropack to 0.2.x (related to #4177) inlining nuxt related stuff

PR #4500, opened 2022-04-21
fix(nuxt): add `app:documentTemplate` to replace old `nitro:document`…

danielroe:
### ❓ Type of change

### 📚 Description

This PR adds back the ability to manipulate the document template removed in https://github.com/nuxt/framework/pull/4187 as part of nitropack migration.

`nitro:document` is currently used in `@nuxtjs/color-mode` and this kind of functionality could be useful in other modules too. (This could also be resolved with another way to manipulate what is rendered in `ssr: false`.)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -88,9 +88,9 @@ const defaultAppTemplate = `
 export const appViewTemplate = {
   filename: 'views/document.template.mjs',
   write: true,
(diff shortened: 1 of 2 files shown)


pi0:
Directly manipulation of document template (which is runtime) seems wrong practice to me. I would introduce a runtime renderer hook and use nitro plugin to manipulate at whatever stage is needed.

danielroe:
Seems reasonable 👍 Will follow up with that approach.

PR #4509, opened 2022-04-21
feat: allow disabling sourcemap generation

danielroe:
### ❓ Type of change

### 📚 Description

This PR disables sourcemap generation for our various transform plugins when building. (I would also be open to adding a `build.sourceMap` option that allows configuring sourcemap generation.)

I've seen one or two reports of memory issues when building, and have previously resolved this with turning off magic string sourcemaps. For context:

* https://github.com/nuxt/bridge/issues/151
* https://github.com/rollup/plugins/issues/1064
* https://github.com/Rich-Harris/magic-string/issues/195

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/build.ts
+++ b/packages/schema/src/config/build.ts
@@ -23,6 +23,15 @@ export default {
     },
   },
   /**
+   * Whether to generate sourcemaps.
+   *
+   * By default, this option is enabled in development.
+   * @version 3
+   */
+  sourcemap: {
+    $resolve: (val, get) => val ?? get('dev')
+  },
+  /**
    * Shared build configuration.
    * @version 2
    * @version 3
@@ -132,7 +141,7 @@ export default {
      * @version 2
      */
     cssSourceMap: {
-      $resolve: (val, get) => val ?? get('dev')
+      $resolve: (val, get) => val ?? get('sourcemap') ?? get('dev')
     },
 
     /**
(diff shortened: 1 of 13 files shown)


pi0:
Do you have any context of the source map memory leak issue? It can happen when enabling development too and we can't just _mask_ the issue this is a nice feature for production build too.

packages/schema/src/config/build.ts
```diff
@@ -22,6 +22,15 @@ export default {
...
     },
   },
+  /**
+   * Whether to generate sourcemaps.
+   *
+   * By default, this option is enabled in development.
+   * @version 3
+   */
+  sourcemap: {
+    $resolve: (val, get) => val ?? get('dev')
```
pi0, line 32: As we discussed, we probably shouldn't _mask_ the memory usage issue otherwise will never find out what is wrong especially since we don't even have a reliable reproduction now and are enabling it for development. Can we instead make sure there is this option and suggest to disable as a workaround for users having trouble with memory?

danielroe: If you like - though I do think after discovering cause of issue in `magic-string`, we should disable this by default when building.

pi0: Even if we could have production sourcemap support for server logs? :)

danielroe: oh no, in _that_ case I'm on board 😁

Issue #4520, opened 2022-04-22
allow manually redirecting to 404 route within a dynamic slug


misaon:
### Environment

```
------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-rc.1`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------
```

### Reproduction

https://stackblitz.com/edit/github-yyumyt?file=pages/404.vue

### Describe the bug

I have defined three pages (routes):
- `car.vue`
- `[slug].vue`
- `404.vue`

If I go to `http://.../hi` => it loads `[slug].vue` => **OK**
If I go to `http://.../car` => it loads `car.vue` => **OK**
If I go to `http://.../404` => it loads `[slug].vue` => **FAIL**

danielroe:
`404.vue` is a special route that handles otherwise unmatched routes: see https://v3.nuxtjs.org/guide/directory-structure/pages#catch-all-route.

misaon:
If I have logic in setup inside [slug].vue where I call useFetch and the api returns 404, I would like to redirect to the /404 route which is not possible now.

misaon:
@danielroe Example: I have a bike shop. A customer accesses the route: http//bike.com => index.vue. The customer wants to view the bike categories http://bike.com/blue-bikes => [slug].vue. The customer selects one of the bikes from the category, http://bike.com/best-blue-bike.html => [slug].html.vue. Everything works great, but how do I catch the condition where I want to redirect to a 404 page?

danielroe:
> which is not possible now.

Are you saying this used to work? Would you provide an example of what you're wanting to do?

(I'm asking as this is a key behaviour and we need to make this as straightforward as we can, so we either need to improve DX or document some examples.)

Aareksio:
I am positive the broad idea of this issue is to ask "how to handle 404 on dynamic path".

```js
  const $route = useRoute()
  const { data } = await useAsyncData(`product:${$route.params.id}`, () => $fetch(`/api/products/${$route.params.id}`))

  // TODO: Display proper 404 when accessing unknown product id
  // Requirements:
  // - Using default error page - big "404" with small "Page not found: /42" - just like native nuxt 404
  // - On all types of navigation:
  //   a) Home -> Product/42 (client)
[...]

PR #4537, opened 2022-04-22
feat(pages)!: explicitly allow optional params with `[[`

danielroe:
Resolves https://github.com/nuxt/framework/issues/3352. Also see https://github.com/nuxt/framework/issues/4520.

### ❓ Type of change

### 📚 Description

Currently routes _ending_ in a dynamic param are marked as optional, e.g.: `~/pages/[slug].vue`.

This PR instead requires double bracket syntax to enable optional params: e.g. `~/pages/[[slug]].vue`. (This also means you can enable optional parameters earlier in the path.)

### 👉 Migration

If you are relying on `[slug]` at the end of a route to generate an optional parameter, you will need to rewrite it as `[[slug]]`.


danielroe:
@pi0 How about we change the default to non-optional at end of route, and use double brackets - e.g. `[[slug]]` - to indicate optionality, which can apply to any level in path?

pi0:
Double bracket seems a nice idea!

Atinux:
I love this idea!

Atinux:
Happy to merge once the documentation is updated. Any blocked @pi0 ?

pi0 (approved):

Issue #4091, opened 2022-04-05
Client side error is not displaying error page


Aareksio:
### Environment

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v16.13.2`
- Nuxt Version:     `3.0.0-27485997.c426f3e`
- Package Manager:  `npm@8.5.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

Repo:
https://github.com/Aareksio/nuxt3-issues/tree/error-handling

Sandbox:
https://codesandbox.io/s/cocky-yalow-tet6fe?file=/app.vue

TL;DR;
```html
<template>
  <h1>Throw Error</h1>
</template>

<script lang="ts" setup>
  import { throwError } from '#app';

  const error = new Error('Error in setup via throwError');
  throwError(error);

  throw new Error('Error in setup via throw');
</script>
```

https://user-images.githubusercontent.com/11166786/161761342-ea9842ce-10e1-4b19-9713-b6c72c96ab4c.mp4

### Describe the bug

Throwing an error client-side either via `throwError` or `throw` does not trigger full page error screen.

According to docs for `throwError`:
> You can call this function at any point on client-side, or (on server side) directly within middleware, plugins or `setup()` functions. It will trigger a full-screen error page (as above) which you can clear with `clearError`.

PR #4539, opened 2022-04-22
feat(nuxt): improve error dx for users

### LINKED ISSUE #4897: allow throwing error within async function
### Environment

https://stackblitz.com/edit/nuxt-starter-4amzzp?file=app.vue

### Reproduction

https://stackblitz.com/edit/nuxt-starter-4amzzp?file=app.vue

### Describe the bug

As with the reproduction, I can't throw an error after calling a server api endpoint

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/4520, resolves #4897

### ❓ Type of change

### 📚 Description

This PR improves the error experience for users:

1. fixes an issue where users who manually throw 404s can't escape from the error page (by clearing error _after_ navigating to new route)
1. renames `throwError` to `showError` and allows graceful fallback when nuxt context is not available
1. allows `showError` to accept `createError` params for better dx
1. allows directly throwing errors created with `createError` within the Vue lifecycle. (👈 This is now the recommended approach.) This is similar to using `createError` in server api routes. On server side these will lead to a full page error, but on client can be handled with error boundaries, `onErrorCaptured`, and so on.

### 📝 Checklist


docs/content/2.guide/2.features/7.error-handling.md
```diff
@@ -82,10 +82,22 @@ This function will return the global Nuxt error that is being handled.
...
 You can call this function at any point on client-side, or (on server side) directly within middleware, plugins or `setup()` functions. It will trigger a full-screen error page (as above) which you can clear with `clearError`.
 
+### Example
+
+```vue [pages/movies/[slug].vue]
+<script setup>
+const route = useRoute()
+const { data: movie } = await useFetch(`/api/movies/${route.params.slug}`)
+if (!movie.value) {
+  throwError({ statusCode: 404, statusMessage: 'Page Not Found' })
```
pi0, line 96: In simple example it is obvious but this at least also helps linters to understand exit path.

```suggestion
  return throwError({ statusCode: 404, statusMessage: 'Page Not Found' })
```

docs/content/2.guide/2.features/7.error-handling.md
```diff
@@ -82,10 +82,22 @@ This function will return the global Nuxt error that is being handled.
 
 ### `throwError`
 
-* `function throwError (err: string | Error): Error`
+* `function throwError (err: string | Error | Partial<H3Error>): Error`
```
pi0, line 85: Even if we use H3 internally, we could keep docs simple:

```suggestion
* `function throwError (err: string | Error | <{ statucCode, statusMessage }>>): Error`
```

docs/content/3.api/3.utils/throw-error.md
```diff
@@ -18,3 +19,20 @@ The thrown error is set in the state using [`useError()`](/api/composables/use-e
...
+
+You can use `createError` from [`h3`](https://github.com/unjs/h3) package to trigger error handling in server API routes.
+
+**Example:**
+
+```js
+import { createError } from 'h3'
+
+export default eventHandler(() => {
+  return createError({
```
pi0, line 33:
```suggestion
  throw createError({
```

packages/nuxt/src/app/composables/error.ts
```diff
@@ -1,14 +1,20 @@
...
   const nuxtApp = useNuxtApp()
   return useState('error', () => process.server ? nuxtApp.ssrContext.error : nuxtApp.payload.error)
 }
 
-export const throwError = (_err: string | Error) => {
+export const throwError = (_err: string | Error | Partial<H3Error>) => {
   const nuxtApp = useNuxtApp()
   const error = useError()
-  const err = typeof _err === 'string' ? new Error(_err) : _err
+  console.log({ _err })
```
pi0, line 12: Debug leftover
```suggestion

```

packages/nuxt/src/app/composables/error.ts
```diff
@@ -1,14 +1,20 @@
...
+  const err = typeof _err === 'string'
+    ? new Error(_err)
+    : _err && typeof _err === 'object' && !(_err instanceof Error)
+      ? createError(_err)
```
pi0, line 16: Why instanceof check is necessary? createError allows artibrary errors

docs/content/3.api/3.utils/throw-error.md
Aareksio: It actually works with `return`, is this not intended?
See: https://stackblitz.com/edit/github-jwfbey?file=server%2Fapi%2Fmovie%2F[id].ts

docs/content/2.guide/2.features/7.error-handling.md
```diff
@@ -82,10 +82,22 @@ This function will return the global Nuxt error that is being handled.
...
 You can call this function at any point on client-side, or (on server side) directly within middleware, plugins or `setup()` functions. It will trigger a full-screen error page (as above) which you can clear with `clearError`.
 
+### Example
+
+```vue [pages/movies/[slug].vue]
+<script setup>
+const route = useRoute()
+const { data: movie } = await useFetch(`/api/movies/${route.params.slug}`)
+if (!movie.value) {
+  return throwError({ statusCode: 404, statusMessage: 'Page Not Found' })
```
pi0, line 96: Question: Do we also allow throwing? I would do that with `throw createError()`. It will also pop out from any nested composable.

docs/content/2.guide/2.features/7.error-handling.md
```diff
@@ -82,10 +82,22 @@ This function will return the global Nuxt error that is being handled.
...
 
 You can call this function at any point on client-side, or (on server side) directly within middleware, plugins or `setup()` functions. It will trigger a full-screen error page (as above) which you can clear with `clearError`.
 
+### Example
+
+```vue [pages/movies/[slug].vue]
+<script setup>
+const route = useRoute()
+const { data: movie } = await useFetch(`/api/movies/${route.params.slug}`)
+if (!movie.value) {
```
misaon, line 95: Why not use error instead of data.value?

Aareksio: https://github.com/nuxt/framework/issues/4520#issuecomment-1106676366

packages/nuxt/src/app/composables/error.ts
danielroe: I'll fix upstream in h3 then - currently we lose the status message when passing an h3error to create error.

pi0: I guess it should preserve it unless instanceof is not working because of different refs to `H3Error` (https://github.com/unjs/h3/blob/915c421a532d5db5e79bd68b5871785a81e34da3/src/error.ts#L28). We also only override when new error has those fields...

docs/content/2.guide/2.features/7.error-handling.md
danielroe: Good point. We can certainly do that. Will amend later.

docs/content/2.guide/2.features/7.error-handling.md
danielroe: No, we don't.

packages/nuxt/src/app/composables/error.ts
danielroe: Yes, it must be different instance then. See https://github.com/nuxt/framework/runs/6131472843?check_suite_focus=true for an example of the issue.

docs/content/2.guide/2.features/7.error-handling.md
Aareksio: Wouldn't it actually be possible to `throw` and use some global error handler to promote such error to _critical_ status? Not being able to exit out of nested logic is painful to work around.

danielroe: Yes, it is possible right now. You just need to add the handler for it via https://v3.nuxtjs.org/guide/features/error-handling/#errors-during-the-vue-rendering-lifecycle-ssr--spa 😉

Aareksio: Exactly, this is already possible in user land. Wouldn't it be nice to provide it by default, out of the box? As you can see @pi0 would like to throw, I and 2 other devs in my team had the same thought when first investigating new error handling.

Also worth noting, [unhandled `throw` overrides `throwError`](https://github.com/nuxt/framework/issues/4091), which may cause unexpected issues when `throwError` is used inside nested logic.

docs/content/2.guide/2.features/7.error-handling.md
```diff
@@ -91,8 +91,8 @@ You can call this function at any point on client-side, or (on server side) dire
 ```vue [pages/movies/[slug].vue]
 <script setup>
 const route = useRoute()
-const { data: movie } = await useFetch(`/api/movies/${route.params.slug}`)
-if (!movie.value) {
+const { data, error } = await useFetch(`/api/movies/${route.params.slug}`)
+if (error.value) {
```
Aareksio, line 95: This actually does not work properly with SSR.
https://stackblitz.com/edit/github-jwfbey-ntts81

image

docs/content/2.guide/2.features/7.error-handling.md
pi0: Indeed we might also do it in another PR. `createError` is basically meant for that. Do indicate this is a user thrown error not an unexpected error to bubble up into a vue fatal error. And we can already control message and code with `throw createError()`.

docs/content/2.guide/2.features/7.error-handling.md
danielroe: Ah yes, I see. Because we're manually throwing an error we're actually re-rendering an error page and we lose the original asyncData error... We'll need to provide a better way to handle that.

packages/nuxt/src/auto-imports/presets.ts
```diff
@@ -16,6 +16,12 @@ export const commonPresets: Preset[] = [
       'isVue2',
       'isVue3'
     ]
+  }),
+  defineUnimportPreset({
+    from: 'h3',
+    imports: [
+      'createError'
```
pi0, line 23: We should add a new local `createNuxtError` utility for future compatibility (it can wrap `createError`)

danielroe: I've proposed instead exposing a `createError` from `#app`, which aligns more with the existing `useError`, `throwError`, and `clearError` utilities.

Rigo-m:
@pi0 is this PR merge-able?

rvmourik:
Would be nice to have this merged, we are launching our new website built with Nuxt 3 and need the ability to throw an 404 page when the API response of the Headless CMS is empty.

Thanks in advance 🙏

Rigo-m:
> Would be nice to have this merged, we are launching our new website built with Nuxt 3 and need the ability to throw an 404 page when the API response of the Headless CMS is empty.
> 
> Thanks in advance 🙏

Same here

docs/content/2.guide/2.features/7.error-handling.md
```diff
@@ -80,11 +80,31 @@ This function will return the global Nuxt error that is being handled.
 ::ReadMore{link="/api/composables/use-error"}
 ::
 
+### `createError`
+
+* `function createError (err: { cause, data, message, name, stack, statusCode, statusMessage }): Error`
```
pi0, line 85: I'm not sure we are handling `cause` currently --however is a good idea to do!--

docs/content/3.api/3.utils/throw-error.md
```diff
@@ -6,10 +6,11 @@ Within your pages, components and plugins you can use `throwError` to throw an e
 
 **Parameters:**
 
-- `error`: `string | Error`
+- `error`: `string | Error | Partial<H3Error>`
```
pi0, line 9: Shall we refer to new `NuxtError`?

packages/nuxt/src/app/composables/error.ts
```diff
@@ -1,14 +1,21 @@
...
 
-export const throwError = (_err: string | Error) => {
+export interface NuxtError extends H3Error {}
+
+export const throwError = (_err: string | Error | Partial<NuxtError>) => {
   const nuxtApp = useNuxtApp()
   const error = useError()
-  const err = typeof _err === 'string' ? new Error(_err) : _err
+  const err = typeof _err === 'string'
+    ? new Error(_err)
```
pi0, line 15: Why not using `createError` here? (next link checks can be also inside `createError` to avoid double wrap)

docs/content/2.guide/2.features/7.error-handling.md
danielroe: ~> https://github.com/unjs/h3/pull/131

packages/nuxt/src/app/composables/error.ts
```diff
@@ -1,14 +1,17 @@
...
 }
 
-export const throwError = (_err: string | Error) => {
+export interface NuxtError extends H3Error {}
+
+export const throwError = (_err: string | Error | Partial<NuxtError>) => {
   const nuxtApp = useNuxtApp()
   const error = useError()
-  const err = typeof _err === 'string' ? new Error(_err) : _err
+  const err = typeof _err === 'string' ? new Error(_err) : createError(_err)
```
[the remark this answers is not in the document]
danielroe: You mean, because `throwError` and `createError` behave differently? Sure, very happy to create PR to h3.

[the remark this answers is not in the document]
danielroe: ~> https://github.com/unjs/h3/pull/132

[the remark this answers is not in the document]
pi0: Merged and released 👍🏼 I think now we can always pipe _err to createError :)

packages/nuxt/src/app/composables/error.ts
```diff
@@ -28,3 +31,9 @@ export const clearError = async (options: { redirect?: string } = {}) => {
   }
   error.value = null
 }
+
+const isNuxtError = (err?: string | object): err is NuxtError => err && typeof err === 'object' && ('__nuxt_error' in err)
```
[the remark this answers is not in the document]
danielroe: I am doubtful of exposing the util as the name may be misleading to users - not every nuxt error will have __nuxt_error set, only server errors or explicitly thrown/created errors: https://github.com/nuxt/framework/blob/bbafd5e75e91ab2387f756b6122316246423bb85/packages/nuxt/src/app/components/nuxt-root.vue#L28-L31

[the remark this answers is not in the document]
pi0: We can avoid documenting it if uncertain but with introducing a new NuxtError entity and `createError` that is exposed to user space, I'm pretty sure will find a use for `isNuxtError` same as one we found for `isH3Error` here :)

packages/nuxt/src/app/composables/error.ts
```diff
@@ -28,3 +31,9 @@ export const clearError = async (options: { redirect?: string } = {}) => {
...
+
+export const createError = (err: string | Partial<NuxtError>): NuxtError => {
+  return isNuxtError(err) ? err : _createError(err)
```
pi0, line 38: Is there any reason we are not setting `__nuxt_error` flag for user-created errors here? `isNuxtError` currently basically bypasses h3 normalization. We can combine more considering NuxtError is extending H3Error.

packages/nuxt/src/app/composables/error.ts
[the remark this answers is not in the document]
pi0: Related: https://github.com/nuxt/framework/pull/4539#discussion_r894574849

packages/nuxt/src/app/composables/error.ts
```diff
@@ -28,3 +31,14 @@ export const clearError = async (options: { redirect?: string } = {}) => {
   }
   error.value = null
 }
+
+export const isNuxtError = (err?: string | object): err is NuxtError => err && typeof err === 'object' && ('__nuxt_error' in err)
+
+export const createError = (err: string | Partial<NuxtError>): NuxtError => {
+  const _err: NuxtError = _createError(err)
+  if (typeof err === 'string' || (!isNuxtError(err) && !err.statusMessage)) {
+    _err.statusMessage = 'Internal Server Error'
```
pi0, line 40: Maybe we can set it in H3 instead of default undefined in H3's create error? Seems more consistent we we always do same behavior...

danielroe: H3 sets it to `H3Error`.

danielroe: ~> https://github.com/unjs/h3/pull/135

pi0: released as 0.7.10 👍🏼

pi0:
Currently, `throwError` and `throw createError` are behaving exactly the same by showing a full-screen error page. I'm thinking following adjustments before adding:

- `throw createError()`
  - Server-side: Full error page
  - Client-side: Let the excitation scope gracefully handles (*)
  - Document and recommend since has the benefit of bubbling up from function/composable to their parent for controlled handling
- `throwError()`
  - Server and Client: Show full error page
  - Adds a prop to NuxtError (`userError`, `fullpageError`, etc) to distingoush
  - Possibly deprecate and rename to a more semantic `showError`
  - Document the behavior and recommend to only use for critical errors 


(*) Graceful client: Showing full-error page for Client-Side, is really unnecessary in many cases. For example, when a refresh button fails, we can avoid bailing the full application into an error state. Using Error boundary is probably best but it is harder to use.

pi0 (approved):

Aareksio:
🔥

Issue #603, opened 2021-09-30
warn users if `asyncData` used without await


antfu:
Follow up: https://github.com/nuxt/framework/pull/579#discussion_r717479594

PR #4549, opened 2022-04-23
docs(api): fix `useAsyncData` example

klevron:
### ❓ Type of change

### 📚 Description

Fix typo and remove pick option since api will return and array of mountains

### DIFF
--- a/docs/content/3.api/1.composables/use-async-data.md
+++ b/docs/content/3.api/1.composables/use-async-data.md
@@ -56,7 +56,7 @@ By default, Nuxt waits until a `refresh` is finished before it can be executed a
 ## Example
 
 ```ts
-const { data, pending, error, refresh } = useAsyncData(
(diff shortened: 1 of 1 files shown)


Aareksio:
Btw this example does not include `await` and as far as I am aware, nowhere in the doc it is stated `useAsyncData` should always be awaited, yet [according to this issue](https://github.com/nuxt/framework/issues/603) this is the case.

docs/content/3.api/1.composables/use-async-data.md
```diff
@@ -58,10 +58,7 @@ By default, Nuxt waits until a `refresh` is finished before it can be executed a
 ```ts
 const { data, pending, error, refresh } = useAsyncData(
```
pi0, line 59:
```suggestion
const { data, pending, error, refresh } = await useAsyncData(
```

PR #4578, opened 2022-04-25
feat(nuxi)!: setup nuxt globally with `nuxt test`

antfu:
### ❓ Type of change

### 📚 Description

It allows users to test their app using `nuxi test` without explicitly add setup in their test files.

### 📝 Checklist


packages/test-utils/src/run.ts
```diff
@@ -21,18 +26,35 @@ export async function runTests (opts: RunTestOptions) {
     process.env.NUXT_TEST_DEV = 'true'
   }
 
+  process.env.NUXT_TEST_OPTIONS = JSON.stringify(opts)
```
pi0, line 29: Shouldn't we cherry-pick options here same as `exposeContextToEnv`? (a shared utility might be useful)

antfu: The options are literal https://github.com/nuxt/framework/blob/80a11210127f1f993a7d9d0204d4b6a3b6bba98d/packages/test-utils/src/run.ts#L4-L15

to me cherry-picking it would not help much but might potentially break the worker thread if it relies on some options

AndrewBogdanovTSS:
Any updates on this PR?

pi0 (approved):

Issue #4497, opened 2022-04-21
When build app useFetch with transform option will return data with value null on client side


lhahut96:
### Environment

Operating System: Windows_NT
Node Version: v16.14.2
Nuxt Version: 3.0.0-rc.1
Package Manager: npm@8.7.0
Builder: vite
User Config: modules
Runtime Modules: @nuxtjs/color-mode@3.0.2
Build Modules: -

### Reproduction

https://github.com/ntt1182/reproduction-nuxt3

### Describe the bug

When using nuxi dev app will run normally without return normal data but when build and start code with node, client-side will return null value data
![image](https://user-images.githubusercontent.com/67304218/164433298-e15ae219-03ba-4819-a14b-2399f9b9869f.png)

danielroe:
The workaround for the moment is to provide a custom key, e.g.:

```js
const { data } = await useFetch('/some/path', {
  transform: data => data,
  key: 'my-key'
})
```

Aareksio:
In relation to #4553, this leads to strange behavior (_bailed hydration?_), _probably_ related to swallowed error. It may be good idea to figure out where the error is caught and ignored, in case similar issue occurs in the future.

<details>
<summary>Dev</summary>

![image](https://user-images.githubusercontent.com/11166786/165073691-5242a02a-1830-4b54-ab25-f8b9b7f6353b.png)

</details>

<details>
<summary>Build + Preview</summary>

![image](https://user-images.githubusercontent.com/11166786/165073589-6f50cc39-def0-46d1-8b9e-58e73fbdf511.png)

</details>

https://stackblitz.com/edit/github-ycnvlb

PR #4590, opened 2022-04-25
fix(nuxt): warn to provide a key for `useFetch` with `transform` option

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/4497

### ❓ Type of change

### 📚 Description

This PR addresses two issues:

1. In production the transform function `.toString()` differs between server/client so the generated key (from `ohash`) doesn't match.
2. When there is a mismatch (no key in payload), then currently we just skip refreshing, when probably we want to re-fetch on client-side.

#### Other options:

1. generate a key at build-time from hash of transform function
3. exclude transform function from hash generation

It feels like the safest option is to require a user to provide a custom key in the event of using `transform`.

### 📝 Checklist


packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -28,6 +28,9 @@ export function useFetch<
   request: Ref<ReqT> | ReqT | (() => ReqT),
   opts: UseFetchOptions<_ResT, Transform, PickKeys> = {}
 ) {
+  if (process.dev && opts.transform && !opts.key) {
+    console.warn('[nuxt] You should provide a key for useFetch when using a custom transform function.')
```
Atinux, line 32:
```suggestion
    console.warn('[nuxt] You should provide a key for `useFetch` when using a custom transform function.')
```

danielroe:
> I think we can exclude options that are not related to fetch like transform from hash transformation. hashing it to distinguish by request.

I would say that transform should make a difference to the hash. Consider:

```js
await useFetch('/api/test', { transform: data => data[0] })
await useFetch('/api/test')
```

These two calls will result in different data, and if we don't enforce a key generation then there's a danger of mismatching payloads.

pi0:
Okay i'm fine with adding warning until find a better way for auto hashing. Can we in the meantime remove `transform` from hash when making warning?

pi0 (approved):

Issue #4235, opened 2022-04-10
nested router - doesnt work


mihail727:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-27460146.2ad93eb`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://stackblitz.com/edit/github-zvowq4-gj7iuh?file=pages/index.vue

### Describe the bug

with latest version of nuxt 3 and follow the [documentation](https://v3.nuxtjs.org/guide/features/server-routes#using-a-nested-router) i cant create nested route
when i'm trying to get data from api i get error: `[Vue Router warn]: No match found for location with path "/test"`

Aareksio:
The error with `handle is not a function` is resolved on latest release. Your reproduction have `h3@0.4.2`, over 20 days old release. With clear install and `h3@7.2.0` it is fixed. However, the nested router does not work.

---

The best I managed to do is changing `data.ts` to:
```diff
import { createRouter } from 'h3';

const router = createRouter();

- router.get('/test', async (req, res) => {
+ router.get('/api/data', async (req, res) => {
  return { msg: 'test' };
});

- export default router;
+ export default router.handler;
```

This serves the response on `/api/data`, treating nested router as normal event handler. As far as I searched, `h3` does not have any special logic for `Router` as event handler, so the nested router approach does not seem to be implemented at all. The closest to handling nested routers I could find is https://github.com/unjs/h3/blob/e529f4ce487fef6e24748081b31bac7f9623f509/src/app.ts#L149-L152, but it does not handle the logic of nested paths and everything else required.

danielroe:
For now, you can workaround this with:

```js
import { createRouter, useBase } from 'h3';

const router = createRouter();

router.get('/', () => 'Hello World');

export default useBase('/api/test', router.handler);
```

mihail727:
> For now, you can workaround this with:
> 
> ```js
> import { createRouter, useBase } from 'h3';
> 
> const router = createRouter();
> 
> router.get('/', () => 'Hello World');
> 
> export default useBase('/api/test', router.handler);
> ```

but how to work with another route?
for example one more `router.post(...)`

ahku:
> For now, you can workaround this with:
> 
> ```js
> import { createRouter, useBase } from 'h3';
[...]

PR #4594, opened 2022-04-25
docs(config): improve `serverMiddleware` jsdocs

danielroe:
### 🔗 Linked issue

resolves #4591

### ❓ Type of change

### 📚 Description

This removes some of the old data from `serverMiddleware` docs. I feel this could be considerably improved, maybe incorporating some of the nice new API docs... Or adding a notice suggesting using auto-detected `~/server/api` etc...?

cc: @clemcode 

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -344,7 +344,7 @@ export default {
    *   // Will register redirect-ssl npm package
    *   'redirect-ssl',
    *   // Will register file from project server-middleware directory to handle /server-middleware/* requires
-   *   { path: '/server-middleware/**', handler: '~/server-middleware/index.ts' },
+   *   { path: '/server-middleware/**', handler: '~/server-middleware/my-middleware' },
    *   // We can create custom instances too, but only in development mode, they are ignored for the production bundle.
    *   { path: '/static2/**', handler: serveStatic(fileURLToPath(new URL('./static2', import.meta.url))) }
    * ]
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -344,9 +344,9 @@ export default {
    *   // Will register redirect-ssl npm package
    *   'redirect-ssl',
    *   // Will register file from project server-middleware directory to handle /server-middleware/* requires
-   *   { path: '/server-middleware', handler: '~/server-middleware/index.js' },
+   *   { path: '/server-middleware/**', handler: '~/server-middleware/index.ts' },
    *   // We can create custom instances too, but only in development mode, they are ignored for the production bundle.
-   *   { path: '/static2', handler: serveStatic(fileURLToPath(new URL('./static2', import.meta.url))) }
+   *   { path: '/static2/**', handler: serveStatic(fileURLToPath(new URL('./static2', import.meta.url))) }
    * ]
    * ```
    *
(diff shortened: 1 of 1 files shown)


packages/schema/src/config/_common.ts
```diff
@@ -344,27 +344,23 @@ export default {
    *   // Will register redirect-ssl npm package
    *   'redirect-ssl',
    *   // Will register file from project server-middleware directory to handle /server-middleware/* requires
-   *   { path: '/server-middleware', handler: '~/server-middleware/index.js' },
+   *   { path: '/server-middleware/**', handler: '~/server-middleware/index.ts' },
    *   // We can create custom instances too, but only in development mode, they are ignored for the production bundle.
-   *   { path: '/static2', handler: serveStatic(fileURLToPath(new URL('./static2', import.meta.url))) }
+   *   { path: '/static2/**', handler: serveStatic(fileURLToPath(new URL('./static2', import.meta.url))) }
```
pi0, line 349: I would rather not to advocate routing features alongside with legacy server middleware

packages/schema/src/config/_common.ts
```diff
@@ -344,27 +344,23 @@ export default {
    *   // Will register redirect-ssl npm package
    *   'redirect-ssl',
    *   // Will register file from project server-middleware directory to handle /server-middleware/* requires
-   *   { path: '/server-middleware', handler: '~/server-middleware/index.js' },
+   *   { path: '/server-middleware/**', handler: '~/server-middleware/index.ts' },
```
pi0, line 347: We can omit extension for being language agnostic:

```suggestion
   *   { path: '/server-middleware/**', handler: '~/server-middleware/my-middleware' },
```

packages/schema/src/config/_common.ts
```diff
@@ -344,27 +344,23 @@ export default {
...
    * form with a specific path.
    *
    * If you pass a string handler, Nuxt will expect that file to export a default function
-   * that handles `(req, res, next) => void`.
+   * produced with `eventHandler` from `h3`.
    *
    * @example
    * ```js
-   * export default function (req, res, next) {
+   * export default eventHandler(event => {
```
pi0, line 361: server-middleware option is basically a legacy option I would mark as deprecated soon. We don't need to update it's example with something it is not mean to be. (BTW for consistency we use `defineEventHandler` across Nuxt docs)

packages/schema/src/config/_common.ts
danielroe: then let's remove this entirely, as it won't work without the change.

pi0: Okay right it seems a bug that we are broke with new handlers for legacy middleware.

packages/schema/src/config/_common.ts
danielroe: Hm. I don't like leaving a legacy example. Shall we just remove and link to https://v3.nuxtjs.org/guide/features/server-routes#server-middleware instead?

pi0: Me either but this IS a legacy option :D Let's introduce new options and deprecate this.

packages/schema/src/config/_common.ts
danielroe: tracked in: https://github.com/nuxt/framework/issues/4235

pi0:
With #4850 this change is not relavant anymore as we register legacy `serverMiddleware` exactly as middleware not route handlers. Also deprecated by #4850 and removed from the docs.

PR #4601, opened 2022-04-25
docs(concepts): change phrasing

sammce:
### ❓ Type of change

### 📚 Description

Changes the phrasing of a sentence on the 'Vue.js development' page in the 'concepts' section to be more grammatically correct.


florian-lefebvre (commented):
Shouldn't it be "which allowed you" instead?

sammce:
My rationale for using 'allows' was because Vue 3 still offers the Options API. If Vue 3 completely deprecated the Options API I would definitely use 'allowed'.

I view the explanation in the docs as an objective definition of what the Options API _**is**_, not was, hence the present tense.

I suppose either would work though.

danielroe (approved):

pi0:
Thanks.

Issue #4259, opened 2022-04-11
Global Middleware on first load both side


shinguweb96:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-27484505.9f6eff7`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `vite`, `privateRuntimeConfig`, `publicRuntimeConfig`, `css`, `buildModules`, `intlify`, `vueuse`
- Runtime Modules:  `-`
- Build Modules:    `@intlify/nuxt3@0.1.10`, `@pinia/nuxt@0.1.8`, `@vueuse/nuxt@8.2.5`
------------------------------

### Reproduction

https://stackblitz.com/edit/github-dypzgk

### Describe the bug

On first load, the global middleware get executed on both side (server and client).
In Nuxt 2, if the app is in universal mode, at first load the middleware get executed only by server side.

shinguweb96:
Hi guys!
Any news?

Atinux:
Hey @shinguweb96 

Could you explain your specific use case to run this global middleware only once?

selucaciu:
@Atinux In Nuxt 2, if the app is in universal mode, at first load the middleware get executed only by server side.
I think there are a bunch of use cases where you would want to execute the middleware only once, like counting the number of pages visited.
I have noticed that also the asyncData behaviour has changed. In nuxt2 returning to the page/route would call async data again but now you need to provide a unique key. For example a use case is that you need to provide a unique key on revisiting the page to be able to refresh the data in asyncData.

pskclub:
Hi guys!
Any news?

pskclub:
> @pskclub You can use `if (process.server)` for example if you want to run it only on server.

i need run only onetime when route change , how to fix that

jdgjxxchy:
> > @pskclub You can use `if (process.server)` for example if you want to run it only on server.
> 
> i only need to run once if the route changes, how can i do that?

I have also ran into this issue. I find a way to judge whether it is running in server side or in client.

1. use `process.client` or `process.server`. if code is running in client side, `process.client` will be true. So if you want to count page enter times, you can add ` if (process.client) ` block.
2. use `useNuxtApp()` . If you need to access nuxtApp by coincidence, you can use `useNuxtApp().ssrContext` to judge, because this property only exist in server side.

PR #4645, opened 2022-04-26
fix(nuxt): process middleware after plugins

### LINKED ISSUE #4703: Route in plugin is always /
### Environment

- Operating System: `Linux`
- Node Version:     `v16.14.1`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-trznob?file=plugins/plugin.ts

### Describe the bug

Hi,

I am trying to get the current path in a plugin using `useRoute()` but it always returns the route with the path `/`.

```
export default defineNuxtPlugin((nuxt) => {
    const route = useRoute()
    console.log(route.path) // always '/'
})
```

My workaround is to use the `app:created` hook in the plugin and call `useRoute()` outside the hook and access the path inside the hook:

```
export default defineNuxtPlugin((nuxt) => {
    const route = useRoute()
    nuxt.hook('app:created', () => {
        console.log(route.path) // works
    })
})
```

If I call `useRoute()` inside the hook I get a nuxt `instance unavailable error`:

```
export default defineNuxtPlugin((nuxt) => {
    nuxt.hook('app:created', () => {
        const route = useRoute() // error
        console.log(route.path)
    })
})
```

Shouldn't the nuxt instance be available inside `app:created` hook?
My goal is to call an API on init in a nuxt module based on the current path. Is there another recommended way to do this?

Best regards
David

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/4257, see also https://github.com/nuxt/framework/issues/4259, resolves #4703

### ❓ Type of change

### 📚 Description

This PR changes middleware behaviour to defer processing middleware until after plugins are initialised. This means that middleware will have access to plugin injections, and that plugins have the chance to add middleware themselves.

~In addition (and this is the breaking change), it skips processing middleware on client-side for server-side-rendered routes, aligning with Nuxt 2.~

**Potential for future improvement** - `.client` middleware that _does_ run on client-side but not on server

**SSR**:
1. set up router + current route (by pushing it)
2. add middleware
2. process plugins
3. replace current route with itself, triggering middleware

**CSR**:
1. set up router + await ready
2. add middleware
3. process plugins
3. replace current route with itself, triggering middleware

### 📝 Checklist


Atinux:
Beautiful work @danielroe ❤️ 

How can we add a test case to confirm the behaviour and avoid any potential regression in the future?

Atinux:
Happy to merge anyway and test internally 😊

pi0:
I'm wondering if we can keep client/server behavior consistent by either executing middleware on initial render or not. If server-only middleware on initial is desired, it can be easily guarded with `if (process.server )`

Atinux:
I am not sure about this, I believe middleware shall only be run once. This is the behavior I have done in Nuxt 2 on purpose.

What is the advantage of rerunning them on client-side?

pi0:
> I believe middleware shall only be run once. 

They are. Like anything else during hydration and same as plugins, they run once on the server and once on the client side.

In general, I don't think there is any advantage that we are even calling middleware on initial rendering. It is a really strange use case that could be easily replicated with a nuxt plugin already and removed all this complexity for pushing route and ordering... But if we are doing this, it should be consistent IMO. We already introduced this change on Nuxt 3 and breaking it once again doesn't makes sense.

danielroe:
At the moment, this PR simply fixes a couple of bugs; it doesn't change the current behaviour, which runs middleware on client _and_ server. It's not meant to make a decision on the question of whether to run middleware on initial client-side load. Let's have that conversation in https://github.com/nuxt/framework/issues/4259; we can always follow up with a further PR reverting [`e14384e` (#4645)](https://github.com/nuxt/framework/pull/4645/commits/e14384e15698bd56768265d42038063d084f7171).

(We could also support  `.client` and `.server` suffixes for middleware and it would be good to have only one breaking change rather than two in a row.)

pi0 (approved):

vis97c:
Why does a "global" middleware run only on the server. If people think they only need it on the server just use a process.server conditional. I have a usecase to run it twice but it seems like it is not possible with the current implementation. @pi0 What a re your thoughts on this?

danielroe:
Please see my immediately preceding comment above. This PR did not end up changing the fact that middleware runs on both client and server.

See https://github.com/nuxt/framework/issues/4259 for an issue requesting this to change and run only once.

PR #4664, opened 2022-04-27
chore: enable typecheck ci for packages

antfu:
Before enabling `strict: true`, we might need to have CI to ensure it works.


pi0:
Thanks!!

PR #4667, opened 2022-04-27
fix(nuxt): remove nuxt2 compat from nitro renderer

danielroe:
### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

This cleans up some leftover compatibility code in the nitro renderer, which was previously necessary for the bridge nitro implementation.


pi0:
I'm going to rework this removing more legacy code at once.

pi0:
I'm going to rework this removing more legacy code at once.

PR #4708, opened 2022-04-29
docs: use `eventHandler` instead of `defineEventHandler`

benjamincanac:
### ❓ Type of change

### 📚 Description

Like defined in Nitro https://nitro.unjs.io/guide/routing.html#usage

### DIFF
--- a/docs/content/2.guide/2.features/9.server-routes.md
+++ b/docs/content/2.guide/2.features/9.server-routes.md
@@ -2,7 +2,7 @@
 
 Nuxt automatically scans files inside the `~/server/api`, `~/server/routes`, and `~/server/middleware` directories to register API and server handlers with HMR support.
 
-Each file should export a default function defined with `defineEventHandler()`.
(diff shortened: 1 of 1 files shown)


manniL:
But the [`h3` JSDocs](https://www.jsdocs.io/package/h3#defineEventHandler) say something different 🤔

benjamincanac:
It is defined as shortcut here: https://github.com/unjs/h3/blob/0692656c59074c3900b92b203151590f4ab8e055/src/event.ts#L32

What is the preferred way in term of documentation @pi0 ?

pi0:
Thanks for the pull request @benjamincanac. I personally love to pick shorter variants but actually main reason that we are still supporting longer `define*` names in h3 and nitropack is Nuxt! After a quick talk with team, we concluded longer syntax is more consistent with other functions such as `defineNuxtPlugin` and `defineNuxtConfig`. So I'm afraid have to close this PR at least for now. But nothing is set in stone and we might change the preferred convention in the docs.

Issue #4661, opened 2022-04-27
Cannot convert object to primitive value at ssrCompile


chakraecho:
### Environment

Nuxt CLI v3.0.0-rc.1                                                                                                  11:41:57
RootDir: /home/jhuang/projets/nuxt3                                                   11:41:59
Nuxt project info:                                                                                                    11:41:59

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.13.1`
- Nuxt Version:     `3.0.0-rc.1`
- Package Manager:  `npm@8.3.0`
- Builder:          `vite`
- User Config:      `components`, `hooks`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

`app.vue`
```vue
<template>
  <div>
    <Test />
  </div>
</template>

<script>
import {defineNuxtComponent} from "#app";
import Test from "./components/test"

export default defineNuxtComponent({
  components: {
    Test
  }
})
</script>
```

`components/test.ts`
```ts
import {defineNuxtComponent} from "#app";

export default defineNuxtComponent({
    name: "TestCompile",
    template: "<div>test</div>"
})
```
`nuxt.config.ts`
```ts

import {defineNuxtConfig} from 'nuxt'

// https://v3.nuxtjs.org/api/configuration/nuxt.config
export default defineNuxtConfig({


    components: {
        global: false
    },
    hooks: {
        'vite:extendConfig': (config, {isClient, isServer}) => {
            if(isClient)
            config.resolve.alias.vue = 'vue/dist/vue.esm-bundler.js';
        },

    },

})
```

- run `npm run build`
- run `npm run preview`
- open  http://0.0.0.0:3000/
- Error in console : 
```
file:///home/jhuang/projets/nuxt3/.output/server/chunks/index.mjs:9869
    return (compileCache[cacheKey] = Function('require', code)(commonjsRequire));
                                     ^

TypeError: Cannot convert object to primitive value
    at Function (<anonymous>)
    at ssrCompile (file:///home/jhuang/projets/nuxt3/.output/server/chunks/index.mjs:9869:38)
    at renderComponentSubTree (file:///home/jhuang/projets/nuxt3/.output/server/chunks/index.mjs:9960:30)
    at renderComponentVNode (file:///home/jhuang/projets/nuxt3/.output/server/chunks/index.mjs:9946:16)
[...]

PR #4762, opened 2022-05-02
feat(nuxt): support vue runtime compiler

huang-julien:
### 🔗 Linked issue

resolve nuxt/nuxt.js#13843 

### ❓ Type of change

### 📚 Description

This PR allow to use vue runtimeCompiler on both server and client side

- add vue.runtimeCompiler boolean option to nuxt/schema
- unmock vue 3 only when runtimeCompiler option is set to true when building and make vue runtime compiler available client side
- with externalVue to true, force trace files used by generated code from the runtime compiler

### 📝 Checklist


pi0:
Hi @huang-julien and thanks for this PR! 💚  I think finally we can add an option to nitro if modification of commonJS plugin options is required. 

I hope we can add runtime support with lighter dependencies. At least having `@vue/devtools-api` doesn't makes sense to me in the bundle for it.

(BTW: You don't need to merge into main on each commit. I will make once rebase before merge)

huang-julien:
> Hi @huang-julien and thanks for this PR! 💚 I think finally we can add an option to nitro if modification of commonJS plugin options is required.
> 
> I hope we can add runtime support with lighter dependencies. At least having `@vue/devtools-api` doesn't makes sense to me in the bundle for it.
> 
> (BTW: You don't need to merge into main on each commit. I will make once rebase before merge)

Hi @pi0  \:wave:  , yes i was thinking about a PR on Nitro to add the option for the commonjs plugin modification instead of just replacing it in the nuxt package. Do you want me to remove the @vue/devtool-api from the bundle through a commit ?

huang-julien:
without the runtimeCompiler : Total size: 3.16 MB (410 kB gzip)
with @vue/devtools : Σ Total size: 5.19 MB (564 kB gzip)
without @vue/devtools : Σ Total size: 5.14 MB (559 kB gzip)

That's quite heavy ...

huang-julien:
linked PR - https://github.com/unjs/nitro/pull/187
the feature will be refactored if the PR is accepted

packages/nuxt/src/core/nitro.ts
```diff
@@ -121,6 +131,16 @@ export async function initNitro (nuxt: Nuxt) {
     nitro.options.rollupConfig.plugins.push(plugin)
   })
 
+  // Enable runtime compiler on build
+  if (nuxt.options.runtimeCompiler && !nuxt.options.dev) {
+    // set vue esm on client
+    nuxt.hook('vite:extendConfig', (config, { isClient }) => {
+      if (isClient) {
+        config.resolve.alias.vue = 'vue/dist/vue.esm-bundler'
```
pi0, line 139: Shall we support client alias for webpack too?

packages/nuxt/package.json
```diff
@@ -36,6 +36,8 @@
     "@nuxt/telemetry": "^2.1.3",
     "@nuxt/ui-templates": "^0.1.0",
     "@nuxt/vite-builder": "^3.0.0-rc.2",
+    "@rollup/plugin-commonjs": "^22.0.0",
```
pi0, line 39: We can drop it now

packages/nuxt/src/core/nitro.ts
huang-julien: i think so, i can't figure out any case where we would need the compiler only on server-side. If the client doesn't have the runtime compiler, it would cause a difference between the dom generated by the ssr and the dom generated by the client. The client will render of the component again but without the runtime compiler, it won't be able to render the template.

packages/nuxt/package.json
huang-julien: oops my bad

huang-julien: looks like it breaks the CI due to yarn.lock

packages/nuxt/src/core/nitro.ts
huang-julien: I just tried with the dev mode. Got a mismatch in dev mode. It needs to be supported client side in dev mode too.
```
[Vue warn]: Hydration node mismatch:
Client vnode: Symbol("Comment") 
- Server rendered DOM: 
<!-- [ -->
 (start of fragment)
 ```

**Comment by huang-julien** [comment]
it should be good now. example [here](https://github.com/huang-julien/nuxt-3/tree/feat/nuxt-ssr-runtime-template-compiler-example)

**Comment by huang-julien** [comment]
@pi0 it should be ok now. [here's](https://github.com/huang-julien/nuxt-3/tree/feat/nuxt-ssr-runtime-template-compiler-example) a working example

File: packages/nuxt/src/core/nitro.ts (hunk 1 above)
Comment (pi0) [reply packages/nuxt/src/core/nitro.ts:139]: Yes, client is fine. What i'm asking is to support **webpack** in addition to **vite** for client builds.

File: packages/schema/src/config/build.ts [hunk 3]
```diff
@@ -638,5 +638,12 @@ export default {
...
     followSymlinks: false
-  }
+  },
+
+  
+  /**
+   * enable runtime compiler on build
+   * @version 3
+   */
+   runtimeCompiler: false
```
pi0, line 648: We might move this option under `vue` namespace

packages/nuxt/src/core/nitro.ts
huang-julien: ah yes sorry, we might need to support it if the builder used is webpack

packages/schema/src/config/build.ts
huang-julien: Since this is used in packages/nuxt/src/core/nitro.ts and packages/schema/src/config/build.ts is considered as "Legacy", can i move it to packages/schema/src/config/nitro.ts instead ?

huang-julien: nervermind, vue namespace is good too :smiley:

huang-julien:
Client builds now supports webpack and the runtimeCompiler option is now under the vue namespace in [packages/schema/src/config/_app.ts](https://github.com/nuxt/framework/commit/7095f92226a5936c26b25410a7e4935961536da3#diff-50493f2f12130c811bb1cab04babc37c120e78ffca1733b75787a41291be2bef)

huang-julien:
@pi0 the h() function is not always used when using the runtime compiler. Devs most of the time just use the template key or write a simple object to define components, should we support it with the runtime compiler ? i can revert the commit if you want

tobiasdiez:
What's the status of this PR? Would be awesome to use the runtime compiler soon!

huang-julien:
@tobiasdiez if you really need it now, this [module](https://www.npmjs.com/package/nuxt-runtime-compiler) can enable it

huang-julien:
investigating https://github.com/nuxt/framework/issues/4661#issuecomment-1182756953

huang-julien:
fixed https://github.com/nuxt/framework/issues/4661#issuecomment-1182756953

Atinux:
Thank you!

It would be nice to document this option somewhere since it is quite useful for CMS integrations using the template prop (see linked issue).

Also, is there a place where we can add a test case for this?

packages/schema/src/config/build.ts
```diff
@@ -266,7 +266,7 @@ export default {
           object: 'src',
           embed: 'src'
         },
-        compilerOptions: { $resolve: (val, get) => val ?? get('vue.compilerOptions') },
+        compilerOptions: { $resolve: (val, get) => val ?? get('vue.compilerOptions') }
```
danielroe, line 269:
```suggestion
        compilerOptions: { $resolve: (val, get) => val ?? get('vue.compilerOptions') },
```

packages/schema/src/config/build.ts
```diff
@@ -638,5 +638,5 @@ export default {
      * @version 2
      */
     followSymlinks: false
-  }
+  },
```
danielroe, line 641:
```suggestion
  }
```

huang-julien:
> Also, is there a place where we can add a test case for this?

Do you mean a unit test or an example ? I've tested the  component with template defined in runtime in the playground [here](https://github.com/huang-julien/nuxt-3/tree/example/runtime-compiler)

Atinux:
I am thinking of a test case here: https://github.com/nuxt/framework/tree/main/test

Any guidance for the best place for it? @danielroe @pi0

pi0:
Testing requires making another fixture currently we only have one. We can add later.

I'm pending to locally test this PR. Some changes can be probably reduced for mocking and saving bundle size with option enabled.

In the meantime you can use [nuxt-runtime-compiler module](https://www.npmjs.com/package/nuxt-runtime-compiler)

huang-julien:
Hi @danielroe , using externalVue is now working but having externalVue set to false does not work anymore (nuxt-edge/rc10). 
`Cannot find module 'D:\GIT\nuxt-3\playground\.output\server\node_modules\estree-walker\dist\esm\estree-walker.js' imported from D:\GIT\nuxt-3\playground\.output\server\chunks\renderer.mjs`

this was working on rc9.

`estree-walker/dist/umd/estree-walker.js` is still sent to `.output` but the esm dist is missing.
 
Unmocking estree-walker does not send a esm version into `.output` anymore.
https://stackblitz.com/edit/nuxt-starter-v4dnrk?file=package.json,app.vue,nuxt.config.ts,pages%2Findex.vue
````json
{
  "main": "./dist/umd/estree-walker.js",
  "module": "./dist/esm/estree-walker.js",
  "exports": {
    "require": "./dist/umd/estree-walker.js",
    "import": "./dist/esm/estree-walker.js"
  }
}
````


Should i open an issue on nitro ?

**Comment by huang-julien** [comment]
@pi0 I think test fixtures for the runtime compiler should be made with externalVue true and false since there was different issues depending on this option.

**Comment by abdonrd** [comment]
Are there plans to merge this PR? Thanks!

**Comment by huang-julien** [comment]
oh i completly forgot about this PR :sweat_smile: i'll add some tests (external vue true -- false) maybe this weekend if have some times, i also need to check if this is still working with nuxt 3.1 being released. meanwhile you can use [nuxt-runtime-compiler](https://www.npmjs.com/package/nuxt-runtime-compiler) which is basically the content of this PR but within a nuxt module

**Comment by abdonrd** [comment]
Thanks @huang-julien! The [nuxt-runtime-compiler](https://www.npmjs.com/package/nuxt-runtime-compiler) works well for us!

One thing...
We are using the `vue.compilerOptions.isCustomElement` option in the `nuxt.config.ts` as mentioned here: https://github.com/vuejs/core/issues/3746#issuecomment-835766808

(Because we are using [Katex](https://github.com/KaTeX/KaTeX))

Could it be that this configuration does not reach the runtime compiler?

**Review by danielroe — COMMENTED** [review]
Would you let me know when this ready to go? And also, would you mind redoing your last merge? I think we've ended up with a bigger diff than intended...

**Comment by huang-julien** [comment]
@danielroe about the options for vue in runtime, how should we do it ? This can be done through a plugin by the user but i think it might be better to be able to use the compatible options used at build time.

**Comment by toniengelhardt** [comment]
Having some trouble getting the module to work, but would be amazing if we could do something like this:

```vue
<script setup lang="ts">
import { compile, h } from 'vue'

const id = 5
const dc = '<div class="some-class"><MyComponent :id="${id}" /></div>'

const RenderContent = () => h(compile(dc))
</script>

<template>
  <div>
    <div>My dynamically generated content:</div>
   <RenderContent />
  </div>
</template>
```

like in these examples: 
- https://stackoverflow.com/a/70151867/1384972
- https://stackoverflow.com/questions/68981411/vue3-how-to-use-vue-compile-in-vue3

abdonrd:
@huang-julien you have an small lint error:
https://github.com/nuxt/framework/actions/runs/4129036194/jobs/7134196036#step:6:14

abdonrd:
@danielroe Is there still interest in merging this? Thank you!

abdonrd:
@huang-julien it seems now the tests do not pass well.

huang-julien:
yes estree walker is not being traced by NFT. Locally a simlink of estree walker is being send to `.output`. Going to check if this is does impact `nuxt-runtime-compiler` as well

huang-julien:
can't find why CI are failing 🤷‍♂️ I can see logs while it does work locally. I'm not even sure if the issue comes from estree walker

packages/nuxt/src/core/nitro.ts
```diff
@@ -163,6 +179,18 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
...
+      dynamicRequireTargets: (!nuxt.options.experimental.externalVue && nuxt.options.vue.runtimeCompiler)
+        ? [
+            'node_modules/**/vue',
+            'node_modules/**/@vue/compiler-core',
+            'node_modules/**/@vue/compiler-dom',
+            'node_modules/**/@vue/compiler-ssr',
+            'node_modules/**/vue/server-renderer',
+            'node_modules/**/estree-walker'
+          ]
+        : []
```
danielroe, line 193: is this change necessary? Tests seem to pass even when this change is omitted.

packages/nuxt/src/core/nitro.ts
```diff
@@ -126,6 +126,18 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
...
+      traceInclude: [
+        // force include files used in generated code from the runtime-compiler
+        ...(nuxt.options.vue.runtimeCompiler)
+          ? [
+              ...nuxt.options.modulesDir.reduce<string[]>((targets, path) => {
+                const serverRendererPath = resolve(path, 'vue/server-renderer/index.js')
+                if (existsSync(serverRendererPath)) { targets.push(serverRendererPath) }
+                return targets
+              }, [])
+            ]
```
danielroe, line 139: it feels like this is only needed because of a bug in the tracing mechanism - it would be good to know if this is an issue with `vercel/nft` or with our externals plugin (cc: @pi0). Maybe because we have both estree-walker v2 and v3 installed?

packages/nuxt/src/core/nitro.ts
huang-julien: This is needed only when externalVue is set to false. I'll add another test :+1:

packages/nuxt/src/core/nitro.ts
huang-julien: I'm not sure if this is a bug. Since it is required by the code generated by the compiler at runtime, `vercel/nft` shouldn't be able to trace it 🤔

danielroe: Hm. But isn't `vue/server-renderer` is required _anyway_ by Nuxt to render the server?

packages/nuxt/src/core/nitro.ts
danielroe: Even when setting `externalVue` to false, the runtime compiler test fixture still seems to work for me without these lines.

autofix-troubleshooter:
Hi! I'm the <a href="https://autofix.ci">autofix logo<strong>autofix.ci</strong></a> troubleshooter bot.

It looks like you correctly set up a CI job that uses the autofix.ci GitHub Action, but the [**autofix.ci GitHub App**](https://github.com/apps/autofix-ci/) has not been installed for this repository. This means that autofix.ci unfortunately does not have the permissions to fix this pull request. If you are the repository owner, please install the app and then restart the CI workflow! 😃

packages/nuxt/src/core/nitro.ts
huang-julien: That's probably because the generated code use `require()` ?

packages/nuxt/src/core/nitro.ts
huang-julien: strange.... it should trigger some issues in build. 
I get a 
`[500] Cannot convert object to primitive value` at `ssrCompile()`

danielroe: in this repository test suite? or in an external project?

huang-julien: wait nvm, I did not see that the boolean has been moved to `experimental`

huang-julien: I remember that this was needed at least for nuxt 3.1 🤔 but it seems that this is not needed anymore

danielroe: wonderful! the best kind of code change...

danielroe (approved):
This looks good to me - I've tested on an external project as well, and all seems good. Very happy to iterate over v3.4 releases and then make stable in `vue.runtimeCompiler` in v3.5.

Thank you so much for your lengthy and good work on this ❤️

packages/nuxt/src/core/nitro.ts
huang-julien: I think this shouldn't be needed when having externalVue set to `true` 🤔

danielroe: perfect!

PR #4791, opened 2022-05-04
fix(nuxt): await asyncData if `ssr: false`

### LINKED ISSUE #4790: When ssr: false, useAsyncData doesn't await fetches
Same problem when nuxt is set to `ssr: false`: await is not waiting for `useFetch()` to actually fetch the data.

_Originally posted by @mario-neuhold in https://github.com/nuxt/framework/issues/4548#issuecomment-1116673434_

**Reproduction**: https://stackblitz.com/edit/github-dcgdwd?file=app.vue

danielroe:
### 🔗 Linked issue

resolves  #4790

### ❓ Type of change

### 📚 Description

We should await `useAsyncData` on initial load if there was no SSR.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -144,7 +144,7 @@ export function useAsyncData<
     if (fetchOnServer && nuxt.isHydrating && key in nuxt.payload.data) {
       // 1. Hydration (server: true): no fetch
       asyncData.pending.value = false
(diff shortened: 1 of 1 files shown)


pi0:
I was thinking if we could make fix more generic and try to refetch if for any reason key is not in the payload (it can be `ssr: false` but also timeout for fallback)

pi0:
I was thinking if we could make fix more generic and try to refetch if for any reason key is not in the payload (it can be `ssr: false` but also timeout for fallback)

PR #4822, opened 2022-05-05
fix(vite): ignore virtual imports that aren't in vfs

derz:
Avoids rollup error `plugin load hook should return a string, a { code, map } object, or nothing/null` when using (custom) virtual imports.

### 🔗 Linked issue

nuxt/nuxt.js#13921

### ❓ Type of change

### 📚 Description

Check if the virtual module exist in nuxt and otherwise just ignore it (aka leave it up for the other loaders to handle that module). Avoids returning `{ code: null, map: null }`.

Resolves nuxt/nuxt.js#13921 

### 📝 Checklist

- [] n/a. I have updated the documentation accordingly.

### DIFF
--- a/docs/content/2.guide/2.features/5.data-fetching.md
+++ b/docs/content/2.guide/2.features/5.data-fetching.md
@@ -118,13 +118,13 @@ watch(posts, (newPosts) => {
 
 Sometimes throughout the course of your user's page visit, you may need to refresh the data loaded from the API. This can happen if the user chooses to paginate, filter results, search, etc.
 
-You can make use of the `refresh()` method returned from the `useAsyncData()` composable to refresh the data with different query parameters:
+You can make use of the `refresh()` method returned from the `useFetch()` composable to refresh the data with different query parameters:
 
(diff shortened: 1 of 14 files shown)


danielroe (approved):
This seems like a good fix to me. Alternatively, we could change the prefix we're using here to `virtual:nuxt` or something.

packages/vite/src/plugins/virtual.ts
```diff
@@ -37,9 +37,11 @@ export default function virtual (vfs: Record<string, string>): Plugin {
     load (id) {
       if (!id.startsWith(PREFIX)) { return null }
       const idNoPrefix = id.slice(PREFIX.length)
-      return {
-        code: vfs[idNoPrefix],
-        map: null
+      if (vfs[idNoPrefix]) {
```
danielroe, line 40: empty files will otherwise not be served

```suggestion
      if (!(idNoPrefix in vfs)) {
```

pi0:
Thanks @derz 💯

PR #4834, opened 2022-05-06
feat(auto-import): custom include option

antfu:
### ❓ Type of change

### 📚 Description

Allowing auto import / components to work with custom extension files.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/components/loader.ts
+++ b/packages/nuxt/src/components/loader.ts
@@ -1,7 +1,7 @@
 import { pathToFileURL } from 'node:url'
 import { createUnplugin } from 'unplugin'
 import { parseQuery, parseURL } from 'ufo'
-import { Component } from '@nuxt/schema'
+import { Component, ComponentsOptions } from '@nuxt/schema'
 import { genDynamicImport, genImport } from 'knitwork'
 import MagicString from 'magic-string'
 import { pascalCase } from 'scule'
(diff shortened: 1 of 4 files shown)


pi0:
The new option looks good to me but can you please elaborate more about "to work with custom extension files" usecase?

antfu:
Here is an example, in UnoCSS's docs, we'd like to use markdown files as Vue components so we installed `vite-plugin-md` to transform md to Vue SFC and let Vue's compiler handle the rest. We could configure Vue by this so it treats `.md` as Vue SFC:

https://github.com/antfu/unocss/blob/9a2bed77e8a3433454aab66f9c3ea2b2bbb7bcf6/interactive/nuxt.config.ts#L33-L35

https://github.com/antfu/unocss/blob/6aa5f9fad9dda65ffceef1455566b4233f7db278/interactive/modules/markdown.ts

With this we could get it work, but the `.md` will not be processed by auto imports thus you lost the capability of using components in the markdown. This PR simply expose those options for that capability.

pi0:
Thanks for explaining @antfu. What do you think to integrate `include` option to auto-include [`options.extensions`](https://v3.nuxtjs.org/api/configuration/nuxt.config#extensions)

antfu:
For API auto import I think it might make sense, but probably not for Components right?

Or should we have `options.vueExtensions` that we also pass to Vue's compiler?

pi0:
> probably not for Components right?

It could be also useful for building components with markdown (https://mdxjs.com/ https://mdx.nuxtjs.org/)

packages/nuxt/src/components/loader.ts
```diff
@@ -10,61 +10,74 @@ interface LoaderOptions {
...
+  return {
+    name: 'nuxt:components-loader',
+    enforce: 'post',
+    transformInclude (id) {
+      if (exclude.some(pattern => id.match(pattern))) {
+        return false
+      }
+      if (include.some(pattern => id.match(pattern))) {
+        return true
+      }
```
pi0, line 29: This could be a nice feature of unplugin if could support `transformInclude` as an object `{ include, exclude }` and generate this logic...

pi0 (approved):

PR #4839, opened 2022-05-06
fix(nuxt): clear errors after navigation

danielroe:
### 🔗 Linked issue

extracted fixes from https://github.com/nuxt/framework/pull/4539

### ❓ Type of change

### 📚 Description

This PR fixes an issue where users who manually throw 404s can't escape from the error page (by clearing error after navigating to new route)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/runtime/router.ts
+++ b/packages/nuxt/src/pages/runtime/router.ts
@@ -107,7 +107,11 @@ export default defineNuxtPlugin(async (nuxtApp) => {
     named: {}
   }
 
-  router.afterEach((to) => {
+  router.afterEach(async (to) => {
+    if (process.client && !nuxtApp.isHydrating) {
+      // Clear any existing errors
+      await callWithNuxt(nuxtApp, clearError)
+    }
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -107,7 +107,11 @@ export default defineNuxtPlugin(async (nuxtApp) => {
     named: {}
   }
 
-  router.afterEach((to) => {
+  router.afterEach(async (to) => {
+    if (process.client && !nuxtApp.isHydrating) {
```
pi0, line 111: Shall we call it after 404 check? Also small perf to check there is an error before attempting to clear it...

pi0: Sure but i mean if we are throwing a new 404, what is the point of clearing the previous one?

danielroe: It keeps the behaviour consistent (we call a hook when error is cleared, which user might want to hook into).

PR #4841, opened 2022-05-06
fix(nuxt): normalise window location for universal router

danielroe:
### ❓ Type of change

### 📚 Description

currently `route.fullPath` is `/` on server and `http://localhost:3000/` on client. This implements same behaviour as on router, which also fixes an issue where baseURL wasn't being correctly handled with universal router:

https://github.com/nuxt/framework/blob/ebc27ce997c80f5c30fc56ad82b02785fdf482bc/packages/nuxt/src/pages/runtime/router.ts#L30-L49

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/plugins/router.ts
+++ b/packages/nuxt/src/app/plugins/router.ts
@@ -1,9 +1,9 @@
 import { reactive, h } from 'vue'
-import { parseURL, parseQuery } from 'ufo'
+import { parseURL, parseQuery, withoutBase } from 'ufo'
 import { createError } from 'h3'
 import { defineNuxtPlugin } from '..'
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/plugins/router.ts
```diff
@@ -85,8 +85,14 @@ interface Router {
   removeRoute: (name: string) => void
 }
 
+function createCurrentLocation (base: string, location: Location): string {
```
pi0, line 88: This utility is any way to get **current** window location. Can be simplified with `getCurrentWindowURL(base?)` maybe? I would unify all with `requrl` but maybe later,,,

pi0 (approved):

PR #4849, opened 2022-05-06
fix(schema): add back `PublicRuntimeConfig` interface for augmentation

### LINKED ISSUE #4550: Typing of useRuntimeConfig.public
### Environment

- Operating System: `Linux`
- Builder:          `vite`
- User Config:      `target`, `bridge`, `vite`, `alias`, `nitro`, `ssr`, `meta`, `css`, `tailwindcss`, `modules`, `watch`, `runtimeConfig`, `build`, `storybook`
- Runtime Modules:  `@nuxtjs/tailwindcss@5.0.2`, `@pinia/nuxt@0.1.8`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-ygh9qg?file=test.ts

### Describe the bug

If you provide a custom typescript declaration of the configuration as described in the [docs](https://v3.nuxtjs.org/guide/features/runtime-config#manually-typing-runtime-config),
```ts
declare module '@nuxt/schema' {
  interface RuntimeConfig {
    apiSecret: string
    public: {
      apiBase: string
    }
  }
}
```
then the return value of `useRuntimeConfig()` correctly has an `apiSecret` of type `string` but `pubic` is still `RuntimeConfigNamespace` as declared in `@nuxt/schema/dist/index.d.ts`, and thus has no `apiBase`. The reason is that typescript allows to add new members via type augmentation, but not to overwrite types.

Similarly, with no explicit declaration, nuxt is able to correctly determine the type of the config and put it in `.nuxt/types/schema.d.ts`, but the type of `public` declared there is ignored and `useRuntimeConfig().public` is still the generic `RuntimeConfigNamespace`.

### Additional context

Moreover, one cannot strictly type the return value of `useRuntimeConfig` to only include certain fields. For example, with the above code `useRuntimeConfig().notExisting` is fine, although its not declared in `RuntimeConfig`.
 
### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #4550

### ❓ Type of change

### 📚 Description

This PR restores the `PublicRuntimeConfig` interface, as a vehicle for augmenting types for runtimeConfig['public'] which otherwise can't really be done. (Otherwise each augmentation would completely override the type of `public`.)

Particularly value your input @pi0 as you touched this last.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -14,7 +14,7 @@ export interface NuxtOptions extends ConfigSchema {
   _layers: ResolvedConfig<NuxtConfig>[]
 }
 
-type RuntimeConfigNamespace = Record<string, any >
+type RuntimeConfigNamespace = Record<string, any>
 
 export interface PublicRuntimeConfig extends RuntimeConfigNamespace { }
 
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -14,17 +14,13 @@ export interface NuxtOptions extends ConfigSchema {
   _layers: ResolvedConfig<NuxtConfig>[]
 }
 
-type RuntimeConfigNamespace = Record<string, any>
+type RuntimeConfigNamespace = Record<string, any >
 
-/** @deprecated use RuntimeConfig interface */
 export interface PublicRuntimeConfig extends RuntimeConfigNamespace { }
 
 /** @deprecated use RuntimeConfig interface */
-export interface PrivateRuntimeConfig extends PublicRuntimeConfig { }
+export interface PrivateRuntimeConfig extends RuntimeConfigNamespace { }
 
-type LegacyRuntimeConfig = PublicRuntimeConfig & Partial<PrivateRuntimeConfig>
-
-export interface RuntimeConfig extends LegacyRuntimeConfig, RuntimeConfigNamespace {
-  app: RuntimeConfigNamespace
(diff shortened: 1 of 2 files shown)


pi0 (approved):

packages/schema/src/types/config.ts
```diff
@@ -16,15 +16,11 @@ export interface NuxtOptions extends ConfigSchema {
...
 type RuntimeConfigNamespace = Record<string, any>
 
-/** @deprecated use RuntimeConfig interface */
 export interface PublicRuntimeConfig extends RuntimeConfigNamespace { }
 
 /** @deprecated use RuntimeConfig interface */
-export interface PrivateRuntimeConfig extends PublicRuntimeConfig { }
-
-type LegacyRuntimeConfig = PublicRuntimeConfig & Partial<PrivateRuntimeConfig>
+export interface PrivateRuntimeConfig extends RuntimeConfigNamespace { }
```
pi0, line 22: Why would we need this? Having a shortcut to augment `public` namespace makes sense but normally we can augment the rest using `RuntimeConfig`

danielroe: I think it is only here to avoid a breaking change, but if we're happy to do so, we can remove the legacy type support as well.

pi0: I'm fine with keeping as deprecated until 3.0.0

pi0 (approved):

PR #4906, opened 2022-05-10
fix(vite, kit, nuxt): generate safe variable names using knitwork

danielroe:
### ❓ Type of change

### 📚 Description

Previously a file like `1.test.global.ts` produced an import name of `1Test` which isn't valid.

### 📝 Checklist


packages/nuxt/src/pages/utils.ts
```diff
@@ -276,7 +276,7 @@ function hasSuffix (path: string, suffix: string) {
 }
 
 export function getImportName (name: string) {
-  return pascalCase(name).replace(/[^\w]/g, r => '_' + r.charCodeAt(0))
+  return pascalCase(name).replace(/[^a-zA-Z]/g, r => '_' + r.charCodeAt(0))
```
pi0, line 279: What about adding a prefix like `_`? Small thing about it, it makes `1Bar` and `1Baz` conflicting

pi0 (approved):

pi0:
This a valid and nice fix but pending to finally refactor import name normalizer out as is a comment requirement in many places.

PR #4949, opened 2022-05-12
docs(api): add docs for `useRouter` composable

Krutie:
### ❓ Type of change

### 📚 Description

Added API documentation for `useRouter` page

### 📝 Checklist


docs/content/3.api/1.composables/use-router.md
```diff
@@ -1,7 +1,91 @@
...
+```html [~/pages/*.vue]
+
+<script setup>
+ const router = useRouter();
+ router.push({ path: "/home" });
+</script>
+
+````
+
+ `useRouter` provides the following helper methods that we can roughly divide into four groups.
```
pi0, line 18: Would be nice to add a readmore linking to `vue-router` docs: https://router.vuejs.org/api/#currentroute

docs/content/3.api/1.composables/use-router.md
```diff
@@ -1,7 +1,91 @@
...
+router.go();
+router.push({ path: "/home" });
+router.replace({ hash: "#bio" });
+````
+
+::ReadMore{link="https://developer.mozilla.org/en-US/docs/Web/API/History"}
 ::
 
-::NeedContribution
+## The ones with the guards
```
pi0, line 49: We probably wouldn't want to document this. Not only because vue-router already has updated docs (https://router.vuejs.org/guide/advanced/navigation-guards.html) but also Nuxt has Route middleware concept that is a wrapper on this which users should normally use which we can link (https://v3.nuxtjs.org/guide/directory-structure/middleware)

docs/content/3.api/1.composables/use-router.md
Krutie: ✅  Link added

docs/content/3.api/1.composables/use-router.md
Krutie: ✅  Navigation guard section revised

Atinux (approved):

pi0 (approved):

Issue #4017, opened 2022-04-01
`useFetch` data return null with mismatching client/request options


misaon:
### Environment

```
------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-27480123.4c77c88`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------
```

### Reproduction

In clean Nuxt project:

```vue
<template>
  <div>
    <NuxtWelcome />
  </div>
</template>

<script lang="ts" setup>
const [{ data: organization }, { data: repos }] = await Promise.all([
  useFetch(`https://api.github.com/orgs/nuxt`),
  useFetch(`https://api.github.com/orgs/nuxt/repos`)
])

console.log(organization.value) // null

const { data: organization } = await useFetch(`https://api.github.com/orgs/nuxt`)
console.log(organization.value) // null
</script>
```

### Describe the bug

useFetch started returning null to the "data" property

danielroe:
You are redeclaring `organization`. Once fixed, I can't replicate this. Would you provide a reproduction if that doesn't fix it for you? 🙏

misaon:
@danielroe Try this:

```
<template>
  <pre>
    {{ xxx }}
  </pre>
  <pre>
    {{ yyy }}
  </pre>
</template>

<script lang="ts" setup>
const [{ data: xxx }, { data: yyy }] = await Promise.all([
  useFetch(`https://jsonplaceholder.typicode.com/todos/1`),
  useFetch(`https://api.github.com/orgs/nuxt/repos`)
])

console.log(xxx.value) // { userId: 1, id: 1, title: 'delectus aut autem', completed: false }
console.log(yyy.value) // null
</script>
```

danielroe:
Also works fine for me. Are you sure you aren't rate limited by GitHub or something?

misaon:
@danielroe 

I've spent time testing the problem now. I am developing a locally based self-hosted certificate. If I use the url address on our production API for `useFetch`, everything works fine. If I give the local URL, it returns null. Also interesting is that `$fetch`, returns **completely correct data** and `useFetch` returns null. therefore, `useFetch` must have extra logic that returns null instead of data in some cases.

```
  const { data: countries } = await useFetch(`https://shop.localhost/api/v2/shop/countries?page=1&itemsPerPage=30`)
[...]

PR #4955, opened 2022-05-12
feat(nuxt): automatically generate unique keys for keyed composables

### LINKED ISSUE #4952: automatically generate unique keys for keyed composables
We have an opportunity to improve DX by adding transform plugins for vite & webpack to automatically generate a fallback key at compile time that can be used if users have not added keys themselves. These keys will be specific to the _file_ the call to the keyed composable is located in.

## `useFetch`

This is particularly relevant, given issues like https://github.com/nuxt/framework/issues/4917 and https://github.com/nuxt/framework/issues/4017. With `useFetch`, we generate key based on options, but sometimes these options may be functions, and then we get mismatches between client/server because the code is transpiled slightly differently and `.toString()` doesn't yield the same result.

### Usage

```vue
<script setup>
// these will now work as expected, though they will not share state with identical call
// to useFetch in another component
const { data } = await useFetch(() => 'https://my.api.com')
const { data } = await useFetch('/api/test', { transform: data => data.test })
</script>
```

##  `useAsyncData`

This allows a convenience factor, where a user can avoid generating keys for one-off calls to `useAsyncData` (or export a composable that auto-generates a shared key used between all calls to it.

### Usage

```vue
<script setup>
// this is automatically keyed to this individual component,
// wherever it is used in the app
const { data } = await useAsyncData(async () => ({ foo: 'bar' }))
</script>
```

## `useState`

This is more DX. It saves users having to create keys for state. (Alternative would be to add an alias for `useState` called `useLocalState` that has this behaviour, and require users create keys for `useState` if we feel it's in any way confusing to allow omitting the key.)

#### Composable

```js
// this will expose a unique state - every time it's imported from this file it will access the same state
export const useMyState = () => useState()
```

#### Component

```vue
<script setup>
// this state will be unique to this component, everywhere it is used in the app
const a = useState()
</script>
```

danielroe:
### 🔗 Linked issue

resolves #4952, resolves https://github.com/nuxt/framework/issues/4917, resolves https://github.com/nuxt/framework/issues/4017

### ❓ Type of change

### 📚 Description

This PR automatically generates fallback keys for `useState`, `useAsyncData`, `useFetch`, `useLazyAsyncData` and `useLazyFetch`. If no key is provided by the user, or one can't be used (if, for example, functions are passed in the options to `useFetch` or `useLazyFetch`) then the fallback key is used instead.

Fallback keys are unique to the file & line number of the call to one of the keyed composables above. (So if a user uses one of them in their own composable, the fallback key will be shared across all possible calls to this composable, rather than being unique to each call.)

I've also updated the types in the documentation for `useAsyncData` and `useFetch` where there were some typos.

~~**Further note**: It would be nice to have native support in `unplugin` for parsing using acorn-compatible AST: https://github.com/unjs/unplugin/issues/90. When that's possible, we can refactor this slightly into a native unplugin rather than just using it as a utility within the webpack builder...~~

### 📝 Checklist

- [ ] examples updated


.eslintrc
```diff
@@ -22,6 +22,7 @@
     "jsdoc/require-param": "off",
     "jsdoc/require-returns": "off",
     "jsdoc/require-param-type": "off",
+    "no-redeclare": "off",
```
pi0, line 25: Why?

docs/content/3.api/1.composables/use-async-data.md
```diff
@@ -5,13 +5,17 @@ Within your pages, components, and plugins you can use useAsyncData to get acces
...
+  handler: (nuxtApp?: NuxtApp) => Promise<DataT>,
+  options?: AsyncDataOptions<DataT>
+): AsyncData<DataT>
 function useAsyncData(
   key: string,
   handler: (nuxtApp?: NuxtApp) => Promise<DataT>,
-  options?: AsyncDataOptions
-): Promise<DataT>
+  options?: AsyncDataOptions<DataT>
+): AsyncData<DataT>
```
pi0, line 16: Type of `AsyncData` is not declared as promise in the docs below. Maybe this could be 

```suggestion
): Promise<AsyncData<DataT>>
```

docs/content/3.api/1.composables/use-async-data.md
```diff
@@ -31,7 +35,7 @@ type DataT = {
 
 ## Params
 
-* **key**: a unique key to ensure that data fetching can be properly de-duplicated across requests
+* **key**: a unique key to ensure that data fetching can be properly de-duplicated across requests. If you do not provide a key, then a key that is unique to the file and line number of the instance of `useAsyncData` will be generated for you.
```
pi0, line 38:
```suggestion
* **key**: a unique key to ensure that data fetching can be properly de-duplicated across requests. If you do not provide a key, then a key that is unique to the file path and line number of the instance of `useAsyncData` will be generated for you.
```

packages/vite/src/plugins/magic-keys.ts
```diff
@@ -0,0 +1,61 @@
+import crypto from 'node:crypto'
```
pi0, line 1: Naming it "maigc keys" is cool but also contextless! Maybe something like `auto-fetch-key` or something like that?

packages/vite/src/plugins/magic-keys.ts
```diff
@@ -0,0 +1,61 @@
...
+  'useState', 'useFetch', 'useAsyncData', 'useLazyAsyncData', 'useLazyFetch'
+]
+
+export const magicKeysPlugin = (options: MagicKeysOptions = {}): Plugin => {
+  return {
+    name: 'nuxt:magic-keys',
+    enforce: 'post',
+    transform (code, id) {
+      const { pathname } = parseURL(decodeURIComponent(pathToFileURL(id).href))
+      if (!pathname.match(/\.([cm][jt]sx?|vue)/)) { return }
```
pi0, line 31: I suppose we do not support `.cjs` parsing (does `.ts` and `.tsx` work with acorn?)

packages/vite/src/plugins/magic-keys.ts
```diff
@@ -0,0 +1,61 @@
...
+      if (!keyedFunctions.some(f => code.includes(f))) { return }
```
pi0, line 32: Perf: We could use a regex (or pre-compile one) to avoid reiterating the string for each keyword)

packages/vite/src/plugins/magic-keys.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface MagicKeysOptions {
+  sourcemap?: boolean
+  useAcorn?: boolean
+}
+
+function createKey (source: string) {
+  const hash = crypto.createHash('md5')
+  hash.update(source)
+  return hash.digest('base64').toString()
+}
```
pi0, line 19: Perf: A 128bit md5 hash is probably most secure but it also adds 20-32 characters for each payload key which is a LOT! We can use short hashes and only for extra security (which chance is close zero since change frequently on similar inputs is high) we can use a top level Map of already used hashes and detect conflicts.

packages/vite/src/plugins/magic-keys.ts
```diff
@@ -0,0 +1,61 @@
...
+          ecmaVersion: 'latest'
+        })
+        : this.parse(script), {
+        enter (node: CallExpression) {
+          if (node.type !== 'CallExpression' || node.callee.type !== 'Identifier') { return }
+          if (keyedFunctions.includes(node.callee.name)) {
+            const end = (node as any).end
+            s.appendLeft(
+              codeIndex + end - 1,
+              (node.arguments.length ? ', ' : '') + "'" + createKey(`${id}-${codeIndex + end}`) + "'"
```
pi0, line 48: We should normalize the id:
- Windows paths (if unplugin already is doing that, is perfect!)
- Make them relative to the rootDir for more predictable builds on different envs

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -48,8 +56,21 @@ export function useAsyncData<
...
+  Transform extends _Transform<DataT> = _Transform<DataT, DataT>,
+  PickKeys extends KeyOfRes<Transform> = KeyOfRes<Transform>
+> (
+  _key: string | ((ctx?: NuxtApp) => Promise<DataT>),
+  _handler?: ((ctx?: NuxtApp) => Promise<DataT>) | AsyncDataOptions<DataT, Transform, PickKeys> | string,
+  _options?: AsyncDataOptions<DataT, Transform, PickKeys> | string,
+  fallback?: string
 ): AsyncData<PickFrom<ReturnType<Transform>, PickKeys>, DataE | null | true> {
+  // eslint-disable-next-line prefer-const
+  let [key, handler, options] = (typeof _key === 'string' ? [_key, _handler, _options] : typeof _options === 'string' ? [_options, _key, {}] : [fallback, _key, _options]) as [string, (ctx?: NuxtApp) => Promise<DataT>, AsyncDataOptions<DataT, Transform, PickKeys>]
```
pi0, line 73: Perf: We don't need array creation and shift on normal `(string, fn, obj?)` usage.

We can use a variable swap for `(fn, obj?)` (is this even valid and handled when magic-keys pluign is not working ?!)

.eslintrc
danielroe: `no-redeclare` isn't TypeScript aware and otherwise has an issue with function overloads.

packages/nuxt/src/app/composables/asyncData.ts
danielroe: Yes, this works even if the plugin isn't working, as long as the user provides a key. If they don't, they will get an error message: 'asyncData key must be a string'.

danielroe: And you're right, we could swap variables around, but we would lose internal typing for those variables by doing it that way.

packages/vite/src/plugins/magic-keys.ts
danielroe: Yes, it works with `.ts` and `.tsx` 👍

packages/vite/src/plugins/magic-keys.ts
pi0: reopened i guess we are still having same logic in new file (https://github.com/nuxt/framework/pull/4955/files#diff-6206f97f029312e11fc580c3a5a948cecae20c211c72ea2315907be576594a91R19)

danielroe: What is the issue?

pi0: Shall we replace it with ohash for shorter implementation and consistency? I'm even up to introduce `safeHash` from `ohash` so we can use it in different places.

danielroe: Sure.

danielroe: ~> https://github.com/unjs/ohash/pull/9

pi0: Thanks! In the meantime, we mighy directly use hash. Chance of collision is almost impossible. Hash function is designed to have gigh entropy with similar (textual) inputs.

danielroe:
@pi0 Let me know when you plan to review this and I will resolve conflicts.

pi0:
You can resolve :)

danielroe:
Redoing merge as we can drop https://github.com/nuxt/framework/pull/5710 in this PR.

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -70,24 +70,19 @@ export function useAsyncData<
...
-    _key: string | ((ctx?: NuxtApp) => Promise<DataT>),
-    _handler?: ((ctx?: NuxtApp) => Promise<DataT>) | AsyncDataOptions<DataT, Transform, PickKeys> | string,
-    _options?: AsyncDataOptions<DataT, Transform, PickKeys> | string,
-    fallback?: string
-  ]
-): AsyncData<PickFrom<ReturnType<Transform>, PickKeys>, DataE | null | true> {
-  if (typeof args[0] !== 'string') { args.unshift(args.pop()) }
+> (...args): AsyncData<PickFrom<ReturnType<Transform>, PickKeys>, DataE | null | true> {
+  const autoKey = typeof args[args.length - 1] === 'string' ? args.pop() : undefined
+  if (typeof args[0] !== 'string') { args.unshift(autoKey) }
```
danielroe, line 75: By removing this final overload in each of these composables we lose typing within our own code, which increases the likelihood of errors internally. Was there a reason you removed it @pi0?

PR #4983, opened 2022-05-14
fix(nuxt): sort routes with dynamic params at the end

### LINKED ISSUE #5367: `[...slug]` pages has higher priority in sub directories
### Environment

```
- Operating System: `Linux`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
```

### Reproduction

https://stackblitz.com/edit/nuxt-starter-r2xyxg

### Describe the bug

Having a `[...slug]` page in a subdirectory (let's say `blog/[...slug].vue` had higher priority than other routes that were defined in the same directory

For example I have `blog/index.vue` and `blog/[...slug].vue`, visiting `http://localhost:3000/blog` always renders `blog/[...slug].vue`.
Non-dynamic routes should have higher priority than dynamic routes like `[...slug]`.

Just to mention, if I move `index.vue` and `[...slug].vue` out of the subdirectory and place them inside `pages` dir, it works as expected.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #5367

### ❓ Type of change

### 📚 Description

Currently if there is a catchall `[...slug].vue` _and_ and `index.vue` in a directory, the index will never be hit. (There are also some other variants of this issue.

We can resolve by sorting lists of routes to put dynamic parameters at the end of the list.

### 📝 Checklist


danielroe:
I can't now reproduce this and wonder if it was originally reported as a side-effect of https://github.com/nuxt/framework/pull/5002.

pi0:
Since we already discussed, isn't it something vue-router has to automatically resolve using route scores? /cc @posva

packages/nuxt/src/pages/utils.ts
```diff
@@ -88,11 +88,14 @@ export function generateRoutesFromFiles (files: string[], pagesDir: string): Nux
     }
 
     parent.push(route)
+    parent.sort((a, b) => normalizePath(a.path).localeCompare(normalizePath(b.path)))
   }
 
   return prepareRoutes(routes)
 }
 
+const normalizePath = (path: string) => path.replace(/^\//, '').replace(/:/, 'z')
```
pi0, line 97: Using normalize might be a little bit misleading since we are normalizing for sort only. Also we might use `~` (asci 126 > z)

packages/nuxt/src/pages/utils.ts
```diff
@@ -88,11 +88,14 @@ export function generateRoutesFromFiles (files: string[], pagesDir: string): Nux
     }
 
     parent.push(route)
+    parent.sort((a, b) => normalizePath(a.path).localeCompare(normalizePath(b.path)))
```
pi0, line 91: A TODO comment (ideally linking to upstream would be nice)

pi0 (approved):

pi0:
@danielroe Can you please check the failed test?

packages/nuxt/src/pages/utils.ts
```diff
@@ -88,11 +88,15 @@ export function generateRoutesFromFiles (files: string[], pagesDir: string): Nux
...
 
     parent.push(route)
+    // TODO: https://github.com/vuejs/router/issues/1435
+    parent.sort((a, b) => getSortablePath(a.path).localeCompare(getSortablePath(b.path)))
   }
 
   return prepareRoutes(routes)
 }
 
+const getSortablePath = (path: string) => path.replace(/^\//, '').replace(/:/, 'z')
```
pi0, line 98: Z is after z. At least a better workaround with localeCompare 
```suggestion
const getSortablePath = (path: string) => path.replace(/^\//, '').replace(/:/, 'Z')
```

PR #5017, opened 2022-05-17
fix(vite-node): do not cache manifest loading

antfu:
### 🔗 Linked issue

With the previous cache change in rc.3, the manifest is always cached with invalidation, causing FOUC and development changes never updates. This PR partially fixed it by opt-out renderer cache in developement. 

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -43,13 +43,15 @@ interface RenderResult {
 }
 
 // @ts-ignore
-const getClientManifest = () => import('#build/dist/server/client.manifest.mjs').then(r => r.default || r)
+const getClientManifest = () => import('#build/dist/server/client.manifest.mjs')
+  .then(r => r.default || r)
+  .then(r => typeof r === 'function' ? r() : r)
 
 // @ts-ignore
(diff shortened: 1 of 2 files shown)


pi0:
This is a nice fix but also means we have to initialize whole renderer instance each time during development mode which is a performance overhead. Maybe we can allow manifest as function within `vue-bundle-renderer`?

antfu:
Yes, we will need https://github.com/nuxt-contrib/vue-bundle-renderer/pull/31, we could have it in `vue-bundle-renderer` first and then I will update this PR

pi0:
Upstream PR released :) https://github.com/nuxt-contrib/vue-bundle-renderer

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -46,15 +48,22 @@ const getSSRRenderer = lazyCachedFunction(async () => {
...
+  const renderer = createRenderer(createSSRApp, {
     clientManifest,
     renderToString,
     publicPath: buildAssetsURL()
   })
+
+  async function renderToString (input, context) {
+    const html = await _renderToString(input, context)
+    // In development with vite-node, the manifest is on-demand and will be available after rendering
+    if (process.env.NUXT_VITE_NODE_OPTIONS) {
```
pi0, line 60: This allows tree-shaking this part:

```suggestion
    if (process.dev && process.env.NUXT_VITE_NODE_OPTIONS) {
```

PR #5022, opened 2022-05-17
feat(nuxt): `navigateTo` supports external redirects

### LINKED ISSUE #5247: Imposible redirect to an external address in page middleware
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-rc.3-27567768.c1f728e`
- Builder:          `vite`
- User Config:      `telemetry`, `modules`, `trpc`, `typescript`, `runtimeConfig`, `css`
- Runtime Modules:  `trpc-nuxt@0.1.23`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-qchxta?file=pages/awesome-nuxt.vue

### Describe the bug

There seems to be no way to redirect to an external resource in the middle of the meddleware. navigateTo` doesn't support external URLs

### Additional context

_No response_

### Logs

_No response_

manniL:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/4997, resolves #5247

### ❓ Type of change

### 📚 Description

Right now, `navigateTo` supports only internal redirects (either via `vue-router` or by using the `baseURL`).
This PR adds rudimentary support for external redirects and protocol checking:

1. It checks if the link passed in is external (if it has a protocol based on `ufo`)
2. If the protocol is `script:`, Nuxt will immediately throw an error on client- and server-side
3. If `allowExternal` is not set to true, then a confirmation dialog will pop up on client-side and an error will be thrown on server-side
4. If it is set to true it'll use `h3` on the server-side or `location.replace`/`location.href = X` on the client-side

Also feel free to tell me if that's the right approach or if we want another function for this behavior 🤔
Doc update will follow when the PR is more mature.

### 📝 Checklist


manniL:
> Should it only allow http/https based urls? Or would mailto and tel also need to work?
> 
> A check that does include mailto and tel.
> 
> ```ts
> try {
> 	url = new URL(string);
> 	//url route
> } catch (_) {
> 	//normal route
> }
> ```

Great idea! Will incorporate that in the code ☺️ 


> Maybe it would be even better to have no check and have a additional entry in the to object so that the following would work:
> 
> ```ts
> navigateTo({url: "https://github.com"});
> ```

That would be one option but would leave the "check" up to the user. IMO the DX is better when Nuxt is doing the check for the user instead of leaving it up to them.

packages/nuxt/src/app/composables/router.ts
```diff
@@ -56,22 +56,56 @@ export interface NavigateToOptions {
...
+
 export const navigateTo = (to: RouteLocationRaw, options: NavigateToOptions = {}): Promise<void | NavigationFailure> | RouteLocationRaw => {
   if (!to) {
     to = '/'
   }
+
+  const path = getPath(to)
+  const isExternal = isExternalLink(path)
+
   if (isProcessingMiddleware()) {
```
danielroe, line 87: I don't believe vue-router supports redirection to an external URL, so we would need to handle this in the same way as below...

```suggestion
  if (!isExternal && isProcessingMiddleware()) {
```

manniL: Sounds good! Added it to the PR 👍

packages/nuxt/src/app/composables/router.ts
```diff
@@ -56,22 +56,56 @@ export interface NavigateToOptions {
...
+  if (typeof to === 'string') {
+    return to
+  }
+  if ('path' in to) {
+    return to.path
+  }
+  return ''
+}
+
+const isExternalLink = (str: string): boolean => {
```
pi0, line 69: We shall use `h3.hasProtocol` to be consistant with `NuxtLink` behavior (https://github.com/nuxt/framework/blob/38a0dddd24658cbbef347c04e6fa20da4486c6fa/packages/nuxt/src/app/components/nuxt-link.ts#L144)

packages/nuxt/src/app/composables/router.ts
```diff
@@ -56,22 +56,56 @@ export interface NavigateToOptions {
   redirectCode?: number
 }
 
+const getPath = (to: RouteLocationRaw): string => {
+  if (typeof to === 'string') {
+    return to
+  }
+  if ('path' in to) {
+    return to.path
```
pi0, line 64: This could be simplified to `return to.path || '/'` also adding fallback for empty string paths.

packages/nuxt/src/app/composables/router.ts
```diff
@@ -56,22 +56,56 @@ export interface NavigateToOptions {
...
+  }
+}
+
 export const navigateTo = (to: RouteLocationRaw, options: NavigateToOptions = {}): Promise<void | NavigationFailure> | RouteLocationRaw => {
   if (!to) {
     to = '/'
   }
-  if (isProcessingMiddleware()) {
+
+  const path = getPath(to)
```
pi0, line 84: Code style: I would call it `const toPath`

pi0:
Thanks for this PR @manniL 💚 In general this is a really nice improvement that we support redirection to external URLs but I'm also little bit concerned about potential security risks implied by combination of this new feature and developer's mistake with a programmatic call. I  suggest some options:

- Explicit protocol validation to avoid things like `script:` 
- A new option `externalURL: ''` / `external: true` and abort when `url` is external with an error to use explicit option for external URL. (We might also use a simple alert/prompt fallback to concenst users about redireting when `url` is implicitly external)

packages/nuxt/src/app/composables/router.ts
manniL: Sadly not due to `path` not being part of the last union of `RouteLocationRaw` (which is `(RouteQueryAndHash & LocationAsRelativeRaw & RouteLocationOptions)`)

I can add the fallback though instead of an empty string 👍

pi0: Regardless we are currently returning `to.path` here without typechecking. If type is an issue, we can force cast it. Or I might misunderstand you... :)

packages/nuxt/src/app/composables/router.ts
manniL: You mean `ufo.hasProtocol`, right?

packages/nuxt/src/app/composables/router.ts
manniL: Will force cast it then 😋

manniL:
> Thanks for this PR @manniL 💚 In general this is a really nice improvement that we support redirection to external URLs but I'm also little bit concerned about potential security risks implied by combination of this new feature and developer's mistake with a programmatic call. I suggest some options:
> 
>     * Explicit protocol validation to avoid things like `script:`
> 
>     * A new option `externalURL: ''` / `external: true` and abort when `url` is external with an error to use explicit option for external URL. (We might also use a simple alert/prompt fallback to concenst users about redireting when `url` is implicitly external)

Thanks for the feedback! 🙏 
I agree regarding some security implications here. DX-wise, the protocol validation part would be more appealing IMO but from a security POV, a new option would make more sense.

What would be your favorite solution here?

pi0:
I would probably do both:

- Protocol restriction in all cases
- Abort SSR / Prompt Client side with `url: 'https://...'`
- Introduce `external: true` or `externalURL: string` (Which one you think is better?)

manniL:
>     * Protocol restriction in all cases

That means not executing any links that start with `script:`. Any other protocols we should consider?

>     * Abort SSR / Prompt Client side with `url: 'https://...'`

By throwing an error with the approriate message, right?

>     * Introduce `external: true` or `externalURL: string` (Which one you think is better?)

I'd probably go with `external: true`. This param won't have any influence when the URL is local and will only matter in case the `path` is actually an external one, right?

pi0:
> That means not executing any links that start with script:. Any other protocols we should consider?

Considering the possibility of native app linking with custom protocols, yes the best we can _exclude_ unsafe `script:` protocol.

> By throwing an error with the appropriate message, right?

Would be an overkill to use good-old [window.confirm](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) as the default fallback? For server yes error is the best!

> I'd probably go with external: true. This param won't have any influence when the URL is local and will only matter in case the path is actually an external one, right?

`external: true` seems good to me too! We can document to use `external` only if you are sure external URLs are safe.

cawa-93:
I will add a little feedback as developer who faced with this problem. In general, reading the documentation, one gets the impression that the `<nuxt-link>`and `navigateTo` work identically. We **can** easily pass the external url to the `<nuxt-link to="https://github.com">`. And I would expect the `navigateTo` to take exactly the same parameters and process them identically

manniL:
Updated the PR with the recommended changes 😋

justin-schroeder:
Until this is merged — I've worked up a small composable function to allow explicit external redirects from route middleware: https://gist.github.com/justin-schroeder/f583797c702155f32c9078d1add2a594

serhez:
Hello! Are there any updates on this? It is quite critical for important workflows like OAuth 2.0 (being able to follow a server redirection to the login form, etc.), unless there is some other way to do it I missed.

OhB00:
Found an interesting workaround for this one, prepending your URL with `//` enables you to perform an external redirect. For example, to redirect to http://google.com use `navigateTo("//google.com")`. This works provided your baseUrl is `/` as ufo will strip a slash from the start of your path.

Works because of parser differentials between the browser and vue-router/ufo.

manniL:
@pi0 Please let me know what is left to do here besides the docs PR ☺️

packages/nuxt/src/app/composables/router.ts
pi0: I see but in this case `LocationAsRelativeRaw` `if ('path' in to)` gets bypassed and we end up with same `/` fallback.

pi0 (approved):

pi0:
Thanks again for PR @manniL and sorry it took long. Finally, I've made few refactors to inline and simplify logic, avoid extra parse calls, improve messages and make the option simpler `{ external: true }`. I hope this changes are fine to you.

packages/nuxt/src/app/composables/router.ts
```diff
@@ -58,26 +58,49 @@ const isProcessingMiddleware = () => {
...
   }
+
   // Client-side redirection using vue-router
+  if (isExternal) {
+    if (options.replace) {
+      location.replace(toPath)
+    } else {
+      location.href = toPath
+    }
+    return Promise.resolve()
```
pi0, line 101: One last thing to ensure: In nuxt 2, we had this problem with external navigations that during navigation, rest of logic goes on. And solution was a custom made error because it was sync. I think we could do something like a never resolving promise that avoids this. (will make a reproduction and implement in later PR btw need more testing)

pi0:
(We also need to document new `external` key 🙈 )

packages/nuxt/src/app/composables/router.ts
```diff
@@ -58,26 +58,49 @@ const isProcessingMiddleware = () => {
...
 
 export const navigateTo = (to: RouteLocationRaw | undefined | null, options: NavigateToOptions = {}): Promise<void | NavigationFailure> | RouteLocationRaw => {
   if (!to) {
     to = '/'
   }
-  // Early redirect on client-side since only possible option is redirectCode and not applied
-  if (process.client && isProcessingMiddleware()) {
+
+  const toPath = typeof to === 'string' ? to : ((to as RouteLocationPathRaw).path || '/')
+  const isExternal = hasProtocol(toPath, true)
```
bars83, line 71: why you check protocol for external? may be I want to redirect within current domain, i.e. respond with headers

```
HTTP/1.1 302 Found
Location: /my-legacy-page
```

pi0: Would you please open an issue with a sandbox to track? 🙏🏼

Issue #5027, opened 2022-05-17
Support auto-imports when using `extends`


danielroe:
When extending projects in `node_modules`, auto-imports are not injected automatically. We can use https://github.com/unjs/unimport/blob/a2a2951c8a62866b7a2f6a88202de62b8d3a7355/src/unplugin.ts#L18 to do so, improving DX.

**Side-note**: as part of https://github.com/nuxt/module-builder we could possibly run a compile-time step to process auto-imports into an explicit import from '#imports'...

antfu:
What's your plan for it? Should we add the include automatically inherit from the `extends`? If so I think it makes sense.

Regards on module builder, I actually think it's better for a library/module to use explicit imports from `#imports` or somewhere rather than relying on the auto-import. But if we want, I think it would be straightforward to implement with unimport.

PR #5042, opened 2022-05-18
fix(nuxt): transform extends layers within `node_modules`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/5027

### ❓ Type of change

### 📚 Description

This PR adds support for auto-imports within extends layers, as well as adding them to `build.transpile`. ~~Incidentally, it also fixes a nasty bug with webpack where we weren't respecting `build.transpile` (I may extract the latter to a separate PR if this PR needs more attention).~~

### 📝 Checklist


packages/webpack/src/presets/nuxt.ts
```diff
@@ -10,6 +10,7 @@ import { vue } from './vue'
 export function nuxt (ctx: WebpackConfigContext) {
   applyPresets(ctx, [
     base,
+    baseTranspile,
```
pi0: I guess it comes from https://github.com/nuxt/framework/pull/5043 might be changed...

packages/nuxt/src/core/nuxt.ts
```diff
@@ -66,6 +66,11 @@ async function initNuxt (nuxt: Nuxt) {
   addVitePlugin(UnctxTransformPlugin(nuxt).vite({ sourcemap: nuxt.options.sourcemap }))
   addWebpackPlugin(UnctxTransformPlugin(nuxt).webpack({ sourcemap: nuxt.options.sourcemap }))
 
+  // Transpile layers within node_modules
+  nuxt.options.build.transpile.push(...nuxt.options._layers.map(
+    i => i.cwd && i.cwd.includes('node_modules') && (isAbsolute(i.cwd) ? relative(nuxt.options.rootDir, i.cwd) : i.cwd)
```
pi0, line 71: - Code style: I would split it into `.filter().map()` 
- Why we need to make paths relative from rootDir?

packages/nuxt/src/auto-imports/module.ts
```diff
@@ -10,15 +11,18 @@ export default defineNuxtModule<Partial<AutoImportsOptions>>({
...
   },
-  defaults: {
+  defaults: nuxt => ({
     presets: defaultPresets,
     global: false,
     imports: [],
     dirs: [],
     transform: {
+      include: nuxt.options._layers.map(
+        i => i.cwd && new RegExp(`(^|\\/)${escapeRE(i.cwd.split('node_modules/').pop())}(\\/|$)`)
```
pi0, line 21: It could be more reliable if we could check against explicit resolved paths in `transformInclude`. Also can't we use `options.build.transpile` here to generate includes?

danielroe: Good idea on resolved paths 👍 

I didn't want all transpiles to be parsed for auto-imports as I think the imports should be explicit within node_modules, apart from _extends_.

danielroe: **Note**: move this logic to nuxt/core

packages/nuxt/src/core/nuxt.ts
```diff
@@ -133,7 +139,14 @@ export async function loadNuxt (opts: LoadNuxtOptions): Promise<Nuxt> {
   // Temporary until finding better placement for each
   options.appDir = options.alias['#app'] = resolve(distDir, 'app')
   options._majorVersion = 3
-  options._modules.push(pagesModule, metaModule, componentsModule, autoImportsModule)
+  options._modules.push(pagesModule, metaModule, componentsModule)
+  options._modules.push([autoImportsModule, {
+    transform: {
+      include: options._layers
+        .filter(i => i.cwd)
+        .map(i => new RegExp(`(^|\\/)${escapeRE(i.cwd.split('node_modules/').pop())}(\\/|$)`))
```
pi0, line 147: For package called `theme`, adding include regex of `(^|\\/)theme(\\/|$)` will match `theme/node_modules/lib/index.js` as well when package-manager is not hosting their deps. Maybe add a lookforward to exclude that? (I might be wrong too about include behavior)

pi0:
@danielroe Can you please check if we also need this for nitro options to include layers?

pi0 (approved):

PR #5043, opened 2022-05-18
fix(webpack): respect `build.transpile`

danielroe:
### ❓ Type of change

### 📚 Description

Extracted from https://github.com/nuxt/framework/pull/5042, this PR respects user's `build.transpile` for forcing transpilation on files in `node_modules`.

### 📝 Checklist


packages/webpack/src/presets/nuxt.ts
```diff
@@ -10,6 +10,7 @@ import { vue } from './vue'
 export function nuxt (ctx: WebpackConfigContext) {
   applyPresets(ctx, [
     base,
+    baseTranspile,
```
pi0, line 13: We could directly use `baseTranspile` in base preset. Also I believe there was some leftover issues that didn't add transpile to the base on initial rewrite.

danielroe: Do you recall what those issues were?

pi0: Not precisely (other than TODO comment with unique in the code). Any way we can test that only I believe it had to be in the base. I exported it back in the time since wasn't sure about legacy+modern transpile but we now effectively have one target.

pi0:
Pending merge to double-check `baseTranspile` util.

Issue #4984, opened 2022-05-14
Cannot set `titleTemplate` as function


heychazza:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v14.19.2`
- Nuxt Version:     `3.0.0-rc.3`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `modules`, `target`, `app`, `content`, `css`, `build`
- Runtime Modules:  `@nuxt/content-edge@2.0.0-27541979.40705ae`
- Build Modules:    `-`
------------------------------

### Reproduction

N/A

### Describe the bug

Cannot set the `titleTemplate` as a function (used for default titles when non existent).

Example:
```js
import { defineNuxtConfig } from 'nuxt'

// https://v3.nuxtjs.org/api/configuration/nuxt.config
export default defineNuxtConfig({
    app: {
        head: {
            titleTemplate: (titleChunk) => {
                return titleChunk ? `${titleChunk} - Site Title` : 'Site Title';
            },
        }
    }
})
```

### Additional context

As far as I know, Nuxt 3 uses vue-meta, which the docs show it supports: https://vue-meta.nuxtjs.org/api/#titletemplate

heychazza:
Furthermore, it’s also referenced in the schema - as found at https://github.com/nuxt/framework/blob/1091d456a8ba6d5d0b57c37bfb959451b0cbce2b/packages/schema/src/types/meta.ts#L25

danielroe:
FYI: Nuxt 3 uses `@vueuse/head`.

However, the issue you raise is correct. `titleTemplate` as a function is not supported in your `nuxt.config`. Instead you can define it as a string - or set it in a plugin. (This is because we are deliberately trying to move functions out of config files and into runtime code.)

A PR to the docs making this clear would be welcome. 🙏

heychazza:
> FYI: Nuxt 3 uses `@vueuse/head`.
> 
> However, the issue you raise is correct. `titleTemplate` as a function is not supported in your `nuxt.config`. Instead you can define it as a string - or set it in a plugin. (This is because we are deliberately trying to move functions out of config files and into runtime code.)
> 
> A PR to the docs making this clear would be welcome. 🙏

Thanks for your reply! With a plugin, how would we set this?

heychazza:
> Like this: https://stackblitz.com/edit/github-d1ngrx?file=app.vue,plugins%2Ftitle.ts

[...]

PR #5064, opened 2022-05-19
fix(nuxt): add default empty title to trigger `titleTemplate`

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/4984

### ❓ Type of change

### 📚 Description

Previously a `titleTemplate` would only be used if there was also a `title` set (even if empty). This PR adds a basic empty title that can be overridden, meaning that titleTemplate works.

### 📝 Checklist


pi0 (approved):
LGTM. Only as discussed with @danielroe it causes an empty `<title>` tag which could be improved.

PR #5077, opened 2022-05-20
docs: Added a Dockerfile to Node.js deployment guide

mishushakov:
### 🔗 Linked issue

There is no issue open

### ❓ Type of change

### 📚 Description

I have added a Dockerfile example to the Node.js deployment guide

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/5.deploy/1.node-server.md
+++ b/docs/content/2.guide/5.deploy/1.node-server.md
@@ -49,6 +49,27 @@ module.exports = {
 }
 ```
 
+### Using Docker
+
+Create a `Dockerfile` with following contents:
+
+```Dockerfile
+# Build
+FROM node:16-alpine AS build
+WORKDIR /build
+COPY . .
+RUN npm i
+RUN npm run build
+
+# Serve
+FROM node:16-alpine AS app
+WORKDIR /app
+COPY --from=build /build/.output .
+
+EXPOSE 3000
+CMD [ "node", "./server/index.mjs" ]
+```
+
 ## Learn more
 
 :ReadMore{link="https://nitro.unjs.io/deploy/node.html" title="the Nitro documentation for node-server preset"}


docs/content/2.guide/5.deploy/1.node-server.md
```diff
@@ -49,6 +49,27 @@ module.exports = {
...
 
+### Using Docker
+
+Create a `Dockerfile` with following contents:
+
+```Dockerfile
+# Build
+FROM node:16-alpine AS build
+WORKDIR /build
+COPY . .
```
Koc, line 60: what about adding only package.json + lock in order to leverage Docker layer caching and stop executing `npm install` all the time even if nothing changed in packages?

sarim:
Maybe add another Dockerfile for yarn? If someone uses docker (like me) then copy paste this dockerfile, it won't have the lock file, resulting in npm recalculate the dependencies again.

mishushakov:
@sarim if you use yarn, npm will detect that and use your `yarn.lock` for dependency resolution

https://blog.npmjs.org/post/621733939456933888/npm-v7-series-why-keep-package-lockjson.html

sarim:
> @sarim if you use yarn, npm will detect that and use your `yarn.lock` for dependency resolution
> 
> https://blog.npmjs.org/post/621733939456933888/npm-v7-series-why-keep-package-lockjson.html
> 

But `COPY package*.json` wont copy yarn.lock.

mishushakov:
aah yes, give me a second

mishushakov:
should be fixed now

sarim:
I'm also using https://github.com/krallin/tini as entrypoint. Otherwise stop signal doesn't get processed properly and container stop takes few seconds instead of instant. That might be a bit out of scope though.

pi0:
Thanks for this PR @mishushakov ❤️ Adding official docker deployment is a most have!

While building using docker is probably best practice, in order to be consistent with other deployment methods, I would include example docker file with second stage that copies `.ouput` directory. This also simplifies complexity with choosing package manager. 

We can add another sub-section below for multi stage build.

---

Worth to mention that we plan to add a Docker preset for nitro soon (https://github.com/unjs/nitro/issues/54). This preset, generates _DockerFile_ and image after `nuxt build` outside of docker. Ideas more than welcome there as well!

Good tip @sarim 👍🏼 I've linked as comment

PR #5078, opened 2022-05-20
fix: Temporary fix for https://github.com/nuxt/framework/issues/4758

### LINKED ISSUE #4758: share useAsyncData state
### Environment

- Operating System: `Linux`
- Package Manager:  `npm@8.1.0`
- Builder:          `vite`
- User Config:      `buildModules`, `modules`, `css`, `env`, `components`, `tailwindcss`, `meta`, `runtimeConfig`
- Runtime Modules:  `-`
- Build Modules:    `@nuxtjs/tailwindcss@5.0.3`, `@pinia/nuxt@0.1.8`

### Reproduction

Example using `useAsyncData` within a composable: https://stackblitz.com/edit/nuxt-starter-bf2vum?file=app.vue
Example using `useAsyncData` directly in components: https://stackblitz.com/edit/nuxt-starter-hhuack?file=app.vue

### Describe the bug

Creating a state using `useAsyncData` within a custom composable, and using the custom composable in 2 different components does not share the state returned in `{data}` correctly.

One of these 2 components already has a state on SSR, while the second component has a `true` pending state and none data for a short time on client site rendering. When im using the composable in a component where both components are childs of that component and awaiting the `fetch` of the composable in the parent component, both child components do have the correct state in the same time thorugh the composable (on page load)

When i use `useAsyncData` directly in two components and assign the same key to `useAsyncData` it looks like the state is shared corrrectly.

In both examples using the `refresh` function within one of the components, not sharing the new collected state between these components.

### Additional context

_No response_

### Logs

_No response_

cherrymarker:
### 🔗 Linked issue

#4758 

### ❓ Type of change

I suppose this is technically a breaking change if applications are dependant on the original broken behaviour. 

### 📚 Description

I have modified useAsyncData to return the correct data associated with a request in the case that another call is made to useAsyncData while the original promise is still pending.

This change allows you to call useAsyncData on a single key, within multiple components, and work with the data on the server side without subsequent calls resolving to pending and null.

Resolves #4758 for now, more robust solution required in future.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -113,10 +113,14 @@ export function useAsyncData<
         }
         asyncData.data.value = result
         asyncData.error.value = null
+
+        return asyncData;
       })
       .catch((error: any) => {
         asyncData.error.value = error
         asyncData.data.value = unref(options.default())
+
+        return asyncData;
       })
       .finally(() => {
         asyncData.pending.value = false
@@ -166,7 +170,7 @@ export function useAsyncData<
   }
 
   // Allow directly awaiting on asyncData
-  const asyncDataPromise = Promise.resolve(nuxt._asyncDataPromises[key]).then(() => asyncData) as AsyncData<DataT, DataE>
+  const asyncDataPromise = Promise.resolve(nuxt._asyncDataPromises[key]).then(x => x) as AsyncData<DataT, DataE>
   Object.assign(asyncDataPromise, asyncData)
 
   return asyncDataPromise as AsyncData<PickFrom<ReturnType<Transform>, PickKeys>, DataE>


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -166,7 +170,7 @@ export function useAsyncData<
   }
 
   // Allow directly awaiting on asyncData
-  const asyncDataPromise = Promise.resolve(nuxt._asyncDataPromises[key]).then(() => asyncData) as AsyncData<DataT, DataE>
+  const asyncDataPromise = Promise.resolve(nuxt._asyncDataPromises[key]).then(x => x) as AsyncData<DataT, DataE>
```
danielroe, line 173:
```suggestion
  const asyncDataPromise = Promise.resolve(nuxt._asyncDataPromises[key]) as AsyncData<DataT, DataE>
```

cherrymarker:
On further inspection this fix is a bit wonky, bear with

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -113,10 +113,14 @@ export function useAsyncData<
         }
         asyncData.data.value = result
         asyncData.error.value = null
+
+        return asyncData;
```
danielroe, line 117:
```suggestion
        return asyncData
```

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -113,10 +113,14 @@ export function useAsyncData<
...
       })
       .catch((error: any) => {
         asyncData.error.value = error
         asyncData.data.value = unref(options.default())
+
+        return asyncData;
```
danielroe, line 123:
```suggestion
        return asyncData
```

danielroe:
@cherrymarker Oops, sorry, missed your comment. Will hold any further comments for now.

cherrymarker:
While this fix "sort of" works, I have realised due to this line there may be problems 
https://github.com/nuxt/framework/blob/d9b6c8a59bd4f1f1f6280e1d55d10ddd477d2eea/packages/nuxt/src/app/composables/asyncData.ts#L170

The original asyncdata is still passed along.

A better solution would be to store the asyncData, and fetch it later IF the promise has already been found.

cherrymarker:
Updated, to a better solution that works both client and server side. 

Wasn't sure about the type definition in `nuxt.ts` but gave it a go.

danielroe:
I wonder if this would be an opportunity to share asyncData across calls, which would also address a related issue where `pending` can get out of sync. Thoughts?

cherrymarker:
I think that the design of useAsyncData is a bit messy and could be simplified by just using async methods.

Instead of `data` and `error` being stored on `nuxt.payload` it would be easier to store all of asyncData on payload instead and retrieve it on a matching key. 

I think instead of having pending as a bool, a Promise that resolves when the handler is finished is far more useful and easier to use.

Atinux:
> I think that the design of useAsyncData is a bit messy and could be simplified by just using async methods.

What do you have in mind?

> Instead of data and error being stored on nuxt.payload it would be easier to store all of asyncData on payload instead and retrieve it on a matching key.

This makes sense since it will keep the scope of each one (`error` shall be used for global error handling)

> I think instead of having pending as a bool, a Promise that resolves when the handler is finished is far more useful and easier to use.

Not 100% sure about this since most people use `v-if="pending"` for state handling.

Issue #4084, opened 2022-04-05
handle vue dependencies in nuxt 3 with externals or 'smart' transpilation


danielroe:
## Summary

At the moment, in both Bridge & Nuxt 3, we transpile `vue`. This decreases the space taken up in our server builds, and also protects against multiple instances of Vue being present in development (or even in build, due to ESM resolution issues).

However, it comes with the issue that if the end user adds any libraries or other files (outside of _their_ project) that import vue, we now have multiple copies of Vue present - both the version that has been bundled and an external version in `node_modules`.

It also means that it is more difficult for users to pick the version of Vue they wish to use in their project (for example, if they want a version with a runtime compiler).

## Possible solutions

1. Users can manually add any such libraries to `build.transpile` - this is our current recommendation.
2. Building on 1, we can also use vite's dependency graph to detect nested dependencies of vue and recommend adding them to `build.transpile`
3. We can fully externalise `vue`, at the cost of greater disk space and the danger of resolving multiple vue versions depending on the ESM status of dependencies.
4. Building on 3, we can also improve our NFT support by adding 'soft-transpilation' to optimize/tree-shake externals.

## Related issues

* https://github.com/nuxt/framework/issues/3817
* https://github.com/nuxt/framework/issues/2530
* https://github.com/nuxt/framework/issues/2326#issuecomment-1026081414
* https://github.com/nuxt/framework/issues/3688
* https://github.com/nuxt/framework/issues/3162
* https://github.com/nuxt/framework/issues/2682

zefman:
Hi @danielroe should adding the library to `build.transpile` always fix the issue? I have a custom component library setup with pnpm alongside Nuxt in a mono repo and still face issues even after adding to the transpile array. The component library is very simple using vite to bundle with vue externalised etc.

dakt:
We have same issue in our project  with dev erroring out:

```
[Vue warn]: inject() can only be used inside setup() or functional components.
[Vue warn]: inject() can only be used inside setup() or functional components.
[Vue warn]: inject() can only be used inside setup() or functional components.
[Vue warn]: Unhandled error during execution of setup function
  at <RouterLink to="typography" activeClass=undefined exactActiveClass=undefined  ... >
[nuxt] [request error] Cannot destructure property 'options' of 'vue.inject(...)' as it is undefined.
```

This is only when we slot <Nuxt* /> components through our UI lib like this:

```html
<UIProvider theme="foo">
[...]

PR #5083, opened 2022-05-20
feat(nuxt): experimental flag to use external vue dependencies

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/4084

### ❓ Type of change

### 📚 Description

This PR starts exploring implementing moving `vue`, `@vue/*` and `vue-router` to externals. As it is, it is not the solution. Although it works, we have (at least) the following issues:

- [ ] multiple copies of imports are included, for almost all `vue` and `@vue` libs, meaning an increase of ~2.2M (688kB -> 2.9M)  in `node_modules` for a minimal project using `pages/` - but note that the **total** `.output/server` size _decreases_ slightly from 3.3M -> 3.2M. (**Note**: we can improve this into a much larger decrease by implementing import conditions upstream in vue.)
- [ ] there are extra, non-treeshaken bits of vue-router that we could prune (such as other forms of router history and all code behind `isBrowser` flag)
- [ ] unneeded _parts_ of the vue api, such as `@vue/devtools-api`, are included in server node_modules

Some ideas to explore:

* 'light' processing of Vue externals - extra compilation steps to strip browser or dev-only code?
* manual post-build step to remove dev cjs builds
* import conditions (development/production) for Vue core libraries

### 📝 Checklist


huang-julien:
HI @danielroe. What if we modify the externals plugin from nitro to avoid sending some files to the final bundle ? Would it work ?  We would be able to exclude the @vue/devtools-api from the server node_modules

by changing [this line](https://github.com/unjs/nitro/blob/a6aaafa57f8e0369cc09a91b5c1532c2ce570738/src/rollup/plugins/externals.ts#L149) to : 
```ts
then((r) => r.filter((file) => file.includes("node_modules") && !excludeExternals.some(name => file.includes(name)));
```


image

packages/nuxt/src/core/nitro.ts
```diff
@@ -58,16 +58,25 @@ export async function initNitro (nuxt: Nuxt) {
     sourcemap: nuxt.options.sourcemap,
     externals: {
       inline: [
-        ...(nuxt.options.dev ? [] : ['vue', '@vue/', '@nuxt/', nuxt.options.buildDir]),
+        ...(nuxt.options.dev
+          ? []
+          : [
+              ...nuxt.options.experimental.externalVue ? [] : ['vue', '@vue/'],
+              '@nuxt/', nuxt.options.buildDir
```
pi0, line 65:
```suggestion
              '@nuxt/',
              nuxt.options.buildDir
```

packages/schema/src/config/experimental.ts
```diff
@@ -17,6 +17,12 @@ export default {
...
      * @see https://vuejs.org/guide/extras/reactivity-transform.html
      */
-    reactivityTransform: false
+    reactivityTransform: false,
+
+    /**
+     * Externalize `vue`, `@vue/*` and `vue-router` when build
+     * @see https://github.com/nuxt/framework/issues/4084
+     */
+    externalVue: false
```
pi0, line 26: What do you think to generalize flag to as `externalDependencies`? So we have the chance to progressively extend it?

pi0: It is short-termed but saw `vue-router` which is not vue exactly. Just taught of using a more generic flag.

danielroe: I'd definitely be up for improving the name. But I guess it's about core vue libraries, and possibly `externalDependencies` is too broad. Can you think of a narrower name?

pi0: External libs/modules might respect same experiment ad well. When vue is externalized, they need to also avoid inline their vue libs right? BTW it doesn't matter that much for naming I'm fine with current one too if you prefer to keep

packages/vite/src/server.ts
```diff
@@ -30,16 +30,21 @@ export async function buildServer (ctx: ViteBuildContext) {
...
+          : {
+              '@vue/reactivity': _resolve(`@vue/reactivity/dist/reactivity.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`),
+              '@vue/shared': _resolve(`@vue/shared/dist/shared.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`),
+              'vue-router': _resolve(`vue-router/dist/vue-router.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`)
+            },
         vue: _resolve(`vue/dist/vue.cjs${ctx.nuxt.options.dev ? '' : '.prod'}.js`)
       }
     },
     ssr: {
+      external: ctx.nuxt.options.experimental.externalVue ? ['#internal/nitro', 'vue', 'vue-router'] : [],
```
pi0, line 47: Why `#internal/nitro` is in the condition? 🤔

danielroe: It was there previously.

danielroe, line 47:
```suggestion
      external: ctx.nuxt.options.experimental.externalVue ? ['#internal/nitro', 'vue', 'vue-router'] : ['#internal/nitro'],
```

pi0 (approved):

PR #5118, opened 2022-05-23
docs(layouts): rework the layout section, add default layout

Xanlantos:
### ❓ Type of change

- [X] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

The default layout was missing and the documentation could be simplified by adding it. See commit msg for a detailed change list. I hope to be of help, but I have never contributed to Nuxt or any other public project. Please feel free to guide me. 

### 📝 Checklist


docs/content/2.guide/3.directory-structure/6.layouts.md
```diff
@@ -8,54 +8,64 @@ head.title: Layouts directory
...
 Unlike other components, your layouts must have a single root element to allow Nuxt to apply transitions between layout changes - and this root element cannot be a `<slot />`.
 ::
 
-## Example: Enabling layouts with `app.vue`
-
-```bash
--| layouts/
----| custom.vue
--| app.vue
+## Example: Enabling the default layout
```
pi0, line 19:
```suggestion
## Example: Enabling the default layout without `~/pages`
```

docs/content/2.guide/3.directory-structure/6.layouts.md
```diff
@@ -8,54 +8,64 @@ head.title: Layouts directory
...
-    Hello world!
+  <NuxtLayout>
+    <NuxtPage/>
   </NuxtLayout>
 </template>
 ```
 
-## Example: setting the layout with `~/pages`
+
+## Example: setting another layout with `~/pages`
```
pi0, line 57:
```suggestion
## Example: setting another layout
```

docs/content/2.guide/3.directory-structure/6.layouts.md
```diff
@@ -8,54 +8,64 @@ head.title: Layouts directory
...
 Unlike other components, your layouts must have a single root element to allow Nuxt to apply transitions between layout changes - and this root element cannot be a `<slot />`.
 ::
 
-## Example: Enabling layouts with `app.vue`
-
-```bash
--| layouts/
----| custom.vue
--| app.vue
+## Example: Enabling the default layout without `~/pages`
```
Atinux, line 19:
```suggestion
## Example: Enabling the default layout without `~/pages`

```

docs/content/2.guide/3.directory-structure/6.layouts.md
```diff
@@ -8,53 +8,72 @@ head.title: Layouts directory
 
 Nuxt provides a customizable layouts framework you can use throughout your application, ideal for extracting common UI or code patterns into reusable layout components.
 
-Layouts are placed in the `layouts/` directory and will be automatically loaded via asynchronous import when used. Layouts are used by setting a `layout` property as part of your page metadata (if you are using the `~/pages` integration), or by using the `<NuxtLayout>` component.
+Layouts are placed in the `layouts/` directory and will be automatically loaded via asynchronous import when used. Layouts are used by setting a `layout` property as part of your page metadata or via the default layout `~/layouts/default.vue` (if you are using the `~/pages` integration), or by using the `<NuxtLayout>` component.
```
danielroe, line 11:
```suggestion
Layouts are placed in the `layouts/` directory and will be automatically loaded via asynchronous import when used. Layouts are used by setting a `layout` property as part of your page metadata (if you are using the `~/pages` integration), or by using the `<NuxtLayout>` component.

If you do not set a layout explicitly, then `~/layouts/default.vue` will be used.
```

danielroe (commented):
What do you think about condensing the examples down to:

1. Layouts without pages
2. Layouts with pages
3. Manual control
4. Changing layout dynamically

I don't think we need separate examples of how to use the default layout if we cover it within the first two examples.

docs/content/2.guide/3.directory-structure/6.layouts.md
Xanlantos: What would be an example for multiple Layouts without pages that is not better done by adapting the app.vue?

- A "dark/light mode" is more of a CSS thing. 
- A simple menu is better done within the app.vue. 
- A login button I rather see in the app.vue
- A Logo is also better to keep it simple and in the app.vue
- A massive complex layout is better split into multiple components anyhow

The only reasonable thing I can come up with is something like:

- two different menus for the same content, i.e. "top aligned menu" vs "sidebar"

I'd say that's a pretty niche thing, and technically you still could do it better with CSS-grid! So If I am not mistaken I'd say Layouts almost always mean you are using pages already. If I am mistaken please tell me.   But if I am right I rather put that first and make a note, towards the end of the page, about how you could do it, if you are not using pages. 

Considering that I would rework it to:

Default Layout -> cause it is the easiest way to get in
Multiple Layouts -> cause that's the next likely thing
Layouts without pages -> cause now we covered the basics and the reader does not get confused if we introduce new concepts
Changing layout dynamically -> I think this first as it is more likely to be needed in my opinion but I am fine swapping it with the last
Manual control -> This feels more like abusing the system and should only be done with proper care maybe even add a warning or at least formulate it as a "technical possibility" and less like an "actual option" an expert can ignore it anyway

Maybe the question, in general, is what are layouts for? Cause technically you can use it in various ways. You could even recreate them with components. But why should we use layouts over that? I don't know the answer on a board Nuxt level. I think layouts are nice to get rid of code duplication and have a proper space for it that is different from components. But honestly, that feels still like a bit of a wishy-washy answer. So if you know maybe you can enlighten me.

pi0: I think it is fair to say we can always replace Layouts with Wrapper dynamic components.

Main point of not using pages/ is to have our own non traditional routing. Imagine a small single page dashboard that has a login state. We can leverage `default` and `guest` layouts (or `default` and `dashboard`) that are automatically switched based on authentication state rather than page route.

docs/content/2.guide/3.directory-structure/6.layouts.md
```diff
@@ -8,53 +8,72 @@ head.title: Layouts directory
...
+```vue [layouts/default.vue]
+<template>
+  <div>
+    Some default Layout shared across all pages
+    <slot />
+  </div>
+</template>
 ```
 
-In your layout files, you'll need to use `<slot />` to define where the content of your layout will be loaded. For example:
```
pi0, line 27: Explaining about `<slot />` was nice why you think should be removed here?

docs/content/2.guide/3.directory-structure/6.layouts.md
```diff
@@ -8,53 +8,72 @@ head.title: Layouts directory
...
-  <NuxtLayout name="custom">
-    Hello world!
+  <NuxtLayout>
+    <NuxtPage/>
   </NuxtLayout>
 </template>
 ```
 
-## Example: setting the layout with `~/pages`
+## Example: setting another layout
```
pi0, line 65: Since we started with two with/without pages variants, it was making sense to emphasis this section also applies to usage with pages.

Otherwise we might move using layouts _without_ pages, to the last and assume the rest are using pages to simplify section.

Xanlantos:
I agree with you on most things and really like your example! If that's how it is supposed to be done then I concur. I have updated it to reflect your comments:

- slot is back, it was a merge error anyway (thanks for spotting it)
- I put pages further back, I think under these circumstances it is best to be introduced later
- I added a sub-section "setting another layout with pages" which we could drop if we like
- I removed "Example" on all headlines as it seems to be only used here and I rather am consistent, but feel free to add them back I don't think they hurt much
- I hope linting and rebase worked *fingers crossed* 

Enjoy reviewing :)

danielroe (approved):

docs/content/2.guide/3.directory-structure/6.layouts.md
```diff
@@ -75,31 +85,36 @@ And in your `app.vue` you need to wrap the `NuxtPage` component with the `NuxtLa
...
+
 ::alert{type=info}
 Learn more about [defining page meta](/guide/directory-structure/pages#page-metadata).
 ::
 
-## Example: manual control with `~/pages`
+## Manual control with `~/pages`
 
-Even if you are using the `~/pages` integration, you can take full control by using the `<NuxtLayout>` component (which is globally available throughout your application), by setting `layout: false`.
+If you are using the `~/pages` integration, you can take full control by setting `layout: false` and then using the `<NuxtLayout>` component within the page.
```
pi0, line 114: @danielroe Shall we move it to last and probably simplify the title? Seems little bit too advanced being amongst the rest of sections.

Xanlantos:
Will this be merged or do I have to do something?

pi0:
Thanks for working on it @Xanlantos @danielroe

Issue #5041, opened 2022-05-18
Route middleware abortNavigation/navigateTo missing options


Rigo-m:
### Environment

n/a

### Reproduction

n/a

### Describe the bug

We should be able to return the error code (e.g.: 404) in abortNavigation, and we should be able to choose the type of redirect in navigateTo (e.g.: 301, 302 etc).

manniL:
`navigateTo` already accepts the status code for the redirect ☺️ 

See https://github.com/nuxt/framework/blob/main/packages/nuxt/src/app/composables/router.ts#L71

Rigo-m:
> `navigateTo` already accepts the status code for the redirect ☺️
> 
> See https://github.com/nuxt/framework/blob/main/packages/nuxt/src/app/composables/router.ts#L71
> 
> For `abortNavigation` it is not available though and I agree, it would be nice to set the status code as `h3`'s [`createError`](https://github.com/unjs/h3/blob/cf2c1a7af7614d00990cc15f6e8468b500134a76/src/error.ts#L26) is accepting it as well.

Missing in docs then :D

noriyuki106:
I think `navigateTo` ignores options (including statusCode), when called inside middleware. is it expected?
https://github.com/nuxt/framework/blob/main/packages/nuxt/src/app/composables/router.ts#L63-L65

PR #5145, opened 2022-05-25
fix(nuxt): respect `navigateTo` options using server middleware in server-side

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/5041#issuecomment-1133861788

### ❓ Type of change

### 📚 Description

On SSR we can directly redirect within middleware when user calls `navigateTo`, which will also mean we can respect the user's options.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -60,7 +60,7 @@ export const navigateTo = (to: RouteLocationRaw, options: NavigateToOptions = {}
   if (!to) {
     to = '/'
   }
-  if (isProcessingMiddleware()) {
+  if (process.client && isProcessingMiddleware()) {
     return to
   }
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/router.ts
```diff
@@ -60,7 +60,7 @@ export const navigateTo = (to: RouteLocationRaw, options: NavigateToOptions = {}
   if (!to) {
     to = '/'
   }
-  if (isProcessingMiddleware()) {
+  if (process.client && isProcessingMiddleware()) {
```
pi0, line 63:
```suggestion
  // Early redirect on client-side since only possible option is redirectCode and not applied
  if (process.client && isProcessingMiddleware()) {
```

pi0 (approved):

PR #5155, opened 2022-05-26
refactor(nuxi): provide better advice on failing clone

harlan-zw:
### 🔗 Linked issue

see https://github.com/nuxt/framework/pull/923 & https://github.com/nuxt/framework/pull/1052

### ❓ Type of change

### 📚 Description

The issue was identified when using `nuxi init`, I modified the arguments incorrectly and got an error that wasn't helpful.

image

The PR https://github.com/nuxt/framework/pull/923 introduced a try/catch for degit cloning, it correctly showed when a user required `git` if they didn't have it installed.

The switch to tiged https://github.com/nuxt/framework/pull/1052 changed what was being caught. Tiged works without git, it will download the tarball if git isn't available. So the error message is actually catching when the repo isn't valid.

The output will now look like this:

image

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -46,7 +46,7 @@ export default defineNuxtCommand({
       await tiged.clone(dstDir)
     } catch (e) {
       if (e.toString().includes('could not find commit hash')) {
-        consola.error(`Failed to clone template from \`${src}\`; please check the repo is valid.`)
+        consola.error(`Failed to clone template from \`${src}\`; please check the repo is valid and that you have installed \`git\` correctly.`)
         process.exit(1)
       }
       throw e
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -46,7 +46,7 @@ export default defineNuxtCommand({
       await tiged.clone(dstDir)
     } catch (e) {
       if (e.toString().includes('could not find commit hash')) {
-        consola.error(`Failed to clone template \`${src}\`, please check the repo is valid.`)
+        consola.error(`Failed to clone template from \`${src}\`; please check the repo is valid.`)
         process.exit(1)
       }
       throw e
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/commands/init.ts
```diff
@@ -46,7 +46,7 @@ export default defineNuxtCommand({
       await tiged.clone(dstDir)
     } catch (e) {
       if (e.toString().includes('could not find commit hash')) {
-        consola.warn('Make sure you have installed `git` correctly')
+        consola.error(`Failed to clone template \`${src}\`, please check the repo is valid.`)
```
danielroe, line 49:
```suggestion
        consola.error(`Failed to clone template from \`${src}\`; please check the repo is valid.`)
```

packages/nuxi/src/commands/init.ts
```diff
@@ -46,7 +46,7 @@ export default defineNuxtCommand({
       await tiged.clone(dstDir)
     } catch (e) {
       if (e.toString().includes('could not find commit hash')) {
-        consola.warn('Make sure you have installed `git` correctly')
+        consola.error(`Failed to clone template from \`${src}\`; please check the repo is valid.`)
```
DamianGlowala, line 49:
```suggestion
        consola.error(`Failed to clone template from \`${src}\`. Please check if the repo is valid.`)
```

harlan-zw: Thanks, I'll leave it as @danielroe's suggestion unless he prefers this

pi0:
Thanks for the changes! I believe we still need an additional hint for checking git being installed since it makes same error.

harlan-zw:
> Thanks for the changes! I believe we still need an additional hint for checking git being installed since it makes same error.

You were right on that, my bad. The testing I did locally was giving me a false positive on the git not being installed.

I've replicated it on [stackblitz](https://stackblitz.com/edit/node-sfaccf?file=index.mjs) and the same catch is thrown for git being missing, so I've updated to message to include the origin git hint.

packages/nuxi/src/commands/init.ts
pi0: Applied via 0ae4546fe136ad9fcad08ee500d2eaef4905c500 (https://github.com/nuxt/framework/pull/5155#discussion_r886368465)

PR #5173, opened 2022-05-27
fix!: default statusCode of `navigateTo` change from 301 to 302

### LINKED ISSUE #5171: `navigateTo` should use 302 instead of 301 as default
I see it's documented as 301 so I guess it might be intentionally and may have some reason behind it. So this is more like a simple RFC to discuss a reasonable default.

I personally found 301 (Move Permanently) is kinda confusing. We use middleware to do redirection dynamically, while ppl could do some conditional checks based on internal or temporary states. A dynamic handler having a permanent effect I think would be less intuitive in most cases.

Additionally, the cache of 301 is handled by the browser and I guess there isn't an easy way to control it on user land, which could also be dangerous across different versions of an app.

When working on Nuxt Hackernews, where we redirect `/` to `/news/1`, the 301 causes my browser always redirect `http://localhost:3000` to `http://localhost:3000/news/1` even when I am working on different apps (which I had to clear my browser history a few times).

From my point of view, 301 (Move Permanently) should be considered as an opt-in optimization to save the bandwidth and we shouldn't make this assumption for users by default. Thus I propose we change the default from `301 (Move Permanently)` to `302 (Moved Temporarily)`

antfu:
### 🔗 Linked issue

close #5171

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/3.directory-structure/7.middleware.md
+++ b/docs/content/2.guide/3.directory-structure/7.middleware.md
@@ -42,7 +42,7 @@ Unlike navigation guards in [the vue-router docs](https://router.vuejs.org/guide
 
 * nothing - does not block navigation and will move to the next middleware function, if any, or complete the route navigation
 * `navigateTo('/')` or `navigateTo({ path: '/' })` - redirects to the given path
-* `navigateTo('/', { redirectCode: 302 })` - Redirects to the given path and will set the redirect code if the redirect happens on the server side (default: 301)
+* `navigateTo('/', { redirectCode: 301 })` - Redirects to the given path and will set the redirect code if the redirect happens on the server side (default: 302)
 * `abortNavigation()` - stops the current navigation
 * `abortNavigation(error)` - rejects the current navigation with an error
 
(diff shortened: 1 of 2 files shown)


docs/content/2.guide/3.directory-structure/7.middleware.md
```diff
@@ -42,7 +42,7 @@ Unlike navigation guards in [the vue-router docs](https://router.vuejs.org/guide
 
 * nothing - does not block navigation and will move to the next middleware function, if any, or complete the route navigation
 * `navigateTo('/')` or `navigateTo({ path: '/' })` - redirects to the given path
-* `navigateTo('/', { redirectCode: 302 })` - Redirects to the given path and will set the redirect code if the redirect happens on the server side (default: 301)
+* `navigateTo('/', { redirectCode: 301 })` - Redirects to the given path and will set the redirect code if the redirect happens on the server side (default: 302)
```
pi0, line 45:
```suggestion
* `navigateTo('/', { redirectCode: 301 })` - redirects to the given path and will set the redirect code to [`301` Moved Permanently
](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) if the redirect happens on the server side
```

docs/content/2.guide/3.directory-structure/7.middleware.md
```diff
@@ -42,7 +42,7 @@ Unlike navigation guards in [the vue-router docs](https://router.vuejs.org/guide
 
 * nothing - does not block navigation and will move to the next middleware function, if any, or complete the route navigation
 * `navigateTo('/')` or `navigateTo({ path: '/' })` - redirects to the given path
```
pi0, line 44:
```suggestion
* `navigateTo('/')` or `navigateTo({ path: '/' })` - redirects to the given path and will set the redirect code to [`302` Found
](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302) if the redirect happens on the server side
```

danielroe (approved):

Atinux (approved):

PR #5227, opened 2022-05-31
feat(nuxt): add `clearNuxtData`

cawa-93:
### 🔗 Linked issue
https://github.com/nuxt/framework/discussions/4636

### ❓ Type of change

### 📚 Description
Resolves #4636

Added a simple helper function that deletes data from the internal cache by key. This method is useful if you want to invalidate the data fetching for another page.

### 📝 Checklist


docs/content/3.api/3.utils/clear-nuxt-cache.md
```diff
@@ -0,0 +1,9 @@
+# `clearNuxtCache`
+
+```ts
+clearNuxtCache(keys?: string | string[])
```
pi0, line 4: It would be better for future compatibility if we introduce as `{ keys: [] }`

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -193,6 +193,29 @@ export function refreshNuxtData (keys?: string | string[]): Promise<void> {
...
+  const _keys = keys ? Array.isArray(keys) ? keys : [keys] : undefined
+  if (_keys === undefined) {
+    return
+  }
+
+  const nuxt = useNuxtApp()
+
+  _keys.forEach((key) => {
+    if (nuxt.payload.data[key] !== undefined) {
+      delete nuxt.payload.data[key]
```
pi0, line 210: Perf: Deleting changes the shape of object and invalidate V8 internal cache. We can instead assign `null` or `undefined`

docs/content/3.api/3.utils/clear-nuxt-cache.md
cawa-93: I'm sorry, but what do you mean? Change make argument an object?
```ts
function clearNuxtCache({keys}: {keys: string | string[]})  {}
```

pi0: Precesely!

pi0:
With the `keys` => `{ keys }` change, implementation looks good to me! Would you mind also adding [API docs](https://v3.nuxtjs.org/api/composables/use-async-data) page under Utils?

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -200,6 +200,29 @@ export function refreshNuxtData (keys?: string | string[]): Promise<void> {
...
+
+  for (const key of _keys) {
+    if (nuxt.payload.data[key] !== undefined) {
+      nuxt.payload.data[key] = undefined
+    }
+
+    if (nuxt.payload._errors[key] !== undefined) {
+      nuxt.payload._errors[key] = undefined
+    }
+  }
```
sasial-dev, line 223: Why do you have to iterate through all the keys?
Can't you just do 
```suggestion
  nuxt.payload.data = undefined
  nuxt.payload._errors = undefined
```

sasial-dev: Or am I missing something stupid here?

cawa-93: Because we should not delete the entire existing cache, but only some parts by keys

pi0:
Thanks for making PR @cawa-93. I've made few refactors to reflect last changes and also cleaning up shared state. Didn't had a change to try it please share feedback if last changes have any issues or could be improved :)

cawa-93:
Uncovered use case: 
Remove cache for several unknown keys. Example: you have app with posts and filter for posts. Each query for posts saved in cache in uniq key, like `posts.list-{param1:1}`, `posts.list-{param2:2}`, `posts.list-{param1:1,param2:3}` (nuxt-trpc works llike that). In this case you may want clear data for all keys `posts.list-*`. 
May be `clearNuxtData` should also accept function to filter keys?
```ts
clearNuxtData(key => key.startsWith('posts.list'))
```
Or provide another method to get all keys in cache and filter it manually:
```ts
clearNuxtData(
  getNuxtDataKeys().filter(key => key.startsWith('posts.list'))
)
```

pi0:
Filter support seems a good idea 👍🏼 When constructing _keys, we can check type and if is function, use Object.keys + filter function. Feel free to make a PR.

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -254,6 +254,28 @@ export function refreshNuxtData (keys?: string | string[]): Promise<void> {
   return useNuxtApp().callHook('app:data:refresh', _keys)
 }
 
+export function clearNuxtData (keys?: string | string[]): void {
+  const nuxtApp = useNuxtApp()
+  const _keys = keys ? Array.from(keys).filter(key => key && typeof key === 'string') : Object.keys(nuxtApp.payload.data)
```
cawa-93, line 259: @pi0 I miss something there? `keys` may be `string` and `Array.from(keys)` return array with letters

pi0: right, my bad! (as i mentioned above, didn't well tested impl)

PR #5274, opened 2022-06-03
fix(nuxt): add default 1px blank favicon gif

danielroe:
### ❓ Type of change

### 📚 Description

This adds a tiny `favicon.ico` to the default nitro stack to prevent the annoying router error message:

```bash
[Vue Router warn]: No match found for location with path "/favicon.ico"
```

It allows overriding by the user creating `~/public/favicon.ico`.

**Note**: we have `serve-placeholder` support in nitro but don't enable it for `/` (probably rightly). Without this PR (even if we enabled serve-placeholder) we'd still be hitting the Nitro runtime on every request.

### 📝 Checklist


pi0:
I believe we are adding fallback (before renderer with known extensions) as a configurable feature of Nitro. This way we can handle other edge cases such as `sw.js` and other 404 public assets. Linked issue: https://github.com/unjs/nitro/issues/94

danielroe:
The merit in adding the default favicon is that it will be added to public folder when built and thereby skip hitting the nitro runtime for deploy targets with a CDN.

Atinux:
I believe this is worth adding as well since it is already a nice fix and feature ready to go and can be refactor with a nitro update 😊

pi0:
With new `defineRenderHandler` in nitro (https://github.com/unjs/nitro/pull/395), we catch favicon.ico 404 (easily improved later for another kind of placeholders). 

I also like the idea of generating a static asset to avoid hitting lambda at all but I think we can use prerender and a new dynamic render handler to implement directly in nitro as another enhancement because this way we can allow custom dynamic handlers that generate the asset on demand.

Issue #5237, opened 2022-06-01
Cannot specify PostCSS config path


ashgo-x:
### Environment

- Operating System: `macOS`
- Node Version:     `v16.15.0`
- Nuxt Version:     `3.0.0-rc.3`
- Package Manager:  `pnpm@7.1.7`
- Builder:          `vite`
- User Config:      `css`, `postcss`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Describe the bug

[nuxt.config#config](https://v3.nuxtjs.org/api/configuration/nuxt.config#config)

There is no option to specify the config path for PostCSS.

danielroe:
Would you explain a little more about your use-case?

ashgo-x:
@danielroe, I have to set PostCSS options in `nuxt.config.ts` currently.

Would be nice if I could have the option to seperate the PostCSS configs in `postcss.config.js`.
Just like how Vite automatically reads the `postcss.config.js`.

pi0:
Hi @ashgo-x. After talking with @danielroe we are finally going to deprecate external config also for webpack. You can always use `nuxt.config` to add custom postcss config or manually import postcss config in top of `nuxt.config` file to provide it. 

Follow up issue to track: https://github.com/nuxt/framework/issues/5368

ashgo-x:
Thanks!

PR #5279, opened 2022-06-03
fix(schema): provide correct `postcss.config` type

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/5237

### ❓ Type of change

### 📚 Description

Was missing curly braces, so custom type wasn't being applied.

### DIFF
--- a/packages/schema/src/config/postcss.ts
+++ b/packages/schema/src/config/postcss.ts
@@ -5,7 +5,7 @@ export default {
   /** @version 3 */
   postcss: {
     /** Path to postcss config file. */
-    /** @type string | false */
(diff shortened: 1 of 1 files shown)


pi0:
I believe same as other packages (vite, webpack, rollup, etc) that we resolve their config internally and allow only extending using nuxt options (`nuxt.config` or modules via hooks), we shall deprecate external postcss config file support (from existing webpack) and also not add it for vite.

/cc @danielroe If you agree, shall we close this?

danielroe:
I think we should find out a bit more about user use-case before finally deciding, but we can close/pend this PR in the mean-time.

PR #5294, opened 2022-06-05
docs: Fix example definition of runtimeConfig variables in nuxtConfig

jonnyparris:
### 🔗 Linked issue

#5056

### ❓ Type of change

### 📚 Description

According to the [test fixtures](https://github.com/nuxt/framework/blob/603f07c3b8f93feefb8c59c5ff3a970a60e38041/test/fixtures/basic/nuxt.config.ts#L14), api for definining runtime config has changed to 2 separate keys for public (clientside) & private (serverside) variables.

### 📝 Checklist

### DIFF
--- a/docs/content/migration/8.runtime-config.md
+++ b/docs/content/migration/8.runtime-config.md
@@ -24,11 +24,11 @@ When referencing these variables within your components, you will have to use th
 import { defineNuxtConfig } from 'nuxt'
 
 export default defineNuxtConfig({
-  runtimeConfig: {
+  privateRuntimeConfig: {
     secretKey: '', // variable that can only be accessed on the server side
-    public: {
-      BASE_URL: process.env.BASE_URL || 'https://nuxtjs.org' // variable that can also be accessed on the client side
-    }
+  },
+  publicRuntimeConfig: {
+    BASE_URL: process.env.BASE_URL || 'https://nuxtjs.org' // variable that can also be accessed on the client side
   },
 })
 ```


pi0:
Hi dear @jonnyparris. What is in the fixtures, is testing the deprecated old option. We moved from `{ publicRuntimeConfig, privateRuntimeConfig }` to `runtimeConfig: { public }` for Nuxt 3 while old options are still supported. Therefore current migration example is correct. Please let me know if like to have more chats about this and keep doing contributions 💚

jonnyparris:
@pi0 I was also facing the same problem outlined in the linked issue #5056 so I think there's a bug with the way it's supposed to work. However I discovered that using `publicRuntimeConfig` still worked so I thought that maybe it was the docs that was lagging behind the code as opposed to the other way around.

...so I guess this PR is actually a +1 on issue #5056 but at least the deprecated version is a useful workaround for now 🤷🏾‍♂️

jonnyparris:
I just tried to re-confirm the bug using the proper `runtimeConfig.public` syntax...and now it's all working fine 🥲  so thanks for shining your general aura of effectiveness this way @pi0 !

jonnyparris:
aha closure finally! Just combed through my commits and found that I initially mis-capitalised it with `runTimeConfig` instead of `runtimeConfig` in my `nuxt.config.ts`.
Weird that typescript doesn't complain about the difference though, right? 🤔

pi0:
Indeed that's not the best DX. We mainly had to remove that guard to allow any keys for modules not explicitly providing that key but it increases chance of typo in config...

PR #5312, opened 2022-06-06
docs: Adding start script to node-server docs

Chippd:
When deploying to a service like Heroku, users will need to define a start script to avoid errors.

### ❓ Type of change

### 📚 Description

When trying to deploy a nuxt app to heroku, I received an error that there was no start script. My change is to update the node server documentation to include this start script. 

This improves the developer experience for Nuxt developers. This change makes it quicker for developers to get something deployed, reducing frustration and increasing satisfaction with Nuxt

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/5.deploy/1.node-server.md
+++ b/docs/content/2.guide/5.deploy/1.node-server.md
@@ -24,6 +24,20 @@ $ node .output/server/index.mjs
 Listening on http://localhost:3000
 ```
 
+## Configuring start script
+
+When deploying your nuxt app to a service like Heroku, you'll need to give it a start script to avoid errors
+```js [ecosystem.config.js]
+"scripts": {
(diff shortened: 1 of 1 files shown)


pi0:
Hi @Chippd and thanks for this PR. For Heroku deployments (and generally other deployments as well), all we need is `.output` directory (not the entire repo) I believe we can autogenerate a corresponding `package.json` file for it or opt-in for Docker before documenting this since in general, is not a good practice including all source code nor needed (related issue for the preset: https://github.com/unjs/nitro/issues/49)

Chippd:
Fair enough, it did feel not-quite-right adding Heroku-specific config here but it should be somewhere, and I couldn't find it easily.

pi0:
Let's track in the upstream issue. Thanks for PR anyways <3

PR #5337, opened 2022-06-07
fix(nuxi): show base url in terminal message

antfu:
### 🔗 Linked issue

Requires https://github.com/unjs/listhen/pull/28

### ❓ Type of change

### 📚 Description

When we configure `app.baseURL`, the terminal still shows the root URL like `http://localhost:3000/`.

This PR requires upstream https://github.com/unjs/listhen/pull/28

### 📝 Checklist


packages/nuxi/src/commands/dev.ts
```diff
@@ -38,6 +38,7 @@ export default defineNuxtCommand({
       port: args.port || args.p || process.env.NUXT_PORT,
       hostname: args.host || args.h || process.env.NUXT_HOST,
       https: Boolean(args.https),
+      baseURL: args.baseURL || process.env.NUXT_BASE_URL,
```
pi0, line 41: If we are to introduce this new flag, I guess we have to pass from args (`--baseUrl`) to the options (`loadNuxt`) as well.

Note: We might rename it to `--config-app-baseURL`. I'm not sure about final convention, but at some point, allowing to override every Nuxt config using env and args, could reduce the chance to deprecate arg. But nothing important.

pi0:
Considering upstream merged, what happens if user is setting base URL directly in `nuxt.config`? I guess first banner would show without and then subsequent reloads add base with override of `showURL`?

antfu:
Are you ok with that we disable `showUrl` option in listen and call it manually after the config resolve?

pi0:
@antfu Any update on this?

antfu:
Shall be ready now.

pi0 (approved):

pi0 (approved):

PR #5376, opened 2022-06-09
fix(nuxi): add client side environment shim types

clarkdo:
### 🔗 Linked issue

Resolve https://github.com/nuxt/framework/issues/5351

### ❓ Type of change

### 📚 Description

The default generated types are for Node.js, so we need to add client side types to support asset imports.

@nuxt/framework I'm just adding 'vite/client' into types instead of checking if build is vite builder because:

  - vite-builder is aleady a direct dep of nuxt https://github.com/nuxt/framework/blob/f5dca6245d97beb4785202a75c0142dec9d2aeab/packages/nuxt/package.json#L38
  - this will also work for other builders like webpack

Another potential solution is adding nuxt own client types which can be based on vite/client, please feel free to drop your suggestions.


clarkdo:
@danielroe @pi0 I've changed to import client types according to builder config, what do you think ?

packages/nuxi/src/utils/prepare.ts
```diff
@@ -72,6 +72,16 @@ export const writeTypes = async (nuxt: Nuxt) => {
 
   await nuxt.callHook('prepare:types', { references, declarations, tsConfig })
 
+  const builder = nuxt.options.builder as ('@nuxt/webpack-builder' | '@nuxt/vite-builder')
```
clarkdo, line 75: @pi0 For the option which is `$resolve`, the type is config value like webpack but not internal value @nuxt/webpack-builder, is there anything I'm missing that can get underlying value ?

pi0:
@clarkdo Do you mind to inline shim types for `~assets` imports? This way we don't have to make a dynamic behavior based on bundle and sure only typing what we support for nuxt.

clarkdo:
Oops, I’ll look into this

PR #5387, opened 2022-06-10
docs: set cookie event

ashgo-x:
### ❓ Type of change

### 📚 Description

`setCookie`'s event parameter should be `req` instead of `res`.

### 📝 Checklist

### DIFF
--- a/docs/content/3.api/1.composables/use-cookie.md
+++ b/docs/content/3.api/1.composables/use-cookie.md
@@ -150,7 +150,7 @@ export default (req, res) => {
   let counter = useCookie(req, 'counter') || 0
 
   // Increase counter cookie by 1
-  setCookie(res, 'counter', ++counter)
+  setCookie(req, 'counter', ++counter)
 
   // Send JSON response
   return { counter }


docs/content/3.api/1.composables/use-cookie.md
```diff
@@ -150,7 +150,7 @@ export default (req, res) => {
   let counter = useCookie(req, 'counter') || 0
 
   // Increase counter cookie by 1
-  setCookie(res, 'counter', ++counter)
+  setCookie(req, 'counter', ++counter)
```
Diizzayy, line 153: the `setCookie` event parameter infact takes either an H3 CompatibilityEvent or the res passed by the legacy handler ( which we advise against using unless unavoidable )

ashgo-x: Got it, thanks!

pi0:
Thanks for PR @ashgo-x and the tip @Diizzayy. I've reopened your PR to update example using event and `dwefineEventHandler()` wrapper. While it works with backward compatibility API, example is still wrong.

PR #5410, opened 2022-06-10
feat(nuxt): typed router

danielroe:
### ❓ Type of change

### 📚 Description

This PR is a prerelease integration of the new typed-routes feature in `vue-router`, for testing purposes. (cc: @posva)

CleanShot 2022-06-04 at 15 41 52

### DIFF
--- a/packages/nuxt/src/app/composables/router.ts
+++ b/packages/nuxt/src/app/composables/router.ts
@@ -1,10 +1,10 @@
-import type { Router, RouteLocationNormalizedLoaded, NavigationGuard, RouteLocationNormalized, RouteLocationRaw, NavigationFailure } from 'vue-router'
+import type { RouterTyped, RouteLocationNormalizedLoaded, NavigationGuard, RouteLocationNormalized, RouteLocationRaw, NavigationFailure } from 'vue-router'
 import { sendRedirect } from 'h3'
 import { joinURL } from 'ufo'
 import { useNuxtApp, useRuntimeConfig } from '#app'
 
 export const useRouter = () => {
-  return useNuxtApp()?.$router as Router
+  return useNuxtApp()?.$router as RouterTyped
 }
 
 export const useRoute = () => {
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -78,6 +78,25 @@ export default defineNuxtModule({
       }
     })
 
+    addTemplate({
+      filename: 'types/router.d.ts',
+      async getContents () {
+        const pages = await resolvePagesRoutes()
+        await nuxt.callHook('pages:extend', pages)
+        const { routes, imports } = normalizeRoutes(pages)
+        return [
+          ...imports,
+          "import { Router, RouterConfig, RouterOptions } from 'vue-router'",
+          `declare const routes = ${routes} as const`,
+          "declare module 'vue-router' {",
+          'export interface Config {',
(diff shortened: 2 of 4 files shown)


packages/nuxt/src/app/composables/router.ts
```diff
@@ -1,10 +1,10 @@
-import type { Router, RouteLocationNormalizedLoaded, NavigationGuard, RouteLocationNormalized, RouteLocationRaw, NavigationFailure } from 'vue-router'
+import type { RouterTyped, RouteLocationNormalizedLoaded, NavigationGuard, RouteLocationNormalized, RouteLocationRaw, NavigationFailure } from 'vue-router'
 import { sendRedirect } from 'h3'
 import { joinURL } from 'ufo'
 import { useNuxtApp, useRuntimeConfig } from '#app'
 
 export const useRouter = () => {
-  return useNuxtApp()?.$router as Router
+  return useNuxtApp()?.$router as RouterTyped
```
victorgarciaesgi, line 7: I think it may work for small apps, but for big projects with a lot of nested routes Typescript will surrender :(

danielroe: Do you have an example for testing? I can take a look at the upstream vue-router types.

packages/nuxt/src/pages/module.ts
```diff
@@ -157,6 +176,7 @@ export default defineNuxtModule({
     nuxt.hook('prepare:types', ({ references }) => {
       references.push({ path: resolve(nuxt.options.buildDir, 'types/middleware.d.ts') })
       references.push({ path: resolve(nuxt.options.buildDir, 'types/layouts.d.ts') })
+      references.push({ path: resolve(nuxt.options.buildDir, 'types/router.d.ts') })
```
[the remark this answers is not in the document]
danielroe: It adds a reference into `nuxt.d.ts` which is sufficient to include the generated file in the TS context.

packages/nuxt/src/app/composables/router.ts
posva: This is something I told him. I think that the feature should be only opt-in because these types are quite complex it gets too slow for large amount of routes (+50). It should be fine with types generated at build though but that will happen later.

packages/nuxt/src/pages/module.ts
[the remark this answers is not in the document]
victorgarciaesgi: Nice!

packages/nuxt/src/app/composables/router.ts
victorgarciaesgi: For build time generated types it's fine, I will look into an project where I have 50+ routes if you want to test.
On my side in `nuxt-typed-router`, for a little app it looks like this https://github.com/victorgarciaesgi/nuxt-typed-router/blob/master/playground/generated/typed-router.d.ts

victorgarciaesgi: I still had problematics about whether the params needed to be optional or not, because with vue-router you're not forced to specify the params for a child route for exemple

danielroe:
Closing as vue-router is no longer exposing these types. Instead we'll follow progress on https://github.com/posva/unplugin-vue-router and will hopefully open another one in future.

packages/nuxt/src/pages/module.ts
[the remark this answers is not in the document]
Rigo-m: I was looking for this the other day to augment app types from within a module. Would be nice if it was an example in the docs

PR #5417, opened 2022-06-11
fix(nuxi): always override `NODE_ENV`

### LINKED ISSUE #5196: NODE_ENV overwriting results in weird errors
### Environment

- Operating System: `Darwin`
- Builder:          `vite`
- User Config:      `ssr`, `runtimeConfig`, `typescript`, `css`, `buildModules`, `intlify`, `alias`
- Runtime Modules:  `-`
- Build Modules:    `@vueuse/nuxt@8.4.2`, `@intlify/nuxt3@0.1.10`

### Reproduction

I stumbled into an issue very similar to the one of https://github.com/nuxt/framework/issues/1913 and https://github.com/nuxt/framework/issues/3730, i.e. overriding `NODE_ENV` environment variable. 
This results in a weird blank page, with no clear info on what happened.

I believe it deserves a warning in the https://v3.nuxtjs.org/guide/features/runtime-config/#environment-variables, it took me a week to figure out that that was the problem.

### Describe the bug

-

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #5196

### ❓ Type of change

### 📚 Description

This adds warnings if user is overriding the correct NODE_ENV for build/dev/analyze. (Other commands like preview, typecheck or prepare I felt did not particularly matter, but happy to expand the warnings.)

### 📝 Checklist


packages/nuxi/src/commands/build.ts
```diff
@@ -13,6 +13,11 @@ export default defineNuxtCommand({
   },
   async invoke (args) {
     process.env.NODE_ENV = process.env.NODE_ENV || 'production'
+
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn(`Building nuxt in \`${process.env.NODE_ENV}\` mode. This is not recommended.`)
```
DamianGlowala, line 18:
```suggestion
      console.warn(`Building Nuxt in \`${process.env.NODE_ENV}\` mode. This is not recommended.`)
```

packages/nuxi/src/commands/analyze.ts
```diff
@@ -16,6 +16,10 @@ export default defineNuxtCommand({
   async invoke (args) {
     process.env.NODE_ENV = process.env.NODE_ENV || 'production'
 
+    if (process.env.NODE_ENV !== 'production') {
+      console.warn(`Building nuxt for analysis in \`${process.env.NODE_ENV}\` mode. This is not recommended.`)
```
DamianGlowala, line 20:
```suggestion
      console.warn(`Building Nuxt for analysis in \`${process.env.NODE_ENV}\` mode. This is not recommended.`)
```

DamianGlowala (changes requested):

pi0:
Thanks for the PR but if we are not recommending custom NODE_ENV, there is no reason to allow overriding it I think. However, it is also worth tracking down what happens with custom env specially for non development builds. It can be simply one or two bugs.

danielroe:
The cause of https://github.com/nuxt/framework/issues/5196 is really a Vite bug (or perhaps it's unfair to say that it's a bug - we could say that Vite does not support _building_ in development mode). Here's how it happens:

1. `options.dev` is set to true if `NODE_ENV` is overridden
2. vitejs/plugin-vue injects HMR code because it's not production: https://github.com/vitejs/vite/blob/main/packages/plugin-vue/src/main.ts#L131-L139
3. vite replaces `import.meta.hot` with `false` because we're building: https://github.dev/vitejs/vite/blob/main/packages/vite/src/node/plugins/define.ts#L32-L45

I would be very happy with _not_ allowing `NODE_ENV` to be overridden, but think if we do then we should definitely warn. What was the reason you originally added it (https://github.com/nuxt/framework/commit/586709a82d7ce45d7cbd528bb7ee002dccae5611)?

pi0:
Thanks for explaining! I think we can then change the strategy to override to what we expect but also warn if user already set a `NODE_ENV` we are overriding and saying hay, we just ignored your env ?

packages/nuxi/src/commands/analyze.ts
```diff
@@ -14,7 +14,11 @@ export default defineNuxtCommand({
     description: 'Build nuxt and analyze production bundle (experimental)'
   },
   async invoke (args) {
-    process.env.NODE_ENV = process.env.NODE_ENV || 'production'
+    if (process.env.NODE_ENV && process.env.NODE_ENV !== 'production') {
+      console.info('Setting NODE_ENV to `production`.')
```
pi0, line 18: Can we please use a warning and also show the user-provided value that is being ignored?

pi0: (this can be refactored to a common utility too I guess now)

packages/nuxi/src/utils/env.ts
```diff
@@ -0,0 +1,8 @@
+export const overrideEnv = (targetEnv: string) => {
+  const currentEnv = process.env.NODE_ENV
+  if (currentEnv && currentEnv !== targetEnv) {
+    console.warn(`Changing NODE_ENV from \`${currentEnv}\` to \`${targetEnv}\`.`)
```
pi0, line 4:
```suggestion
    console.warn(`Changing \`NODE_ENV\` from \`${currentEnv}\` to \`${targetEnv}\`.`)
```

pi0 (approved):

Issue #5414, opened 2022-06-11
dev server hangs indefinitely with css in nuxt.config.ts


jiblett1000:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.14.0`
- Nuxt Version:     `3.0.0-rc.3-27581242.8831e4a`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `typescript`, `css`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/jiblett1000/inter-test

### Describe the bug

`yarn build` works fine but the dev server hangs indefinitely when importing css files in `nuxt.config.ts`. Removing the css import from the config fixes the dev server hang.

danielroe:
@jiblett1000 Does this still occur with the latest nuxt edge release? I think it might also have been resolved by https://github.com/nuxt/framework/pull/5409.

jiblett1000:
@danielroe yep, still happening for me with 3.0.0-rc.3-27581242.8831e4a

jiblett1000:
@danielroe @pi0 Works great now. Thanks!

PR #5421, opened 2022-06-11
fix(nuxt): exclude non-cwd auto-imports and don't process _all_ files

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/5414

### ❓ Type of change

### 📚 Description

This fixes a regression introduced by https://github.com/nuxt/framework/pull/5042, by:

1. only adding layers within `node_modules` to autoImports include.
2. not force-transforming all files in `include` but only the ones that match the test for file time

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/auto-imports/transform.ts
+++ b/packages/nuxt/src/auto-imports/transform.ts
@@ -15,13 +15,8 @@ export const TransformPlugin = createUnplugin(({ ctx, options, sourcemap }: {ctx
       const exclude = options.transform?.exclude || [/[\\/]node_modules[\\/]/]
       const include = options.transform?.include || []
 
-      // Custom includes
-      if (include.some(pattern => id.match(pattern))) {
-        return true
-      }
-
-      // Exclude node_modules by default
-      if (exclude.some(pattern => id.match(pattern))) {
+      // Custom includes - exclude node_modules by default
+      if (exclude.some(pattern => id.match(pattern)) && !include.some(pattern => id.match(pattern))) {
         return false
       }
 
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/auto-imports/transform.ts
```diff
@@ -15,13 +15,8 @@ export const TransformPlugin = createUnplugin(({ ctx, options, sourcemap }: {ctx
       const exclude = options.transform?.exclude || [/[\\/]node_modules[\\/]/]
       const include = options.transform?.include || []
 
-      // Custom includes
-      if (include.some(pattern => id.match(pattern))) {
```
pi0, line 19: Shouldn't we keep this check after exclude? (ie: explicit include without explicit exclude). It is current behavior before (https://github.com/nuxt/framework/pull/5042/files#diff-1edf1ef633d6d36d0410fe36249ca18b1f5fd1a58b658296696698c2fc542ee9L18)

danielroe: No, because we exclude node_modules.

pi0: Then maybe reverse?

- Include checks:
  - include `node_modules/<layer>/<!node_modules>`
- Exclude checks:
  - exclude `node_modules`

Combined method with && in this PR works but makes it unable to explicitly _include_ a pattern.

danielroe: The point is that we still need to perform checks on the file extension.

pi0: Previously we had two checks for exlude list and include list option. We are now checking for all excludes only if they are not included.... Anyway let's merge and see any upcoming limitations with this. Options are mostly internal now.

pi0 (approved):

Issue #4598, opened 2022-04-25
Navigation fails if pages do not have single root element


henningko:
### Environment

  * Operating System: `macOS 12.3.1`

  * Node Version:     `v16.14.2`

  * Nuxt Version:     `3.0.0-rc.1`

  * Package Manager:  `yarn@1.23.0`

  * Builder:          `vite`

  * User Config:      `-`

  * Runtime Modules:  `-`

  * Build Modules:    `-`

### Reproduction
Create a new Nuxt3 app with the following files:

`app.vue`
```vue
<template>
  <div>
    <NuxtLayout>
      <NuxtPage />
    </NuxtLayout>
  </div>
</template>
```

`pages/index.vue`
```vue
<template>
  <div>Home</div>
  <NuxtLink to="/thankyou">thanks</NuxtLink>
</template>
```

`pages/thankyou.vue`
```vue
<template>
  <div>
    <h1>thank you</h1>
  </div>
</template>

```
### Describe the bug

When navigating from `/` to `/thankyou` using `NuxtLink` (as well as `navigateTo` and `router.push`, the result is an empty page.

When wrapping the content of `pages/index.vue` into a single div, the problem disappears.

No error messages are given.

Aareksio:
Reproduction: [StackBlitz](https://stackblitz.com/edit/github-6syimm)

![image](https://user-images.githubusercontent.com/11166786/165155470-c5ed60d6-56fd-4de0-890f-b79b0be01e14.png)

Documentation ([source](https://v3.nuxtjs.org/guide/directory-structure/pages#usage)):
> Pages **must have a single root element** to allow route transitions between pages. (HTML comments are considered elements as well.)

Couldn't this be caught by Nuxt and surfaced with better error message? Perhaps adding yellow alert in docs may be helpful too.

henningko:
Shoot, yes, totally missed that in the docs (perhaps also because I understand transitions to be animated).

IIRC, in it would throw an error in Nuxt2—so no (explicit) error/warning makes it hard to understand what's going on.

Issue #3587, opened 2022-03-10
onMounted hook running before DOM is available


fanckush:
### Environment

- Operating System: `Darwin`
- Node Version:     `v17.3.1`
- Nuxt Version:     `3.0.0-27447229.20f3171`
- Package Manager:  `yarn@1.22.17`
- Bundler:          `Vite`
- User Config:      `vite`, `build`, `ssr`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Not needed. The Bug is very clear

### Describe the bug

The screenshot below says it all. On first load (refresh) everything is fine and I'm able to run `querySelector` and access the DOM just fine. However if I navigate to a different route then come back I get `null`.
``` vue
<template>
  <div class="TheAdvantages">
    <ul>
      <li class="first" @click="$router.push('')"> I'm first </li>
    </ul>
  </div>
</template>

<script setup lang="ts">
onMounted(() => {
  console.log('onMonuted Hook:')

  console.log(
    "document.querySelector('.first'): ",
    document.querySelector('.first')
  )

  setTimeout(() => {
    console.log('setTimeout 0ms')
    console.log(
      "document.querySelector('.first'): ",
      document.querySelector('.first')
    )
  }, 0)

})
```

<img width="584" alt="image" src="https://user-images.githubusercontent.com/6278713/157617245-898d6400-b18b-4ed9-9f27-cc90883d82a5.png">

fanckush:
For anyone else having this issue until it's resolved you can use `setTimeout` 

Instead of:

``` js
onMounted(() => {
... code that needs DOM // WON'T WORK
})
```

Do this:
``` js
onMounted(() => {
  setTimeout(() => {
    ... code that needs DOM // WORKS!! 
  }),
    0
})

```

bcspragu:
I just ran into the same thing, working with the PayPal SDK. Hard refreshes work fine, and soft navigation works fine as well, but the second navigation calls `onMounted` before the page is rendered. When I drop a `debugger` in, I can see I'm still clearly on the previous page (where I clicked the link), and so the target element doesn't exist.

fanckush:
@the-m-git could you provide a snippet?

danielrjames:
I had a similar issue where I could not call .focus() on a template ref in onMounted(). I found that the setTimeout hack does work, but it's pretty messy, so I'm not really a fan of that.

[...]

PR #5469, opened 2022-06-15
feat(nuxt): add warning in dev mode if layouts/pages do not have a single root node

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/4598, resolves https://github.com/nuxt/framework/issues/6476

### ❓ Type of change

### 📚 Description

This is a persistent issue for users and this change aims to improve dx in dev mode by proactively warning about components that render a comment (ie. a fragment).

~~@antfu Can you think of a better way of checking this than inspecting the el with a `setTimeout`?~~

~~**Note**: this will only provide the warning on first render at the moment, rather than on subsequent navigation. Will need to look into how that can be done.~~

### 📝 Checklist


antfu:
This is the method I could think of as well. Would `nextTick`, or `onMounted` be better than `setTimeout`?

danielroe:
I couldn't think of a way to access `onMounted` within the render function but feel free to implement if you have an idea - it would be more idiomatic (except note https://github.com/nuxt/framework/issues/3587 - 🤦)!

danielroe:
This will only provide the warning on first render at the moment, rather than on subsequent navigation. Will need to look into how that can be done. (Marking as draft for that reason.)

packages/nuxt/src/app/components/layout.ts
```diff
@@ -16,6 +16,18 @@ export default defineComponent({
   setup (props, context) {
     const route = useRoute()
 
+    let vnode: VNode
+    let _layout: string | false
+    if (process.dev && process.client) {
+      onMounted(() => {
```
pi0, line 22: Can we somehow expose it as internal util to share with pages?

danielroe: Hm. I've extracted it out to test, and it results in a net increase in LOC (+10) to handle the edge cases (e.g. page message depends on vnode, whereas layout has to be set. It's also a little trickier to ensure it's tree-shaken. But doable if you think the benefits are worth it.

(Personally I would wait until we use it somewhere else before creating shared util.)

pi0 (approved):

PR #5505, opened 2022-06-17
fix(vite): disable hmr in build

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/5196

### ❓ Type of change

### 📚 Description

Vite tests whether hmr is truthy in deciding whether to inject certain code. We should disable this in build mode as a matter of best practice, although there shouldn't be any active bugs as we are [now overriding NODE_ENV](https://github.com/nuxt/framework/pull/5417).

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -79,12 +79,14 @@ export async function bundle (nuxt: Nuxt) {
           watch: {
             ignored: isIgnored
           },
-          hmr: {
-            // https://github.com/nuxt/framework/issues/4191
-            protocol: 'ws',
-            clientPort: hmrPort,
-            port: hmrPort
-          },
+          hmr: nuxt.options.dev
+            ? {
+                // https://github.com/nuxt/framework/issues/4191
+                protocol: 'ws',
+                clientPort: hmrPort,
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Nice find! Should we also apply for other options such as `watch`?

packages/vite/src/vite.ts
```diff
@@ -76,15 +76,15 @@ export async function bundle (nuxt: Nuxt) {
...
-            ignored: isIgnored
-          },
-          hmr: {
-            // https://github.com/nuxt/framework/issues/4191
-            protocol: 'ws',
-            clientPort: hmrPort,
-            port: hmrPort
-          },
+          watch: nuxt.options.dev ? { ignored: isIgnored } : {},
+          hmr: nuxt.options.dev
```
pi0, line 80: Refactor suggestion: What if after constructing initial object, we directly nullify some dev-only options like `ctx.config.server.watch = false`. This way we can make code simpler with less inline `?` condition and also add a clear comment above why we are doing this.

pi0 (approved):

antfu (approved):

Issue #5535, opened 2022-06-20
moduleContainer addlayout not working


savyjs:
### Environment

Nuxt CLI v3.0.0-rc.4                                                                                                                                                            19:15:25
RootDir: nuxt3-test                                                                                                                                        19:15:57
Nuxt project info:                                                                                                                                                              19:15:57

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v16.13.0`
- Nuxt Version:     `3.0.0-rc.4`
- Package Manager:  `yarn@1.22.17`
- Builder:          `vite`
- User Config:      `modules`, `content`
- Runtime Modules:  `@nuxt/content@2.0.1`, `~/modules/test`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-dgnpdx-tuc35v?file=pages%2Findex.vue,nuxt.config.ts,modules%2Ftest%2Fpackage.json,modules%2Ftest%2Fsrc%2Fmodule.ts

### Describe the bug

@nuxt/kit moduleContainer addLayout does not work

### Additional context

Invalid layout `name-of-layout` selected.

PR #5537, opened 2022-06-20
feat(kit): add `addLayout`

danielroe:
### ❓ Type of change

resolves https://github.com/nuxt/framework/issues/5535

### 📚 Description

This adds a Nuxt 3 implementation for `this.addLayout` and also exports it as a standalone util. Happy to go in the other direction and reconsider the standalone utility (or even deprecate the module container util), but if we have it, we might as well make it available and working on nuxt 3.

### 📝 Checklist


pi0:
I think this PR is introducing a new feature :)

packages/kit/src/layout.ts
```diff
@@ -0,0 +1,32 @@
...
+      logger.warn(`Duplicate layout registration, "${layoutName}" has been registered as "${layout}"`)
+    }
+    nuxt.options.layouts[layoutName] = `./${filename}`
+    if (name === 'error') {
+      this.addErrorLayout(filename)
+    }
+  } else {
+    // Nuxt 3 adds layouts on app
+    nuxt.hook('app:templates', (app) => {
+      app.layouts[layoutName] = {
```
pi0, line 26: Should we make behavior consistent for warning about duplicate/overriding?

packages/kit/src/layout.ts
```diff
@@ -0,0 +1,32 @@
...
+    // Nuxt 2 adds layouts in options
+    const layout = nuxt.options.layouts[layoutName]
+    if (layout) {
+      logger.warn(`Duplicate layout registration, "${layoutName}" has been registered as "${layout}"`)
+    }
+    nuxt.options.layouts[layoutName] = `./${filename}`
+    if (name === 'error') {
+      this.addErrorLayout(filename)
+    }
+  } else {
```
pi0, line 23: Code style: We can early return here (or better separate Nuxt 2 logic to internal addNuxt2Layout)

packages/kit/src/layout.ts
```diff
@@ -0,0 +1,32 @@
...
+  const layoutName = name || parse(src || filename).name
+
+  if (isNuxt2(nuxt)) {
+    // Nuxt 2 adds layouts in options
+    const layout = nuxt.options.layouts[layoutName]
+    if (layout) {
+      logger.warn(`Duplicate layout registration, "${layoutName}" has been registered as "${layout}"`)
+    }
+    nuxt.options.layouts[layoutName] = `./${filename}`
+    if (name === 'error') {
```
pi0, line 20: Should we backport this too to Nuxt 3?

Atinux: I think warning would be best since is was not great to have an error layout.

Would be nice to expose another util for the error handler.

packages/kit/src/layout.ts
Atinux: It would be nice to test if I have a layout in my personal project named `layouts/foo.vue` and check if the module is overwriting it or the opposite.

I believe layouts added by modules has less priority than user defined.

packages/kit/src/layout.ts
```diff
@@ -0,0 +1,41 @@
...
+import { join, parse, relative } from 'pathe'
+import { isNuxt2 } from './compatibility'
+import { useNuxt } from './context'
+import { logger } from './logger'
+import { addTemplate } from './template'
+
+export function addLayout (tmpl: NuxtTemplate, name?: string) {
+  const nuxt = useNuxt()
+  const { filename, src } = addTemplate(tmpl)
+  const layoutName = name || parse(src || filename).name
```
pi0, line 11: Shall we normalize the layout name? (casing)

packages/kit/src/layout.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  // Nuxt 3 adds layouts on app
+  nuxt.hook('app:templates', (app) => {
+    if (layoutName in app.layouts) {
+      const relativePath = relative(nuxt.options.srcDir, app.layouts[layoutName].file)
+      return logger.warn(
+        `Not overriding \`${layoutName}\` (provided by \`~/${relativePath}\`) with \`${src || filename}\`.`
+      )
+    }
+    app.layouts[layoutName] = {
```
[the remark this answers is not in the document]
danielroe: A user might legitimately have an layout _named_ error in Nuxt 3.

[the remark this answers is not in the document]
pi0: I see but I found it confusing with the behavior of Nuxt 2 and for migration. Maybe warning for a while until projects migrate to Nuxt 3 until a point in 3.x release line?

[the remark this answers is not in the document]
danielroe: I would suggest just adding it to the migration docs?

pi0 (approved):

sppmstar:
when will this update be published ?

Atinux:
You can already play with it using the edge channel @sppmstar : https://v3.nuxtjs.org/guide/going-further/edge-channel

sppmstar:
That worked! thanks a lot :)

sppmstar:
So, the method will add a layout with a generated key which looks like this:

```
'layouts-editor-layout-c-142c-378': {
    name: 'AsyncComponentWrapper',
    __asyncLoader: [Function: load],
    __asyncResolved: [Getter],
    setup: [Function: setup]
}
```
What I don't understand is : it works if i specify `layouts-editor-layout-c-142c-378` in definePageMeta, but I find this not super intuitive... is there a way we can specify the layout name somehow ? My implementation code looks like this:

```
const layoutsDir = fileURLToPath(new URL('./layouts', import.meta.url))
const tpl = addTemplate(resolve(layoutsDir, 'editor-layout.vue'))
addLayout(tpl)
```

PR #5553, opened 2022-06-21
fix(head): inject `<script>` in rendered HTML when `ssr: false`

### LINKED ISSUE #5492: <script> injected in <head> tag cannot be executed before application code
### Environment

Nuxt project info:                                                                                                                                                                                                  11:29:52

- Operating System: `Darwin`
- Package Manager:  `npm@6.14.15`
- Builder:          `vite`
- User Config:      `typescript`, `modules`, `ssr`, `app`, `css`, `vite`
- Runtime Modules:  `@pinia/nuxt@0.1.9`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-ztv7ch-3v2mxy?file=nuxt.config.ts

### Describe the bug

As described here: https://github.com/nuxt/framework/discussions/5476. I need to inject a script in the `<head>` tag, to be able to set some runtime configuration on the global `window` object.

I'm injecting the script using the `nuxt.config.ts`:
```ts
  ssr: false, 
  typescript: {
    shim: false,
  },
  modules: ["@pinia/nuxt"],
  app: {
    head: {
      link: [{ rel: "icon", type: "image/x-icon", href: "/favicon.ico" }],
      script: [
        {
          // injects the config based on the environment
          src: "/static/config/config.js",
        },
      ],
    },
  },
  vite: {
    plugins: [svgLoader()],
  },
});
```

`config.js`:
```
window.config = {
  ENV: "dev",
  API: {
    EXPERIENCE_RAW_SERVICE_BASE_URL: "https://experience-raw-service.dev.company.com",
  },
};
```

The new [meta module](https://github.com/nuxt/framework/tree/main/packages/nuxt/src/head) injects this script at runtime (as you can see, there is no `<script>` tag in the html output):
```html
<!DOCTYPE html>
<html >

<head >
  <link rel="stylesheet" href="/_nuxt/entry.983043ac.css">
</head>

<body >
  <div id="__nuxt"></div><script>window.__NUXT__={serverRendered:false,config:{public:{},app:{baseURL:"\u002F",buildAssetsDir:"\u002F_nuxt\u002F",cdnURL:""}}}</script><script type="module" src="/_nuxt/entry-3549318c.mjs"></script>
</body>

</html>
```

The problem is that this script should be executed before anything else. But as you can see in the [repro](https://stackblitz.com/edit/github-ztv7ch-3v2mxy?file=package.json,shims.global.d.ts,types%2FConfigEnv.ts,nuxt.config.ts,public%2Fconfig%2Fconfig.js) project, if I'm trying to access `window.config` in a Plugin, I get `undefined`.
This is because some of the application code is run before the `config.js` script is injected in the `<head>`.

There should be a way to run this script before anything else in the page.

**Note**: This issue is related to "client side only" projects (`ssr: false`), setting `ssr: true` solves this issue.

### Additional context

_No response_

### Logs

_No response_

JulianMar:
### 🔗 Linked issue

#5492 

### ❓ Type of change

### 📚 Description

Add global Header to prerenderd HTML on spa mode.
This is just a first draft, so I would welcome some feedback. For example how to better handle the creation of the meta object, as I have seen there are two libraries for generating the meta object.

Resolves #5492 

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -24,8 +24,8 @@ interface RenderResult {
     bodyAttrs: string,
     headAttrs: string,
     headTags: string,
-    bodyScriptsPrepend : string,
-    bodyScripts : string
+    bodyScriptsPrepend: string,
+    bodyScripts: string
   }>
 }
 
@@ -77,7 +77,7 @@ const getSPARenderer = lazyCachedFunction(async () => {
       entryFiles = clientManifest.initial.map(file => ({ file }))
     }
 
-    return Promise.resolve({
+    const rendered: RenderResult = {
       html: '<div id="__nuxt"></div>',
       renderResourceHints: () => '',
       renderStyles: () =>
@@ -93,7 +93,13 @@ const getSPARenderer = lazyCachedFunction(async () => {
             return `<script ${isMJS ? 'type="module"' : ''} src="${buildAssetsURL(file)}"></script>`
           })
           .join('')
-    })
+    }
+
+    if (config.meta) {
+      rendered.meta = config.meta
+    }
+
+    return Promise.resolve(rendered)
   }
 
   return { renderToString }
@@ -163,7 +169,7 @@ async function renderHTML (payload: any, rendered: RenderResult, ssrContext: Nux
   })
 }
 
-function lazyCachedFunction <T> (fn: () => Promise<T>): () => Promise<T> {
+function lazyCachedFunction<T> (fn: () => Promise<T>): () => Promise<T> {
   let res: Promise<T> | null = null
   return () => {
     if (res === null) {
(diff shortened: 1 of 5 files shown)


Atinux:
Thank you for the PR, indeed this is a feature we are missing from Nuxt 2. I will let @pi0 and @danielroe check about the implementation details if there is not a better way than hooking into `nitro:config`.

Also something I would like to explore later if the ability to render an "App shell" or rendering `app.vue` to get the meta data defined in this component.

pi0:
> You want to decouple the head implementation from vue. Does that mean you just want to remove the plugins and do most of it server side or just don't attach it to the nuxt instance? I am just curious about how this will work together, and I am not that deep in nuxt yet :D

The idea is to split `@vueuse/head` into a generic JS lib for head generation and vue bindings to the Nuxt core. This way we have a stable head API for nuxt using same logic of common lib that can power ideas such as this and upcoming `@nuxt/script` module. Probably more possibilities. Essentially any head logic that needs to live outside of (Vue)SSR lifecycle.

pi0:
Hi @JulianMar. I've added a quick workaround here https://github.com/nuxt/framework/issues/5492#issuecomment-1213419740. I've to close PR since is stalled but will do my best to deliver unjs util ASAP making it easier for integration. In the meantime, you can create it using a nitro plugin (please ping if need help on how)

And once again, thanks for contributing to Nuxt <3

PR #5630, opened 2022-06-27
perf(nuxt)!: remove legacy app context

### LINKED ISSUE #5021: Reduce compatibility layer via feature flag
Hello 👋 

right now, the current build size of a "minimal" Nuxt application is [roughly 90kB](https://github.com/nuxt/framework/discussions/5016) including Vue, Nuxt, a custom error page, etc.
One of the included parts is the compatibility layer to Nuxt2. This is very helpful when migrating applications but might not be needed (in this extent) for new applications.

Thus, I'd suggest to add a feature flag that slims down the compatibility layer to a minimum and reducing the filesize.

In a more sophisticated approach, auto detecting which parts of the compat layer are needed and disabling those that are not used could be a viable way as well.

PS: As @danielroe highlighted `window.$nuxt` might be necessary for DevTools/VueTelescope detection and might be kept (as minimal as possible).

danielroe:
### 🔗 Linked issue

resolves #5021

### ❓ Type of change

### 📚 Description

We are currently including a legacy plugin that creates a proxy converting the Nuxt 3 app interface to a Nuxt 2 compatible context. This is probably not needed in majority of new apps, but it is still included in bundle.

This PR completely removes the legacy context feature.

### 👉 Migration

If you are using plugins (perhaps injected by a module) that are relying on the Nuxt 2 format of `(context, inject) => void` then it needs to be updated to the new Nuxt 3 format of `(nuxtApp) => void`: https://v3.nuxtjs.org/guide/directory-structure/plugins#creating-plugins

Moreover, if you are relying on the globally available legacy-format `window.$nuxt` you will need to inject it yourself.

### 📝 Checklist


pi0:
Do you have a benchmark of prod bundle size impact and percentage savings? It is considerably large, I see no reason to even keep backward-compatible code into the final release.

danielroe:
This PR will mean by default this plugin is not included in the bundle. The impact was ~4/92 kB of uncompressed JS (saving approx 4%). Worth mentioning that previously because the components plugin was not wrapped in defineNuxtPlugin, by default the legacy app plugin was always running.

pi0:
Then let's remove it as soon as possible to avoid misleading behavior. 4.5Kb is considerable (although gzip is probably much less in entry bundle). I initially added it for vue devtool and with intention of supporting Nuxt 2 compatible modules injecting a plugin but today it seems we need to migrate them anyway. It wouldn't worth to continue experimenting...

danielroe:
Sounds good to me! Will update 👍

danielroe:
**Note to self**: an example of a plugin that _does_ require this is https://github.com/moritzsternemann/vue-plausible/blob/main/src/nuxt-plugin.ts.

pi0 (approved):

PR #5688, opened 2022-07-01
feat(nuxt): server-only components

danielroe:
### 🔗 Linked issue

- Pending: https://github.com/nuxt/framework/pull/5689 (Implements: https://github.com/nuxt/framework/discussions/666)
- Resolves https://github.com/nuxt/framework/issues/648

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

#### Summary

This PR adds support for:

1. rendering individual components through a special GET request to `/__nuxt_render` (implementing RFC https://github.com/nuxt/framework/discussions/666)
2. registering and rendering server components with `.server.vue` suffix (implementing https://github.com/nuxt/framework/issues/648), including injecting scripts + styles from this render back into client-side DOM. As part of this, we _only_ build CSS from server-components into client build (which is required), meaning the code from server-components won't be present in your client bundle.

CleanShot 2022-07-01 at 11 10 13

#### Next steps

- [ ] The key next step is to review to see if we want to make any changes to RFC or spec.
- [ ] Update documentation with details about server (+ client) components

#### Future improvement ideas

1. 'Batching' server component updates
2. Detect 'dependencies' of server components so entire state doesn't need to be pushed
3. Detect 'effects' of server components so entire state doesn't need to be returned from renderer
4. Move server-only component CSS out of manifest (but this can also be handled as part of solution to https://github.com/nuxt/framework/issues/4475)

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #5689, opened 2022-07-01
feat(nuxt): server component islands 

danielroe:
### 🔗 Linked issue

- Implements: https://github.com/nuxt/framework/discussions/666

### ❓ Type of change

### 📚 Description

This PR adds support for rendering individual components through a special GET request to `/__nuxt_render` (implementing RFC https://github.com/nuxt/framework/discussions/666).

In addition to providing the basis for server-only components (https://github.com/nuxt/framework/pull/5688) it can also provide a way to render individual components for future testing or devtools functionality.

**Note** that this only supports components that are already globally registered in the server.

If you encounter any issues with rendering async components in this way, please also see https://github.com/vuejs/core/issues/6207.

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -111,6 +111,7 @@ const getSPARenderer = lazyCachedFunction(async () => {
 export default eventHandler(async (event) => {
   // Whether we're rendering an error page
   const ssrError = event.req.url?.startsWith('/__nuxt_error') ? useQuery(event) : null
+  const customRender = event.req.url?.startsWith('/__nuxt_render') ? useQuery(event) : null
```
pi0, line 114: We shall support POST as well. Both for security aspect (not serializing event to the URL that is logged and potentially cached in CDN) and size limitation.

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -122,8 +123,11 @@ export default eventHandler(async (event) => {
     runtimeConfig: useRuntimeConfig(),
     noSSR: !!event.req.headers['x-nuxt-no-ssr'],
     error: ssrError,
+    render: customRender
+      ? { components: JSON.parse(customRender.components as string || '[]') }
```
pi0, line 127: Maybe one time parse (with destr) and type customRender context?

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -148,18 +152,40 @@ export default eventHandler(async (event) => {
...
 
+  await renderMeta(rendered, ssrContext)
+
+  // Render server components
+  if (ssrContext.teleports?.['render-target-0']) {
+    const components = Object.entries(ssrContext.teleports)
+      .filter(([key]) => key.startsWith('render-target'))
+      .map(([, value]) => ({ html: value.replace(/<!--teleport anchor-->$/, '') }))
+
+    event.res.setHeader('Content-Type', 'application/json')
```
pi0, line 163: Doesn't h3 set this?

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -148,18 +152,40 @@ export default eventHandler(async (event) => {
...
+
+    const result: ComponentRenderResult = {
+      rendered: components,
+      state: ssrContext.payload.state,
+      style: rendered.renderStyles() + (ssrContext.styles || ''),
```
pi0, line 168: styles?

packages/nuxt/src/pages/module.ts
```diff
@@ -76,7 +76,12 @@ export default defineNuxtModule({
         const pages = await resolvePagesRoutes()
         await nuxt.callHook('pages:extend', pages)
         const { routes, imports } = normalizeRoutes(pages)
-        return [...imports, `export default ${routes}`].join('\n')
+        return [
+          ...imports,
+          `const routes = ${routes}`,
+          "if (process.server) { routes.push({ name: '_nuxt_render', path: '/__nuxt_render', component: { render: () => '' } }) }",
```
pi0, line 82: https://github.com/nuxt/framework/pull/5689/files#r911861340 then we won't need this anymore :)

danielroe: We could take the URL as part of the request, but that makes it harder to cache...

pi0: Shouldn't pages/vue-router bypassed whwn rendering individual?

danielroe: You mean, that particular plugin should be bypassed? Well, it's tricky. I think we should either:

1. We can either say that server-only components _cannot_ reference router (or any plugins at all?)
2. We can include optionally include URL in the request and set this for the purposes of rendering the component, e.g. a breadcrumb component could be rendered in this way, or a markdown component that relies on the URL to get the correct content....

pi0: We can push url to state context and rely on universal routing utils. But it theb reduces caching possibility (alternative is passing route as explicit prop as best practices and document it)

danielroe: Agreed about passing route as explicit prop. So we skip routing but run other plugins? (Seems good to me.)

pi0: We might find other cases. What do you think of providing a flag to the app context (like `isIndividualRender`) and plugins such as router can self-disable for thee new mode?

danielroe: Great! Implemented. And added a test suite too.

pi0: Would be nice if we introduce it as an app flag (it can infer from route for now) instead of composable.

danielroe:
FYI: I'll be refactoring this to support only explicit island components rather than any global components. That change is pending merge of https://github.com/nuxt/framework/pull/6070.

danielroe:
This regexp does not match global components.

pi0:
To improve caching possibilities and isolation, I've made some refactors. New `/__nuxt_island` endpoint accepts `name` (required), `props` and `format` queries and returns constructed HTML object with only different that rendered body, is (teleported) island component. This way we can apply same render: transforms.

Example endpoints (working with `yarn nuxt dev test/fixtures/basic`)
- http://localhost:3000/__nuxt_island/?name=PureComponent
- http://localhost:3000/__nuxt_island/?name=PureComponent&format=html

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    async function fetchComponent () {
+      nuxtApp[pKey] = nuxtApp[pKey] || {}
+      if (!nuxtApp[pKey][hashId.value]) {
+        nuxtApp[pKey][hashId.value] = _fetchComponent().finally(() => {
+          delete nuxtApp[pKey][hashId.value]
+        })
+      }
+      const res = await nuxtApp[pKey][hashId.value]
+      injectHead(res.html)
```
danielroe, line 50: I would check that this doesn't lead to a memory leak when components are updated on client-side. `useHead` is a composable and in general I would expect that it be used within a setup function, rather than programmatically to insert elements into the head.

danielroe, line 50: I would check that this doesn't lead to a memory leak when components are updated on client-side. `useHead` is a composable and in general I would expect that it be used within a setup function, rather than programmatically to insert elements into the head.

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (process.server || !nuxtApp.isHydrating) {
+      await fetchComponent()
+    }
+
+    if (process.client) {
+      watch(props, debounce(fetchComponent, 100))
+    }
+
+    return () => createStaticVNode(html.value, 1)
```
danielroe, line 62: I believe this will cause hydration issues if the component doesn't have a single root node.

packages/nuxt/src/components/plugins/server-components.ts
```diff
@@ -1,36 +0,0 @@
...
-}
-
-export const stripServerComponentsPlugin = createUnplugin((options: StripServerComponentsPluginOptions) => {
-  return {
-    name: 'nuxt:components:strip-server-components',
-    enforce: 'post',
-    transformInclude (id) {
-      const components = options.getComponents()
-      return components.some(c => c.filePath === id && c.mode === 'server' && !components.some(o => o.mode === 'client' && o.pascalName === c.pascalName))
-    },
```
danielroe: For any server component implementation (I realise with islands, it's a little more complex), it will be important to be able to tree-shake the actual code for the component, or we'll end up shipping it in both bundles. Would you explain a little about why you've completely removed the plugin?

danielroe (commented):
(I'm sure you're already looking at resolving the test failures.)

In general, lots of really nice changes.

My main worry is that this PR was meant to focus on the rendering side of things and leave discussion of component implementation to https://github.com/nuxt/framework/pull/5688. So I deliberately didn't address that stage of implementation, and it feels like these changes are now being made quite quickly at the end stage of a long review process. Would love the chance to look through them in more detail, _alongside_ the server-component implementation.

pi0:
> My main worry is that this PR was meant to focus on the rendering side of things and leave the discussion of component implementation to https://github.com/nuxt/framework/pull/5688.

Server-only components proposed in #5688 are a really nice addition to build on top of component islands API and looking forward to it 🔥

Islands API not only allows implementing server-only components but also being able to fetch components from remote sources in runtime (not same server) and more possibilities for testing, playground, etc. So I think `<NuxtIsland>` is more relevant to the scope of initial PR. We could try different ideas and improvements in parallel. Originally referred to this as selective-rendering/isolated-rendering in my RFC and with intention of making an external library to be integrated with Nuxt but adding to nuxt became more straightforward. Last changes are closer to what I had in mind and hope compatible with server-only components impl.

packages/nuxt/src/app/components/nuxt-island.ts
pi0: Yes, we enforce one root element for islands.

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -0,0 +1,89 @@
...
+  const tags: {tag: string, attrs: Record<string, string>}[] = []
+  for (const tagMatch of html.matchAll(HTML_TAG_RE)) {
+    const attrs = {} as Record<string, string>
+    for (const attraMatch of tagMatch.groups!.rawAttrs.matchAll(HTML_TAG_ATTR_RE)) {
+      attrs[attraMatch.groups!.name] = attraMatch.groups!.value
+    }
+    tags.push({ tag: tagMatch.groups!.tag, attrs })
+  }
+  return tags
+}
```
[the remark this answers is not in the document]
pi0: I will replace with head manager right afterwards. As explained in other comment it is experimental and _parsing_ should be universal.

packages/nuxt/src/app/components/nuxt-island.ts
```diff
@@ -0,0 +1,89 @@
...
+    function _fetchComponent () {
+      // TODO: Validate response
+      return $fetch<NuxtIslandResponse>(`${props.name}:${hashId.value}`, {
+        baseURL: '/__nuxt_island',
+        params: {
+          ...props.context,
+          props: props.props ? JSON.stringify(props.props) : undefined
+        }
+      })
+    }
```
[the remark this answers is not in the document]
pi0: Probably not. It is similar to `ClientOnly` vs `.client`. Server only components could probably have better embedding method.

packages/nuxt/src/components/plugins/server-components.ts
pi0: Islands need to be always available in the server bundle (request is dynamic). They are not meant to be directly referenced in code `<ComponentName>` but `<NuxtIsland name="">`. Server-only PR could register them to be directly referenced as component name and provide wrapper that handles rendering of them in server/client.

danielroe: Do you mean for islands to be included in client bundle?

packages/nuxt/src/app/components/nuxt-island.ts
pi0: Partially fixed by https://github.com/nuxt/framework/pull/5689/commits/22dfbcdf883a10e3ea457242ab4bd685fbaac8ce at least injects 1 time extra (might be solvable too) but ultimately agree we should avoid `useHead`

packages/nuxt/src/components/plugins/server-components.ts
pi0: I mean register a placeholder that renders them as server component probably not by default adding to client bundle (at least this PR is not doing this either)

packages/nuxt/src/app/components/nuxt-island.ts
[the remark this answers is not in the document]
pi0: Updated. Moved parser to renderer (later dedicated server plugin). This way we return minimal json ready to be rendered.

packages/nuxt/src/components/plugins/server-components.ts
danielroe: If they are not part of the client bundle then the CSS that renders them will not be produced.

If they _are_ part of the client bundle then they won't be fully isolated.

This was one of the main tasks of this plugin - to ensure a _partial_ build that includes CSS but not full JS. Would love to hear if you have another solution.

test/basic.test.ts
```diff
@@ -482,3 +484,181 @@ describe('app config', () => {
...
+      }</pre></div>"
+    `)
+    expect(result.state).toMatchInlineSnapshot(`
+      {
+        "$shasRouter": true,
+      }
+    `)
+
+    // TODO: Test response differes!
+    // expect(result.html.head.join(' ')).toMatch(/PureComponent[^"']*.css/)
```
danielroe, line 654:
```suggestion
    expect(result.html.head.join(' ')).toMatch(/PureComponent[^"']*.css/)
```

- [ ] resolve before merging

See https://github.com/nuxt/framework/pull/5689#discussion_r949365675. I suspect this is because (as commented there) the CSS is not being generated in client side build.

packages/nuxt/src/components/plugins/server-components.ts
pi0: Server should be able to extract and generate partial CSS for island components. I'm aware it is currently broken.

test/basic.test.ts
pi0: https://github.com/nuxt/framework/pull/5689#discussion_r949578267 i'm aware it is broken but main issue is that styles are not currently properly inlined from server bundle while they should. It is not unique issue to islands.

danielroe: I believe this is the intended behaviour from vite-plugin-vue, and not broken.

It would certainly be a nice enhancement to have the option to render inlined styles, of course, albeit at the cost of an increased server bundle size.

pi0: Let's track in #6755 but if it is actually behavior/limitation of `vite-plugin-vue` for server build, it means it is not ready for server-only components because transform plugin is essentially making partial use of client build to make styles working.

danielroe:
@pi0 Just want to check in on this. What are the next steps in the path to merging this PR? Is there anything I can do?

pi0:
Working inlined styles (https://github.com/nuxt/framework/issues/6755) to have a working MVP.

Sunhat:
#6755 Merged 😏

danielroe:
@pi0 What do you think about merging in behind an experimental flag?

Atinux:
Is it possible to update the Readme to give an example on how to use `<NuxtIsland>` component? Shall we start writing a page in the docs regarding the Islands component and the current state?

danielroe:
This PR was originally meant to be just the rendering method, with https://github.com/nuxt/framework/pull/5688 to implement the actual component to render the result. Somehow `<NuxtIsland />` was added to this PR, which overlaps somewhat with the server component wrapper in that PR. Once this is merged, it should be fairly quick to update that PR to implement proper server components + add documentation for `<NuxtIsland />`. (Or I could merge that one into this and add docs.) wdyt @Atinux?

Atinux:
Tell me if I am wrong, but `<NuxtIsland>` is a shortcut for the endpoint to render HTML from a component right?

#5688 goal is to support `.server.vue` suffix that will leverage `<NuxtIsland>` under the hood?

If so, I think we can instead merge this PR in that state and improve #5688 which already has a great description!

danielroe:
Exactly!

danielroe:
@pi0 What's the status of this PR?

pi0:
Update: This PR is really in a good state and is ready to be landed as an experimental feature. I've also worked on inlined styles to be working on production (7c66c41554ecd2fca5d512a1f32c45348f7cf312). 

Landing this feature for 3.1.0 alongside `.server` components. (Landing for 3.0.0 is really hard since we plan release today sorry for delays).

josueayala27:
🫶🏻🫶🏻🫶🏻 @danielroe

PR #5695, opened 2022-07-02
docs(examples): add pinia examples to docs

valcosmos:
### ❓ Type of change

### 📝 Checklist


examples/other/pinia/nuxt.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineNuxtConfig } from 'nuxt'
+
+// https://v3.nuxtjs.org/api/configuration/nuxt.config
+export default defineNuxtConfig({
+  buildModules: [
```
vencho-mdp, line 5: AFAIK, `buildModules` is deprecated, so I think `modules` option should be used instead.

valcosmos: Thank you very much, it was my oversight, and I have fixed it.😊

danielroe:
Thanks for your work on this! However, I'd suggest that we focus on making sure that this example is in the pinia docs instead. Might also be worth adding some information in the migration guide about how to migrate a store from Nuxt 2 -> Nuxt 3?

pi0:
Thanks for PR @valcosmos 

I agree with you @danielroe that module examples are probably to be in their own docs but pinia is kinda store replacement. I wouldn't mind having an exception for it :)

valcosmos:
> Thanks for your work on this! However, I'd suggest that we focus on making sure that this example is in the pinia docs instead. Might also be worth adding some information in the migration guide about how to migrate a store from Nuxt 2 -> Nuxt 3?
> 
> WDYT @pi0?

Hello, Thanks for u reply!  Should I close this PR?

pi0:
Let's close it for now. We might add pinia as default store implementation for Nuxt later on. Currently there are no external module integrations in examples dir.

Thanks again for contributing @valcosmos <3

PR #5738, opened 2022-07-05
feat(nuxt): share `asyncData` between calls

OhB00:
Had a look at https://github.com/nuxt/framework/issues/4758 and https://github.com/nuxt/framework/pull/5078 and made some changes to improve readability, and fix the race condition mentioned in the issue.

I have also implemented the change suggested in the other pull request to share `asyncData` between calls.

### 🔗 Linked issue

#4758 

### ❓ Type of change

### 📚 Description

1. I have refactored `useAsyncData` to use more async await, this (hopefully) makes it much more readable + manageable.
2. I have eliminated some dead code from various code paths, `_nuxtOnBeforeMountCbs` are no longer touched at all server side, data caching is handled top level rather than within refresh, 
3. `refresh` was simplified with the concept of 'tagging along', if a promise is found for a key, use that instead of calling the handler again.
4. `asyncData` is sync'd between calls, for example, if you had a `useUserData` composable, that displays user data in two separate components, calling refresh will change both.
5. As a result of the previous point various bugs related to no sync are resolved (#4758 and likely others)
6. Added a `force` option to refresh, this seems to make more sense then what was provided in the documentation originally, this may be a breaking change.
7. Added some tests for `useAsyncData`, as it's fairly easy to break.
8. Refreshing on the server side updates the cache, this seemed logical.

### 📝 Checklist

### DIFF
--- a/docs/content/3.api/1.composables/use-async-data.md
+++ b/docs/content/3.api/1.composables/use-async-data.md
@@ -51,7 +51,7 @@ Under the hood, `lazy: false` uses `<Suspense>` to block the loading of the rout
 * **refresh**: a function that can be used to refresh the data returned by the `handler` function
 * **error**: an error object if the data fetching failed
 
-By default, Nuxt waits until a `refresh` is finished before it can be executed again. Passing `true` as parameter skips that wait.
+By default, Nuxt waits until a `refresh` is finished before it can be executed again. Passing `{ force: true }` as parameter skips that wait.
 
 ## Example
 
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -33,7 +33,8 @@ export interface AsyncDataOptions<
 }
 
 export interface RefreshOptions {
-  _initial?: boolean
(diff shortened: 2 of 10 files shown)


antfu (changes requested):
Please revert the formatting changes before we could proceed reviewing.

OhB00 (commented):
Reverted formatting changes

pi0:
Thanks for your nice PR @OhB00 and sorry it took long to proceed. 

I have splitted your changes into smaller PR and issues (next times, please consider doing this. this way we can quickly iterate over smaller changes)

1. I prefer to avoid using await in this particular utility because we only have current component/nuxt context available until first await and this utility returns a hybrid promise interface
2. Merged via #7056
3. Iterating over with some refactors.
4. (**main fix**) Done via #7055 -- probably needs iteration to handle edge cases like ones @danielroe mentioned in https://github.com/nuxt/framework/pull/7055#discussion_r958315070
5. (same as 4)
6. Moved to discussion #7058
7. Merged via #7055
8. (same as 3)

Thanks again!

PR #5748, opened 2022-07-06
fix(nuxt): `useRequestHeaders` type should allow for undefined values

### LINKED ISSUE #5746: useRequestHeaders typing incorrect
### Environment

- Operating System: `Linux`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

[GitHub repo link](https://github.com/maxtomczyk/use-request-headers-repro)

To reproduce the bug just start the reproduction project and run `curl localhost:3000`

### Describe the bug

Type for example header cookie is string:

image

However, if I hit localhost:3000 with basic curl request (without `cookie` header set) cookie will be undefined, so the typing is correct only if header will be provided. I think it's important to fix `useRequestHeaders` return type as it can be misleading in some cases (for example kubernetes liveness probe as it happened to me).

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #5746

### ❓ Type of change

### 📚 Description

This explicitly adds `undefined` as a possible value of the return type. This will only have effect if users are using TypeScript in strict mode, as is best practice.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/ssr.ts
+++ b/packages/nuxt/src/app/composables/ssr.ts
@@ -3,8 +3,8 @@ import type { CompatibilityEvent } from 'h3'
 import { useNuxtApp } from '#app'
 import { NuxtApp } from '#app/nuxt'
 
-export function useRequestHeaders<K extends string = string> (include: K[]): Record<K, string>;
-export function useRequestHeaders (): Readonly<Record<string, string>>;
+export function useRequestHeaders<K extends string = string> (include: K[]): Record<K, string | undefined>
+export function useRequestHeaders (): Readonly<Record<string, string | undefined>>
 export function useRequestHeaders (include?) {
   if (process.client) { return {} }
   const headers: Record<string, string | string[]> = useNuxtApp().ssrContext?.event.req.headers ?? {}


packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -3,8 +3,8 @@ import type { CompatibilityEvent } from 'h3'
 import { useNuxtApp } from '#app'
 import { NuxtApp } from '#app/nuxt'
 
-export function useRequestHeaders<K extends string = string> (include: K[]): Record<K, string>;
-export function useRequestHeaders (): Readonly<Record<string, string>>;
+export function useRequestHeaders<K extends string = string> (include: K[]): Record<K, string | undefined>
+export function useRequestHeaders (): Readonly<Record<string, string | undefined>>
```
pi0, line 7: I might be wrong but isn't fallback an empty object? [src](https://github.com/nuxt/framework/blob/aa9950cfb352060b3c2294beb7b015465705d5d8/packages/nuxt/src/app/composables/ssr.ts#L9)

danielroe: Yes. That is what this types.

PR #5785, opened 2022-07-08
docs: improve runtimeConfig example

duvior:
Made the nuxt.config.ts example using .env, a little more explicit.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/2.features/10.runtime-config.md
+++ b/docs/content/2.guide/2.features/10.runtime-config.md
@@ -54,9 +54,9 @@ NUXT_PUBLIC_API_BASE=https://nuxtjs.org
 ```ts [nuxt.config.ts]
 export default defineNuxtConfig({
   runtimeConfig: {
-    apiSecret: process.env.NUXT_API_SECRET,
+    apiSecret: process.env.NUXT_API_SECRET || '', // .env value + an empty string as a fallback value
     public: {
-      apiBase: process.env.NUXT_PUBLIC_API_BASE,
+      apiBase: process.env.NUXT_PUBLIC_API_BASE || 'someBase', // Or .env value + a fallback value
       apiPublicKey: 'somePublicApiKey', // Or a default value
     }
   },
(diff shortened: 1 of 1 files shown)


danielroe (changes requested):
This won't work as written. The point being made above is:

> For this to work, you _must_ have a fallback value (which can just be an empty string) defined in your `nuxt.config`.

We're trying to avoid the situation where someone things that just because they have `apiSecret: process.env.NUXT_API_SECRET` defined in the config, that this will therefore means it gets overriden in production.

Instead, it must be _at least_: `apiSecret: process.env.NUXT_API_SECRET || ''`.

We also want to make it clear that the name of the environment variable is determined by the key within runtimeConfig - e.g. `apiSecret` becaomes `NUXT_API_SECRET` rather than by the environment variable passed as the default variable.

Any thoughts about how to make the above clearer in the example would be welcome.

docs/content/2.guide/2.features/10.runtime-config.md
```diff
@@ -54,9 +54,10 @@ NUXT_PUBLIC_API_BASE=https://nuxtjs.org
 ```ts [nuxt.config.ts]
   runtimeConfig: {
-    apiSecret: '',
+    apiSecret: process.env.NUXT_API_SECRET || '', // .env value + an empty string as a fallback value
```
danielroe, line 57:
```suggestion
    apiSecret: '', // can be overridden by NUXT_API_SECRET environment variable
```

docs/content/2.guide/2.features/10.runtime-config.md
```diff
@@ -54,9 +54,10 @@ NUXT_PUBLIC_API_BASE=https://nuxtjs.org
...
     public: {
-      apiBase: '', // Or a default value
+      apiBase: process.env.NUXT_PUBLIC_API_BASE || 'someBase', // Or .env value + a fallback value
+      apiPublicKey: 'somePublicApiKey', // Or a default value
```
danielroe, line 60:
```suggestion
      apiBase: '', // can be overridden by NUXT_PUBLIC_API_BASE environment variable
```

danielroe (commented):
Let's go with @pi0's suggestion, which I think is clearer.

danielroe (approved):

Issue #1865, opened 2021-11-11
Runtime.ImportModuleError - netlify deploy 


productdevbook:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.13.0`
- Nuxt Version:     `3.0.0-27277330.202617b`
- Package Manager:  `yarn@1.22.15`
- Bundler:          `Vite`
- User Config:      `buildModules`, `build`
- Runtime Modules:  `-`
- Build Modules:    `@pinia/nuxt@0.1.5`
------------------------------

### Describe the bug

https://v3nuxt.netlify.app

<img width="1287" alt="CleanShot 2021-11-11 at 17 33 30@2x" src="https://user-images.githubusercontent.com/38668796/141315700-5cf252bf-66c9-4414-b2ba-bc58c9def2ad.png">

### Reproduction

https://github.com/productdevbook/nuxt3-app

J-Kern:
This is a duplicate of #1826 .

Please refer to this answer from the [netlify-team-account](https://github.com/nuxt/framework/issues/1826#issuecomment-966190438).

PR #5796, opened 2022-07-09
feat(nuxt): nested layout files

DannyAndres:
This PR resolves the issue [#5793](https://github.com/nuxt/framework/issues/5793). A developer will be able to create multiple folders in /layouts and then access them in pages.

This feature was heavily inspired in the PR [Adds support for folders in /layouts #1865
](https://github.com/nuxt/nuxt.js/pull/1865) for Nuxt 2 and works in the exact same way.

For instance, if in /layouts there are 2 folders like:

- layouts
  - folder1
  - folder2

the developer is able to create files which are named the same but they are in different folders.
Like folder1/default.vue and folder2/default.vue and then access those layouts in /pages using following rule.

`folder_in_layouts` + `layout-file-name`


DannyAndres:
In advance sorry for the extra edition, was done by `npx yarn lint --fix`

danielroe (changes requested):
Nice idea! If you get a chance, would you please revert the spacing/formatting changes to help us as we review this?

DannyAndres:
Sure, I will send another commit directly from main and manually lint the corresponding files

jjjrmy:
bump, really interested in this feature

PR #5797, opened 2022-07-09
docs: use `defineLazyHandler` in wasm example

valcosmos:
### ❓ Type of change

### 📝 Checklist


examples/experimental/wasm/server/api/sum.ts
```diff
@@ -1,7 +1,9 @@
-import { useQuery, lazyHandle } from 'h3'
+import { useQuery, defineLazyHandler } from 'h3'
 
-export default lazyHandle(async () => {
-  const { exports: { sum } } = await loadWasmInstance(
+export default defineLazyHandler(async () => {
+  const {
+    exports: { sum }
+  } = await loadWasmInstance(
```
danielroe, line 6: Let's keep formatting.

```suggestion
  const { exports: { sum } } = await loadWasmInstance(
```

examples/experimental/wasm/server/api/sum.ts
```diff
@@ -1,7 +1,9 @@
-import { useQuery, lazyHandle } from 'h3'
+import { useQuery, defineLazyHandler } from 'h3'
```
danielroe, line 1: `useQuery` is auto-imported.

```suggestion
import { defineLazyHandler } from 'h3'
```

danielroe (commented):
Thanks!

Would you also update the return so it's not using `req` but `event`?

valcosmos:
> Thanks!
> 
> Would you also update the return so it's not using `req` but `event`?



> Thanks!
> 
> Would you also update the return so it's not using `req` but `event`?



Thank you for your reply, and I have replaced req with event.

valcosmos:
> Thanks!
> 
> Would you also update the return so it's not using `req` but `event`?

Thank you for your reply, and I have replaced req with event.

valcosmos:
> Thanks!
> 
> Would you also update the return so it's not using `req` but `event`?

Thanks for your reply, and I have replaced req with event.

pi0 (approved):

pi0:
Thanks!

PR #5809, opened 2022-07-10
optimization: we can run one single check for both

modex98:
in js we have `null == undefined`

### ❓ Type of change

### 📝 Checklist


DreadIsBack:
Maybe just write 'if(!value)'?

pi0:
Thanks for PR but we tend to avoid less strict check such as `==` for code readability. Difference is not noticable.

PR #5836, opened 2022-07-11
fix(nuxt): set unctx instance in callWithNuxt

### LINKED ISSUE #5740: using try/catch with an async method inside a middleware causes a `nuxt instance unavailable` error.
### Environment

- Operating System: `Linux`
- Node Version:     `v17.9.0`
- Builder:          `vite`
- User Config:      `modules`, `runtimeConfig`, `autoImports`
- Runtime Modules:  `@nuxtjs/tailwindcss@5.1.3`
- Build Modules:    `-`

### Reproduction

**Visit the `/secret` page to see the error:**

https://stackblitz.com/edit/nuxt-starter-elpq43?file=middleware%2Fauth.ts

### Describe the bug

I'm trying to write an async middleware. However, when catching the rejected promise an error is thrown by nuxt when using `navigateTo()`:

```ts
async function fetchUser() {
  throw new Error();
}

export default defineNuxtRouteMiddleware(async (to, from) => {
  let user;

  try {
    user = await fetchUser();
  } catch (e) {
    user = null;
  }

  if (!user) return navigateTo('/');
});
```

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve #5740

### ❓ Type of change

### 📚 Description
server-side, set the nuxt instance in unctx if ctx.use() return null since async middlewares cannot retrieve the vue vm when calling useNuxtApp().

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/nuxt.ts
+++ b/packages/nuxt/src/app/nuxt.ts
@@ -205,6 +205,10 @@ export function defineNuxtPlugin<T> (plugin: Plugin<T>) {
 export function callWithNuxt<T extends (...args: any[]) => any> (nuxt: NuxtApp | _NuxtApp, setup: T, args?: Parameters<T>) {
   const fn = () => args ? setup(...args as Parameters<T>) : setup()
   if (process.server) {
+    if (!nuxtAppCtx.use()) {
+      // set instance to unctx, #5740
+      nuxtAppCtx.set(nuxt)
+    }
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for the PR but i'm afraid this is not a solution.

`nuxtAppCtx.callAsync` temporarily sets the active context while executing Nuxt plugins in the server-side. Permanently setting context causes sharing state between two requests.

PR #5859, opened 2022-07-12
feat(nuxt): config options for default keepalive, page & layout transitions

### LINKED ISSUE #5729: allow setting default route transition (page/layout) properties globally
currently it's only possible to set transition props on a per route basis with `definePageMeta`, or via something like this:

```js
  modules: [
    // ...
    function (options, nuxt) {
      nuxt.hook('pages:extend', (pages) => {
        for (const page of pages) {
          page.meta = {
            layoutTransition: false,
            pageTransition: false,
          }
        }
      })
    }
  ],
})
```

It would be better to support configuring this so it can easily be disabled or set on a project-wide base

danielroe:
### 🔗 Linked issue

resolves #5729

### ❓ Type of change

### 📚 Description

This allows support for configuring page & layout transition default values, as well as keepalive. (Otherwise there is the tedious task of doing so for _every_ individual file.)

In order to do this, we now inject `#build/app.config.mjs` to automatically expose all values within the `app` key in Nuxt config, rather than creating ever-more virtual files for particular use-cases. (We would, for example, have had to do this for layouts and page transitions _separately_ otherwise.) This means we were able to drop `meta.config.mjs` and in general is a useful pattern for future `app` configuration. 

### 📝 Checklist


packages/nuxt/src/core/templates.ts
```diff
@@ -74,6 +74,34 @@ export const serverPluginTemplate = {
...
+
+<body \${params.BODY_ATTRS}>\${params.BODY_PREPEND}
+  \${params.APP}
+</body>
+
+</html>\`
+`
+  }
+}
+
```
danielroe, line 104: Merge issue?

```suggestion
```

pi0: My bad. 33912d4a79839f4ecf5bf0a28ef31c985e7800ef

pi0:
Thanks for rebasing. Shall we type `appConfig.nuxt` from schema now?

Atinux:
Beautiful work @danielroe ❤️ 

Shall we document more to use `app.config.ts` for such configurations from now on?

packages/schema/src/types/config.ts
```diff
@@ -65,10 +65,11 @@ export interface RuntimeConfig extends PrivateRuntimeConfig, RuntimeConfigNamesp
 export interface AppConfigInput extends Record<string, any> {
   /** @deprecated reserved */
   private?: never
-  /** @deprecated reserved */
-  nuxt?: never
+  nuxt?: DeepPartial<Omit<ConfigSchema['app'], 'baseURL' | 'buildAssetsDir' | 'assetsPath' | 'cdnURL'>>
```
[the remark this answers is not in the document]
danielroe: Good point. A further comment - do we want to nest all of the nuxt config under `nuxt`? Or do you think we can pull `head` out as a top-level key?

pi0:
> Shall we document more to use app.config.ts for such configurations from now on?

Currently usage of `.nuxt` in `app.config` is kinda internal. For some, we might have built-time optimization opportunities (like **not bundling transition bytes**) while if we make users able to set via `app.config`, we cannot do this anymore. I suggest to remove types from `AppConfigInput` as well.

packages/schema/src/types/config.ts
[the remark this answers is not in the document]
pi0: Yes, we should progressively move them to `head` and as @Atinux mentioned, support them for actual user. I would suggest to do this in a later PR for user-input app config from `app.config` (we might be careful about build-time optimization and what can be there. `head` is safer probably...)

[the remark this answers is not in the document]
pi0: I consider `nuxt` more an internal namespace we use to transfer config and persist.

packages/schema/src/types/config.ts
```diff
@@ -71,4 +73,13 @@ export interface AppConfigInput extends Record<string, any> {
...
-export interface AppConfig { }
+export interface NuxtAppConfig {
+  head: MetaObject
+  layoutTransition: boolean | TransitionProps
+  pageTransition: boolean | TransitionProps
+  keepalive: boolean | KeepAliveProps
+}
+
+export interface AppConfig {
+  nuxt: NuxtAppConfig
```
pi0, line 84: Last change request!! (sorry for splitting comments). Shall we use `_nuxt` (while keeping `nuxt` reserved) for now? I'm worried if users still start depending on this new interface it might get broken for them as we refactor for top-level API unless we add backward compatibility for an option just added.

pi0 (approved):

pi0:
(fixture seems broken)

danielroe:
Indeed, that is one of the key use-cases of this PR. It is possible to set them to false and it will disable keepalive, page & layout transitions.

packages/nuxt/src/pages/runtime/page.ts
```diff
@@ -35,9 +36,9 @@ export default defineComponent({
 
           const key = generateRouteKey(props.pageKey, routeProps)
 
-          return _wrapIf(Transition, routeProps.route.meta.pageTransition ?? defaultPageTransition,
-            wrapInKeepAlive(routeProps.route.meta.keepalive, isNested && nuxtApp.isHydrating
-            // Include route children in parent suspense
+          return _wrapIf(Transition, routeProps.route.meta.pageTransition ?? appConfig._nuxt.pageTransition,
```
pi0, line 39: For this, we could have a bundler flag (like `NUXT_PAGE_TRANSITION`) that _replaces_ value to `false` on build time otherwise setting app config to `false` won't be able to tree-shake...

danielroe: Yes. This PR does not enable tree-shaking out; it just sets _default_ behaviour. We can't tree shake this way as user may wish to enable transitions/keepalive on a per-page basis.

danielroe: I would love to be able to tree-shake out the Transition/Keepalive components but will it would need a different approach than this PR.

pi0: That's true. We use appConfig currently only to specify _default_ but could be also used to specify feature flags. (swapping condition order when value is false to _always_ disable transitions which I think would make more sense for false handling)

pi0:
Thinking more about approaches, i think for specific nuxt bundled config, using an internal template with named exports (`#build/nuxt.config.mjs` ?) could be safer since we probably don't need nor could support HMR for these configs and it makes it consistent with feature plans being able to treeshake at build time (appConfig is totally runtime-dynamic. we cannot depend on it for this unless passing another build flag)

packages/nuxt/src/pages/runtime/page.ts
danielroe: Maybe support an additional option, e.g. `keepalive: 'disabled'` as the user may simply wish to disable them by default but not for all routes?

danielroe:
Okay - so just to confirm, we'll refactor back to original approach of this PR, but using `nuxt.config.mjs` filename?

pi0 (approved):
LGTM.

PR #5900, opened 2022-07-14
feat(nuxt): allow `getRouteFromPath` to use objects

### LINKED ISSUE #5518: `vue-router`'s functions on single paged apps should still allow the use of objects
If you were to have a single-page app (an app with only an `app.vue` file, no pages directory), vue-router limits itself to what it can do since there is really only one route. However, if you were to use one of its functions like `$router.push` or `$router.replace`, single-page apps are limited to only using strings as the parameter. This does make sense since there is only one route, but objects should still be an option for these functions. Objects are useful for query and hashes for single page apps, which is something I think should be permitted.

The following example is not possible currently for single-page apps, even though, I think it should be:
```js
$router.push({
  query: {
    id: 123,
  },
})
```
It beats it string form in readability:
```js
$router.push("?id=123");
```

TheColaber:
### 🔗 Linked issue

#5518

### ❓ Type of change

### 📚 Description

Allows the `getRouteFromPath` function to accept objects. I convert the object to a URL first by using the `stringifyParsedURL` in `ufo` and the code proceeds as normal. The playground has an example of a single-page app using an object to set the path, query, and hash. The issue is described in #5518.

This is my first time playing around with something like this, so I know there will probably be many things that need to be changed or this being closed entirely.

Resolves #5518

### 📝 Checklist


playground/app.vue
```diff
@@ -8,5 +8,13 @@
   </div>
 </template>
 
-<style scoped>
-</style>
+<script setup>
```
pi0, line 11: Would you please revert changes from PR?

danielroe (approved):
Nice enhancement! ❤️

PR #5916, opened 2022-07-15
docs(runtime-config): clarify how environment variables are loaded

ennioVisco:
This change tries to clarify some approximate usage of the word runtime in explaining how the runtimeConfig API works.

### 🔗 Linked issue

It relates to issue #5903, and, to discussion #5024.

### ❓ Type of change

### 📚 Description

In particular: 
- clarifies what happens at runtime, and what at compile/build time
- clarifies the purpose of `dotenv` (remarking that it also works in SSG)

### 📝 Checklist


docs/content/2.guide/2.features/10.runtime-config.md
```diff
@@ -35,14 +35,16 @@ console.log(runtimeConfig.public.apiBase)
 The most common way to provide configuration is by using [Environment Variables](https://medium.com/chingu/an-introduction-to-environment-variables-and-how-to-use-them-f602f66d15fa).
 
 ::alert{type=info}
-Nuxt CLI has built-in [dotenv](https://github.com/motdotla/dotenv) support.
+Nuxt CLI has built-in [dotenv](https://github.com/motdotla/dotenv) support, allowing the adoption of `.env` files also outside of a node environment (e.g. in SSG).
```
pi0, line 38: This should make it more clear that it is during generation.

```suggestion
Nuxt CLI has built-in [dotenv](https://github.com/motdotla/dotenv) when building, generating, and also in development mode.
```

docs/content/2.guide/2.features/10.runtime-config.md
```diff
@@ -35,14 +35,16 @@ console.log(runtimeConfig.public.apiBase)
...
 
-However, **after your project is built**, you are responsible for setting environment variables when you run the server - your `.env` file will not be read at this point. For example, you could pass the environment variables as arguments using the terminal `DATABASE_HOST=mydatabaseconnectionstring node .output/server/index.mjs`.
+Alternatively, you could pass the environment variables as arguments using the terminal, e.g. `DATABASE_HOST=mydatabaseconnectionstring node .output/server/index.mjs`.
+
+However, at **runtime** (i.e. while the application is running), your initial configuration files will not be read anymore, and only the current value accessible via the `runtimeConfig` API will be accessible.
 ::
 
-Runtime config values are automatically replaced by matching environment variables at runtime. For this to work, you _must_ have a fallback value (which can just be an empty string) defined in your `nuxt.config`.
+Runtime config values are automatically replaced by matching environment variables. For this to work, you _must_ have a fallback value (which can just be an empty string) defined in your `nuxt.config`.
```
pi0, line 47: It is actually at runtme. 

```suggestion
Runtime config values are automatically replaced by matching environment variables at runtime. For this to work, you _must_ have a fallback value (which can just be an empty string) defined in your `nuxt.config`.
```

ennioVisco: > It is actually at runtme.

But in other parts of the docs you are saying that nuxt will restart when these are updates (see https://v3.nuxtjs.org/guide/directory-structure/nuxt.config#runtimeconfig).
This means that the change _is tracked_ at runtime, but the value _**is not**_ available at runtime

pi0: Yes, that how it works for development mode to restart Nuxt and updating `process.env` from `.env`. Still overriding happens in runtime for all modes. Meaning they are replaced from latest and live value of `process.env`. Regardless of `process.env` being populated from `.env` or an explicit env you set.

ennioVisco: > Yes, that how it works for development mode to restart Nuxt and updating `process.env` from `.env`. Still overriding happens in runtime for all modes. Meaning they are replaced from latest and live value of `process.env`. Regardless of `process.env` being populated from `.env` or an explicit env you set.

I see. Then it should be clear that the restart is happening only in develop mode (although it's still not clear to me why a restart is needed)

pi0: Makse sense. Maybe it can be appended to the end of line "In addition to any process environment variable..." that "when updating `.env` in development mode, Nuxt instance is automatically restarted to apply new values to the `process.env`" (or something like that)

ennioVisco: > Makse sense. Maybe it can be appended to the end of line "In addition to any process environment variable..." that "when updating `.env` in development mode, Nuxt instance is automatically restarted to apply new values to the `process.env`" (or something like that)

Sounds good to me, do you want me to write or will you do it?

pi0: Feel free to edit.

danielroe (approved):

docs/content/2.guide/2.features/10.runtime-config.md
```diff
@@ -35,11 +35,15 @@ console.log(runtimeConfig.public.apiBase)
 The most common way to provide configuration is by using [Environment Variables](https://medium.com/chingu/an-introduction-to-environment-variables-and-how-to-use-them-f602f66d15fa).
 
 ::alert{type=info}
-Nuxt CLI has built-in [dotenv](https://github.com/motdotla/dotenv) support.
+Nuxt CLI has built-in [dotenv](https://github.com/motdotla/dotenv) support in development mode and when running `nuxi build` and `nuxi generate`.
 
-In addition to any process environment variables, if you have a `.env` file in your project root directory, it will be automatically loaded into `process.env` and accessible within your `nuxt.config` file and modules.
+In addition to any process environment variables, if you have a `.env` file in your project root directory, it will be automatically loaded **at build/dev/generate time**, and any environment variables set there will be accessible within your `nuxt.config` file and modules.
```
pi0, line 40:
```suggestion
In addition to any process environment variables, if you have a `.env` file in your project root directory, it will be automatically loaded **at build, dev, and generate time**, and any environment variables set there will be accessible within your `nuxt.config` file and modules.
```

pi0 (approved):
Thanks!

PR #5929, opened 2022-07-15
fix(kit): only call `viteExtendConfig` callback once if possible

danielroe:
### ❓ Type of change

### 📚 Description

There's an outstanding issue with kit RC 5 that may break plugins that rely on being shared between client + server vite configs. (see https://github.com/unocss/unocss/pull/1276 for an upstream hotfix). This PR ensures that if client/server flags aren't set, we will call the function only once.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/build.ts
+++ b/packages/kit/src/build.ts
@@ -99,6 +99,11 @@ export function extendViteConfig (
     return
   }
 
+  if (options.server !== false && options.client !== false) {
+    // Call fn() only once
+    return nuxt.hook('vite:extend', ({ config }) => fn(config))
+  }
+
   nuxt.hook('vite:extendConfig', (config, { isClient, isServer }) => {
     if (options.server !== false && isServer) {
       return fn(config)


packages/kit/src/build.ts
```diff
@@ -99,6 +99,11 @@ export function extendViteConfig (
     return
   }
 
+  if (options.server !== false && options.client !== false) {
+    // Call fn() only once
+    return nuxt.hook('vite:extend', ({ config }) => fn(config))
```
pi0, line 104: What about evaulationg fn only once in extendConfig hook? Result would be same but makes order the same

danielroe: Not sure I understand. extendConfig hook is called twice.

pi0: Hmm right. It is just strange because this way we do a lifecycly change when flag is set (without and for both: before resolving server/client and with: after). Thinking of a solution to avoid this inconsistency.

danielroe: Personally, I think it makes sense to call it twice always. Only issue is projects that rely on implicit shared plugin context.

pi0: For calling twice, kit utiliy can be called twices explicitly. These flags was kinda confusing we introduced...

danielroe: Well, hook with same name is also called twice.

pi0: Internally we use hooks yes. Usage doesn't seems so when using one extend.

danielroe: 🤷

pi0: BTW an issue could be nice to track this :)

danielroe: you're asking me to create an issue?

pi0: We already have PR...

danielroe: I'm sorry, what do you mean?

pi0 (approved):
LGTM as workaround. To discuss later about possible behavior changes for client+server mode.

PR #5930, opened 2022-07-15
fix(nuxt): lazy load and tree-shake error templates

danielroe:
### ❓ Type of change

### 📚 Description

Currently `error-dev` is being bundled and not tree-shaken. This change uses a dynamic import for dev errors which allows vite to tree-shake it out of the final build.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-error-page.vue
+++ b/packages/nuxt/src/app/components/nuxt-error-page.vue
@@ -37,10 +37,10 @@ const description = error.message || error.toString()
 const stack = process.dev && !is404 ? error.description || `<pre>${stacktrace}</pre>` : undefined
 
 // TODO: Investigate side-effect issue with imports
-const get404Template = () => import('@nuxt/ui-templates/templates/error-404.vue')
-const getErrorTemplate = process.dev
-  ? () => import('@nuxt/ui-templates/templates/error-dev.vue').then(r => r.default || r)
-  : () => import('@nuxt/ui-templates/templates/error-500.vue').then(r => r.default || r)
+const _Error404 = defineAsyncComponent(() => import('@nuxt/ui-templates/templates/error-404.vue'))
+const _Error = process.dev
+  ? defineAsyncComponent(import('@nuxt/ui-templates/templates/error-dev.vue'))
+  : defineAsyncComponent(import('@nuxt/ui-templates/templates/error-500.vue'))
 
-const ErrorTemplate = defineAsyncComponent(is404 ? get404Template : getErrorTemplate)
+const ErrorTemplate = is404 ? _Error404 : _Error
 </script>
--- a/packages/nuxt/src/app/components/nuxt-error-page.vue
+++ b/packages/nuxt/src/app/components/nuxt-error-page.vue
@@ -39,8 +39,8 @@ const stack = process.dev && !is404 ? error.description || `<pre>${stacktrace}</
 // TODO: Investigate side-effect issue with imports
 const _Error404 = defineAsyncComponent(() => import('@nuxt/ui-templates/templates/error-404.vue'))
 const _Error = process.dev
-  ? defineAsyncComponent(import('@nuxt/ui-templates/templates/error-dev.vue'))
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/components/nuxt-error-page.vue
```diff
@@ -38,5 +37,5 @@ const statusMessage = error.statusMessage ?? (is404 ? 'Page Not Found' : 'Intern
 const description = error.message || error.toString()
 const stack = process.dev && !is404 ? error.description || `<pre>${stacktrace}</pre>` : undefined
 
-const ErrorTemplate = is404 ? Error404 : process.dev ? ErrorDev : Error500
+const ErrorTemplate = is404 ? Error404 : process.dev ? await import('@nuxt/ui-templates/templates/error-dev.vue').then(r => r.default || r) : Error500
```
pi0, line 40: It is strange that unused import is not tree-shaken. I suspect a side effect is being detected that avoids this. (Generally, async loading error componens is good idea)

danielroe: Perhaps there is a side-effect. Normally vite vue plugin marks defineComponent as PURE though, I think 🤔

I can also lazy load error component in nuxt-root in same PR if you think it worth it.

kevinmarrec: Also, we may consider having the 3 Error components sharing same css to prevent 3 similar scoped styles in bundle.

pi0: Lazy loading error compoent alone would be nice solution. 

This workaround would hid any other possible root causes...

@kevinmarrec Nice idea. Currently build process in nuxt/ui generates them separately but we can improve by refactoring shared styles css chunk.

danielroe: This workaround is still necessary for now. Lazy loading error component won't remove the duplicate styles from css.

pi0: An issue would be nice BTW to track root causes :)

pi0, line 40:
```suggestion
// TODO: Investigate tree-shaking issues
const ErrorTemplate = is404 ? Error404 : process.dev ? await import('@nuxt/ui-templates/templates/error-dev.vue').then(r => r.default || r) : Error500
```

packages/nuxt/src/app/components/nuxt-error-page.vue
```diff
@@ -38,5 +36,11 @@ const statusMessage = error.statusMessage ?? (is404 ? 'Page Not Found' : 'Intern
 const description = error.message || error.toString()
 const stack = process.dev && !is404 ? error.description || `<pre>${stacktrace}</pre>` : undefined
 
-const ErrorTemplate = is404 ? Error404 : process.dev ? ErrorDev : Error500
+// TODO: Investigate side-effect issue with imports
+const _Error404 = defineAsyncComponent(() => import('@nuxt/ui-templates/templates/error-404.vue'))
+const _Error = process.dev
+  ? defineAsyncComponent(import('@nuxt/ui-templates/templates/error-dev.vue'))
+  : defineAsyncComponent(import('@nuxt/ui-templates/templates/error-500.vue'))
```
kevinmarrec, line 43:
```suggestion
  ? defineAsyncComponent(() => import('@nuxt/ui-templates/templates/error-dev.vue'))
  : defineAsyncComponent(() => import('@nuxt/ui-templates/templates/error-500.vue'))
```

No ?

kevinmarrec (changes requested):

packages/nuxt/src/app/components/nuxt-error-page.vue
pi0: Nice find!

danielroe: Yes 👍

pi0 (approved):

kevinmarrec (approved):

PR #5938, opened 2022-07-16
fix(vite): use shared watcher instance

### LINKED ISSUE #2047: HMR does not pick up updates to data
### Environment

Running on WSL but have had the same issue in a containerized application..

Nuxt CLI v3.0.0-27288761.d3fb512                                                                                                                                                                           
RootDir: ...                                                                                                                                                   
Nuxt project info:                                                                                                                                                                                         
- Operating System: `Linux`
- Nuxt Version:     `3.0.0-27288761.d3fb512`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/radusuciu/nuxt3-app

repro

### Describe the bug

Updated data is not picked up by HMR. Updating the template is. Data updates are reflected after a full refresh only.

Potentially related to #1036 but I don't see any errors and it isn't sporadic - it just doesn't work even though Vite does seem to pick up the updates judging by the "Vite server built in XXms" logs messages and "[vite] hot updated: /app.vue" output in the console.

### Additional context

I have no issues with HMR in nuxt2, nor if I use vite by itself.

### Logs

_No response_

pi0:
### 🔗 Linked issue

Resolves #2047

### ❓ Type of change

### 📚 Description

In platforms other than macOS (Windows, Linux, and WSL2) when two vite server watchers are running (client and bundle), only the last one (server) receives watcher events causing HMR to be broken. This is likely an issue with chokidar (without `fsevents` that is used for macOS) within vite. (https://github.com/nuxt/framework/issues/2047#issuecomment-1186034999)

This PR tried to work around this by using a shared watcher and close another one. Since both vite servers (client and server) use exactly the same chokidar params, it should be safe also more performant.

[PR is playground for now not working]

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite.ts
+++ b/packages/vite/src/vite.ts
@@ -129,6 +129,20 @@ export async function bundle (nuxt: Nuxt) {
       .then(() => logger.info(`Vite ${env.isClient ? 'client' : 'server'} warmed up in ${Date.now() - start}ms`))
       .catch(logger.error)
   })
+
+  // Use single watcher instance shared between client and server
+  // https://github.com/nuxt/framework/issues/2047
+  // TODO: Not needed for MacOS
+  let viteWatcher
+  nuxt.hook('vite:serverCreated', async (server: vite.ViteDevServer) => {
+    if (viteWatcher) {
+      await server.watcher.close()
+      server.watcher = viteWatcher
+    } else {
+      viteWatcher = server.watcher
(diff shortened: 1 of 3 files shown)


pi0:
Seems wrong solution. Issue is with vite itself.

PR #5998, opened 2022-07-19
feat(nuxi): support `--ui-templates` arg

kvcvc:
### 🔗 Linked issue

#5997

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/3.api/5.commands/dev.md
+++ b/docs/content/3.api/5.commands/dev.md
@@ -1,7 +1,7 @@
 # `nuxi dev`
 
 ```{bash}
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for this pull request @kvcvc 👍🏼 

I can understand you would like to customize the loading page and it makes sense as an opt-in feature since we also had it for Nuxt 2 in nuxt/screens.

However, since we are using `@nuxt/ui-templates` in several places, we need a better configuration interface instead of new CLI flag. PS: Exports of `@nuxt/ui-templates` can change over the time.

Let's keep tracking this feature request in #5997. I will try to offer a better solution to implement.

PR #6010, opened 2022-07-20
fix(nuxt)!: change `statusCode`  type to number for error page template

danielroe:
### ❓ Type of change

### 📚 Description

We are currently providing errorCode as string, which is counter intuitive. I think it should probably be a number instead. However, this is a breaking change if people are checking it directly.

### 📝 Checklist


pi0:
Maybe merge this all together with #4539?

danielroe:
Do you want to do that rather than merge this immediately?

PR #6029, opened 2022-07-21
fix(kit, nuxt): always sort globby results

antfu:
### ❓ Type of change

### 📚 Description

The order of the result from `globby` is not reliable. This will potentially cause the order of installed plugins/modules to be unstable across multiple builts or on different machines. We should always sort them to make it stable.

### 📝 Checklist


pi0 (approved):
Nice!

PR #6030, opened 2022-07-21
fix(nuxt): use relative path to generate plugin variables

antfu:
### ❓ Type of change

### 📚 Description

To make the build hash consistent and avoid leaking the full path of user workspace.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/internal/template.ts
+++ b/packages/kit/src/internal/template.ts
@@ -33,7 +33,7 @@ const importSources = (sources: string | string[], root: string, { lazy = false 
   const imports: string[] = []
   for (const src of sources) {
     const path = relative(root, src)
-    const variable = genSafeVariableName(path) + hash(path)
+    const variable = genSafeVariableName(path).replace(/_(45|46|47)/g, '_') + '_' + hash(path)
     variables.push(variable)
     imports.push(lazy
       ? `const ${variable} = ${genDynamicImport(src, { comment: `webpackChunkName: ${JSON.stringify(src)}` })}`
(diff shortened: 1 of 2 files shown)


packages/kit/src/internal/template.ts
```diff
@@ -23,16 +24,25 @@ export async function compileTemplate (template: NuxtTemplate, ctx: any) {
...
-    if (lazy) {
-      return `const ${genSafeVariableName(src)} = ${genDynamicImport(src, { comment: `webpackChunkName: ${JSON.stringify(src)}` })}`
-    }
-    return genImport(src, genSafeVariableName(src))
-  }).join('\n')
+  const variables: string[] = []
+  const imports: string[] = []
+  sources.forEach((src) => {
+    const path = relative(root, src)
+    const variable = genSafeVariableName(path)
```
pi0, line 35: Since we are stipping the full path, now there is a chance two generated names overlap. We can add `hash(path)` to the end to make it safe. `../..` are same in different envs and not included in variable name but affect the hash.

packages/kit/src/internal/template.ts
```diff
@@ -23,16 +24,25 @@ export async function compileTemplate (template: NuxtTemplate, ctx: any) {
...
   }
-  return sources.map((src) => {
-    if (lazy) {
-      return `const ${genSafeVariableName(src)} = ${genDynamicImport(src, { comment: `webpackChunkName: ${JSON.stringify(src)}` })}`
-    }
-    return genImport(src, genSafeVariableName(src))
-  }).join('\n')
+  const variables: string[] = []
+  const imports: string[] = []
+  sources.forEach((src) => {
```
pi0, line 33: Code style: Can we please use `for of`?

pi0 (approved):

PR #6035, opened 2022-07-21
fix(nuxt): always write nitro types when building

### LINKED ISSUE #6024: auto import is not working on server directory codes when build
### Environment

- Package Manager:  `npm@8.14.0`
- Builder:          `vite`
- User Config:      `typescript`, `build`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

add this to nuxt.config.ts

`typescript: {
    strict: true,
    typeCheck: true,
  },`

npm run build is fail with below message

`ERROR(TypeScript)  Cannot find name 'defineEventHandler'.`

### Describe the bug

ERROR(TypeScript)  Cannot find name 'defineEventHandler'.

1 | export default defineEventHandler(async (event) => {
   |                ^^^^^^^^^^^^^^^^^^
2 |     return 'test !'
3 | })
4 |

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #6024

### ❓ Type of change

### 📚 Description

If users are checking types on build, we should write `nitro.d.ts` then as well.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -141,7 +141,7 @@ export async function initNitro (nuxt: Nuxt) {
 
   // Add typed route responses
   nuxt.hook('prepare:types', async (opts) => {
-    if (nuxt.options._prepare) {
+    if (nuxt.options._prepare || !nuxt.options.dev) {
       await scanHandlers(nitro)
       await writeTypes(nitro)
     }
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/nitro.ts
```diff
@@ -141,7 +141,7 @@ export async function initNitro (nuxt: Nuxt) {
 
   // Add typed route responses
   nuxt.hook('prepare:types', async (opts) => {
-    if (nuxt.options._prepare) {
+    if (nuxt.options._prepare || !nuxt.options.dev) {
```
pi0, line 144: Do we have `dev: true` + `_prepare: true` situation? If not maybe we can simplify condition?

danielroe: We also want to write types in development.

pi0 (approved):

Issue #6058, opened 2022-07-22
Duplicated path prefix with static rendering and RC6


apss-pohl:
### Environment

-----------------------------
- Operating System: `Linux`
- Node Version:     `v16.15.0`
- Nuxt Version:     `3.0.0-rc.6-27641313.416f98b`
- Package Manager:  `yarn@3.2.1`
- Builder:          `vite`
- User Config:      `srcDir`, `ssr`, `app`, `publicRuntimeConfig`, `nitro`, `alias`, `typescript`, `hooks`, `build`, `css`, `buildModules`, `telemetry`, `intlify`, `dayjs`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `@pinia/nuxt@0.3.0`, `@intlify/nuxt3@0.2.3`, `@vueuse/nuxt@8.9.4`, `@nuxtjs/dayjs@1.4.1`
------------------------------

### Reproduction

Run a static build with : nuxi generate

### Describe the bug

Duplicated prefix in the .output/public/index.html
```ts
<head >
  <link rel="stylesheet" href="./_nuxt/_nuxt/entry.d20389ae.css">
</head>
```

danielroe:
Thanks @apss-pohl! Do you have a custom `app.buildDir`?

apss-pohl:
https://stackblitz.com/edit/github-c88iyh?file=package-lock.json,package.json&theme=light

run yarn dev
then you can cat dist/index.html and find something like:
<link rel="modulepreload" href="./_nuxt/_nuxt/entry.9fef1731.mjs" as="script" crossorigin>

Root cause is setting "NUXT_APP_BASE_URL=./"

nikolas-psaroudakis-rokoko:
This happens with the webpack builder as well

danielroe:
@nikolas-psaroudakis-rokoko Would you open a new issue with a reproduction?

nikolas-psaroudakis-rokoko:
I'll open a new issue in the evening, until then I would like to describe that with the webpack builder, it seems to be worse than with vite.

The generated css, js files are placed in `.output/public/_nuxt/_nuxt` but the generated `index.html` file points to `_nuxt` instead of `_nuxt/_nuxt` ( which seems to be the case with vite), so the site turns out to be broken

PR #6063, opened 2022-07-22
fix(vite): ignore baseurl when pruning manifest

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/6058

### ❓ Type of change

### 📚 Description

We shouldn't prefix RE with baseURL - only the buildAssetsDir should be removed from these filenames.

### 📝 Checklist


Atinux:
Cannot we add a test case for the fix issue to avoid any future regression?

packages/vite/src/server.ts
```diff
@@ -18,7 +18,7 @@ export async function buildServer (ctx: ViteBuildContext) {
   const _resolve = id => resolveModule(id, { paths: ctx.nuxt.options.modulesDir })
   const serverConfig: vite.InlineConfig = vite.mergeConfig(ctx.config, {
     base: ctx.nuxt.options.dev
-      ? joinURL(ctx.nuxt.options.app.baseURL, ctx.nuxt.options.app.buildAssetsDir)
+      ? joinURL(ctx.nuxt.options.app.baseURL.replace(/^\./, '') || '/', ctx.nuxt.options.app.buildAssetsDir)
```
pi0, line 21: We need to handle base url with dot like `.foo/` too.

danielroe: No, not here. Such a baseURL would in fact be `/.foo` and not caught by this rule. A URL starting with `.` is fully disallowed by vite, and would lead to the following error message:

```
The value can only be an absolute URL, ./, or an empty string.
```

pi0: It might be a bug in vite as well. /.app/ is valid. Still can we please make regex more precise to ./?

pi0 (approved):

Issue #6064, opened 2022-07-22
Checker is not works when app is in sub directory


mhfeizi:
Checker is not works when app is in sub directory:

https://github.com/nuxt/framework/blob/eff6b0c4caf7a4daff7bf0f17599caf8246b1a91/packages/vite/src/server.ts#L93

danielroe:
Would you provide a reproduction? 🙏

danielroe:
Here are two working examples:

* https://stackblitz.com/edit/github-c1nhqh
* https://stackblitz.com/edit/github-c1nhqh-imr9pw

pi0:
(for the context this issue was happening only for production build when `vue-tsc` CLI is used)

PR #6069, opened 2022-07-22
fix(vite): resolve `tsconfig.json` for vue-tsc type checker

mhfeizi:
### 🔗 Linked issue
https://github.com/nuxt/framework/issues/6064

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/vite/src/server.ts
+++ b/packages/vite/src/server.ts
@@ -108,8 +108,11 @@ export async function buildServer (ctx: ViteBuildContext) {
   // Add type-checking
   if (ctx.nuxt.options.typescript.typeCheck === true || (ctx.nuxt.options.typescript.typeCheck === 'build' && !ctx.nuxt.options.dev)) {
     const checker = await import('vite-plugin-checker').then(r => r.default)
-    serverConfig.plugins.push(checker({ vueTsc: { tsconfigPath: await resolveTSConfig(ctx.nuxt.options.rootDir)
- } }))
+    serverConfig.plugins.push(checker({
+      vueTsc: {
+        tsconfigPath: await resolveTSConfig(ctx.nuxt.options.rootDir)
+      }
+    }))
   }
 
   await ctx.nuxt.callHook('vite:extendConfig', serverConfig, { isClient: false, isServer: true })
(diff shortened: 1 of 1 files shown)


packages/vite/src/server.ts
```diff
@@ -107,7 +107,7 @@ export async function buildServer (ctx: ViteBuildContext) {
   // Add type-checking
   if (ctx.nuxt.options.typescript.typeCheck === true || (ctx.nuxt.options.typescript.typeCheck === 'build' && !ctx.nuxt.options.dev)) {
     const checker = await import('vite-plugin-checker').then(r => r.default)
-    serverConfig.plugins.push(checker({ vueTsc: true }))
+    serverConfig.plugins.push(checker({ vueTsc: { root: ctx.nuxt.options.rootDir, tsconfigPath: "tsconfig.json" } }))
```
danielroe, line 110: Nice find. Would there be any merit in pointing to the generated tsconfig? We're assuming that the user has a `tsconfig` in their root directory. In a monorepo we might have a tsconfig higher up, possibly?

danielroe, line 110:
```suggestion
    serverConfig.plugins.push(checker({ vueTsc: { root: ctx.nuxt.options.rootDir, tsconfigPath: 'tsconfig.json' } }))
```

mhfeizi: without `tsconfigPath` not works :

https://github.com/fi3ework/vite-plugin-checker/blob/f0f8413e6bb28e8c59456e0ad752070d2899598e/packages/vite-plugin-checker/src/checkers/vueTsc/main.ts#L143

pi0: Indeed we should assume there is a suer specified tsconfig. Pointing to generated can be unwanted if user extends but adds custom tsconfig as well the this.

danielroe: My point is that it might not be in that location.

Maybe we can use `pkg-types` utility `findNearestTSConfig ` to resolve the closest tsconfig ourselves and then split it into rootDir + `tsconfig.json`?

pi0: Makes sense  But If lookup is typescript behavior, shall we improve this improve to vue-tsc? Considering root is now provided.

danielroe:
@mhfeizi So I can test this PR, would you be able to give a reproduction of the original issue?

pi0 (approved):

PR #6070, opened 2022-07-22
feat(nuxt): add support for `components/global`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/5164

### ❓ Type of change

### 📚 Description

This PR:
1. Cleans up component templates and creates two different templates, for client/server modes, which should fix any issues with client/server components being imported in the wrong context. This will be needed for server components. So importing from `#components` on client-side will no longer pull in server-only components or vice versa.
2. Supports registering global components with `~/components/global` or `MyComponent.global.ts`. It does not change how they are made available; it's still via async registration.
3. Adds a couple of test cases for the above.

I'm happy to split out 1 & 2 if that would be helpful in reviewing it. Let me know if so.

### 📝 Checklist


danielroe:
The order is important.

packages/nuxt/src/components/module.ts
```diff
@@ -30,7 +32,10 @@ export default defineNuxtModule<ComponentsOptions>({
         return dir.map(dir => normalizeDirs(dir, cwd)).flat().sort(compareDirByPathLength)
       }
       if (dir === true || dir === undefined) {
-        return [{ path: resolve(cwd, 'components') }]
+        return [
+          { path: resolve(cwd, 'components') },
+          { path: resolve(cwd, 'components/global'), global: true }
```
danielroe, line 37: The order is important, because we skip directories that are included _within_ previous directories.

danielroe:
```suggestion
          { path: resolve(cwd, 'components/global'), global: true },
          { path: resolve(cwd, 'components') }
```

pi0 (approved):

PR #6082, opened 2022-07-23
fix(vite): allow overriding client hmr options

### LINKED ISSUE #6078: Vite HMR user settings not being respected
Hey guys! I'm excited to get using Vite 3 but I've noticed that [this PR](https://github.com/nuxt/framework/pull/5398) actually seems to overwrite user Vite settings (as set in nuxt.config).

I think perhaps the configs are provided [here](https://github.com/nuxt/framework/blob/b8329535e7112323edee1d4b6012ad3bf02b13f7/packages/vite/src/client.ts#L24) in the reverse order...

Hopefully it's something just this simple 😅

_Originally posted by @tylerforesthauser in https://github.com/nuxt/framework/issues/5398#issuecomment-1192973131_

danielroe:
### 🔗 Linked issue

resolves #6078

### ❓ Type of change

### 📚 Description

This fixes a regression in upgrade to Vite 3. Our client config should only contain what we want to override the basic user config, and HMR is probably in the list of things that should be configurable.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/client.ts
+++ b/packages/vite/src/client.ts
@@ -4,7 +4,6 @@ import vuePlugin from '@vitejs/plugin-vue'
 import viteJsxPlugin from '@vitejs/plugin-vue-jsx'
 import type { Connect } from 'vite'
 import { logger } from '@nuxt/kit'
-import { getPort } from 'get-port-please'
 import { joinURL, withLeadingSlash, withoutLeadingSlash, withTrailingSlash } from 'ufo'
 import escapeRE from 'escape-string-regexp'
 import { cacheDirPlugin } from './plugins/cache-dir'
@@ -16,11 +15,6 @@ import { devStyleSSRPlugin } from './plugins/dev-ssr-css'
 import { viteNodePlugin } from './vite-node'
 
 export async function buildClient (ctx: ViteBuildContext) {
-  const hmrPortDefault = 24678 // Vite's default HMR port
-  const hmrPort = await getPort({
-    port: hmrPortDefault,
-    ports: Array.from({ length: 20 }, (_, i) => hmrPortDefault + 1 + i)
-  })
   const clientConfig: vite.InlineConfig = vite.mergeConfig(ctx.config, {
     experimental: {
       renderBuiltUrl: (filename, { type, hostType }) => {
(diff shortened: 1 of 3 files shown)


packages/vite/src/server.ts
```diff
@@ -94,6 +94,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       }
     },
     server: {
+      hmr: false,
```
pi0, line 97: Isn't this change enough to disable HMR for server? Your refactor to have hmrPort fix in the client was making sense...

danielroe: The issue wasn't that it was enabled for server. The issue was that moving the port -> client file was overriding any user config for this setting.

pi0: Can we add _defaults_ as first param to `mergeConfig` in client and explicitly disable (override) `server.hmr` for server?

danielroe: If we swap the order we will need to move a fair chunk of the config out and manually override (e.g. `process.client`) after the merge.

pi0: We can keep scope of this PR to the regression fix for hmr port as is :)

The general ability to override vite config is something we can improve later and I would like to also discuss more because some overrides are probably unsafe and for the purpose of workarounds, there are hooks.

danielroe: So I don't think I understand you then. `vite.mergeConfig` only takes two options - defaults + overrides. In this case, the defaults is the user config + global defaults. My point is that if we swap them, as you are suggesting(?), we may have to manually override the resulting config. I am sounding a note of caution on that approach.

pi0: Maybe for a client, we can add port assignment to (former) L80 after merging using defu for explicit control?

danielroe: Okay - what do you think of this?

packages/vite/src/client.ts
```diff
@@ -82,6 +72,20 @@ export async function buildClient (ctx: ViteBuildContext) {
     clientConfig.server.hmr = false
   }
 
+  if (clientConfig.server.hmr !== false) {
+    const hmrPortDefault = 24678 // Vite's default HMR port
+    const hmrPort = await getPort({
+      port: hmrPortDefault,
+      ports: Array.from({ length: 20 }, (_, i) => hmrPortDefault + 1 + i)
+    })
+    const userHMRConfig = clientConfig.server.hmr === true ? {} : clientConfig.server.hmr
```
pi0, line 81: Tip: Defu/Object assignment handled non objects (`a=true; b={...a} // = {}`)

danielroe: Yes, I'm aware. It's like this for TS.

pi0 (approved):

PR #6119, opened 2022-07-25
fix(nuxi): load `.env` file before starting dev server

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/6118

### ❓ Type of change

### 📚 Description

We load `.env` for `nuxi preview` and even later in the process, so it seems reasonable to read it in before starting the dev server, allowing host/port to be defined there when developing.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -1,9 +1,11 @@
+import { existsSync } from 'node:fs'
 import { resolve, relative, normalize } from 'pathe'
 import chokidar from 'chokidar'
 import { debounce } from 'perfect-debounce'
 import type { Nuxt } from '@nuxt/schema'
 import consola from 'consola'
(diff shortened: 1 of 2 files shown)


packages/nuxi/src/commands/dev.ts
```diff
@@ -33,6 +35,12 @@ export default defineNuxtCommand({
       return currentHandler ? currentHandler(req, res) : loadingHandler(req, res)
     }
 
+    const rootDir = resolve(args._[0] || '.')
+
+    if (existsSync(resolve(rootDir, '.env'))) {
+      process.env = await loadDotenv({ cwd: rootDir, fileName: '.env', env: process.env })
```
pi0, line 41: loadDotenv should check for existance. Also we can use `setupDotenv` that does the polyfill.

pi0 (approved):

Issue #6140, opened 2022-07-26
Get warning when v-on:error event handler is set to NuxtErrorBoundary 


hecateball:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.11.0`
- Nuxt Version:     `3.0.0-rc.6`
- Package Manager:  `npm@8.6.0`
- Builder:          `vite`
- User Config:      `meta`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

Create a new Nuxt project and replace `app.vue` with following code:

```
<template>
  <NuxtErrorBoundary @error="() => console.error('ERROR!')">
    <h1>Error Sample</h1>
  </NuxtErrorBoundary>
</template>
```
 
Run the application and check console log.

### Describe the bug

Get following Vue warning (because `emits` declaration is missing at NuxtErrorBoundary component):

```
[Vue warn]: Extraneous non-emits event listeners (error) were passed to component but could not be automatically inherited because component renders fragment or text root nodes. If the listener is intended to be a component custom event listener only, declare it using the "emits" option.
```

hecateball:
I've created PR to fix this issue. Please have a check!

PR #6141, opened 2022-07-26
fix(nuxt): add emits.error declaration to NuxtErrorBoundary component

hecateball:
### 🔗 Linked issue

#6140 Get warning when v-on:error event handler is set to NuxtErrorBoundary

### ❓ Type of change

### 📚 Description

add `emits.error` declaration to NuxtErrorBoundary to prevent prevent getting warning message.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-error-boundary.ts
+++ b/packages/nuxt/src/app/components/nuxt-error-boundary.ts
@@ -2,6 +2,11 @@ import { defineComponent, ref, onErrorCaptured } from 'vue'
 import { useNuxtApp } from '#app'
 
 export default defineComponent({
+  emits: {
+    error (_error: unknown) {
+      return true
+    }
+  },
   setup (_props, { slots, emit }) {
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/components/nuxt-error-boundary.ts
```diff
@@ -2,6 +2,11 @@ import { defineComponent, ref, onErrorCaptured } from 'vue'
 import { useNuxtApp } from '#app'
 
 export default defineComponent({
+  emits: {
+    error (_error: unknown) {
```
hecateball, line 6: payload type is `unknown` here just because the error given to `onErrorCaptured` callback is actually unknown.
https://vuejs.org/api/composition-api-lifecycle.html#onerrorcaptured

*any value pass the runtime validation.

danielroe (approved):
Nice catch and solution 👌

pi0 (approved):

PR #6154, opened 2022-07-26
fix(nuxt): clear `vite-node` runner cache after each render

antfu:
### 🔗 Linked issue

#5018

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Otherwise, the cache will be preseved across builds

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #6156, opened 2022-07-26
perf(vite-node): only invalidate cache for user code

antfu:
### 🔗 Linked issue

Continues from #6154

### ❓ Type of change

### 📝 Checklist


packages/vite/src/runtime/vite-node.mjs
```diff
@@ -22,6 +22,11 @@ export default async (ssrContext) => {
   process.server = true
   render = render || (await runner.executeFile(viteNodeOptions.entryPath)).default
   const result = await render(ssrContext)
-  runner.moduleCache.clear()
+  // reset cache for non-node-modules
+  for (const key of runner.moduleCache.keys()) {
+    if (!key.includes('/node_modules/')) {
```
pi0, line 27: Note: Does this mean we need a full restart when changing in node_modules? (client build should be updated I guess with hmr and changing in node_modules...)

antfu: This is how Vite makes the assumption. In the client, it pre-bundles the lib under node_modules and requires a full start with force cache clear to make the changes take effect.

PR #6165, opened 2022-07-26
feat(nuxt): remove wrapper from client only components

### LINKED ISSUE #5867: defineExpose doesn't work with client-only SFCs
### Environment

Nuxt CLI v3.0.0-rc.4
RootDir: /Volumes/SuperData/Sites/reelcrafter/v2-presentation-nuxt3
Nuxt project info: 

- Operating System: `Darwin`
- Package Manager:  `yarn@3.2.1`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-ud1mt5?file=app.vue,components%2FTest.client.vue

### Describe the bug

I'm using `defineExpose` to expose functions to the parent component. This works fine, except with `.client.vue` SFCs.

Template refs to client-only components also look much different in the console:

CleanShot 2022-07-12 at 19 05 45

in comparison to client + server components:

CleanShot 2022-07-12 at 19 07 05

huang-julien:
### 🔗 Linked issue
resolve #5867

### ❓ Type of change

### 📚 Description

Before, the createClientOnly returned a new component wrapping the component with ".client" in its name. So when we were setting a ref on `<HelloWorld ref="helloworld" />` the ref assigned was on the `<ClientOnly />` component but not on  `<HelloWorld />`. 

````md
- App
  - ClientOnly (has the ref)
    - HelloWorld.client
````

This PR make `createClientOnly()` modifying the component instead of wrapping it. The setup hook return an additionnal `mounted ` `ref()` . It also override the `render()` function to conditionally render the component if `mounted` value is set to true

````md
- App
  - HelloWorld.client (has the ref)
````

### 📝 Checklist

### CODE REVIEW

**Comment by huang-julien** [comment]
@danielroe it should be good now, i've tested on a component with a `<script setup>` and on a `<script>`.

```vue
<template>
  <div>
    hellow world
    <button @click="add">
      {{ count }}
    </button>
  </div>
</template>

<script setup lang="ts">
const count = ref(0)

const add = () => count.value++

defineExpose({ add, count })
</script>
````
````vue
<template>
  <div>
    hellow world
    <button @click="add">
      {{ count }}
    </button>
  </div>
</template>

<script  lang="ts">

export default {
  setup () {
    const count = ref(0)

    const add = () => count.value++
    return {
      count, add
    }
  }
}
</script>
````

File: packages/nuxt/src/app/components/client-only.mjs [hunk 1]
```diff
@@ -19,18 +19,22 @@ export default defineComponent({
...
-    inheritAttrs: false,
-    setup (_props, { attrs, slots }) {
+  const { setup } = component
+  return {
+    ...component,
+    setup (props, ctx) {
+      const res = setup?.(props, ctx) || {}
+
       const mounted = ref(false)
       onMounted(() => { mounted.value = true })
```
danielroe, line 29: Might be worth renaming this to something like `__mounted` to decrease likelihood of collision with user return from setup.

danielroe (approved):
This is a nice fix ❤️ 

I've made a couple of minor changes - feel free to check and let me know if you agree.

huang-julien:
@danielroe only the setup state was missing, which was the "res". That mean we don't even have to use a collision-resistant name

packages/nuxt/src/app/components/client-only.mjs
```diff
@@ -19,18 +19,19 @@ export default defineComponent({
...
-      return () => {
-        if (mounted.value) {
-          return h(component, attrs, slots)
-        }
-        return h('div', { class: attrs.class, style: attrs.style })
-      }
+
+      return Promise.resolve(setup?.(props, ctx) || {})
+        .then((setupState) => {
+          const render = setupState && typeof setupState === 'function' ? res : component.render
```
danielroe, line 31:
```suggestion
          const render = setupState && typeof setupState === 'function' ? setupState : component.render
```

danielroe:
Would you take a look at stateful use, and test on both build + dev mode? It's not quite there yet, and I wonder if we may have to migrate back to using a render function in some cases.

Modify the example component `ClientWrapped.client.vue` to:

```vue
<script setup lang="ts">
function exposedFunc () {
  console.log('ok')
}

const state = ref('component')
defineExpose({ exposedFunc })

await new Promise(resolve => setTimeout(resolve, 300))

onMounted(() => { console.log('mounted') })
</script>

<template>
  <div>
    client-only {{ state }}
  </div>
</template>
```

huang-julien:
> Would you take a look at stateful use, and test on both build + dev mode? It's not quite there yet, and I wonder if we may have to migrate back to using a render function in some cases.
> 
> Modify the example component `ClientWrapped.client.vue` to:
> 
> ```
> <script setup lang="ts">
> function exposedFunc () {
>   console.log('ok')
> }
> 
> const state = ref('component')
> defineExpose({ exposedFunc })
> 
> await new Promise(resolve => setTimeout(resolve, 300))
> 
> onMounted(() => { console.log('mounted') })
> </script>
> 
> <template>
>   <div>
>     
>     client-only {{ state }}
>   </div>
> </template>
> ```
> 
> See [SFC playground](https://sfc.vuejs.org/#eNpVUj1v3DAM/SusFvuAsx2g2+ESoEO7FejQUYtj0z6lFilI9F0Dw/89lH0Jko1fenx8T4v5EUJ9ndGczDl10QWBhDIHmFoaH62RZM2TJecDR4EFIg5HYPrNMwn2R+hxcIQ//wdOCCsMkT0UildYsgQwzNSJYwLcJvpfmkN5gCU3O6bEE9YTj2XB/4qDpTU/y3WlIa0gPOaNZdGxEiAkyUOfd5bLF+hV25baW+sECG/wR/k4nYqom64K95TP++s88izv1SN8f3g4bC8/LiuVpA4vX0n6vVkctkXnZldM9dFE0IdJKWsGcO7ddQs0/FZVIBeXoGsJnlERvddLsAdHjdIAYZDoxhEj9G4YMGo3TwU3IbQhRG67CyaoqjtkNzkdqZimV1iWu1Kraqfbmn3zufngY45mt6/ybahfEpOavRlg7w31+LRbkmvqXs6tuYiEdGqaNHT5i7ykmuPYaFRHlUE1rDH56jnyLWFUYGsyhNJYzfoGELbRaQ==).

i could fix the dev-build issue, but i need more testing to verify as many cases a possible, will probably commit tonight

huang-julien:
i could fix the build-dev issue, i set this PR to draft  to do more testing. There might be some refactor that can be done on the fix

huang-julien:
@danielroe The fix is working good, however since we return a render function in the setup , the devtool cannot access to the data returned by the setup. The workaround is to define the render function on the component in dev and to return a render function in the setup only in production. The issue is already known at  https://github.com/vuejs/devtools/issues/1295 

everything wrapped by an process.dev if statement will be stripped from the build right ?

danielroe:
This seems to be working very nicely, though I think we've lost the attrs at build-time.

huang-julien:
yes, it needs a fix. I've finnally understand when does the component have the render function. Only <script setup> SFC are compiled into setup only component. So we need to keep the render function if the the original component have one

huang-julien:
checks
- [x] modify the component to conditionally render if mounted
- [x] keep the context/props intact
- [x] keep original component structure (name, hrmId, render?, setup)
- [x] keep component ref accessible   

done on

- [x] SFC script setup
- [x] SFC script
- [] .ts/js component (with runtime compiler)

danielroe:
Currently if users pass `class`/`style` attributes the following error is shown:
```
Extraneous non-props attributes (style, class) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.
```

huang-julien:
> Currently if users pass `class`/`style` attributes the following error is shown:
> 
> ```
> Extraneous non-props attributes (style, class) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.
> ```
> 
> Would you explain more about the oldChildren null issue you unearthed?



This was happening only in build, not sure why but at the setup render function, in some components, if we don't wrap the component render [here](https://github.com/nuxt/framework/pull/6165/files#diff-ce55bf5a46f331c65de3586dec9caf71aba63f13092305fb7e22f9bc9592fe55R51) with a Fragment, the component is not being re-rendered when mounted. And in some cases there's a oldChildren is null error triggered by vue. I'll try to reproduce it.

danielroe:
I'm sure you're looking into it, but just to clarify - we're currently getting attrs applied before mount, but not afterwards.

huang-julien:
[here(https://github.com/huang-julien/nuxt-3/tree/example/oldChildren-is-null-client-only-wrapper) is the reproduction

huang-julien:
> I'm sure you're looking into it, but just to clarify - we're currently getting attrs applied before mount, but not afterwards.

yes, just fixed. Sorry for taking so much of your time for a simple PR :pray: :cry:

danielroe (approved):
This is great - looks good in my testing! Not a simple PR by any means. Well done.

huang-julien:
> > Currently if users pass `class`/`style` attributes the following error is shown:
> > ```
> > Extraneous non-props attributes (style, class) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would you explain more about the oldChildren null issue you unearthed?
> 
> This was happening only in build, not sure why but at the setup render function, in some components, if we don't wrap the component render [here](https://github.com/nuxt/framework/pull/6165/files#diff-ce55bf5a46f331c65de3586dec9caf71aba63f13092305fb7e22f9bc9592fe55R51) with a Fragment, the component is not being re-rendered when mounted. And in some cases there's a oldChildren is null error triggered by vue in build and dev. I'll try to reproduce it.

This is likely a Vue issue [stackblitz](https://stackblitz.com/edit/vitejs-vite-n192fc?file=src%2Fcomposable%2FuseClientOnly.ts)

packages/nuxt/src/app/components/client-only.mjs
```diff
@@ -19,18 +19,38 @@ export default defineComponent({
 })
 
 export function createClientOnly (component) {
+  const { setup, render: _render, template: _template } = component
+  if (_render) {
+    // override the component render (non <script setup> component)
+    component.render = (ctx, cache, props, state, data, options) => {
+      return state.mounted$
+        ? h(Fragment, null, [h(_render(ctx, cache, props, state, data, options, ctx), ctx.$attrs ?? ctx._.attrs)])
```
pi0, line 27: @danielroe Is it expected that we have another `ctx` as last argument?

packages/nuxt/src/app/components/client-only.mjs
```diff
@@ -19,18 +19,38 @@ export default defineComponent({
...
-      onMounted(() => { mounted.value = true })
-      return () => {
-        if (mounted.value) {
-          return h(component, attrs, slots)
-        }
-        return h('div', { class: attrs.class, style: attrs.style })
-      }
+    ...component,
+    setup (props, ctx) {
+      const mounted$ = ref(false)
```
pi0, line 40: Can we call it something like `$_mounted` or any specific reason `$` moved to the last?

packages/nuxt/src/app/components/client-only.mjs
pi0: (trying with rest operator, seems working fine 860c716c6a66179299f818ad2002b1612aa65a87)

packages/nuxt/src/app/components/client-only.mjs
pi0: pushed a refactor to use `__mounted__`. vue uses `$` for normal internal/injections and `_` is usually used for private names. I think it should be safe but happy to revert if there was a specific reason for naming convention.

pi0: Oh i see in browser logs. Reverted 👍🏼

packages/nuxt/src/app/components/client-only.mjs
danielroe: checking

danielroe: Looks good in my testing - nice catch.

pi0 (approved):
Thanks for working on this!

ffxsam:
Thank you for this! @huang-julien @danielroe @pi0

PR #6176, opened 2022-07-27
fix(schema): exclude functions from `DeepPartial`

### LINKED ISSUE #5477: `nuxt.config`'s `vue` types are incorrect
### Environment

- Builder:          `vite`
- User Config:      `modules`, `css`, `typescript`, `build`, `runtimeConfig`, `vue`
- Runtime Modules:  `@nuxtjs/tailwindcss@5.1.2`, `floating-vue/nuxt`, `@pinia/nuxt@0.1.9`, `@vueuse/nuxt@8.6.0`
- Build Modules:    `-`

### Reproduction

[StackBlitz](https://stackblitz.com/edit/github-pc4kbz?file=nuxt.config.ts&view=editor)

### Describe the bug

Types do not seem to be correct for the `vue` property in the `nuxt.config` file.

image

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #5477

### ❓ Type of change

### 📚 Description

An early return for functions solves some issues with functional types (including missing types for `hook`, for example). There were also a couple of incorrect types in `_app`.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/_app.ts
+++ b/packages/schema/src/config/_app.ts
@@ -13,7 +13,7 @@ export default {
      * Properties that will be set directly on `Vue.config` for vue@2.
      *
      * @see [vue@2 Documentation](https://v2.vuejs.org/v2/api/#Global-Config)
-     * @type {import('vue/types/vue').VueConfiguration}
+     * @type {typeof import('vue/types/vue').VueConfiguration}
      * @version 2
      */
     config: {
@@ -23,7 +23,7 @@ export default {
     /**
      * Options for the Vue compiler that will be passed at build time
      * @see [documentation](https://vuejs.org/api/application.html#app-config-compileroptions)
-     * @type {import('@vue/compiler-core').CompilerOptions}
+     * @type {typeof import('@vue/compiler-core').CompilerOptions}
      * @version 3
      */
     compilerOptions: {}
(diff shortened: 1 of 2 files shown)


packages/schema/src/config/_app.ts
```diff
@@ -13,7 +13,7 @@ export default {
      * Properties that will be set directly on `Vue.config` for vue@2.
      *
      * @see [vue@2 Documentation](https://v2.vuejs.org/v2/api/#Global-Config)
-     * @type {import('vue/types/vue').VueConfiguration}
+     * @type {typeof import('vue/types/vue').VueConfiguration}
```
pi0, line 16: Does it makes sense to automatically infer this in untyped?

pi0 (approved):

packages/schema/src/config/_app.ts
danielroe: The issue is that we have a two stage build here.

pi0: Sure. I mean if untyped can add `typeof` prefix to `import` in `@type` annotations.

pi0: (anyway, this fix makes sense. Was thinking of a way to avoid future issues like this)

PR #6191, opened 2022-07-27
fix(vite, webpack): generate composable keys based on order

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/6178

### ❓ Type of change

### 📚 Description

Introduced in #4955, we were creating hashed based on code index. But this differs between server + client for vue components (among others) due to differences in the code generated in the two environments. However, I think the _order_ of keyed composables is more solid and likely to remain fixed between client/server.

Other approaches I also tried:

1. using sourcemaps to get original LOC - I think the best approach conceptually. This did not match however (1 line out), surprisingly, due to a bug (I think) in vite's vue plugin - I'll need to investigate further and report. Additionally, we don't yet have isomorphic support for sourcemaps in unplugin (https://github.com/unjs/unplugin/issues/146) so it would add some complexity for webpack.

2. moving the key generation 'up' to earlier in the pipeline (via `enforce: pre`). Also much more reliable as an approach. This would work except that acorn can't handle TS. I could add an additional esbuild transform in before parsing with acorn, but it felt like this would be too heavy.

### 📝 Checklist


pi0 (approved):
Nice approach!

PR #6193, opened 2022-07-27
docs: Update middleware example redirect syntax

jonnyparris:
...because navigateTo didn't work for me when deploying as a static site, but returning the target path string directly did work

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/3.directory-structure/8.middleware.md
+++ b/docs/content/2.guide/3.directory-structure/8.middleware.md
@@ -29,7 +29,7 @@ export default defineNuxtRouteMiddleware((to, from) => {
   if (to.params.id === '1') {
     return abortNavigation()
   }
-  return navigateTo('/')
+  return '/'
 })
 ```
 


danielroe:
You're right, returning the string directly will work, and is a good workaround. But our intend is that `navigateTo` should work also, and be more semantic. Would you raise a bug report with a reproduction so we can address the underlying issue? 🙏

jonnyparris:
I'm struggling to create a minimal workaround of this 😩  but the bug is still hitting my primary app so I'm working on a "top-down" approach removing dependencies until the bugs stops.
Meanwhile rc8 was released which I thought might include an indirect fix but I can't test yet because of #6495 .
I'll update when I can

pi0:
@jonnyparris Do you mind to make a an issue with reproduction to track? 🙏🏼  Indeed this has to be solved to allow `navigateTo` usage.

jonnyparris:
Just fyi - I indirectly fixed my particular navigation issue with thanks to workarounds in this issue thread https://github.com/nuxt-modules/supabase/issues/28

PR #6201, opened 2022-07-28
fix(nuxt): don't call renderMeta if it is not defined

danielroe:
### ❓ Type of change

### 📚 Description

`.renderMeta` is not defined in SPA mode. I think we should also serialize `app.head` for use in this context, but that can be a follow-up PR.

### 📝 Checklist


jkneepkens:
This bug completely destroyed our dev and build step, as we have ssr set on false in the config. Hopefully it can be merged and deployed quickly!

Atinux (approved):

Issue #5508, opened 2022-06-17
css files are being loaded as script module in dev with `ssr: false` (webpack)


gsmeira:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.15.1`
- Nuxt Version:     `3.0.0-rc.4`
- Package Manager:  `yarn@1.22.19`
- Builder:          `webpack`
- User Config:      `builder`, `ssr`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

- `npx nuxi init nuxt-app`
- `cd nuxt-app`
- `yarn add -D @nuxt/webpack-builder`
- add to `nuxt.config.js`: `builder: 'webpack'` / `ssr: false`
- `yarn dev`

### Describe the bug

```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/css"
```

I tested using vite and works fine, but when I set the builder to webpack all css files are being loaded as an script module.

![Screenshot from 2022-06-17 15-18-49](https://user-images.githubusercontent.com/1690400/174355770-87d67bd7-bada-4d5b-b4c0-c61b02c514b8.png)
![Screenshot from 2022-06-17 15-18-36](https://user-images.githubusercontent.com/1690400/174355780-e3892bd6-bdda-4ada-81b1-36466d44edd0.png)

bachilli:
Same here.

When I open my browser at http://localhost:3000/, this happen:

> Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/css". Strict MIME type checking is enforced for module scripts per HTML spec.

BurakGur:
Same issue but i use vite

bachilli:
Any updates about this issue?

danielroe:
reproduction: https://stackblitz.com/edit/github-bu3umw.

chengzhuo5:
这个问题后面还有进展吗

mannok:
same here. Will it be fixed?

Issue #4552, opened 2022-04-23
Assets in the Vite `.output/public/_nuxt/manifest.json` file with `"isEntry": true"` flag is not preloaded with `<link rel="preload">`


mrauhu:
### Environment

```
------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-rc.1`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------
```

### Reproduction

https://stackblitz.com/edit/github-ndsjqb?file=app.vue

### Describe the bug

Hello.

This issue is based on:

* Bug in the `vue-bundle-renderer`
https://github.com/nuxt-contrib/vue-bundle-renderer/pull/28

* Default value of a `shouldPreload` option in the `vue-bundle-renderer` is missing `font` and `image` types:
https://github.com/nuxt-contrib/vue-bundle-renderer/pull/29

* We don't have ability to override a default function of the `shouldPreload` option via `render.bundleRenderer.shouldPreload` option in `nuxt.config.ts`:
https://github.com/nuxt/framework/blob/0d85c9ef77360fc1e56450039dc8436bd2390c69/packages/schema/src/config/render.ts#L13
  * because config options is not passed to the `createRender()` function in the Nitro Renderer:
  https://github.com/nuxt/framework/blob/0eebb0a390c53c05c5bcbcb4fbda64d59a8945b6/packages/nuxt/src/core/runtime/nitro/renderer.ts#L25-L25

Best wishes,
Sergey.

### Additional context

See also: https://github.com/nuxt/framework/discussions/2291.

### Logs

```shell
npm run build
cat .output/public/_nuxt/manifest.json
```

```json
{
  "node_modules/nuxt/dist/app/entry.mjs": {
    "file": "entry-2aec98a4.mjs",
    "src": "node_modules/nuxt/dist/app/entry.mjs",
    "isEntry": true,
    "css": [
      "entry.4964cb3c.css"
    ],
    "assets": [
      "open-sans-cyrillic-ext-400-normal.e1e96c44.woff2",
      "open-sans-all-400-normal.c2e761ea.woff",
      "open-sans-cyrillic-400-normal.f3794e9a.woff2",
      "open-sans-greek-400-normal.e471f953.woff2",
      "open-sans-hebrew-400-normal.651bfd6f.woff2",
[...]

Issue #4475, opened 2022-04-20
Nuxt 3 prefetch all assets from every pages and every layouts.


kstraszewski:
### Environment

Nuxt CLI v3.0.0-rc.0
Nuxt project info:

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.14.0`
- Nuxt Version:     `3.0.0-rc.0`
- Package Manager:  `npm@8.3.1`
- Builder:          `vite`
- User Config:      `css`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/kstraszewski/nuxt3-playground/tree/prefetching-all-assets

Also we have qa version of our project here:
https://b2c.qa2.lendi.pl/design

### Describe the bug

Nuxt after build prefetch all the assets like CSS, fonts, media, imgs from all the pages and all the layouts. As you can see in the repo example you have `/pageA` which implement `layout-a`, with some video in the pageA.vue. Also, you have `/pageB` which implements `layout-b` which loads popular [@mdi/font](https://www.npmjs.com/package/@mdi/font). 

The problem I have is it doesn't matter on which page you go nuxt always loads all the assets. When you go to `/pageA` you will see the mdi fonts from pageB and layoutB also are prefetching by the nuxt:

PageA:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/10195995/164307955-94b3b43f-61a0-424f-a60c-b2041eb642bf.png">

PageB:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/10195995/164308103-7a653d1b-6ba9-432a-88cd-d19e07db14f4.png">

Both pages Loads a lot.

My guess is that vite builder prefetch every URL that he find in CSS or template.

### Additional context

It makes nuxt unusable in production mode because It hurts LCP very badly:

As you can see on the example below we have huge LCP (I've made some tricks witch mdi/fonts, but this is very ugly solution) with fully loaded mdi/fonts I've received 27 sec of LCP.

https://pagespeed.web.dev/report?url=https%3A%2F%2Fb2c.qa2.lendi.pl%2Fdesign

Foddy:
I am facing the same issue. Also not sure what causes this… :S

kkauper:
I can also report the same issue. Would be nice to have a solution for production soon!

tomekhub:
I'm facing the same issue :/

AleksanderWojtkowski:
Same here ...

[...]

PR #6210, opened 2022-07-28
feat(nuxt): update to `vue-bundle-renderer@0.4`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/4475, resolves https://github.com/nuxt/framework/issues/4552, resolves https://github.com/nuxt/framework/issues/5508

### ❓ Type of change

### 📚 Description

This PR migrates fully to the vite manifest format, normalizing the webpack manifest at build time rather than runtime.

It also adds a nitro hook (`nuxt:render:options`) which allows to customise `shouldPrefetch`, `shouldPreload` and `manifest`.

Finally, we run the actual renderer for SPA mode (but once only, with the result then reused for future requests), which avoids us having to rewrite renderer code within nuxt.

The PR is in draft to give visibility to the corresponding changes in the framework that will match https://github.com/nuxt-contrib/vue-bundle-renderer/pull/30. However, tests will fail until we can actually upgrade to a released version in this branch.

**Note**: I also move `@vite/client` to `@vite/client.mjs` so the bundle renderer can load it as a module. An alternative would be _defaulting_ to module behaviour for JS, which might be a better approach, but that will need to be implemented in https://github.com/nuxt-contrib/vue-bundle-renderer/pull/30.

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -69,7 +74,20 @@ const getSSRRenderer = lazyCachedFunction(async () => {
...
+  const manifest = await getClientManifest()
+
+  // Allow customizing options for renderer
+  const nitroApp = useNitroApp()
+  const options = {
+    manifest,
+    renderToString: () => '<div id="__nuxt"></div>',
+    buildAssetsURL
+  }
+  nitroApp.hooks.callHook('nuxt:render:options', options)
```
pi0, line 86: (general note: We should await on `callHook`. Hookable hooks are awaitable by default)

Also what are possible usecases are you thinking to use this hook? I think we can reiterate and think better about customization interface.

packages/vite/src/manifest.ts
```diff
@@ -1,27 +1,32 @@
...
-    async: [],
-    modules: {}
+  const devClientManifest: Manifest = {
+    '@vite/client': {
+      isEntry: true,
+      file: '@vite/client',
+      css,
+      // placeholders for future use
+      assets: [],
+      dynamicImports: [],
```
pi0, line 20: Can normalizer add empty fields?

packages/nuxt/src/core/runtime/nitro/renderer.ts
danielroe: This is about users providing `shouldPreload` and `shouldPrefetch`.

danielroe:
TODO items from my perspective:

- [ ] decide on whether we keep `nuxt:render:options` in this PR
- [ ] release `vue-bundle-renderer` and update resolution in this PR

Then I think it's ready to leave draft.

packages/nuxt/src/core/runtime/nitro/renderer.ts
pi0: Sure. We can add it back. I need to try new renderer more. We might find sexier interface than vue2 shouldPreload/prefetch callbacks.

pi0:
- Released `vue-bundle-renderer@0.4.0`
- You can remove customization hook for this PR 👍🏼

dungonfq:
Awesome work @danielroe @pi0 🙌🏻 🎉 Thank u a lots for these PRs, we'd waited for it for a long time.

When will we have documentation about how to customize `shouldPrefetch`, `shouldPreload` and `manifest`?

danielroe:
@dungonfq The ability customise the renderer options was [removed shortly before merge](https://github.com/nuxt/framework/pull/6210#discussion_r937950236) for faster iteration.

I will update the description.

dungonfq:
@danielroe could u give me where & how to use these new options `shouldPrefetch`,` shouldPreload` and `manifest`?

At first, I thought I can config them in [nuxt.config.ts](https://v3.nuxtjs.org/api/configuration/nuxt.config).

danielroe:
It is not yet possible.

dungonfq:
Thank @danielroe , my last confusion:

So, after u merge this PR. The prefetch issue will be fixed automatically, right?
https://github.com/nuxt/framework/issues/4475

What I expected is all my modules weren't prefetched unnecessarily. Then when I load a page, only the required components/modules, which the page is using, will be fetched.

danielroe:
@dungonfq I don't know what you are asking, but if you are experiencing a new issue, please check that it is not resolved on the edge channel, and then open a new issue and I'll look into it 🙏

kosmeln:
Hello @danielroe . 
Would you help me to understand if this ticket resoled https://github.com/nuxt/framework/issues/4475 issue? I'm still experiencing it after upgrading to RC-7, but was thinking if this is something you mentioned that was branched of for the next release.

danielroe:
@kosmeln Yes, it should have resolved it. (But see https://github.com/nuxt/framework/issues/4475#issuecomment-1162818358.)

If you are experiencing a bug, please raise a new issue with a reproduction. :pray:

PR #6215, opened 2022-07-29
fix(vite): fix distDir resolution

### LINKED ISSUE #6202: vite-node breaks with commit d15c472
### Environment

`v3.0.0-rc.6-27648546.d15c472`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-hk7jvp?file=package.json,nuxt.config.ts

Try moving back to 3.0.0-rc.6-27647394.5e62e0f in package.json and reinstalling dependencies, the project will work fine

### Describe the bug

```
[h3] [unhandled] H3Error: Cannot find module '/Users/yaelguilloux/Code/nuxt/docus/node_modules/@nuxt/vite-builder/dist/chunks/runtime/client.manifest.mjs' imported from /Users/yaelguilloux/Code/nuxt/docus/.nuxt/dist/server/client.manifest.mjs
    at createError (file:///Users/yaelguilloux/Code/nuxt/docus/node_modules/h3/dist/index.mjs:191:15)
    at Server.nodeHandler (file:///Users/yaelguilloux/Code/nuxt/docus/node_modules/h3/dist/index.mjs:381:21) {
  statusCode: 500,
  fatal: false,
  unhandled: true,
  statusMessage: 'Internal Server Error'
}
```

This gets shown if booting with vite-node, site works fine if disabled

### Additional context

_No response_

### Logs

_No response_

antfu:
### 🔗 Linked issue

close #6202

### ❓ Type of change

### 📚 Description

The cause of #6202 is that the `dirs.ts` module being bundled into `chunks/index.js` changes its relative path. Move it to `utils` to make it consistent across dev and build.

Also taking the chance to improve the code splitting by dynamically importing for optional plugins. Previously `vite-node` and `rollup-plugin-visualizer` are always imported regardless if the feature is enabled or not.

### 📝 Checklist


danielroe (approved):
❤️

packages/vite/src/utils/dirs.ts
```diff
@@ -0,0 +1,5 @@
+import { fileURLToPath } from 'node:url'
+import { resolve } from 'pathe'
+
+export const distDir = resolve(fileURLToPath(import.meta.url), '../..')
```
pi0, line 4: This file was intentionally in top-level to match source and generated chunk path.

antfu: Yes, but unbuild always extracts it into a chunk even when I have directly imported it in `index.ts`. Or maybe we need another top-level export `dirs` to make sure it's always on top-level?

antfu: It was a non-problem because we used to have only one file for this package.

pi0: I see. The only issue is that is approach can be easily broken if chunk generation changes with unbuild>rollup.

Currently, we use this workaround for the main nuxt pkg: https://github.com/nuxt/framework/blob/5a042cd69b8f1461d309edf4b59cd1690493dcaf/packages/nuxt/src/dirs.ts#L5

antfu: Ok, updated

packages/vite/src/client.ts
```diff
@@ -58,7 +56,9 @@ export async function buildClient (ctx: ViteBuildContext) {
         rootDir: ctx.nuxt.options.rootDir,
         buildAssetsURL: joinURL(ctx.nuxt.options.app.baseURL, ctx.nuxt.options.app.buildAssetsDir)
       }),
-      viteNodePlugin(ctx)
+      ctx.nuxt.options.experimental.viteNode
+        ? await import('./vite-node').then(r => r.viteNodePlugin(ctx))
```
pi0, line 60: Since this is going to be the default option, I think we can safely import it normally when removed experimental flag. Difference is small but rollup this way can optimize startup time to add nested dependencies to top level `dist/index.mjs`

pi0 (approved):

Issue #5018, opened 2022-05-17
Vite Node stability


antfu:
- [x] FOUC on every request since rc.3 - #5017
- [x] FOUC on initial request - https://github.com/nuxt-contrib/vue-bundle-renderer/pull/31
- [x] HMR stability - https://github.com/nuxt/framework/pull/6154
- [ ] https://github.com/nuxt/framework/pull/6153

antfu:
I will send a final PR to improve the HMR then we and enable it in the next version

antfu:
I think the combination of #6154 and #6153 will solve most of the issues. I would have one more release as experimental to be sure then it should be ready.

Issue #3998, opened 2022-03-31
[NITRO] WARN  [worker] listen EADDRINUSE: address already in use


misaon:
### Environment

```
------------------------------
- Operating System: `Linux`
- Node Version:     `v17.8.0`
- Nuxt Version:     `3.0.0-27477996.646c2f6`
- Package Manager:  `yarn@3.2.0`
- Builder:          `vite`
- User Config:      `modules`
- Runtime Modules:  `normalizedModule()`, `normalizedModule()`
- Build Modules:    `-`
------------------------------
```

### Reproduction

Hapen randomly

### Describe the bug

Sometimes I get this error during development in the console. It seems completely random but often.

### Additional context

Nuxt runs in my clean Docker container:

```
FROM node:17.8.0-alpine3.15 as nuxtBuild

WORKDIR /app

COPY ./client ./

RUN yarn config set npmRegistryServer https://registry.npmmirror.com/ \
    && yarn install \
    && yarn run build

FROM node:17.8.0-alpine3.15

WORKDIR /app

COPY ./client/package.json /app/package.json
COPY ./docker-entrypoint.sh /usr/local/bin/oxy-entrypoint
COPY --from=nuxtBuild /app/.output /app/.output

RUN chmod +x /usr/local/bin/oxy-entrypoint

ENV NUXT_HOST=0.0.0.0

ENTRYPOINT ["oxy-entrypoint"]
```

Entrypoint:

```
#!/bin/sh
set -e

if [ "$NODE_ENV" = "production" ]; then
    yarn run start
else
    yarn install
    yarn run dev
fi
```

### Logs

```shell
WARN  [worker] listen EADDRINUSE: address already in use /tmp/nitro/worker-74-2.sock
 
   at Server.setupListenHandle [as _listen2] (node:net:1355:21)
   at listenInCluster (node:net:1420:12)
   at Server.listen (node:net:1519:5)
   at .nuxt/nitro/index.mjs:140:8
   at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
   at async Promise.all (index 0)
   at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
   at async loadESM (node:internal/process/esm_loader:88:5)
   at async handleMainPromise (node:internal/modules/run_main:61:12)
```

stillhart-at-bechterew:
Same here, so I choose to just delete the sockets when it happens. Not nice but at least an easy work around. 
```
[...]

PR #6217, opened 2022-07-29
feat(schema, vite)!: enable `vite-node` by default

antfu:
### 🔗 Linked issue

#5018

### ❓ Type of change

### 📝 Checklist


packages/vite/src/vite.ts
```diff
@@ -16,6 +16,7 @@ import { composableKeysPlugin } from './plugins/composable-keys'
 export interface ViteOptions extends InlineConfig {
   vue?: Options
   ssr?: SSROptions
+  viteNode?: boolean
```
pi0, line 19: Can we please change option to `serverBundler: 'vite-node' | 'legacy'`? This reflects the fact it is only for server and allows any new mechanism in the future as same option.

packages/schema/src/config/experimental.ts
```diff
@@ -30,5 +25,20 @@ export default {
...
+
+    /**
+     * Use vite-node for on-demand server chunk loading
+     * 
+     * @deprecated use `vite.viteNode`
+     */
+    viteNode: {
+      $resolve: (val) => {
+        if (val === true)
+          console.warn('`vite-node` is now enabled by default. You can safely remove `experimental.viteNode` from your config.')
```
pi0, line 37: This is nice with experimental notice. Can we also support `EXPERIMENTAL_VITE_NODE` and also code style: `if (...) {} else {}`

misaon:
This will broke all guys which use docker for develop in Nuxt:
https://github.com/nuxt/framework/issues/3998#issuecomment-1184281675

misaon:
@pi0 I can create a sandbox where the bug can be simulated. However, as has already been written in the thread, the bug manifests itself randomly. It doesn't seem to have an obvious cause, it's just that during development (editing components, pages, composables) a situation arises where the bug appears. In our team, where we develop in Docker projects built on Nuxt, the problem happens randomly to everyone.

We have found that if this error appears in the console, we have to make a change in some vue file (sometimes several times) to get the watcher to start again and work properly. We use the simplest form of Dockerfile

```dockerfile
FROM node:18.6.0-alpine3.15 as nuxtBuild

WORKDIR /app

COPY ./ ./

RUN yarn install \
    && yarn run build

FROM node:18.6.0-alpine3.15

ENV NODE_OPTIONS='--max-http-header-size=32768'
ENV HOST=0.0.0.0
WORKDIR /app

COPY --from=nuxtBuild /app/.output /app/.output
```

It is possible that this is the fault of the Alpine distribution itself, but it is on this distribution that NodeJS officially publishes its docker images.

pi0:
I can confirm the same issue. It seems not even limited to docker but any Linux/Windows env, HMR watcher is ignored for server and only applied to the client, therefore, having hydration mismatch.

I've converted PR to draft until we solve this. 

It might be reload issue #2047 since now client HMR on linux works and Server not (in current nuxt, it is otherwise). But also trying `CHOKIDAR_USEPOLLING=1` I still have issue.

misaon:
> 

Good work @pi0! ❤️

packages/schema/src/types/config.ts
```diff
@@ -45,4 +45,9 @@ export interface ViteConfig extends ViteUserConfig {
    * @see https://github.com/vitejs/vite/tree/main/packages/plugin-vue
    */
   vue?: VuePluginOptions
+  /**
+   * Bundler for dev time server-side rendering.
+   * @default 'vite-node'
+   */
+  serverBundler?: 'vite-node' | 'legacy'
```
danielroe, line 52: Maybe:


```suggestion
  devBundler?: 'vite-node' | 'legacy'
```

To make it clear we are not talking about anything runtime in built app.

pi0 (approved):
LGTM at current status. I cannot reproduce (at least stabily) HMR issues. I think we can move it forward to test on edge until next rc.

PR #6227, opened 2022-07-29
perf(vite): remove vite warm up

antfu:
### ❓ Type of change

### 📚 Description

This reduces the start-up time of `nuxt.com` repo with `viteNode: true` (`nuxi dev` to first render on the browser)
**from 60s to 40s** on my M1 Max Macbook Pro.

Reasoning:
- Vite will now do warm-up for sync deps automatically.
- Our warm-up uses `server.transformRequest(url)` for both client/server. Where the server bundler actually needs `server.transformRequest(url, { ssr: true })`, which is actually using a different plugin pipeline and caches. 

### 📝 Checklist


pi0:
> Vite will now do warm-up for sync deps automatically.

Does Vite also warmups the entrypoint in background? (Same for `vite-node` with `ssr` condition)

> Our warm-up uses server.transformRequest(url) for both client/server. Where the server bundler actually needs server.transformRequest(url, { ssr: true }), which is actually using a different plugin pipeline and caches.

Nice find and probably to late to fix if warmup is not relevant anymore 😅

packages/vite/src/vite.ts
```diff
@@ -2,14 +2,13 @@ import * as vite from 'vite'
...
 import type { InlineConfig, SSROptions } from 'vite'
-import { logger, isIgnored } from '@nuxt/kit'
+import { isIgnored } from '@nuxt/kit'
 import type { Options } from '@vitejs/plugin-vue'
 import replace from '@rollup/plugin-replace'
 import { sanitizeFilePath } from 'mlly'
 import { buildClient } from './client'
 import { buildServer } from './server'
 import virtual from './plugins/virtual'
-import { warmupViteServer } from './utils/warmup'
```
pi0, line 12: If this is not relevant anymore I think we can remove util as well. Only I want to make sure vite and vite-node does the same because in past, without entrywarmup we were simply _delaying_ it until first request.

antfu:
I also tried to only warm up the entry (and let Vite crawl the deps), it ends up around 50s start-up time. I guess it might be because it occupied the execution time of other parts. I could do a more detailed benchmarking, and probably try delaying the warm-up with a timeout.

pi0:
The suggested method to benchmark:
- Cleanup cache with `nuxi cleanup` each time
- In parallel initiate a fetch to `http://localhost:300` and start `nuxi dev`
  - Measure time to first byte for SSR perf
- Open a new browser window ensuring no other windows are open
  - Measure time to load for entrypoint scripts

PR #6220, opened 2022-07-29
fix(schema): add typings for `vite.vue` options

danielroe:
### 🔗 Linked issue

resolves #6218

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This adds typings for `vite.vue` to the schema.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #6230, opened 2022-07-29
chore(schema): update vite config

pi0:
### ❓ Type of change

### 📚 Description

Add leftovers from #6220

### 📝 Checklist

### DIFF
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -7,7 +7,7 @@ type DeepPartial<T> = T extends Function ? T : T extends Record<string, any> ? {
 
 /** User configuration in `nuxt.config` file */
 export interface NuxtConfig extends DeepPartial<Omit<ConfigSchema, 'vite'>> {
-  // Avoid  DeepPartial for vite config interface (#4772)
+  // Avoid DeepPartial for vite config interface (#4772)
   vite?: ViteConfig
   [key: string]: any
 }
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -7,7 +7,7 @@ type DeepPartial<T> = T extends Function ? T : T extends Record<string, any> ? {
 
 /** User configuration in `nuxt.config` file */
 export interface NuxtConfig extends DeepPartial<Omit<ConfigSchema, 'vite'>> {
-  // Avoid  DeepPartial for vite config inteface (https://github.com/nuxt/framework/pull/4772)
+  // Avoid  DeepPartial for vite config interface (#4772)
   vite?: ViteConfig
   [key: string]: any
 }
--- a/packages/schema/src/types/config.ts
+++ b/packages/schema/src/types/config.ts
@@ -1,13 +1,14 @@
 import { ConfigSchema } from '../../schema/config'
 import type { ResolvedConfig } from 'c12'
-import { UserConfig } from 'vite'
-import { Options as VuePluginOptions } from '@vitejs/plugin-vue'
+import type { UserConfig as ViteUserConfig } from 'vite'
+import type { Options as VuePluginOptions } from '@vitejs/plugin-vue'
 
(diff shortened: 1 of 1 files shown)


packages/schema/src/types/config.ts
```diff
@@ -1,13 +1,14 @@
...
-import { Options as VuePluginOptions } from '@vitejs/plugin-vue'
+import type { UserConfig as ViteUserConfig } from 'vite'
+import type { Options as VuePluginOptions } from '@vitejs/plugin-vue'
 
 type DeepPartial<T> = T extends Function ? T : T extends Record<string, any> ? { [P in keyof T]?: DeepPartial<T[P]> } : T
 
 /** User configuration in `nuxt.config` file */
 export interface NuxtConfig extends DeepPartial<Omit<ConfigSchema, 'vite'>> {
-  vite?: ConfigSchema['vite']
+  // Avoid  DeepPartial for vite config inteface (https://github.com/nuxt/framework/pull/4772)
```
pi0, line 10:
```suggestion
  // Avoid  DeepPartial for vite config interface (#4772)
```

packages/schema/src/types/config.ts
```diff
@@ -1,13 +1,14 @@
...
-import { Options as VuePluginOptions } from '@vitejs/plugin-vue'
+import type { UserConfig as ViteUserConfig } from 'vite'
+import type { Options as VuePluginOptions } from '@vitejs/plugin-vue'
 
 type DeepPartial<T> = T extends Function ? T : T extends Record<string, any> ? { [P in keyof T]?: DeepPartial<T[P]> } : T
 
 /** User configuration in `nuxt.config` file */
 export interface NuxtConfig extends DeepPartial<Omit<ConfigSchema, 'vite'>> {
-  vite?: ConfigSchema['vite']
+  // Avoid  DeepPartial for vite config interface (#4772)
```
danielroe, line 10:
```suggestion
  // Avoid DeepPartial for vite config interface (#4772)
```

packages/schema/src/types/config.ts
```diff
@@ -1,13 +1,14 @@
...
-import { UserConfig } from 'vite'
-import { Options as VuePluginOptions } from '@vitejs/plugin-vue'
+import type { UserConfig as ViteUserConfig } from 'vite'
+import type { Options as VuePluginOptions } from '@vitejs/plugin-vue'
 
 type DeepPartial<T> = T extends Function ? T : T extends Record<string, any> ? { [P in keyof T]?: DeepPartial<T[P]> } : T
 
 /** User configuration in `nuxt.config` file */
 export interface NuxtConfig extends DeepPartial<Omit<ConfigSchema, 'vite'>> {
-  vite?: ConfigSchema['vite']
```
danielroe, line 10: Why this change? Seems like this would override the schema types entirely.

pi0: They are same reference now with new ViteConfig interface.

danielroe: So why not keep it as `ConfigSchema['vite']`? It would be confusing for a future contributor to edit schema and not have it take effect.

packages/schema/src/types/config.ts
```diff
@@ -1,13 +1,14 @@
...
+  // Avoid DeepPartial for vite config interface (#4772)
+  vite?: ViteConfig
```
danielroe, line 11:
```suggestion
  vite?: ConfigSchema['vite']
```

packages/schema/src/types/config.ts
pi0: Actually, I was confused at first too what is the difference between these two and had to check. 

Also as I can understand, we do not infer this type from schema defaults. Two NuxtConfig and NuxtOptions refer to the same custom interface.

danielroe: I think it is better to keep as ConfigSchema. Logically we are modifying it only to apply DeepPartial to _part_ of it. So both forks should keep reference to ConfigSchema, even if it ultimately resolves to ViteConfig. It is more consistent.

PR #6232, opened 2022-07-29
fix(webpack): output `.mjs` to use crossorigin preloads

danielroe:
### ❓ Type of change

### 📚 Description

We are outputting `.js` files for webpack, which means we also don't add the `crossorigin` attribute when rendering. Because of a typo in our tests we weren't pick up on this.

Bug shows up with following warnings (+ double fetches of all webpack JS assets):

```
A preload for 'http://localhost:3000/_nuxt/59e3a34.js' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.
The resource http://localhost:3000/_nuxt/59e3a34.js was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
```

### 📝 Checklist

### DIFF
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -42,7 +42,7 @@ export async function expectNoClientErrors (path: string) {
   const { pageErrors, consoleLogs } = await renderPage(path)
 
   const consoleLogErrors = consoleLogs.filter(i => i.type === 'error')
-  const consoleLogWarnings = consoleLogs.filter(i => i.type === 'warn')
+  const consoleLogWarnings = consoleLogs.filter(i => i.type === 'warning')
 
   expect(pageErrors).toEqual([])
   expect(consoleLogErrors).toEqual([])
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -42,7 +42,7 @@ export async function expectNoClientErrors (path: string) {
   const { pageErrors, consoleLogs } = await renderPage(path)
 
   const consoleLogErrors = consoleLogs.filter(i => i.type === 'error')
-  const consoleLogWarnings = consoleLogs.filter(i => i.type === 'warning')
+  const consoleLogWarnings = consoleLogs.filter(i => ['warning', 'warn'].includes(i.type))
 
   expect(pageErrors).toEqual([])
   expect(consoleLogErrors).toEqual([])
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -42,7 +42,7 @@ export async function expectNoClientErrors (path: string) {
   const { pageErrors, consoleLogs } = await renderPage(path)
 
   const consoleLogErrors = consoleLogs.filter(i => i.type === 'error')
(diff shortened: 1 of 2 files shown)


pi0:
Would it be fixed with upcoming vue-bundle-renderer? At some point we have to switch back to `.js` for client module assets. Considering CDN mime support.

test/utils.ts
```diff
@@ -42,7 +42,7 @@ export async function expectNoClientErrors (path: string) {
   const { pageErrors, consoleLogs } = await renderPage(path)
 
   const consoleLogErrors = consoleLogs.filter(i => i.type === 'error')
-  const consoleLogWarnings = consoleLogs.filter(i => i.type === 'warn')
+  const consoleLogWarnings = consoleLogs.filter(i => i.type === 'warning')
```
pi0, line 45: Consola's type is `warn` not warning. Could you trace why type is with longer type?

danielroe:
No, the vue-bundle-renderer PR doesn't change this behaviour when rendering `crossorigin` attribute. (We _can_ change it there to default to module - do you think that's a better option?)

test/utils.ts
danielroe: It is coming from playwright.

CleanShot 2022-07-29 at 13 25 56

pi0:
This PR seems good to go until we do this but yes I think it is a viable option to support `type:module` instead of depending on extension for module detection if possible.

danielroe:
As a top-level option to `createRenderer` (defaulting to true)?

test/utils.ts
danielroe: I can add the `warn` condition as well if you think conceivably the type could be this also? I think we mock consola with console in build, right?

pi0:
> As a top-level option to createRenderer (defaulting to true)?

I believe could be in manifest.

test/utils.ts
pi0: My bad. If types are comming from browser, yes make sense we had to type internal array to avoid this mistake. later this util could be generalized and let's remember to include both variants (or map playwright type to be consistent with logging infra)

danielroe:
~> https://github.com/nuxt-contrib/vue-bundle-renderer/issues/32.

test/utils.ts
```diff
@@ -42,7 +42,7 @@ export async function expectNoClientErrors (path: string) {
   const { pageErrors, consoleLogs } = await renderPage(path)
 
   const consoleLogErrors = consoleLogs.filter(i => i.type === 'error')
-  const consoleLogWarnings = consoleLogs.filter(i => i.type === 'warn')
+  const consoleLogWarnings = consoleLogs.filter(i => ['warning', 'warn'].includes(i.type))
```
pi0, line 45:
```suggestion
  const consoleLogWarnings = consoleLogs.filter(i => i.type === 'warning')
```

pi0 (approved):

PR #6240, opened 2022-07-29
feat(kit): support `followSymbolicLinks` option for  `resolveFiles`

Diizzayy:

### 📚 Description

This PR allows module authors to opt-out of following symbolic links when using `resolveFiles`, as there are many cases where this current default isn't the desired.

### DIFF
--- a/packages/kit/src/resolve.ts
+++ b/packages/kit/src/resolve.ts
@@ -151,7 +151,7 @@ async function isDirectory (path: string) {
   return (await fsp.lstat(path)).isDirectory()
 }
 
-export async function resolveFiles (path: string, pattern: string | string[], opts?: { followSymbolicLinks?: boolean }) {
+export async function resolveFiles (path: string, pattern: string | string[], opts: { followSymbolicLinks?: boolean } = {}) {
   const files = await globby(pattern, { cwd: path, followSymbolicLinks: opts?.followSymbolicLinks || true })
   return files.map(p => resolve(path, p)).filter(p => !isIgnored(p)).sort()
 }
--- a/packages/kit/src/resolve.ts
+++ b/packages/kit/src/resolve.ts
@@ -151,7 +151,7 @@ async function isDirectory (path: string) {
   return (await fsp.lstat(path)).isDirectory()
 }
 
-export async function resolveFiles (path: string, pattern: string | string[], followSymbolicLinks: boolean = true) {
-  const files = await globby(pattern, { cwd: path, followSymbolicLinks })
+export async function resolveFiles (path: string, pattern: string | string[], opts?: { followSymbolicLinks?: boolean }) {
+  const files = await globby(pattern, { cwd: path, followSymbolicLinks: opts?.followSymbolicLinks || true })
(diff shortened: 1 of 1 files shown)


packages/kit/src/resolve.ts
```diff
@@ -151,7 +151,7 @@ async function isDirectory (path: string) {
   return (await fsp.lstat(path)).isDirectory()
 }
 
-export async function resolveFiles (path: string, pattern: string | string[]) {
-  const files = await globby(pattern, { cwd: path, followSymbolicLinks: true })
+export async function resolveFiles (path: string, pattern: string | string[], followSymbolicLinks: boolean = true) {
```
pi0, line 154: Can you please make 3rd argument an object `{ followSymbolicLinks }`

Diizzayy:
>  In the future, please consider opening an issue describing improvement and enhancement 🙏🏼 

Sorry about that, I had begun writing an issue when I realized that this was the underlying cause.
I'll be sure to keep this in mind going forward. I jumped the gun here.

> Also what is the current use case you have for this utility for skipping link resolution

@pi0 When using pnpm workspaces which uses symlinks between packages, you'll find that based on how the resolveFiles utility is used in a module, it's default behavior ( following symlinks ) can result in scanning deeply nested files in `node_modules` that match the pattern ( which may not be the desired ). This can also lead to an _almost_ infinite loop of resolving files when a symlinked package depends on another that also contains matching files (which is where I encountered it).

packages/kit/src/resolve.ts
```diff
@@ -151,7 +151,7 @@ async function isDirectory (path: string) {
   return (await fsp.lstat(path)).isDirectory()
 }
 
-export async function resolveFiles (path: string, pattern: string | string[]) {
-  const files = await globby(pattern, { cwd: path, followSymbolicLinks: true })
+export async function resolveFiles (path: string, pattern: string | string[], opts?: { followSymbolicLinks?: boolean }) {
+  const files = await globby(pattern, { cwd: path, followSymbolicLinks: opts?.followSymbolicLinks || true })
```
pi0, line 155: when `followSymbolicLinks` is `false`, it will evaluate to `false || true` (true). 
```suggestion
  const files = await globby(pattern, { cwd: path, followSymbolicLinks: opts?.followSymbolicLinks ?? true })
```

packages/kit/src/resolve.ts
```diff
@@ -151,7 +151,7 @@ async function isDirectory (path: string) {
   return (await fsp.lstat(path)).isDirectory()
 }
 
-export async function resolveFiles (path: string, pattern: string | string[]) {
-  const files = await globby(pattern, { cwd: path, followSymbolicLinks: true })
+export async function resolveFiles (path: string, pattern: string | string[], opts?: { followSymbolicLinks?: boolean }) {
```
pi0, line 154:
```suggestion
export async function resolveFiles (path: string, pattern: string | string[], opts: { followSymbolicLinks?: boolean } = {}) {
```

pi0 (approved):

PR #6251, opened 2022-07-29
feat(nuxt): add `fetchpriority` attribute and literal typings for meta components

sanjaiyan-dev:
Added fetchpriority attr in script and link component.


danielroe (approved):

DamianGlowala:
Shall we narrow down the accepted values for the `fetchpriority` attribute, as per https://wicg.github.io/priority-hints/#definitions?

packages/nuxt/src/head/runtime/components.ts
```diff
@@ -70,6 +70,7 @@ export const Script = defineComponent({
       default: undefined
     },
     defer: Boolean,
+    fetchpriority: String,
```
DamianGlowala, line 73: https://wicg.github.io/priority-hints/#definitions
```
type FetchPriority = 'high' | 'low' | 'auto'
```

```suggestion
    fetchpriority: String as PropType<FetchPriority>,
```

packages/nuxt/src/head/runtime/components.ts
```diff
@@ -118,6 +119,7 @@ export const Link = defineComponent({
     as: String,
     crossorigin: String,
     disabled: Boolean,
+    fetchpriority: String,
```
DamianGlowala, line 122:
```suggestion
    fetchpriority: String as PropType<FetchPriority>,
```

DamianGlowala (changes requested):

DamianGlowala:
Great enhancement! I have closed #6262 in favor of your changes. There are two remarks that I've got:

- Why is boolean accepted for `crossorigin`? Some counterexamples:
    - https://github.com/posthtml/htmlnano/pull/79
    - https://github.com/nuxt/vue-meta/issues/447
- `x-ua-compatible` appears to be missing in `httpEquiv`, as per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-http-equiv.

DamianGlowala:
All I am thinking is whether the following would be preferable:

```
crossorigin: String as PropType<CrossOrigin>
```

Let's wait for others to express their thoughts, not sure which way to go :)

packages/nuxt/src/head/runtime/components.ts
```diff
@@ -167,7 +168,7 @@ export const Base = defineComponent({
   props: {
     ...globalProps,
     href: String,
-    target: String,
+    target: String as PropType<Targets>,
```
DamianGlowala, line 171: I think you meant `Target` (singular form) 👀

packages/nuxt/src/head/runtime/components.ts
```diff
@@ -1,40 +1,56 @@
...
+import { useHead } from "./composables";
+import type {
+  Props,
+  FetchPriority,
+  CrossOrigin,
+  HTTPEquiv,
+  ReferrerPolicy,
+  LinkRelationship,
+  Target,
+} from "./types.ts";
```
danielroe, line 12: Love the direction of this PR. Would you revert the various style/lint changes that have been made by your IDE?

danielroe (changes requested):
Love the direction of this PR. Would you revert the various style/lint changes that have been made by your IDE?

sanjaiyan-dev:
Made suggested change (not sure) 😄

packages/nuxt/src/head/runtime/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  | 'next'
+  | 'pingback'
+  | 'preconnect'
+  | 'prefetch'
+  | 'preload'
+  | 'prerender'
+  | 'prev'
+  | 'search'
+  | 'stylesheet'
+  | string
```
danielroe, line 45: **A note** - no autocompletion will be provided for `LinkRelationship` and `Target` as providing `string` as a type effectively removes the string literal types.

danielroe (approved):
Really nice enhancement - thanks! ❤️

packages/nuxt/src/head/runtime/types.ts
pi0: Does it makes sense to keep the types then?

danielroe: Yes, we could equally remove them without changing user experience.

danielroe: Okay - small change keeps the autocompletion 👍

pi0 (approved):
Thanks!

sanjaiyan-dev:
> Thanks!

My pleasure 💪🙌

packages/nuxt/src/head/runtime/types.ts
DamianGlowala: Nice one! An interesting approach to preserve autocompletion by treating the `(string & {})` as 'distinguishable' :) As a fun fact, it's also covered by the `LiteralUnion` type in the `type-fest` lib:

https://github.com/sindresorhus/type-fest/blob/main/source/literal-union.d.ts

PR #6275, opened 2022-08-01
fix(nuxt): separate routes for different suspense forks

### LINKED ISSUE #6274: route updates before suspense resolves
Currently using `useRoute` (or `$route`) within a page means that some unexpected and counter-intuitive behaviours can occur when changing routes.

For example, content within a page that is dependent on the route can update _before_ the new page swaps in, resulting in a strange flash of outdated data, or a network request that returns an invalid result (for example, when `route.params.slug` goes to undefined). This is more evident the longer it takes to resolve async setups, so when network speed is lower. In some cases that also mean that invalid data was getting cached within `useAsyncData` responses.

The issue is that `<Suspense>` creates two trees of vnodes, and needs to resolve the hidden tree before swapping it in. But it does not freeze reactivity on the route, so the 'visible' page reacts to the change of route before the new route page actually appears on the page.

**Reproduction**: https://stackblitz.com/edit/github-qwr9bd

CleanShot 2022-08-01 at 09 50 46

danielroe:
### 🔗 Linked issue

resolves #6274

### ❓ Type of change

### 📚 Description

This PR provides the current route directly at the point of suspense, meaning within the context of a suspense fork, the route will not change until suspense has resolved. We also update the default `useRoute` behaviour to mean that the route will not change outside of a page context until suspense has resolved.

**Context**: https://github.com/nuxt/framework/pull/4124, https://github.com/nuxt/framework/issues/4122, https://github.com/nuxt/framework/pull/4092.

---

**Previous behaviour**:

CleanShot 2022-08-01 at 09 50 46

**New behaviour**:

CleanShot 2022-08-01 at 09 51 37

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/runtime/router.ts
+++ b/packages/nuxt/src/pages/runtime/router.ts
@@ -77,7 +77,6 @@ export default defineNuxtPlugin(async (nuxtApp) => {
   })
 
   // Allows suspending the route object until page navigation completes
-  // https://github.com/vuejs/vue-router-next/blob/master/src/router.ts#L1192-L1200
   const _route = shallowRef(router.resolve(initialURL) as RouteLocation)
   const syncCurrentRoute = () => { _route.value = router.currentRoute.value }
   nuxtApp.hook('page:finish', syncCurrentRoute)
@@ -89,7 +88,7 @@ export default defineNuxtPlugin(async (nuxtApp) => {
     }
   })
 
-  // https://github.com/vuejs/vue-router-next/blob/master/src/router.ts#L1192-L1200
+  // https://github.com/vuejs/router/blob/main/packages/router/src/router.ts#L1225-L1233
   const route = {}
   for (const key in _route.value) {
     route[key] = computed(() => _route.value[key])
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -76,15 +76,10 @@ export default defineNuxtPlugin(async (nuxtApp) => {
...
-  // https://github.com/vuejs/vue-router-next/blob/master/src/router.ts#L1192-L1200
-  const route = {}
-  for (const key in router.currentRoute.value) {
-    route[key] = computed(() => router.currentRoute.value[key])
-  }
-
   // Allows suspending the route object until page navigation completes
-  const _activeRoute = shallowRef(router.resolve(initialURL) as RouteLocation)
-  const syncCurrentRoute = () => { _activeRoute.value = router.currentRoute.value }
+  // https://github.com/vuejs/vue-router-next/blob/master/src/router.ts#L1192-L1200
```
posva, line 80: what is the correct line at https://github.com/vuejs/router/blob/main/packages/router/src/router.ts#L1192-L1200?

danielroe: Updated to https://github.com/vuejs/router/blob/main/packages/router/src/router.ts#L1225-L1233

Atinux (approved):
Looks good to me ❤️

pi0 (approved):
Tested locally. All seems great!

PR #6281, opened 2022-08-01
feat(vite-node): clear console when rendering error get fixed

### LINKED ISSUE #4787: Clear console on HMR
Currently, the terminal only outputs the warnings and errors but has no feedback on HMR events. Making it stays on the previous error even when the app is back to working normally.

antfu:
### 🔗 Linked issue

fix #4787

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/vite/src/runtime/vite-node.mjs
+++ b/packages/vite/src/runtime/vite-node.mjs
@@ -1,5 +1,6 @@
 import { ViteNodeRunner } from 'vite-node/client'
 import { $fetch } from 'ohmyfetch'
+import consola from 'consola'
 import { getViteNodeOptions } from './vite-node-shared.mjs'
 
 const viteNodeOptions = getViteNodeOptions()
@@ -15,6 +16,7 @@ const runner = new ViteNodeRunner({
 })
 
 let render
+let previousError = null
 
 export default async (ssrContext) => {
   // Workaround for stub mode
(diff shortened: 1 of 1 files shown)


Atinux:
How do you test it locally?

antfu:
You can add a `throw new Error` in `<script setup>` to trigger SSR error and then remove it.

pi0:
Nice idea to only clear console when there are errors but i think we can do it with main consola instance wrapping console in nuxi to catch all sources of warning and error.

As for this PR, can we show a message when we detect vite-node assets are changed in next render? Currently there is no watcher and message like client-side indicating HMR happened or legacy vite builder. Do we still have a watcher with vite-node ?

Atinux:
I may do something wrong but I don't see the clear (also got the server stuck at one point):

Uploading Screenshot 2022-08-02 at 17.12.17.mp4…

packages/vite/src/runtime/vite-node.mjs
```diff
@@ -28,5 +30,12 @@ export default async (ssrContext) => {
       runner.moduleCache.delete(key)
     }
   }
+  if (ssrContext.error) {
+    previousError = ssrContext.error
+  } else if (previousError) {
+    previousError = null
+    consola.clear()
```
pi0, line 37: This doesn't works to clear screen. `consola.clear` was legacy name for `consola. removeReporter`. 

Also we probably shouldn't clear screen from runtime.

pi0:
Thanks to #6343, we can now show accurate message on HMR. Moving log enhancenment to #6351. And lets track clear on rebuild experience in https://github.com/nuxt/framework/issues/4787. I think we could handle it on nuxi level.

PR #6293, opened 2022-08-02
fix(nuxt): render head scripts that use `body: true`

harlan-zw:
### 🔗 Linked issue

@vueuse/head PR: https://github.com/vueuse/head/pull/67

issues: https://github.com/nuxt/framework/issues/2374, https://github.com/nuxt/framework/issues/5014, https://github.com/nuxt/framework/issues/3130

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

With nuxt using the latest version of @vueuse/head, it already has support for appending script and links to the end of the body tag with the `body: true` option. 

However due to a mismatch with property names of NuxtMeta and @vueuse/head nothing is rendered, this fixes that and adds some small doc and tests around the feature

Note: This does not work in components, this will need to be looked at in a separate PR as there is some extra complexity involved

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #6296, opened 2022-08-02
fix(nuxt): avoid duplicate titleTemplate

harlan-zw:
### 🔗 Linked issue

Related PR: https://github.com/vueuse/head/commit/1382c43c5efd9d00efb17fefefa732febbea2d80

### ❓ Type of change

- [x] ⚠️ Breaking change (VERY MINOR) (fix or feature that would cause existing functionality to change)

### 📚 Description

When working on https://github.com/nuxt/framework/pull/6293, I updated the @vueuse/head package to get the [dedupe fix](https://github.com/vueuse/head/commit/c02a96fd2489e74a1bf80ceb64f128f544629b4a). 

In updating, I noticed that all titles had duplicate template processing. For example `Home - My Title`, became `Home - My Title - My Title`. This was due to @vueuse/head introducing its own [titleTemplate](https://github.com/vueuse/head/commit/1382c43c5efd9d00efb17fefefa732febbea2d80) feature.

I figured the best course of action would be to adopt the native titleTemplate logic which, as far as I tested, includes the same feature set. 

All seems to work well, however, there is one very minor "breaking" change. In the documentation, it recommended setting `titleTemplate` to `undefined` to disable it. This is not possible with these changes, so the doc is updated to use `null` instead. The @vueuse/head package ignores null values so we need to cast it to an empty string.

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/2.features/4.head-management.md
+++ b/docs/content/2.guide/2.features/4.head-management.md
@@ -47,7 +47,7 @@ The `titleTemplate` can either be a string, where `%s` is replaced with the titl
 </script>
 ```
 
-Now, if you set the title to `My Page` with `useHead` on another page of your site, the title would appear as 'My Page - Site Title' in the browser tab. You could also pass `undefined` to default to the site title.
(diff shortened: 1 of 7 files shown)


pi0 (approved):

pi0:
Thanks!

PR #6308, opened 2022-08-02
fix(nuxt): generate `200.html` and `404.html` spa fallback files

pi0:
### ❓ Type of change

### 📚 Description

As a regression of https://github.com/unjs/nitro/issues/166, we were generating `dist/200/index.html` instead of `dist/200.html` for spa fallbacks. This makes expected output.

### 📝 Checklist


manniL:
This still works only for `ssr: false`, right?

pi0:
@manniL Yes 👍 We need  nitro route rules as discussed to to selectively set spa mode on these routes when generating with ssr enabled.

PR #6333, opened 2022-08-03
feat(nuxt): `app.config` with hmr and reactivity support

### LINKED ISSUE #5918: Editable runtime config with HMR
Currently, providing [runtimeconfig](https://v3.nuxtjs.org/guide/features/runtime-config) from `nuxt.config`, `.env` and environment variables is possible but DX is not great because for each change, we need to restart Nuxt instance and there is no browser HMR.

We can introduce new `app.config.[ext]` (2) file that basically extends `runtimeConfig.app` namespace but is directly loaded by bundler (vite/webpack/nitro) and accepts hot module replacements and exposed as `useAppConfig()` / `$appConfig` (1). This file can also have more JS native types (non POJO).

```js
import { defineAppConfig } from 'nuxt/app'

export default defineAppConfig ({
  title: 'My Website',
  seo: {
    ogImage: '/og/card.png'
  },
  theme: {
    primaryColor: '#ababab'
  }
})
```

**Typing:**

Same as RuntimeConfig, types can be inferred directly from config and extended from `NuxtAppConfig` interface from `@nuxt/schema` by modules and theme layers

**HMR support:**

Since the idea is to add a virtual template and dependency to the bundlers, HMR is expected to work out of the box. But we can make it even better to toggle reactivity for `useRuntimeConfig()` ref in app.

---

1. Discussing with @Tahul, alongside with this feature, we might also introduce injected `$appConfig` for easier usage and access within templates.

2. Alongside with this feature, discussing with and by experience @nuxt/content-module team, there is a very common pattern to have a _theme_ configuration. I personally vote for the simplicity of having one `app.config` and `$appConfig` interface with `theme` subkey. But we can also introduce `theme.config` and `useThemeConfig()`/`$themeConfig` namespaced shortcuts from app config.

pi0:
### 🔗 Linked issue

Closes #5918

### ❓ Type of change

### 📚 Description

This PR adds support for (multi layer) `app.config` file with true HMR support and reactivity/editable on both client and server side.

A diversion from the original proposal is that I've decided to keep it separated from runtimeConfig at least for now since runtime config is globally frozen on server side and injection method of appConfig is foundomantelay different (bundler vs nitro-provided). This means `appConfig` and `runtimeConfig.app` are not the same.

Playground is edited to try new feature.

TODO:
- [x] Working Vite HMR (*)
- [x] Test Webpack HMR 
- [x] ~~Test deep assign~~ (seems not required)
- [x] `appConfig` in `nuxt.config` for type support and easy module integration
- [x] Programmatic way to add additional `app.config` files (possible by modifying `appConfig` or `app:resolve` hook)
- [x] Add tests
- [x] ~~Inject `$appConfig`~~ (A plugin can easily inject reactive shortcut. Delegating addition to next steps)
- [x] Infer types from `app.config` for `useAppConfig()`
- [x] Add docs (API, Guide and File)
- [ ] Add module author guide (adding after trying on edge

(*) When testing, I've found that virtual files do not support HMR

### 📝 Checklist


packages/nuxt/src/core/templates.ts
```diff
@@ -176,6 +176,18 @@ export const useRuntimeConfig = () => window?.__NUXT__?.config || {}
 `
 }
 
+export const appConfigTemplate: NuxtTemplate = {
+  filename: 'app.config.mjs',
+  write: true,
+  getContents: ({ app }) => {
+    return `
+import defu from 'defu'
+${app.configs.map((id, index) => `import ${`cfg${index}`} from ${JSON.stringify(id)}`).join('\n')}
```
danielroe, line 185: Maybe use `genImport`?

packages/nuxt/src/app/config.ts
```diff
@@ -0,0 +1,36 @@
+import type { AppConfig, RuntimeConfig } from '@nuxt/schema'
+import { reactive } from 'vue'
+import { useNuxtApp } from './nuxt'
+// @ts-ignore
+import __appConfig from '#build/app.config.mjs'
+
+export const _appConfig = __appConfig as AppConfig
+
+export function useRuntimeConfig (): RuntimeConfig {
```
Atinux, line 9: deprecated I guess 👅

pi0: Haha. Probably too early but maybe at somepoint once we could cover all usages and also support nitro and environment variables to migrate to new appConfig. (also private runtime config is needed)

packages/schema/src/types/config.ts
```diff
@@ -29,10 +40,17 @@ export interface RuntimeConfig extends PrivateRuntimeConfig, RuntimeConfigNamesp
...
-  vue?: VuePluginOptions
+// -- App Config --
+
+// Reserved namespaces to avoid breaking changes if Nuxt starts to use them
+// type ReservedAppConfigKeys =
+//   'baseURL' | 'buildAssetsDir' | 'cdnURL' |
+//   'public' | 'private' | 'app'| 'config' | 'head' |
+//   'nuxt' | 'nitro' | 'internal' | 'runtime'
+
+type AppConfigNamespace = Record<string, any> // & Partial<Record<ReservedAppConfigKeys, never>>
```
pi0, line 51: @danielroe Do you know a way to prevent using these keys but also not show in IDE auto complete?

playground/app.config.ts
```diff
@@ -0,0 +1,13 @@
+// import { defineAppConfig } from '#imports'
```
pi0, line 1: /cc @antfu I've to also check on edge with external project but if you try this on playground, when a a file is imported from a virtual, auto imports and `#imports` do not work for it...

danielroe: We would have to at least emit a type to match the virtual import.

packages/schema/src/types/config.ts
danielroe: Let me see. To be clear, we _don't_ want them to show up in IDE auto-complete when entering in config, but we _do_ want them showing up when consuming the config, right?

playground/app.config.ts
pi0: No. Virtual import is internally used for `app/config.ts`. Issue is that, when a virtual file imports a user-space file, it can't use any auto imports (`defineAppConfig` becomes `undefined`)

packages/schema/src/types/config.ts
danielroe: I'll look at this. Worst case scenario, it auto completes but if they use it they get a type error and an explanation.

pi0: Behavior by enabling comments: (indeed it auto completes and then nags about not using them :D)

image

danielroe: I would suggest also deprecating, so it will draw a line through them in the dropdown. That will require explicit itemising of keys rather than just a key signature.

packages/nuxt/src/core/templates.ts
```diff
@@ -194,16 +193,24 @@ export const useRuntimeConfig = () => window?.__NUXT__?.config || {}
...
 const inlineConfig = ${JSON.stringify(nuxt.options.appConfig, null, 2)}
 
-${app.configs.map((id, index) => `import ${`cfg${index}`} from ${JSON.stringify(id)}`).join('\n')}
-export default defu(${['{}'].concat(app.configs.map((_id, index) => `cfg${index}`)).concat(['inlineConfig']).join(', ')})
+${app.configs.map((id, index) => `import ${`cfg${index}`} from ${JSON.stringify(id.replace(/(?<=\w)\.\w+$/g, ''))}`).join('\n')}
+const resolvedConfig = defu(${[].concat(app.configs.map((_id, index) => `cfg${index}`)).concat(['inlineConfig']).join(', ')})
+
+export default resolvedConfig
+
+type ResolvedConfig = typeof resolvedConfig
```
pi0, line 209: Writing template file was a temporary workaround for vite issue. I think would be better if we explicitly generate a dedicated `.d.ts` file for type support

danielroe: Perfect.

danielroe: Refactored. I'll let you excise the workaround!

test/fixtures/basic/app.config.ts
```diff
@@ -1,10 +1,12 @@
 // TODO
 // import { defineAppConfig } from '#imports'
-import type { AppConfig } from '@nuxt/schema'
 
-export default <AppConfig>({
+// For now we can't use AppConfig to type this
```
pi0, line 4: Shall we create _AppConfig to resolve this issue?

danielroe: Typing it as anything results in losing type inference. The solution is either not to type it, or use a define helper.

Atinux:
It would be nice to also test it in https://github.com/nuxt/framework/tree/main/examples/advanced/config-extends to check the multi layer support.

pi0:
@Atinux It is supported by default (main layer is part of other layers :))

packages/nuxt/src/app/config.ts
```diff
@@ -0,0 +1,53 @@
...
+  const nuxtApp = useNuxtApp()
+  if (!nuxtApp._appConfig) {
+    nuxtApp._appConfig = reactive(_appConfig) as AppConfig
+  }
+  return nuxtApp._appConfig
+}
+
+// HMR Support
+if (process.dev) {
+  function applyHMR (newConfig: AppConfigInput) {
```
danielroe, line 24: This should be the resolved config:

```suggestion
  function applyHMR (newConfig: AppConfig) {
```

danielroe: ... not that makes a huge difference as it's internal!

pi0:
HMR seems unstable with vite-node but is probably not directly related to this PR. We can test better on edge.

PR #6352, opened 2022-08-04
fix(nuxt): better equality check for json cookies

### LINKED ISSUE #5262: Cookie gets rewritten on read (using useCookie)
### Environment

- Builder:          `vite`
- User Config:      `css`, `build`, `buildModules`, `vite`, `typescript`
- Runtime Modules:  `-`

### Reproduction

`useCookie("x-abc").value`

Cookie value: `{"accessToken":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"}`

Make request with the following cookie (it must be valid JWT token)

### Describe the bug

Once useCookie recieves valid JWT cookie Nuxt decides to overwrite it on cookie read (and adds set-cookie header in response). If cookie value is a simple string bug is not reproduced

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #5262

### ❓ Type of change

### 📚 Description

In case of JSON cookies, we can perform a basic JSON.stringify equality test. I think we're not worried with deep equality or order because we are really testing to see if a write-operation has been performed on the cookie.

### 📝 Checklist


pi0 (approved):

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -72,3 +72,12 @@ function writeServerCookie (event: CompatibilityEvent, name: string, value: any,
     appendHeader(event, 'Set-Cookie', serializeCookie(name, value, opts))
   }
 }
+
+function isEqualValue (val1: unknown, val2: unknown) {
+  try {
+    if (typeof val1 === 'object' && typeof val2 === 'object') {
```
pi0, line 78: What happens if type of both is not object? I think in that case we need to fallback to eqeqeq check (numbers for example)

DamianGlowala: What about https://lodash.com/docs/4.17.15#isEqual?

pi0: Definitely not loadash in 2022  😆 BTW you made me think, we could use [ohash.objectHash](https://github.com/unjs/ohash#objecthashobject-options) which does deep and stable serialization.

danielroe: If type of either is not an object it does  back to eqeqeq check.

pi0: https://github.com/unjs/ohash#isequalobj1-obj2-options

Issue #6087, opened 2022-07-23
Missing "./dist/app/entry" export in "nuxt3" package 


aligzl:
### Environment

Operating System: Windows 11
Node Version: v16.15.1
Nuxt Version:3.0.0-rc.6-27641313.416f98b
Package Manager: yarn@1.22.19

Runtime Modules: -
Build Modules: - pinia tailwindcss

### Reproduction

I cant provide Reproduction this happens only windows

### Describe the bug

Cant start the nuxt after rc-6 on windows os

### Logs

```shell
ERROR  Missing "./dist/app/entry" export in "nuxt3" package                                                                              16:24:22

  at bail (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:33603:8)
  at resolve (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:33680:10)
  at resolveExports (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:34417:12)
  at resolveDeepImport (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:34435:31)
  at tryNodeResolve (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:34160:20)
  at Context.resolveId (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:33897:32)
  at Object.resolveId (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:35209:55)
  at async Context.resolve (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:34974:23)
  at async Object.resolveId (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:35209:32)
  at async ModuleGraph.resolveUrl (/D:/git/honeymoon/honeymoon-frontend/node_modules/vite/dist/node/chunks/dep-1513d487.js:57784:26)


 WARN  [SSR] Error transforming D:/git/honeymoon/honeymoon-frontend/node_modules/nuxt/dist/app/entry: Missing "./dist/app/entry" export in "nuxt3" p
ackage

[...]

PR #6355, opened 2022-08-04
fix(vite): use url for entry on windows

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/6087

### ❓ Type of change

### 📚 Description

This resolves the entry path via a file URL on Windows to skip vite trying to resolve it as an import from `nuxt3` directly.

### 📝 Checklist


packages/vite/src/vite.ts
```diff
@@ -84,6 +86,10 @@ export async function bundle (nuxt: Nuxt) {
     )
   }
 
+  if (isWindows) {
+    ctx.config.resolve.alias['/entry.mjs'] = pathToFileURL(ctx.config.resolve.alias['/entry.mjs']).href
```
pi0, line 90: Any reason not doing this by default for all platforms?

packages/vite/src/dev-bundler.ts
```diff
@@ -40,21 +40,17 @@ async function transformRequest (opts: TransformOptions, id: string) {
...
+  if (id && !id.startsWith('/@fs/') && id.startsWith('/')) {
     // Relative to the root directory
     const resolvedPath = resolve(opts.viteServer.config.root, '.' + id)
     if (existsSync(resolvedPath)) {
       id = resolvedPath
     }
   }
 
+  // On Windows, we prefix absolute paths with `/@fs/` to skip node resolution algorithm
+  id = id.replace(/^\/?(?=\w:)/, '/@fs/')
```
pi0, line 52: - Is there an upstream issue in vite to track or it is related to internal use?
- Can we consistantly add `/@fs` prefix always?

danielroe: We can always add `/@fs/` for absolute URLs, but some incoming requests _look_ like absolute URLs but are meant to be relative to the root directory, which is why I have only added it for drive letters/windows.

I'm not aware of an upstream vite issue. I can try to create reproduction but suspect the issue is our dev-bundler usage. (No fix was needed for vite-node.)

pi0: Okay thanks for explaining. Then let's keep as workaround and iterate over. We need to verify it on edge before release anyway :)

pi0 (approved):
<3

PR #6368, opened 2022-08-04
refactor(nuxt): enable strict type checking

antfu:
### ❓ Type of change

### 📚 Description

The first step migrating the entire repo to `strict: true`

- Added temporary `tsconfig.json` to mark the scope of strict - later after we migrate all modules to strict, we could then remove all the sub tsconfig and turn the root level `strict: true`
- `wrapInRef` is redundant - ref already does that: https://github.com/vuejs/core/blob/4d7803ed28fb67d45a83d3500f5407754e65bf64/packages/reactivity/src/ref.ts#L90-L95
- Not only would the strictness help us internally, but also for the generated types consumed by users. Previously the return type of `AsyncData` does not indicate the data could possibly be `null`, this PR also catches that.

### 📝 Checklist


packages/nuxt/src/app/entry.ts
```diff
@@ -35,7 +35,7 @@ if (process.server) {
       await nuxt.hooks.callHook('app:created', vueApp)
     } catch (err) {
       await nuxt.callHook('app:error', err)
-      ssrContext.error = ssrContext.error || err
+      ssrContext!.error ||= err
```
pi0, line 38: Node 14 is not supported. Let's delay using this syntax.

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -176,15 +176,21 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
           // converts `""` to `null` to prevent the attribute from being added as empty (`rel=""`)
           : firstNonUndefined<string | null>(props.rel, options.externalRelAttribute, href ? DEFAULT_EXTERNAL_REL_ATTRIBUTE : '') || null
 
-        const navigate = () => navigateTo(href, { replace: props.replace })
+        const navigate = () => {
+          if (href) {
+            navigateTo(href, { replace: props.replace })
```
pi0, line 181: `navigateTo` returns a value (Promise) in some cases. We need to return it.

packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -176,15 +176,21 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
         if (props.custom) {
-          if (!slots.default) { return null }
+          if (!slots.default) {
+            return null
+          }
           return slots.default({
             href,
             navigate,
-            route: router.resolve(href),
+            route: href ? router.resolve(href) : undefined,
```
pi0, line 193: Can you please double check for behavior of `routerResolve(undefined)`? Does it returns undefined?

antfu: This part confuses me, it seems `href` could be `null` while those functions do not seem to support them. I will check

antfu: image

Just tried, `router.resolve(null)` will result in this error. Probably a potential bug that we never hit. What do you think the best approach for this?

packages/nuxt/src/core/app.ts
```diff
@@ -77,7 +81,8 @@ export async function resolveApp (nuxt: Nuxt, app: NuxtApp) {
 
   // Resolve layouts/ from all config layers
   app.layouts = {}
-  for (const config of nuxt.options._layers.map(layer => layer.config)) {
+  for (const config of getLayerConfigs(nuxt)) {
+    if (!config.srcDir) { continue }
```
antfu, line 85: I am not sure about this. Currently, both of these are optional on type:

```ts
nuxt.options._layers.map(layer => layer.config?.srcDir?.startsWith())
//                                       ^       ^
```

While the usage seems to assume them to be not nullable. Is there a missing logic for the runtime, or is there a misalignment in the type?

pi0: You can assume both exist 👍🏼  (we filter layers in nuxi check code)

pi0: I think we should fix the type that this prop exists (btw pleas double check code i'm not 100% sure)

pi0: Just checked the code. We make sure layers have srcDir: https://github.com/nuxt/framework/blob/cb98c8b9216cb731dc90dcc59195db4d8c244374/packages/kit/src/loader/config.ts#L27

pi0 (approved):

pi0:
This is a nice initiative 🚀 Merging to avoid stalling PR. There has been some leftovers also in packages/nuxt which we can improve type coverage and fix errors.

I've triple checked but there might be regressions with changes. /cc @danielroe heads up

@antfu Can we please allow unplugin define to make input options mandatory? It can help to properly type them. 🙏🏼

antfu:
@pi0 Published `unplugin@0.9.1` support required options

PR #6389, opened 2022-08-05
fix(nuxt): use payload error state as source of truth

### LINKED ISSUE #6386: Client-only plugins swallow errors with `ssr: true`
### Environment

- Operating System: `Linux`
- Builder:          `vite`
- User Config:      `ssr`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-dmtnav?file=app.vue

### Describe the bug

A client-only plugin silently breaks app when running `ssr: true`.

- server plugin + ssr: true = verbose error
- client plugin + ssr: false = verbose error
- client plugin + ssr: true = silent error

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #6386

### ❓ Type of change

### 📚 Description

This resolves an inconsistency caused by the fact that if there was no error on server render, we completely ignore `payload.error` on client. Instead, we can use this as the source of truth instead of duplicating the error into the state as well.

### 📝 Checklist


packages/nuxt/src/app/composables/error.ts
```diff
@@ -1,9 +1,16 @@
 import { createError as _createError, H3Error } from 'h3'
-import { useNuxtApp, useState } from '#app'
+import { computed } from 'vue'
+import { useNuxtApp } from '#app'
 
 export const useError = () => {
   const nuxtApp = useNuxtApp()
-  return useState('error', () => process.server ? nuxtApp.ssrContext.error : nuxtApp.payload.error)
+  const source = process.server ? nuxtApp.ssrContext : nuxtApp.payload
```
pi0, line 7: It seems there would be still places of inconsistency (payload/ssrContext). Can you please track what places reference to `ssrContext.error`? They could migrate to `ssrContext.payload.error` directly and `useError` using `toRef(payload, 'error')`

danielroe: We do need to keep ssrContext.error to distinguish a freshly thrown error from an error being handled with `error.vue`. But using `toRef` is spot on 👌

pi0: Can we maybe distinguish with a `ssrError` flag? Two sources with same name seems confusing and error-prune.

danielroe: Well, the payload gets constructed later on. Passing in a pre-filled payload.error is a little bit bigger change. Do you still want to proceed with `ssrError` flag + `ssrContext.payload.error`? Happy to implement.

pi0: I think makes sense to be with the same change but happy to do it in another PR.

danielroe: Up to you. I will do it, but it's possible I won't get around to it until tomorrow. Feel free to merge this in the mean time if you want.

packages/nuxt/src/app/nuxt.ts
```diff
@@ -70,6 +71,14 @@ interface _NuxtApp {
     data?: Record<string, any>
     state?: Record<string, any>
     rendered?: Function
+    error?: Error | {
+      url: string
+      statusCode: string
+      statusMessage: string
+      message: string
+      description: string
```
pi0, line 79: Isn't `description` optional? (Also don't we have an exported type for object error already? 🙈 )

pi0 (approved):
LGTM overall.

packages/nuxt/src/app/nuxt.ts
danielroe: We are always setting the value of description: https://github.com/nuxt/framework/blob/3a822c71772ceb22110e92af9379bacb8634900a/packages/nuxt/src/core/runtime/nitro/error.ts#L16-L18

I'm not sure what you mean by exported type, but if it's `ParsedError`, that is not exported and also differs somewhat from this error object:

https://github.com/unjs/nitropack/blob/cb6a84c37fc1e4686b1e98a7320e44b1d7ed90d9/src/runtime/error.ts#L7-L13

Issue #6411, opened 2022-08-07
Payload extraction support


pi0:
In Nuxt 2, we introduced a major enhancement to static generation referred to as [Full Static Generation](https://nuxtjs.org/announcements/going-full-static/) inspired by [DreaMinder/nuxt-payload-extractor](https://github.com/DreaMinder/nuxt-payload-extractor). You can read more about the history of initial decisions in nuxt/rfcs#22 which finally landed by https://github.com/nuxt/nuxt.js/pull/6159 in 2.13.0 and being available by opt-in option `target: 'static'`. 

In Nuxt 3, thanks to nitro we can move payload extraction to the next level to not only support it for a static generation but also the hybrid static mode and even prefetching payload of server-rendered components. We initially tested this enhanced functionality using Nuxt bridge ([src](https://github.com/nuxt/bridge/blob/53a1a03b218a40e6f35bcb14d96bdc6bf03ef99b/src/runtime/nitro/renderer.ts#L196)) and @danielroe made a POC in [danielroe/nuxt-full-static](https://github.com/danielroe/nuxt-full-static) for Nuxt 3 userland implementation.

Using this issue to track the progress of the core implementation roadmap.

- [x] Initial POCs (bridge and nuxt3)
- [ ] Add payload chunk endpoint and JSONP utils to Nuxt3
- [ ] Nitro prerender hints for payload extraction
- [ ] Restore payload on navigation with full static enabled
  - https://github.com/danielroe/nuxt-full-static/blob/main/src/runtime/plugin.client.ts
- [ ] Integrate with NuxtLink to prefetch payload of next pages
  - https://github.com/nuxt/framework/pull/4329
- [ ] Deep integration with `useSyncData` and `useState` for cache invalidation
- [ ] Avoid double pre-rendering for page+payload (unique to nuxt 3 design)

Next steps:

There are still unresolved challenges we have left from Nuxt 2. It is worth listing them and possibly addressing them.

- Initial state (useStore and useState) can be differentiated to only send payload changes
- Static routes manifest (can be supported via universal route caching rules)
- Version change detection on new deployment to reload the page

gkweb:
@danielroe @pi0 - Is this something that will be added to bridge?

pi0:
[...]

PR #6455, opened 2022-08-09
feat(nuxt): payload rendering support

pi0:
### 🔗 Linked issue

#6411

### ❓ Type of change

### 📚 Description

This PR:

- Adds new extension to renderer that by adding `/_payload.js` to the end of the URL, returning only payload chunk
- Adds `loadPayload(url, force?)` composable
  - Client-side: Uses [browser dynamic import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) to import and cache payload
  - Server-side: No-op and returns null 
- Adds `prefetchPayload(url)` composable
  - Server-side: Add link with [`modulepreload` rel](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/modulepreload)
  - Client-side: ~~calls `usePayload(url)` to fill in the cache~~ Same as server adds `modulepreload` to trigger browser prefetching in background
- When using `nuxt generate` or `prefetchPayload(url)`, it detect rendering mode based on headers and sets the` x-nitro-prerender` hint to prerender `/[route]/_payload.js` file
- Add `prerenderedAt` to the payload when using prerendering or nuxt generate
- Add `isPrerendered` composable (check active payload)

Testing:
- `yarn nuxt build test/fixtures/basic`
- `node test/fixtures/basic/.output/server/index.mjs`
- http://localhost:3000/random/a

Todo:
- [x] Cache prerender payload to avoid duplicate render
- [x] Replace payload from prerendered links to an external file
- [x] Add prerender header for prerendered pages
- [x] Load payload on navigation
- [x] Payload filtering for unassigned keys
- [x] Assign state on navigation
- [x] Run basic tests 

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -101,10 +101,20 @@ const getSPARenderer = lazyCachedFunction(async () => {
...
-  const url = ssrError?.url as string || event.req.url!
+  let url = ssrError?.url as string || event.req.url!
+
+  // Whether we are rendering payload response
+  let payloadReq = false
+  if (url.endsWith(PAYLOAD_FILE_NAME)) {
+    url = url.substring(0, url.length - PAYLOAD_FILE_NAME.length) || '/'
+    event.req.url = url
+    payloadReq = true
+  }
```
danielroe, line 117: Loving it!

Would be nice to allow a way to hook into this logic so a module could render other formats of response. (Such as, for example, web snapshots, JSON, or something else.) Or so we could conditionally enable/disable this behaviour if user did not have a use-case for JSONP.

In other words, making renderer more composable.

We could also extract ssr error rendering in almost exactly same kind of way (as a different render format) which would then allow us to let users take full control over the error handling (e.g. customising JSON response, or headers of HTML response).

pi0: We have `render:response` hook to fully customize it. But would be also nice idea, later on, refactor the renderer into smaller utils.

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -133,6 +143,20 @@ export default defineRenderHandler(async (event) => {
     throw ssrContext.payload.error
   }
 
+  // Payload render
+  if (payloadReq) {
+    const response: RenderResponse = {
+      body: `__NUXT_JSONP__("${url}", ${devalue(ssrContext.payload)})`,
```
danielroe, line 149: Do you mind if I raise the question of why we are rendering JSONP?

* if it's about bypassing same origin policies, then I think that's not relevant now because most of our scripts are loaded via fetch and fall foul of the same potential policies
* possibly size? devalue might be able to render slightly smaller payloads?

I also have a small question about whether JSON might be slightly more performant and require less in the client-side code (in terms of JSONP parsing, for example).

pi0: There is a LONG discussion you can find in the linked issue from Nuxt 2. But let's also discuss it over our call :)

packages/nuxt/src/app/composables/payload.ts
```diff
@@ -0,0 +1,47 @@
...
+  const parsed = parseURL(url)
+  if (parsed.search) {
+    throw new Error('Payload URL cannot contain search params: ' + url)
+  }
+  return joinURL(parsed.pathname, '_payload.js')
+}
+
+async function _importPayload (payloadURL: string) {
+  if (process.server) { return null }
+  const { default: payload } = await import(/* @vite-ignore */ payloadURL) as { default: any }
```
Atinux, line 45: We may want to handle the error and return null to avoid any crash?

packages/nuxt/src/app/composables/utils.ts
```diff
@@ -1,3 +1,10 @@
 import { isRef, ref, Ref } from 'vue'
+import { useNuxtApp } from '#app'
 
 export const wrapInRef = <T> (value: T | Ref<T>) => isRef(value) ? value : ref(value)
+
+export function isPrerender () {
```
Atinux, line 6: isPrerendered?

packages/nuxt/src/app/composables/payload.ts
pi0: Normaly this shouldn't happen because unlike Nuxt 2, we actually can **render** payload so even if it is not pre-rendered, it is expected import works.

packages/schema/src/config/router.ts
```diff
@@ -154,7 +154,7 @@ export default defineUntypedSchema({
    * linked page when the `<nuxt-link>` is visible in the viewport, making instant navigation.
    * @version 2
    */
-  prefetchPayloads: true,
+   prefetchPayloads: true,
```
danielroe, line 157:
```suggestion
  prefetchPayloads: true,
```

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -233,3 +277,23 @@ async function renderInlineStyles (usedModules: Set<string> | string[]) {
...
 }
+
+function renderPayloadResponse (ssrContext: NuxtSSRContext) {
+  return <RenderResponse> {
+    body: `export default ${devalue(splitPayload(ssrContext).payload)}`,
+    statusCode: ssrContext.event.res.statusCode,
+    statusMessage: ssrContext.event.res.statusMessage,
+    headers: {
+      'Content-Type': 'text/javascript;charset=UTF-8',
+      'X-Powered-By': 'Nuxt'
```
danielroe, line 288:
```suggestion
      'content-type': 'text/javascript;charset=UTF-8',
      'x-powered-by': 'Nuxt'
```

danielroe (commented):
❤️ This looks great! Would be nice to see some tests as well.

test/fixtures/basic/pages/random/[id].vue
```diff
@@ -0,0 +1,49 @@
...
+    </button>
+  </div>
+</template>
+
+<script setup lang="ts">
+const route = useRoute()
+
+const pageKey = 'rand_' + route.params.id
+
+const { data: randomNumbers, refresh } = await useFetch('/api/random', { key: pageKey as string })
```
Comment (danielroe) [inline test/fixtures/basic/pages/random/[id].vue:38]: This seems to be making API call to `/api/random` at runtime _as well as_ request for payload - is that intended?

Comment (pi0) [reply test/fixtures/basic/pages/random/[id].vue:38]: Strange. Will double check.

danielroe:
Mind if I push a test suite later this evening before merging?

test/fixtures/basic/pages/random/[id].vue
Comment (danielroe) [reply test/fixtures/basic/pages/random/[id].vue:38]: ~> https://github.com/nuxt/framework/pull/6455/commits/bb00934ee30b2a5c43ae3363f4efe618cf006eee

danielroe (approved):
Looks good to me!

Some suggestions for follow-up PRs:

* enabling static in _dev_ mode without the method I used in tests - user may have an API that can't be called from browser or relies on server-only api keys
* hybrid mode (as you mention)
* `process.env.prerender` - presumably we'll set this somehow in nitropack whilst prerendering?
* would be nice to deduplicate payloads - e.g. if you hard load `/random/a` then it should 'fill' the payload and not redownload it when navigating from a -> b -> a.
* [another note: would be nice to have support for testing generated or hybrid sites in test-utils - cc: @antfu]

pi0:
Thanks for helping to add tests @danielroe ❤️ Regarding your notes:

> enabling static in dev mode without the method I used in tests - user may have an API that can't be called from browser or relies on server-only api keys

Later, with route rules, we could predict which routes are prerendered/static to adjust behavior. But note that payload rendering is not only for static mode (ie: server API keys available during prerendering phase. that should work in production too and with API fallback)

> process.env.prerender - presumably we'll set this somehow in nitropack whilst prerendering?

It is already added and released from nitropack which I've used. If it doesn't work, not only is a performance issue but a **security issue** because an attacker could fill the memory of server by abusing payload cache. I plan to improve it later with a shared LRU.

> would be nice to deduplicate payloads - e.g. if you hard load /random/a then it should 'fill' the payload and not redownload it when navigating from a -> b -> a.

Not quite sure to understand this. Payload state is fresh per page and payload URL is unique. We don't download by with but use dynamic import. Browsers have caching mechanism for esm imports and don't redownload. (unless using new hash to force download a new payload)

> would be nice to have support for testing generated or hybrid sites in test-utils

Do you mind to open an issue to elaborate more about your idea about possible utils?

test/basic.test.ts
```diff
@@ -586,6 +586,55 @@ describe('app config', () => {
...
+    const page = await createPage()
+    const requests = [] as string[]
+    page.on('request', (req) => {
+      requests.push(req.url().replace(url('/'), '/'))
+    })
+    await page.goto(url('/random/a'))
+    await page.waitForLoadState('networkidle')
+
+    // TODO: We should not prefetch the payload we are displaying
+    // expect(requests).not.toContain('/random/a/_payload.js')
```
pi0, line 607: @danielroe  This is expected behavior. We had been doing this in Nuxt 2 as well as Nuxt 3 to _preload_ scripts of the same page in the head loading them parallel to page loading. Payload of the page is necessary for hydration.

danielroe: Is it also present in the HTML?

pi0: That's the whole point to hint the browser preload script before running scripts :)

danielroe: My point is that if `process.env.prerender` is not set, then we have the payload present in the HTML:

```html
<script>window.__NUXT__={data:{$frand_a:[5312,4359,1601,5527,5942,4502,2971,9429,8146,1745]},state:{"$srandom:rand_a":70,"$srandom:default":95},_errors:{},serverRendered:true,config:{public:{testConfig:123},app:{baseURL:"\u002F",buildAssetsDir:"\u002F_nuxt\u002F",cdnURL:""}},prerenderedAt:1662805411150}</script>
```

I am aware that when we are prerendering the pages then it is split out and needs to be in the requests list. But in this test case (ie. server-rendered) we have a _duplicate_ request to `/random/a/_payload.js`.

pi0: I _guess_ this is mutually exclusive. SSR: Full payload and no `_payload.js` hint | Prerender: `_payload.js`. They both don't happen at the same time in same page... But yeah in testing it is confusing because we pretender but fetch SSR

danielroe: Makes sense - it's a build vs dev difference in the test suite. (It's not a bug in the implementation.) Would be worth updating test with additional assertions to ensure correct behaviour. I can do it later, either before or after this merges (but not today).

PR #6487, opened 2022-08-10
refactor(nuxt)!: remove `null` handling for `titleTemplate`

danielroe:
### 🔗 Linked issue

follow on from #6296

### ❓ Type of change

### 📚 Description

It seems we have a little leftover `titleTemplate` handling. I think we don't need this - was there a reason we left it in?

### 👉 Migration

If you were previously using `titleTemplate: null` you should update this to `titleTemplate: ''`.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/head/runtime/lib/vueuse-head.plugin.ts
+++ b/packages/nuxt/src/head/runtime/lib/vueuse-head.plugin.ts
@@ -15,20 +15,10 @@ export default defineNuxtPlugin((nuxtApp) => {
     headReady = true
   })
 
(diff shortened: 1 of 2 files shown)


pi0:
Have you checked notes in https://github.com/nuxt/framework/pull/6296#issue-1325689204? I belive it was to avoid a breaking change for null handling

danielroe:
Yes, I spotted that. I think an acceptable change for RC7 if you agree, but we'll need to mention in release notes.

pi0 (approved):

PR #6042, opened 2022-07-21
feat(nuxt): `app:rendered` and `app:response` hooks

pi0:
### 🔗 Linked issue

Resolves #5534 (compatible with #5565)

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR adds new `nuxt:app:rendered` and `nuxt:app:response` nitro hooks and also provides same render context to the existing `app:rendered` hook usable in Nuxt plugins.

A small additional charge, `X-Powered-By: Nuxt` header is now added to the default response. It can be disabled with nitro plugin and `nuxt:app:response` hook.

Introducing external types for nitro in subsequent PR ~> #6170

**Example usage:** (nitro server plugin)

```js
export default defineNitroPlugin((nitroApp) => {
  nitroApp.hooks.hook('nuxt:app:rendered', ({ ssrContext, html }) => { console.log({ html }) })
  nitroApp.hooks.hook('nuxt:app:response', ({ response }) => { console.log(response) })
})
```

**Example usage:** (nuxt plugin)

```js
export default defineNuxtPlugin((nuxtApp) => {
  nuxtApp.hook('app:rendered', ({ ssrContext, html }) => {
    console.log({ html })
  })
})
```

Example output:

```js
{
  html: {
    htmlAttrs: [ 'data-head-attrs=""' ],
    head: [
      '<title></title><meta charset="utf-8"><meta name="viewport" content="width=device-width, initial-scale=1"><meta name="head:count" content="2">',
[...]

PR #6521, opened 2022-08-11
fix(nuxt)!: allow `app:rendered` to modify ssr context and add `render:html` 

### LINKED ISSUE #6518: ssrContext effectively read-only in app:rendered hook 3.0.0-rc.7
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0-rc.7`
- Package Manager:  `npm@8.13.2`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-2wanhk?file=plugins/test.client.ts

### Describe the bug

Not sure if this is a bug or if it is intentional. However, in the previous version of the app:rendered hook the ssrContext was mutable.

Using the app:rendered hook you can access the HTML including ssrContext, however, you cannot mutate its value. It seems like the HTML is already generated at this point, and no longer mutable.

`nuxtApp.hook('app:rendered', ({ssrContext}) => {
    ssrContext.payload.state.test = 'test';
  });`

`console.log(window.__NUXT__.state.test) -> undefined`

### Additional context

_No response_

### Logs

_No response_

pi0:
### 🔗 Linked issue

Resolves #6518

### ❓ Type of change

- [x 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

We were calling `app:rendered` hook just after ssr rendering is being done in Nuxt 3 <= 3.0.0-rc.6. #6042 changed this behavior in v3.0.0-rc.7 that `app:rendered` is being called with new `rendered` context to have access to constructed html but this means, it is too late to modify payload. 

This PR reverts `app:rendered` back to previous behavior with new limited `{ ssrContext }` context only. (breaking change)

Nitro plugins have access to the final rendered context via `render:html` (breaking change: `nuxt:app:rendered` renamed for nitro plugins to be simpler, consistent with `render:response` hook and not confused with `app:rendered` lifecycle hook)

### 📝 Checklist


danielroe (approved):

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -137,42 +135,38 @@ export default defineRenderHandler(async (event) => {
   const renderedMeta = await ssrContext.renderMeta?.() ?? {}
 
   // Create render context
-  const rendered: NuxtRenderContext = {
-    ssrContext,
```
danielroe, line 141: It feels like `ssrContext` would be quite a useful thing to have in this hook. We could mark it in the types as Readonly if that is the main issue.

pi0: Keeping ssrContext at this point can be source of bad assumptions and possible breaking changes in the future because we already _consumed_ it. I've added `{ event }` in last commit. `event.context` could be used to pass any context from app if needed.

test/fixtures/basic/nuxt.d.ts
```diff
@@ -0,0 +1,8 @@
+// This file is auto generated by `nuxt prepare`
+// Please do not manually modify this file.
+
+/// <reference path=".nuxt-node/nitro.d.ts" />
+/// <reference types="nuxt3" />
+/// <reference path=".nuxt-node/components.d.ts" />
+/// <reference path=".nuxt-node/auto-imports.d.ts" />
+export {}
```
danielroe, line 8: this should likely be gitignored.

packages/nuxt/src/core/runtime/nitro/renderer.ts
pi0: Other benefits: At allows garbage collecting ssrContext at this stage by not depending on it and in the future, could be standarized interface in nitro for other integrations depending on html modification.

test/fixtures/basic/nuxt.d.ts
pi0: Haha good catch :D

packages/nuxt/src/core/runtime/nitro/renderer.ts
danielroe: I don't think the perf benefit of moving GC up/down a few lines makes much difference. More, that things like payload/state/used components/error information are all accessible on `ssrContext`. Even if the rendering of the Vue app is done, the rendering of the response hasn't yet finished. Users could be forced to awkward workaround of setting ssrContext data on `event` in previous hook in order to consume it in this one.

Marking it as readonly in hook would give a signal to users users that modifying it doesn't have any effect.

pi0: Rendering of the response is not entirely within vue renderer. And no one should set ssrContext to event context (this is kinda point of not passing it!)

Do you think of any use case right now needs direct access to ssrContext here? (BTW we could always add it to second argument bag `{ event, ssrContext }` if needed)

matthewhardern: An excellent use case is with web component libraries like lit.dev as it would allow getting the state object to the lit element renderer and then returning the modified HTML as part of the `render:html` hook

danielroe: That's very helpful, thanks. Would you open a new issue or discussion so we can track, maybe with some example code? I imagine this could get lost in the comments of a merged PR.

PR #6546, opened 2022-08-12
fix(vite, nuxt): resolve relative to `srcDir` rather than `rootDir`

danielroe:
### ❓ Type of change

### 📚 Description

With a custom `srcDir`, various things were not working, including scanning of new components.

This also makes a couple of changes to vite-node, to match the non-vite node implementation where `root` is set to `srcDir` and not `rootDir` but would value @antfu's verdict on these.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/runtime/vite-node.mjs
+++ b/packages/vite/src/runtime/vite-node.mjs
@@ -7,7 +7,7 @@ import { getViteNodeOptions } from './vite-node-shared.mjs'
 const viteNodeOptions = getViteNodeOptions()
 
 const runner = new ViteNodeRunner({
-  root: viteNodeOptions.rootDir,
+  root: viteNodeOptions.root,
   base: viteNodeOptions.base,
   async fetchModule (id) {
     return await $fetch('/module/' + encodeURI(id), {
--- a/packages/vite/src/runtime/vite-node.mjs
+++ b/packages/vite/src/runtime/vite-node.mjs
@@ -7,7 +7,7 @@ import { getViteNodeOptions } from './vite-node-shared.mjs'
 const viteNodeOptions = getViteNodeOptions()
 
 const runner = new ViteNodeRunner({
-  root: viteNodeOptions.root,
+  root: viteNodeOptions.root, // Equals to Nuxt `srcDir`
   base: viteNodeOptions.base,
   async fetchModule (id) {
     return await $fetch('/module/' + encodeURI(id), {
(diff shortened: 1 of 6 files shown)


packages/vite/src/vite-node.ts
```diff
@@ -143,7 +143,7 @@ export async function initViteNodeServer (ctx: ViteBuildContext) {
   // Serialize and pass vite-node runtime options
   const viteNodeServerOptions = {
     baseURL: `${protocol}://${host}:${port}/__nuxt_vite_node__`,
-    rootDir: ctx.nuxt.options.rootDir,
+    root: ctx.nuxt.options.srcDir,
```
antfu, line 146: I guess would also need to set the root for the Vite server to ensure the path resolution's consistency

https://github.com/nuxt/framework/blob/b0bf25cf51e390ffbc15bdac1a895b8a433f5dd8/packages/vite/src/vite.ts#L34

Currently we seems to use the implicit `process.cwd()`

antfu:
```suggestion
    root: ctx.ssrServer.config.base.root,
```

After that we might reuse the root option from the vite server dierectly

danielroe: Interesting. I think we are setting it here, right?

https://github.com/nuxt/framework/blob/e55938fe6ec8e71800a1c8bc6842a379183556f0/packages/schema/src/config/vite.ts#L14-L17

pi0 (approved):

packages/vite/src/vite-node.ts
antfu: Oh yeah, then it totally makes sense to use `srcDir` 👍

pi0: They are safe references. I think it is better to have single source of trust `srcDir` referencing everywhere.

danielroe: Okay! 👍

packages/vite/src/runtime/vite-node.mjs
```diff
@@ -7,7 +7,7 @@ import { getViteNodeOptions } from './vite-node-shared.mjs'
 const viteNodeOptions = getViteNodeOptions()
 
 const runner = new ViteNodeRunner({
-  root: viteNodeOptions.rootDir,
+  root: viteNodeOptions.root,
```
pi0, line 10:
```suggestion
  root: viteNodeOptions.root, // Equal to Nuxt `srcDir`
```

PR #6595, opened 2022-08-13
fix(nuxi): update `server.port` and `server.host` with listener info

pi0:
### ❓ Type of change

### 📚 Description

Currently, vite-node depends on `server.port` for listener. It works for default port and also when using `PORT` environment variable. But in case of random or fallback port, it fails. This PR updates those from real value in listener

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -1,3 +1,4 @@
+import type { AddressInfo } from 'node:net'
 import { resolve, relative, normalize } from 'pathe'
 import chokidar from 'chokidar'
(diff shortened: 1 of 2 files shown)


danielroe (approved):
Not tested locally yet but seems like a good hotfix. ❤️

Later, might be nice to have some other way of getting access to dev server and dev server options without mutating Nuxt config.  Possibly as part of a dev event bus that could be used for dev error handling, etc?

pi0:
> Possibly as part of a dev event bus that could be used for dev error handling, etc?

It is also accessible using `listen` hook to modules and in dev.

Issue #6365, opened 2022-08-04
Duplicate response when calling a composable with useLazyFetch inside


lucassimines:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-rc.6`
- Package Manager:  `npm@8.8.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://stackblitz.com/edit/github-16xwyo?file=app.vue,composables%2FuseApiFetch.ts

### Describe the bug

The bug is that, when using `useLazyFetch` inside a composable, and then call multiple different requests using the `custom function`, it returns the same data as the first request.

It is not generating an unique `key`

aerksn:
any workaround for this one?

manniL:
> any workaround for this one?

The workaround is to set the key manually for now (different for each usage where state shouldn't be shared)

aerksn:
> > any workaround for this one?
> 
> The workaround is to set the key manually for now (different for each usage where state shouldn't be shared)

cool, thanks for quick reply! :)

Issue #6629, opened 2022-08-15
support computed keys for data fetching composables


danielroe:
There are a number of cases where users want to pull data in (often based on route parameters) and dynamically update this data as things change. (For example, https://github.com/nuxt/framework/issues/6604, https://github.com/nuxt/framework/issues/5993, https://github.com/nuxt/framework/issues/6361.)

It would be useful to add support for computed keys in `useFetch` and `useAsyncData` so that the original data fetched isn't overwritten but rather a new key/asyncdata value is accessed based on the computed key. This would also have the effect of meaning that `initialCache` would not have to be set to `false` to enable updates on route navigation.

pi0:
One major issue with supporting computed keys, is that we might endup with (mostly client-side) memory leaks when input dynamically changes. There might be better ways to overcome with limitations of dynamic route keys (like using a subkey). Let's wait for it until fixing asyncData shared state.

PR #6632, opened 2022-08-15
fix(nuxt)!: include request url and params in `useFetch` key

### LINKED ISSUE #6361: Page Route not refresh the template request data
### Environment

Nuxt CLI v3.0.0-rc.6 
- Builder:          `vite`
- User Config:      `components`, `modules`, `css`, `plugins`
- Runtime Modules:  `@nuxtjs/tailwindcss@5.0.4`, `@nuxtjs/color-mode@3.0.3`, `nuxt-lodash@2.2.0`
- Build Modules:    `-`

### Reproduction

npm run dev

[stackblitz link](https://stackblitz.com/edit/nuxt-starter-uwg9ug)

### Describe the bug

Page Route not refresh the template request data
When I link to jump to another page
 I don't have this problem when I use version 3.0.0-rc.1
Upgrade to 3.0.0-rc.6 There will be this problem
useFetch request { data }  data  not refresh (Always the first requested data)
Is the method changed?
I use usefetch to request

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #6361, resolves https://github.com/nuxt/framework/issues/6365

### ❓ Type of change

### 📚 Description

A regression of https://github.com/nuxt/framework/pull/4955 - we were not using the URL as a key if it was provided. This PR uses the url as default key (including params and baseURL if provided).

There is still room for future improvement here via https://github.com/nuxt/framework/issues/6629, for example, and I'll be following up with some more PRs targeting specific issues.

### 📝 Checklist


pi0:
Any update on this @danielroe ? Last time we discussed i think we decided to combination of auto key and hash.

danielroe:
@pi0 Sure! Would you summarise what the issues are from your point of view, apart from hashing the url key?

pi0:
Async data key to be `{autokey}-{hash(input)}`

danielroe:
How about instead: `options.key || ((no functional options) ? hash(url + options) : autoKey)`?

The reason is that any hash of input options is going to mismatch between server + client if any of those options are functions.

pi0:
By hash i mean url+query only not all options.

seems good idea to allow entirely overriding with custom key 👍

danielroe:
excellent

packages/nuxt/src/app/composables/fetch.ts
```diff
@@ -44,7 +46,7 @@ export function useFetch<
   arg2?: string
 ) {
   const [opts = {}, autoKey] = typeof arg1 === 'string' ? [{}, arg1] : [arg1, arg2]
-  const _key = opts.key || autoKey
+  const _key = opts.key || (typeof request === 'string' ? hash(withBase(unref(opts.baseURL) || '', withQuery(request as string, unref(opts.params) || {}))) : autoKey)
```
pi0, line 49: (didn't locally tested) We can pass raw object to hash function. I guess also withBase not needed we mainly need digest.

```suggestion
  const _reqHash = typeof request === 'string' ? hash([unref(opts.baseURL)]) : ''
  const _key = opts.key || (autoKey + '-' + (typeof request === 'string' ? hash([request, unref(opts.baseURL), unref(opts.params)]) : ''))
```

danielroe: updated. though seems like this approach would prevent sharing useFetch state between pages?

pi0 (approved):

examgoal:
The fetch options key not working anymore! I think it broke it, please check again

manniL:
@examgoal Please create a new issue with reproduction ☺️

Issue #1950, opened 2021-11-16
require is not defined


pixelpaulaus:
### Environment

latest versions

### Reproduction

` <img :src="require('~/assets/image.png')">`  anywhere in vue component.

### Describe the bug

require is not defined when using it such as....

` <img :src="require('~/assets/image.png')">`

### Logs

```shell
at $id_9513b4f8 (file://./.nuxt/dist/server/server.mjs:10135:22)
at __instantiateModule__ (file://./.nuxt/dist/server/server.mjs:10596:9)
at __ssrLoadModule__ (file://./.nuxt/dist/server/server.mjs:10543:25)
at ssrImport (file://./.nuxt/dist/server/server.mjs:10568:13)
at $id_45a9f067 (file://./.nuxt/dist/server/server.mjs:9928:37)
at __instantiateModule__ (file://./.nuxt/dist/server/server.mjs:10596:9)
at __ssrLoadModule__ (file://./.nuxt/dist/server/server.mjs:10543:25)
at ssrImport (file://./.nuxt/dist/server/server.mjs:10568:13)
at $id_7689e89d (file://./.nuxt/dist/server/server.mjs:1935:37)
at async __instantiateModule__ (file://./.nuxt/dist/server/server.mjs:10596:3)
```

danielroe:
If you are using vite then `require` will not work; it's webpack-only. Instead, just use:

```vue
<template>
  <div class="app">
    <img src="~/assets/image.png">
  </div>
</template>
```

pixelpaulaus:
thats the weird thing, i am not using vite O_o
and i am wanting to use  require from a variable value.

danielroe:
@pixelpaulaus You can set `vite: false` in your `nuxt.config`.

kayoderock:
Yeah @pixelpaulaus, I had this same issue, how can I make a dynamic src using Vite?

kayoderock:
Yeah found it https://vitejs.dev/guide/assets.html#new-url-url-import-meta-url - Vite docs

```
imgUrl () {
    return new URL(`../assets/${this.icon}.svg`, import.meta.url).href
  }
```

tsulatsitamim:
> Yeah found it https://vitejs.dev/guide/assets.html#new-url-url-import-meta-url - Vite docs
> 
> ```
> imgUrl () {
>     return new URL(`../assets/${this.icon}.svg`, import.meta.url).href
>   }
> ```

not work in ssr

[...]

PR #6635, opened 2022-08-15
[doc] Add how to use assets with dynamic names

tobiasdiez:
### 🔗 Linked issue

Based on https://github.com/nuxt/framework/issues/1950#issuecomment-1192122525

### ❓ Type of change

### 📚 Description

Add how to dynamically link assets.
Caveat: It doesn't work using the syntax `~/assets/...` instead of `/assets/...`, and it sadly doesn't work in production. So I guess I'm missing something.

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/2.features/3.assets.md
+++ b/docs/content/2.guide/2.features/3.assets.md
@@ -39,6 +39,13 @@ For example, referencing an image file that will be processed if a build tool is
 </template>
 ```
 
+This also works if the path to the image is dynamic:
+```vue [app.vue]
+<template>
+  <img :src="`/assets/${dynamic_image_name}`" alt="Discover Nuxt 3" />
+</template>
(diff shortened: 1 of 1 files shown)


Aareksio:
`/assets` being available in dev is a Vite quirk. Adding this to documentation is misleading. That said, documentation could use an example of how to use dynamic imports from assets to replace ``require(`~/assets/{name}`)`` from the old days.

pi0:
@tobiasdiez Can you please open a new issue with reproduction? We might try to investigate and fix support for dynamic assets. (But generally you can always use `public/` directory too)

tobiasdiez:
@pi0 Done: https://github.com/nuxt/framework/issues/7121!

The public folder works in some situation, but not if you want to have some optimization etc going on during the build process.

Issue #6638, opened 2022-08-15
optimisations for prefetching chunks


danielroe:
Currently dynamic imports include things like routes, error components, and any global components (so if using with `@nuxt/content`). I am _very_ concerned to decrease HTML and number of prefetched resources but if we simply disable prefetching, sites will seem much slower.

From my point of view, before removing current prefetch behaviour, we need the ability to make some finer-grained decisions.

A set of suggestions:

- [ ] `prefetch` support via NuxtLink (https://github.com/nuxt/framework/pull/4329)
- [ ] programmatic prefetching of routes: https://github.com/nuxt/framework/issues/3649
- [ ] remove page chunks from dynamic imports in manifest, or smartly add 'sibling' pages
- [ ] programmatic prefetching of global components
- [ ] (optionally) add a plugin to update manifest so that the components that _might_ use global components are the ones that have the dynamic import registered
- [ ] add hook for user to manipulate client manifest
- [ ] (possibly) add ability for user to customise `vue-bundle-renderer` options - see https://github.com/nuxt/framework/pull/6210#discussion_r937950236.

kosmeln:
Hello @danielroe .
I'm not sure if this is related to this improvement or I should create a separate issue. Please let me know.
I've created a [test project](https://codesandbox.io/s/nuxt3-playground-rc-8-8xvt77?file=/pages/red-page.vue) to visualise how pre-fetching stylesheets may affect the way page looks.
There are 2 layouts - red and blue (under /layouts). If I open red-page component (https://8xvt77.sse.codesandbox.io/red-page), blue layout - that is being pre-fetched - affects the way page looks - and h2 that should be black displays in blue instead.
The example is oversimplified but we have a real use case of this being a problem for our project.

pi0:
@kosmeln Please ensure to use scoped styles (`<style scoped>`) or selector prefixes per parent/layout. You can use a special CSS class in layouts that enable them otherwise, loading a layout can make unpredictable behavior. See this fork: https://codesandbox.io/s/nuxt3-playground-rc-8-forked-6l6uot?file=/layouts/blue.vue:122-134

pi0:
@danielroe As we discussed, we can move both of first items in parallel.

Build-time and manifest based page prefetching is probably something we don't want to do in Nuxt 3 since was always tricky in Nuxt 2 when number of pages increases. Only reliable way to predict next pages is runtime rendering. Both Server-Side and Client-Side of `NuxtLink` can smartly add prefetches.

[...]

PR #6661, opened 2022-08-16
feat(nuxt): allow programmatically prefetching global components

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/6638

### ❓ Type of change

### 📚 Description

~~This adds a new hook (`components:prefetch`) that allows prefetching of global components (Pascal case required, to avoid runtime addition of scule utility). We could also expose a utility like `nuxtApp.prefetch` or standalone `prefetchComponent` that uses this hook under the hood.~~

This PR exposes new `preloadComponents` utility that allows prefetching of global components (Pascal case required, to avoid runtime addition of scule utility). 

Also an alias `prefetchComponents` is exposed which does same as `preloadComponents` but as we progress to make better implementation based on manifest maps, users can already opt-in to use semantically write utility.

### 📝 Checklist


packages/nuxt/src/components/templates.ts
```diff
@@ -40,7 +40,23 @@ const components = ${genObjectFromRawEntries(globalComponents.map((c) => {
...
 
+function prefetchAsyncComponent (component) {
+  if (component?.__asyncLoader && !component.__asyncResolved) {
+    return component.__asyncLoader()
+  }
+}
+
 export default defineNuxtPlugin(nuxtApp => {
+  if (process.client) {
+    nuxtApp.hook('components:prefetch', async (components) => {
```
pi0, line 51: It would be nice if we instead expose a standalone composable like `prefetchComponent` that allows prefetching in any part of lifecycle.

packages/nuxt/src/app/composables/prefetch.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Prefetch a component or components that have been globally registered
+ * with the Nuxt component integration.
+ *
+ * @note This does not work for components you have globally registered yourself.
+ * @param components Pascal-cased name or names of components to prefetch
+ */
+export const prefetchComponents = (components: string | string[]) => {
+  if (process.server) { return }
+
+  return useNuxtApp().callHook('components:prefetch', components)
```
pi0, line 13: If we could expose async chunks manifest from manifest, composable could directly implement prefetchAsyncComponent without depending on hook

danielroe: I think it would be a bit more fragile to do that. Vite already manages fetching requests/asset URLs, etc.

danielroe: We can add type support for this composable (nice idea!) without changing the implementation.

It feels like it will be duplicated work to add such a manifest + composable... I am worried about negative performance implications, but perhaps you have a reason for doing it this way?

pi0: In fact, it could save bundle size since composable can be tree-shaken but plugin bytes are always there.

danielroe: You're right, it could be tree-shaken. 👍  But if the user does use it, or is using a library like `@nuxt/content`, then we would have to inline a subset of the manifest. If you still think it's worth it, I can refactor. (Benefit would be a consistent implementation for layouts/pages/components).

pi0: I think it would be better with composable. The plugin includes the object of dynamic components anyway as manifest.

pi0 (approved):
LGTM. As discussed with @danielroe we should iterate over this to make use of build manifest to generate actual `<link>` tags and also implement server-side rendering hints (currently not needed with #6876)

Issue #6670, opened 2022-08-16
Create project hashes in buildDir


pi0:
Context: In nuxt 2.13, after the introduction of full-static mode, we introduced a new build-time optimization to reduce CI time (static generation) by avoiding a rebuild if project hashes were not different but only a new prerendering was needed (https://github.com/nuxt/nuxt.js/pull/7712 - [lastest impl](https://github.com/nuxt/nuxt.js/blob/86570cf6ab82cd7f58fb7973bd4c43a55307175b/packages/cli/src/utils/generate.js#L167))

We shall backport this feature to Nuxt 3 which is not only useful for "smart rebuild" functionality but also automatically cleaning up cache directories on project hash changes (#6668) and also (partially) generate it with `nuxt dev`.

PR #6672, opened 2022-08-16
feat(nuxi): auto cleanup with project manifest changes

### LINKED ISSUE #6668: clean up up cache dirs on `rootDir`/hash change
we've had a number of issues tied back somehow to cache not being properly invalidated, such as when a project directory is moved or renamed, a transform is updated within nuxt, or a downstream package is updated. These seem to tie back to invalid caches.

Vite already invalidates its cache in certain conditions:
https://github.com/vitejs/vite/blob/ed0286186b24748ec7bfa336f83c382363a22f1d/packages/vite/src/node/optimizer/index.ts#L1034-L1055

This issue is to track the possible causes of cache failure and invalidate them ourselves.

Possibly related: https://github.com/nuxt/framework/issues/6623

---

Current solution: Run `npx nuxi-edge@latest cleanup` after dependency upgrades and before `nuxt dev` manually to ensure caches are cleaned up.

pi0:
### 🔗 Linked issue

Resolves #6668 and adds basic manifest implementing #6670

### ❓ Type of change

### 📚 Description

When a major change in a project happens (such as updating nuxt or renaming directory) we most likely need to invalidate caches to avoid unpredictable behavior. This PR does two things:

- Write a basic `nuxt.json` to `buildDir`/`.nuxt` only for `nuxi dev` (1)
- Checks manifest hash (full hash) against the previous build and call `nuxi cleanup` in case of hash mismatches.

Remarks:
- I've not included more extensive manifest info and hashes unless we find initial change is not enough. As mentioned by @danielroe in #6668, bundlers (vite and webpack) also handle cache invalidation for dependencies
- We probably need to generate two separate manifests for dev and prod (https://github.com/nuxt/framework/issues/6671)
- It only invalidates caches for `nuxi dev` and not `nuxi build` because i don't believe we saw any cache issue on `nuxi build`. We can add later on to both commands and utils are already refactored to prepare for this.

(1) Sample `.nuxt/nuxt.json`:

```json
{
  "_hash": "AQBCqiYugT",
  "project": {
    "rootDir": "/Users/pooya/Code/framework/playground"
  },
  "versions": {
    "nuxt": "3.0.0-rc.8"
  }
}
```

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -11,6 +11,7 @@ import { writeTypes } from '../utils/prepare'
 import { loadKit } from '../utils/kit'
 import { importModule } from '../utils/cjs'
 import { overrideEnv } from '../utils/env'
+import { writeNuxtManifest, loadNuxtManifest, cleanupNuxtDirs } from '../utils/nuxt'
 import { defineNuxtCommand } from './index'
 
 export default defineNuxtCommand({
@@ -73,6 +74,13 @@ export default defineNuxtCommand({
         currentNuxt = await loadNuxt({ rootDir, dev: true, ready: false })
         if (!isRestart) {
           showURL()
+        }
+
+        // Check if we need cache invalidation
+        const previousManifest = await loadNuxtManifest(currentNuxt.options.buildDir)
+        const newManifest = await writeNuxtManifest(currentNuxt)
+        if (previousManifest && newManifest && previousManifest._hash !== newManifest._hash) {
+          await cleanupNuxtDirs(currentNuxt.options.rootDir)
         }
 
         await currentNuxt.ready()
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -76,7 +76,7 @@ export default defineNuxtCommand({
           showURL()
         }
 
-        // Check if we need cache invalidation
+        // Write manifest and also check if we need cache invalidation
         if (!isRestart) {
           const previousManifest = await loadNuxtManifest(currentNuxt.options.buildDir)
           const newManifest = await writeNuxtManifest(currentNuxt)
--- a/packages/nuxi/src/commands/dev.ts
+++ b/packages/nuxi/src/commands/dev.ts
@@ -77,10 +77,12 @@ export default defineNuxtCommand({
         }
 
         // Check if we need cache invalidation
-        const previousManifest = await loadNuxtManifest(currentNuxt.options.buildDir)
-        const newManifest = await writeNuxtManifest(currentNuxt)
-        if (previousManifest && newManifest && previousManifest._hash !== newManifest._hash) {
-          await cleanupNuxtDirs(currentNuxt.options.rootDir)
+        if (!isRestart) {
+          const previousManifest = await loadNuxtManifest(currentNuxt.options.buildDir)
+          const newManifest = await writeNuxtManifest(currentNuxt)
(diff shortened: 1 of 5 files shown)


packages/nuxi/src/commands/dev.ts
```diff
@@ -75,6 +76,13 @@ export default defineNuxtCommand({
           showURL()
         }
 
+        // Check if we need cache invalidation
+        const previousManifest = await loadNuxtManifest(currentNuxt.options.buildDir)
+        const newManifest = await writeNuxtManifest(currentNuxt)
+        if (previousManifest && newManifest && previousManifest._hash !== newManifest._hash) {
+          await cleanupNuxtDirs(currentNuxt.options.rootDir)
+        }
```
danielroe, line 84: Do you think we need to do this within the refresh loop? I imagine you are thinking if project dependencies are updated while dev server is running? Or might it be enough to do it at start of dev command?

danielroe: If you are imagining runtime updating deps, then is there any merit in watching `package.json`? I don't think we can directly add to this line as it's relative to `srcDir`.

https://github.com/nuxt/framework/blob/75e3be9d0eefabf6d65463c6457349670bb32e02/packages/nuxi/src/commands/dev.ts#L105

packages/nuxi/src/utils/nuxt.ts
```diff
@@ -0,0 +1,58 @@
...
+  return manifest
+}
+
+export async function loadNuxtManifest (buildDir: string): Promise<NuxtProjectManifest | null> {
+  const manifestPath = resolve(buildDir, 'nuxt.json')
+  const manifest: NuxtProjectManifest | null = await fsp.readFile(manifestPath, 'utf-8')
+    .then(data => JSON.parse(data) as NuxtProjectManifest)
+    .catch(() => null)
+  return manifest
+}
```
danielroe, line 58: Maybe make it more explicit?

```suggestion
export async function writeNuxtManifest (nuxt: Nuxt): Promise<NuxtProjectManifest> {
  const manifest = resolveNuxtManifest(nuxt)
  const manifestPath = resolve(nuxt.options.buildDir, 'nuxt.manifest.json')
  await fsp.mkdir(dirname(manifestPath), { recursive: true })
  await fsp.writeFile(manifestPath, JSON.stringify(manifest, null, 2), 'utf-8')
  return manifest
}

export async function loadNuxtManifest (buildDir: string): Promise<NuxtProjectManifest | null> {
  const manifestPath = resolve(buildDir, 'nuxt.manifest.json')
  const manifest: NuxtProjectManifest | null = await fsp.readFile(manifestPath, 'utf-8')
    .then(data => JSON.parse(data) as NuxtProjectManifest)
    .catch(() => null)
  return manifest
}
```

packages/nuxi/src/commands/dev.ts
pi0: Indeed at the moment doesn't benefits reloads. Will make it once.

packages/nuxi/src/utils/nuxt.ts
pi0: Nitro uses the same `nitro.json` format and we also generate `dist/module.json` for module-builder output. "Manifest" itself is an unexplicit name IMO to assign (ie: what is exactly manifest).

packages/nuxi/src/commands/dev.ts
pi0: b1e86568897748f17d462faa76f02673c96b0c64

PR #6683, opened 2022-08-16
fix(nuxt): pass and format vite-node build errors 

antfu:
### ❓ Type of change

### 📚 Description

To pass down the nested `h3` app's error to be handled by the main one. `nodeHandler` comes with its own error handling (https://github.com/unjs/h3/blob/b95ab25575e1a6c022ff40cfd731d27cb735f621/src/app.ts#L57), changed to `handler` so it let the error flow down. 

### 📝 Checklist


pi0:
Thanks for PR @antfu. Please use description for next PRs even if we internally discussed 🙏🏼

pi0:
Locally trying, these changes causes the server to be completely stuck now instead of "Internal Server Error".

antfu:
Added, thanks!

pi0:
Checking more locally, it seems passing main handler is not enough. Vite node (client) throws nice object with nice trace info (id, plugin, message, name, stack, loc, pluginCode, frame). Passing it up to the dev server handler, won't help. But we could pass it _down_ to the vite-node fetch client. There I tried to restore main error object from HTTP response and throw it in `fetchModule` of viteNodeRunner. But it will be only handled when we call `render`...

pi0:
BTW comparing to legacy bundler, we were never handling build errors from server side but using an empty module. They were being then handled on client-side.

I've just pushed https://github.com/nuxt/framework/pull/6683/commits/808fd5bf9043c0e6d713a91fe4ff9b3ede1a6252 to show how similar effect works.

image

packages/vite/src/vite-node.ts
```diff
@@ -120,12 +120,18 @@ function createViteNodeMiddleware (ctx: ViteBuildContext, invalidates: Set<strin
       if (moduleId === '/') {
         throw createError({ statusCode: 400 })
       }
-      const module = await node.fetchModule(moduleId) as any
+      const module = await node.fetchModule(moduleId).catch((err) => {
+        throw createError({
+          statusCode: 500,
+          statusMessage: `[vite-node] Cannot fetch module ${moduleId}`,
+          data: err
```
antfu, line 127: Do we need a more specific name other than `data`? Like `transformError` or something?

Other than that this approach looks good to me 👍

pi0:
I had been testing giving error data for few cases. There is really no standard for errors coming from vite, rollup plugins and parsers such as babel used for sfc. I've made a temporary normalizer on best efforts and compatible with default error template but we should make something more standard for normalization later.

Script:

image

Template:

image


Style:

image

PR #6685, opened 2022-08-16
refactor(nuxt, kit): improve type strictness

antfu:
### ❓ Type of change

### 📝 Checklist


packages/kit/src/ignore.ts
```diff
@@ -11,7 +11,7 @@ export function isIgnored (pathname: string): boolean {
 
   // Happens with CLI reloads
   if (!nuxt) {
-    return null
+    return false
```
pi0, line 14: Context: This was to distinguish situations where `isIgnored` **cannot** identify ignoring from when it detects file is actually ignored.

pi0 (approved):
Nice changes. Thanks <3

PR #6697, opened 2022-08-17
fix(nuxt): fallback to static error page on server error

antfu:
### 🔗 Linked issue

Part of #6683

### ❓ Type of change

### 📚 Description

When there is a server error, the `/__nuxt_error` might also fail to render, we fallback to use static HTML render in this case.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/error.ts
+++ b/packages/nuxt/src/core/runtime/nitro/error.ts
@@ -1,6 +1,7 @@
 import { withQuery } from 'ufo'
 import type { NitroErrorHandler } from 'nitropack'
 import type { H3Error } from 'h3'
+import { importModule } from '@nuxt/kit'
 import { normalizeError, isJsonRequest } from '#internal/nitro/utils'
 
 export default <NitroErrorHandler> async function errorhandler (error: H3Error, event) {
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -13,7 +14,7 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error,
     statusCode,
     statusMessage,
     message,
-    description: process.env.NODE_ENV === 'development' && statusCode !== 404
+    stack: process.env.NODE_ENV === 'development' && statusCode !== 404
```
antfu, line 17: Not sure the reason initially, but this change makes the error page:

Before:
image

After:
image

pi0: `description` field is used in error templates for title. I think we can now adjust it to error message for dev.

pi0:
Trying locally, the production build errors makes an infinity loop. Investigating... 

image

packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -1,7 +1,6 @@
 import { withQuery } from 'ufo'
 import type { NitroErrorHandler } from 'nitropack'
 import type { H3Error } from 'h3'
-import { importModule } from '@nuxt/kit'
```
pi0: @danielroe Do you have any idea why this wasn't caught by import protection?

pi0 (approved):

PR #6708, opened 2022-08-17
fix(nuxi): properly detect hash and tag for upgrade changelog

harlan-zw:
### ❓ Type of change

### 📚 Description

When upgrading RC versions, the Changelog link is assuming that nuxt was installed using a version with a git commit as the suffix. 

For versions that don't have this, only the last digit is used, so the user gets messages like this:

```
ℹ Cleaning up generated nuxt files and caches...                                                                                                                                                20:39:36
ℹ Upgraded nuxt version: 3.0.0-rc.8                                                                                                                                                         20:39:36
✔ Successfully upgraded nuxt from 3.0.0-rc.6 to 3.0.0-rc.8                                                                                                                                    20:39:36
ℹ Changelog: https://github.com/nuxt/framework/compare/6...8
```

This PR fixes the behaviour and links the changelog correctly.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/upgrade.ts
+++ b/packages/nuxi/src/commands/upgrade.ts
@@ -21,6 +21,16 @@ async function getNuxtVersion (paths: string | string[]): Promise<string|null> {
   }
 }
 
+const nuxtVersionToGitIdentifier = (version: string) => {
+  const parts = version.split('.')
+  // match the git identifier in the release, for example: 3.0.0-rc.8-27677607.a3a8706
+  if (parts.length > 4) {
+    return parts.pop()
+  }
+  // match github tag, for example 3.0.0-rc.8
+  return `v${version}`
+}
+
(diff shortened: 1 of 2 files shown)


packages/nuxi/src/commands/upgrade.ts
```diff
@@ -21,6 +21,16 @@ async function getNuxtVersion (paths: string | string[]): Promise<string|null> {
   }
 }
 
+const nuxtVersionToGitIdentifier = (version: string) => {
+  const parts = version.split('.')
+  // match the git identifier in the release, for example: 3.0.0-rc.8-27677607.a3a8706
+  if (parts.length > 4) {
```
pi0, line 27: For detecting edge, we shall use a more specific regex (because `-rc.8` will be removed at some point). Something like `\.([0-9a-f]{7})$` might work.

pi0 (approved):

Issue #4325, opened 2022-04-13
'setupDevtoolsPlugin' is not exported by `unenv`


billymaulana:
### Environment

NUXT 3 : 3.0.0-27497374.59593a0
NODE : 16.4.2
pnpm": "^6.32.3"

### Reproduction

.

### Describe the bug

when normaly build its fine.
<img width="870" alt="image" src="https://user-images.githubusercontent.com/13347455/163191690-78da4ea7-0795-4496-b709-41b5a5c71664.png">

### Additional context
my nuxt config not set preset.
when im build NITRO_PRESET=cloudflare pnpm build. 

### Logs
Error like this:
```
 ERROR  Rollup error: 'setupDevtoolsPlugin' is not exported by node_modules/.pnpm/unenv@0.4.6/node_modules/unenv/runtime/mock/proxy.mjs, imported by node_modules/.pnpm/pinia@2.0.13_d3195e5b5ca92757b2890fe178df3bd1/node_modules/pinia/dist/pinia.mjs

 ERROR  'setupDevtoolsPlugin' is not exported by node_modules/.pnpm/unenv@0.4.6/node_modules/unenv/runtime/mock/proxy.mjs, imported by node_modules/.pnpm/pinia@2.0.13_d3195e5b5ca92757b2890fe178df3bd1/node_modules/pinia/dist/pinia.mjs

  at error (node_modules/.pnpm/rollup@2.70.1/node_modules/rollup/dist/es/shared/rollup.js:1784:30)
  at Module.error (node_modules/.pnpm/rollup@2.70.1/node_modules/rollup/dist/es/shared/rollup.js:12345:16)
  at Module.traceVariable (node_modules/.pnpm/rollup@2.70.1/node_modules/rollup/dist/es/shared/rollup.js:12703:29)
  at ModuleScope.findVariable (node_modules/.pnpm/rollup@2.70.1/node_modules/rollup/dist/es/shared/rollup.js:11333:39)
  at FunctionScope.findVariable (node_modules/.pnpm/rollup@2.70.1/node_modules/rollup/dist/es/shared/rollup.js:6301:38)
  at ChildScope.findVariable (node_modules/.pnpm/rollup@2.70.1/node_modules/rollup/dist/es/shared/rollup.js:6301:38)
  at Identifier.bind (node_modules/.pnpm/rollup@2.70.1/node_modules/rollup/dist/es/shared/rollup.js:7592:40)
[...]

PR #6713, opened 2022-08-17
fix(nuxt): use `vue-devtools-stub` to mock `@vue/devtools-api` for both cjs + esm

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/4325

### ❓ Type of change

### 📚 Description

When `@vue/devtools-api` is imported in an ESM module (as in pinia), our cjs proxy doesn't work. Thanks to @pi0, we now have a standalone stub we can use that works in both ESM + CJS contexts.

### 📝 Checklist

### DIFF
--- a/test/fixtures/basic/pages/index.vue
+++ b/test/fixtures/basic/pages/index.vue
@@ -24,9 +24,7 @@
 import { setupDevtoolsPlugin } from '@vue/devtools-api'
 import { useRuntimeConfig } from '#imports'
 
-if (useRuntimeConfig().enableDevtools) {
-  setupDevtoolsPlugin()
-}
+setupDevtoolsPlugin()
 
 const config = useRuntimeConfig()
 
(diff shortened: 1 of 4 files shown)


test/fixtures/basic/pages/index.vue
```diff
@@ -21,8 +21,13 @@
 </template>
 
 <script setup>
+import { setupDevtoolsPlugin } from '@vue/devtools-api'
 import { useRuntimeConfig } from '#imports'
 
+if (useRuntimeConfig().enableDevtools) {
```
pi0, line 27: Do we have this flag in fixture?

pi0, line 27: Do we have `enableDevtools` in fixture? (i think could be always tested...)

danielroe: now passing options - should succeed 👍

PR #6735, opened 2022-08-18
fix(vite): dedupe `vue` in client bundle

### LINKED ISSUE #6720: Slots break extend, if extend-folder is side by side with project-folder
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.19.0`
- Package Manager:  `npm@6.14.16`
- Builder:          `vite`
- User Config:      `extends`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/fabianwohlfart/test__nuxt-3-extend

```
cd extend
npm install

cd base
npm install
npm run build
npm run preview
```

### Describe the bug

If you move the extend-folder out of the project-folder, to be a separate repo, and use `<slots>` it throws an error regarding `isCE` and `Hydration`. Only in build mode.

```
entry.4c2179d9.mjs:1 TypeError: Cannot read properties of null (reading 'isCE')
    at Q (TestButton.b9bd506a.mjs:1:1182)
    at Proxy.<anonymous> (TestButton.b9bd506a.mjs:1:5029)
    (...)

entry.4c2179d9.mjs:1 Hydration completed but contains mismatches.

```

### Additional context

I want to have my extension as a separate repo to use it for multiple projects.
Therefor I moved the extension folder out of the root folder.
The extend-folder also needs a `node_modules` as it is separated from the projects-folder.
I assume this is the problem, as it is _another_ nuxt instance?

Removing the `<slot>` works, as does including the extend-folder inside the projects-folder without it's own  `node_modules`.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #6720, also see https://github.com/nuxt/framework/issues/4084

### ❓ Type of change

### 📚 Description

This enables vite's dedupe functionality for vue, where we definitely want to ensure we have a single instance bundled. A previous issue with this plugin (where it emitted CJS code) seems to have been resolved in Vite 3.

**Edit**: enabling only for client.

~~It injects the following code in the server bundle:~~

https://github.com/vitejs/vite/blob/9e9cd23763c96020cdf1807334368e038f61ad01/packages/vite/src/node/plugins/ssrRequireHook.ts#L53-L71

~~We can alternatively enable this only for the client bundle. WDYT @pi0?~~

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -26,10 +26,7 @@ export default {
       })
     },
     resolve: {
-      extensions: ['.mjs', '.js', '.ts', '.jsx', '.tsx', '.json', '.vue'],
-      resolve: {
-        dedupe: ['vue']
-      }
+      extensions: ['.mjs', '.js', '.ts', '.jsx', '.tsx', '.json', '.vue']
     },
     publicDir: {
       $resolve: (val, get) => val ?? resolve(get('srcDir'), get('dir').public)
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -26,7 +26,10 @@ export default {
       })
     },
     resolve: {
-      extensions: ['.mjs', '.js', '.ts', '.jsx', '.tsx', '.json', '.vue']
+      extensions: ['.mjs', '.js', '.ts', '.jsx', '.tsx', '.json', '.vue'],
+      resolve: {
+        dedupe: ['vue']
+      }
     },
     publicDir: {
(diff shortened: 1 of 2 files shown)


pi0 (approved):
seems a good option to try

pi0:
Just checked SSR plugin. It seems modifying Node.js `Module` behavior. Probably safer to enable for client bundle only. Nitro should properly does it.

pi0 (approved):

danielroe:
Putting this here for future reference: https://github.com/vitejs/vite/issues/7454#issuecomment-1090007684

PR #6768, opened 2022-08-19
feat(nuxt): `imports.autoImport` option to disable auto-imports

### LINKED ISSUE #6428: allow disabling auto imports
### Describe the feature

First mentioned here by @nikolay-zoteev and @cawa-93 in https://github.com/nuxt/framework/discussions/2271.

I would like to be able to disable auto import, so I can make explicit imports from respective vendor library required in the codebase.

```
import { ref } from '#imports'
```
Would become:
```
import { ref } from 'vue'
```

Additional, from https://github.com/nuxt/framework/discussions/2271#discussioncomment-2906232:
> I still need a reliable and documented way to turn off auto-import.
> 
> * Auto-import does not work for the server part
> * Auto-import doesn't work for everyone. You still have business logic modules, helpers, utilities, services that you have to import on your own (actually IDE will do everything for you).
> * Only top level files can be auto-imported.
> * Auto-import of asynchronous components (such as those created by [`defineAsyncComponent`](https://vuejs.org/api/general.html#defineasynccomponent)) is not possible.
> 
> In addition
> 
> * Auto-import complicates IDE code analysis (jet brains have issues with deep import analysis)
> * It takes twice as long to go from calling a component to declaring it

### Additional information

- [ ] Would you be willing to help implement this feature?

### Final checks

antfu:
…import

### 🔗 Linked issue

fix #6428

### ❓ Type of change

### 📚 Description

Since Nuxt uses `#imports` internally and modules would also need it, we shouldn't completely disable the transform, but instead, we could make implicit imports possible to opt-out.

### 📝 Checklist


docs/content/2.guide/1.concepts/4.auto-imports.md
```diff
@@ -55,3 +55,17 @@ Nuxt exposes every auto-import with the `#imports` alias that can be used to mak
...
 ```
+
+## Disable Auto-imports
+
+In case you want to disable auto-imports, you can set `autoImports.enabled` to `false`.
+
+```ts [nuxt.config.ts]
+  autoImports: {
+    enabled: false
```
pi0, line 66: What do you think of calling option `implicit: false` or `global: false` (since we still have _auto imports_ integration with `#imports`.

antfu: Yeah I kinda thought about that. My idea was:

- `implicit` might be a bit less intuitive as not everyone needs to know the exits of explicit imports. Most of the time I feel the "Auto import" should refer to implicit auto import.
- We picked the scope name of `autoImports` (with we went a bit beyond that tbh 😄), `autoImports.enabled = false` sounds just like we are disabled the "auto" part. While the explicit import technically is not that "Auto" 🤣

For more semantic meaning maybe we could reconsider the `autoImports` scope naming.

But that said, I think `implicit: false` could also be a good name.

danielroe: What about `mode: 'inject' | 'import'`?

darthf1:
Wow, thanks for looking into this, and so soon :)

So, If i understand everything correctly, when the implicit auto import is disabled, `#import` keeps working, because it is used in modules, e.g. like [Pinia](https://github.com/vuejs/pinia/blob/v2/packages/nuxt/src/runtime/plugin.vue3.ts#L2).

Can I still use `#import` in my own app? Or is that alias removed from the `./.nuxt/tsconfig.json`? My preference would be that I cannot use `#import` in my own app, and force users to use `import { ref } from 'vue'` instead of `import { ref } from '#imports'`. But I'm not sure if this would be possible  with the previous point.

What happens with imports from `nuxt/dist/app`? Do I use `import { defineNuxtPlugin } from 'nuxt/dist/app'`, or is this what the `#app` alias is used for? Is `#app` being kept, or do i need to `import { defineNuxtPlugin } from 'nuxt'` after this PR?
```ts
readonly defineNuxtComponent: UnwrapRef<typeof import('../../node_modules/nuxt/dist/app')['defineNuxtComponent']>
readonly defineNuxtLink: UnwrapRef<typeof import('../../node_modules/nuxt/dist/app')['defineNuxtLink']>
readonly defineNuxtPlugin: UnwrapRef<typeof import('../../node_modules/nuxt/dist/app')['defineNuxtPlugin']>
readonly defineNuxtRouteMiddleware: UnwrapRef<typeof import('../../node_modules/nuxt/dist/app')['defineNuxtRouteMiddleware']>
```

Again, thanks for looking into this. Looking forward to use Nuxt 3 with this feature.

docs/content/2.guide/1.concepts/4.auto-imports.md
pi0: I like it but inject mode _also_ allows import (ie: it is not switch between them but switch auto)

antfu:
> My preference would be that I cannot use #imports in my own app

I would suggest setting an ESLint rule for that if you strictly want that.

> Do I use import { defineNuxtPlugin } from 'nuxt/dist/app', or is this what the #app alias is used for? Is #app being kept, or can I do import { defineNuxtPlugin } from 'nuxt' after this PR?

It's recommended to use `#app` (or `#imports`) whenever you can. `#app` refers to the runtime utilities that Nuxt generates. It's different from importing `nuxt`, which is the tool itself. Ideally, I would never import from `nuxt` except in config and programmatic usage IIUC.

docs/content/2.guide/1.concepts/4.auto-imports.md
danielroe: Then what about something like: 'auto' | 'explicit' ?

pi0: Still auto means (auto + explicit). More i'm thinking, we could rename option to `imports`, and then support `imports.autoImport: false`.

rvanlaak:
Found this PR while doing package selection for choosing Nuxt as possible framework for a future project. What I'm focussing on; finding a framework that adopts common coding best practices related to SOLID principles. In what situation would the concept of `#imports` match with SOLID principles? 

The great news is that [auto-imports](https://v3.nuxtjs.org/guide/concepts/auto-imports/) seem to make it drastically easier for beginning developers to build stuff, but also can be considered as a direct violation to the dependency inversion priniciple. Having a huge `#imports` pool of  unclear and opinionated classes will lead to tight coupling to the framework. Auto imports seem to introduce black compiler magic that make it unclear what code you actually are relying on, and will in turn make debugging harder. How can a developer do dependency inversion or guarantee LSP on one of the framework's services?

The feature in this PR thereby possibly could be considered as a must-have before the major release? Has it been considered in what way the introduction of auto-imports would (also negatively) impact the framework future? An interesting topic to read about is how in the land of PHP the Laravel framework Facades do opinionate development and thereby made two camps.

So, although the topic obviously might be out of scope of this PR, the question to ask is in what way Nuxt can on the one hand simplify development by supporting features like auto-imports, while on the other hand drive adoption of commonly accepted development principles like decent dependency injection. It is not strange that DI containers like [InversifyJS do so well on Github with 9.2k stars](https://github.com/inversify/InversifyJS/blob/master/wiki/classes_as_id.md). [PSR-11](https://www.php-fig.org/psr/psr-11/) is a great example of what Nuxt could also do. 

What about, instead of silently enabling auto-imports and forcing engineers to use auto-imports, as a framework giving the developer freedom of choice? Nuxt coud support the embedding of a DI container in Nuxt, while also giving them the other option to enable auto-imports? It could be as simple as a container interface or facade, and a toggle in the Nuxt config.

antfu:
This PR is now on top of https://github.com/nuxt/framework/pull/6768, and renamed to `imports.autoImport`

rvanlaak:
You probably wanted to reference to another PR here, given your comment mentions this same PR.

DaniFoldi:
Probably #6864 is the correct one

antfu:
@rvanlaak Thanks for the DI idea, but I guess it might a bit off-topic here. It would be great if you could open up a new discussion with a more detailed example/approach that you think would be useful. So we could discuss there.

pi0:
@antfu Do you mind to rebase PR?

pi0 (approved):

PR #6776, opened 2022-08-19
feat(nuxt): allow `useAppConfig` with specific key

pi0:
### ❓ Type of change

### 📚 Description

Context: #6333

`useAppConfig` returns a reactive object but it is super easy to loose reactivity when restructuring it's interface like `const theme = useAppConfig().theme` is not reactive anymore to accept HMR.

This PR adds a simple addition using computed for one top-level key.

Todo:
- [ ] Type support
- [ ] Reactivity without `.value`

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/config.ts
+++ b/packages/nuxt/src/app/config.ts
@@ -1,5 +1,5 @@
 import type { AppConfig } from '@nuxt/schema'
-import { reactive, computed } from 'vue'
+import { reactive, toRef } from 'vue'
 import { useNuxtApp } from './nuxt'
 // @ts-ignore
 import __appConfig from '#build/app.config.mjs'
@@ -13,7 +13,7 @@ export function useAppConfig (key?: string): AppConfig {
     nuxtApp._appConfig = reactive(__appConfig) as AppConfig
   }
   if (key) {
-    return computed(() => nuxtApp._appConfig[key])
+    return reactive(toRef(nuxtApp._appConfig, key))
   }
   return nuxtApp._appConfig
 }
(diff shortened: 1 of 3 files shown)


Atinux:
One tricky way here is that when using a key, we will need to use `.value` right?

antfu:
Do you mean `toRef(nuxtApp._appConfig, key)`?

pi0:
> One tricky way here is that when using a key, we will need to use .value right?

Indeed both computed and toRef make access harder with `.value`...

pi0:
Chaining `reactive(toRef(` _seems_ to work pefectly fine and without `.value` (but catch is, users should obviously do this for object keys)

pi0:
TODO: Type support

packages/nuxt/src/app/config.ts
```diff
@@ -1,17 +1,20 @@
...
 export const _getAppConfig = () => __appConfig as AppConfig
 
-export function useAppConfig (): AppConfig {
+export function useAppConfig (key?: string): AppConfig {
   const nuxtApp = useNuxtApp()
   if (!nuxtApp._appConfig) {
     nuxtApp._appConfig = reactive(__appConfig) as AppConfig
   }
+  if (key) {
+    return reactive(toRef(nuxtApp._appConfig, key))
```
antfu, line 16: I suspect this would work. Wrapping a ref with `reactive()` will return as-is.

I am not very sure about this feature. Kinda babysitting TBH. It's basically the limitation of reactive objects, same as you can't restructure Components' props - stating `useAppConfig` returns a reactive object should be enough.

I feel the best solution to improve this is the Reactivity Transform and the later https://github.com/vuejs/core/pull/5856

pi0:
With fix in #6788, HMR issue should be solved. Let's revise this idea later.

PR #6780, opened 2022-08-19
Add docker example

516310460:
Help users with docker deployment

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/5.deploy/3.docker.md
+++ b/docs/content/2.guide/5.deploy/3.docker.md
@@ -1,0 +1,42 @@
+# Deployment Docker
+
+## Dockerfile
+
+```Dockerfile
+FROM node:17-alpine
+
+RUN mkdir -p /usr/src/nuxt3-app
+WORKDIR /usr/src/nuxt3-app
+COPY package.json /usr/src/nuxt3-app/package.json
+COPY . .
+
+## When packaging on docker
+# RUN npm install pnpm -g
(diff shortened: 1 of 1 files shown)


danielroe:
This is great! We have currently not provided a Dockerfile in the docs so far with the aim of instead providing a Docker preset that could even generate a file for users. https://github.com/unjs/nitro/issues/54. @pi0 would be better able to comment, but it's possible we might defer adding this to the docs until we have a first-class nitro docker preset.

In terms of feedback on the Dockerfile, I would also advise not including the `~/node_modules` folder in the image but only `~/.output` as it should have all its dependencies already installed there.

516310460:
> This is great - thank you!
> 
> We have currently not provided a Dockerfile in the docs so far with the aim of instead providing a Docker preset that could even generate a file for users. [unjs/nitro#54](https://github.com/unjs/nitro/issues/54). @pi0 would be better able to comment, but it's possible we might defer adding this to the docs until we have a first-class nitro docker preset.
> 
> In terms of feedback on the Dockerfile, I would also advise not including the `~/node_modules` folder in the image but only `~/.output` as it should have all its dependencies already installed there.

I commented on the dependency file. There may be users who want to use it. By default, no dependent files will appear. 'is used Output 'file

pi0:
Thanks for PR @516310460 <3 As daniel mentioned, we might introduce docker preset with proper nitro support, multi-stage build and only including `.output`.

PR #6861, opened 2022-08-23
Modify `pages/index.vue` and content

516310460:
Add `tsx` and `template`, and change the content to `Welcome to nuxt3`

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/docs/content/2.guide/3.directory-structure/10.pages.md
+++ b/docs/content/2.guide/3.directory-structure/10.pages.md
@@ -19,8 +19,27 @@ Pages are Vue components and can have the `.vue`, `.js`, `.jsx`, `.ts` or `.tsx`
 ::code-group
 
 ```vue [pages/index.vue]
+<script lang="tsx" setup>
+const welcome = () => {
+  return <span>Welcome </span>
+};
+const nuxt3 = () => {
+  return <span>nuxt3</span>
+};
+const content = () => {
+  return (
+    <div>
+      {welcome()}
+      <span>to </span>
+      {nuxt3()}
+    </div>
+  );
+};
+</script>
 <template>
-  <h1>Index page</h1>
+  <div>
+    <content />
+  </div>
 </template>
 ```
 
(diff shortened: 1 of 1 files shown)


danielroe (commented):
I think we want to be careful using to give JSX examples anywhere but in the advanced or examples sections. Here's a section from the Vue docs:

> Vue recommends using templates to build applications in the vast majority of cases. However, there are situations where we need the full programmatic power of JavaScript. That's where we can use the render function.
>
> https://vuejs.org/guide/extras/render-function.html#render-functions-jsx

516310460:
> I think we want to be careful using to give JSX examples anywhere but in the advanced or examples sections. Here's a section from the Vue docs:
> 
> > Vue recommends using templates to build applications in the vast majority of cases. However, there are situations where we need the full programmatic power of JavaScript. That's where we can use the render function.
> > https://vuejs.org/guide/extras/render-function.html#render-functions-jsx

Okey. maybe i am a `tsx` paranoid

danielroe:
Understood! I still think it's worth documenting, but maybe in an example, or in an advanced section?

516310460:
This example should appear in the `vue.js` document, but unfortunately I checked all the information, it seems that there is no. I think it's very useful because it provides `tsx` and `template` Flexibility to mix and match. I don't know if you like this usage

pi0:
> This example should appear in the vue.js document, but unfortunately I checked all the information, it seems that there is no

Indeed. This is because supporting `.tsx` is not official for Vue.js - while possible -. We could have an example in examples/ directory to show how it works.

516310460:
> We could have an example in examples/ directory to show how it works.

Which directory should it be in `examples/`?

pi0:
`examples/advanced/tsx` i would suggest :)

516310460:
> `examples/advanced/tsx` i would suggest :)

Great. I've already figured out how to write it. There should be components and pages, and various ways to use TSX

516310460:
Link: https://github.com/nuxt/framework/pull/6870

PR #6890, opened 2022-08-24
feat(nuxt): fallback error handler for CSR error pages

yunyuyuan:
### 🔗 Linked issue

[#6706](https://github.com/nuxt/nuxt.js/issues/14618)

### ❓ Type of change

### 📚 Description

Added an CSR error fallback handler to use static js templates from `@nuxt/ui-templates/templates/error-500.vue` as fallback.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-root.vue
+++ b/packages/nuxt/src/app/components/nuxt-root.vue
@@ -1,7 +1,12 @@
 <template>
   <Suspense @resolve="onResolve">
-    <ErrorComponent v-if="error" :error="error" />
-    <App v-else />
+    <NuxtErrorBoundary>
+      <ErrorComponent v-if="error" :error="error" />
+      <App v-else />
+      <template #error="{ errMsg = error }">
+        <FallbackErrorComponent :status-message="errMsg" />
+      </template>
+    </NuxtErrorBoundary>
   </Suspense>
 </template>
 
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -44,21 +44,22 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error,
 
   // HTML response (via SSR)
   const isErrorPage = event.req.url?.startsWith('/__nuxt_error')
-  let html = !isErrorPage ? await $fetch(withQuery('/__nuxt_error', errorObject)).catch(() => null) : null
+  const html = !isErrorPage ? await $fetch(withQuery('/__nuxt_error', errorObject)).catch(() => null) : null
 
+  // Not necessary because there have <NuxtErrorBoundary> in nuxt-root.vue,it can handle errorComponent
   // Fallback to static rendered error page
-  if (!html) {
```
pi0, line 50: This is necessary because we are not only handling Vue SSR errors but also errors happening from API. And SSR Error page could fail itself.

yunyuyuan: @pi0 Got it,thanks for your review❤,I've removed this.
Is that commit of adding `<NuxtErrorBoundary>` here correct?

danielroe:
Due to the repository migration, I'm closing this here - please feel free to reopen on https://github.com/nuxt/nuxt. However, I think that using `@nuxt/ui-templates` will probably be too costly in terms of size, so it may be worth discussing about the right approach in the linked issue first.

Issue #6832, opened 2022-08-22
`app.config` improvements


pi0:
Following up #5918 (editable `app.config`), notes in initial implementation #6333 and ideas from @Tahul in #6779, i've created this issue to track some potential improvements before releasing this new feature in next RC. (currently support is available on edge. Docs: https://v3.nuxtjs.org/guide/features/app-config)

- [x] Initial support (#6136)
- [ ] ~~support `useConfig(key)` (#6776)~~ 
- [x] Support HMR after destructing objects with deep assignment (#6788)
- [ ] Support global `$appConfig` for templates (?)
- [ ] Use `defu.arrayFn` to allow _overrriding_ arrays in multi-layer format
- [ ] Use untyped to (auto)generate types from `appConfig` instead of defu
- [ ] Support `_schema` (global config) [needs new issue to track]

Atinux:
I like the `$appConfig` for templates, could we also expose `$runtimeConfig` at the same time?

pi0:
However we have `$runtimeConfig.public`, exposing it could lead to new potential security issues using `$runtimeConfig` directly in templates that can render a token. Instead we can start migrating and advocating safer app.config for (public) and template usages.

brownsugar:
Hope app.config will support auto typing too, currently only appConfig in nuxt.config can be auto typed.

danielroe:
@brownsugar What do you mean by that? Would you give an example?

brownsugar:
> @brownsugar What do you mean by that? Would you give an example?

@danielroe Hi,

If I specified the app configs in `app.config.ts`:
``` ts
export default defineAppConfig({
  serverURL: ''
})
```

The `inlineConfig` is not auto generated in `.nuxt/types/app.config.d.ts`.

<img width="643" alt="image" src="https://user-images.githubusercontent.com/2935980/193727202-db20adad-4d48-4408-bff5-9c68621ffd25.png">

``` ts
const { serverURL } = useAppConfig()
// serverURL = any
```

After I move them to `appConfig` in `nuxt.config.ts`:

``` ts
export default defineNuxtConfig({
  appConfig: {
    serverURL: ''
  }
})
```

The `inlineConfig` is auto generated now.

<img width="603" alt="image" src="https://user-images.githubusercontent.com/2935980/193727520-a3811c61-6969-49ab-81da-16dc8b5c3efd.png">

``` ts
const { serverURL } = useAppConfig()
// serverURL = string
```

[...]

PR #6905, opened 2022-08-24
feat(nuxt): `app.config` improvements

Tahul:
### 🔗 Linked issue

#6832

### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

Resolves:
- [X] Use defu.arrayFn to allow overrriding arrays in multi-layer format
   - In c54292b17f56609e81fecea29eafcbf01f303ea9 I implemented it how it was made in [@nuxt-themes/kit](https://github.com/nuxt-themes/config/blob/main/src/utils.ts#L47-L53)
   - In 1b07424dce819355e2d479bf73639cb70e2e3289 I implemented it using `arrayFn` as suggested in #6832
      - Leads to strange behaviors with HMR (array keys going `null` when going from a play array to function format)
   - I would say that overwriting is **expected** behavior in this kind of config, and that `function` format seem to introduce a bit of overhead for the user to understand when reading config files.
- [X] Fixed an issue in which deep keys were not properly deleted on HMR, as seen [here](https://github.com/nuxt/framework/blob/main/packages/nuxt/src/app/config.ts#L24-L28).
	- This portion of codes properly deepAssign on changes, but does not delete existing keys
	- [Reproduction](https://stackblitz.com/edit/nuxt-starter-rsskjr?file=package.json,app.vue,app.config.ts)
- [X] Implement [updateAppConfig](https://github.com/nuxt/framework/commit/bdd222a85341de5cccb18cd29002efe30629e33c#diff-421426fc478e91dfcb94a9b4a10fa8c38e6c30a0546606fe7a6baaf05837adf8R45-R51) composable

Few notes about this:

I've seen appConfig merging occurs in built template [here](https://github.com/nuxt/framework/blob/main/packages/nuxt/src/core/templates.ts#L205-L218), considering that, what is the proper way to implement features that needs to have access to the fully merged app config file inside Nuxt?

In #6832, there is mentions to:
> - Use untyped to (auto)generate types from appConfig instead of defu
> - Support _schema (global config) [needs new issue to track]

I would love to help on these issues but need guidance on how to resolve the whole appConfig object from Nuxt side, in order to generate typings from it.

### 📝 Checklist

- [X] I have linked an issue or discussion.

### DIFF
--- a/docs/content/3.api/1.composables/update-app-config.md
+++ b/docs/content/3.api/1.composables/update-app-config.md
@@ -1,0 +1,20 @@
+# `updateAppConfig`
+
+::StabilityEdge
+::
+
+Updates [app config](/guide/features/app-config):
+
+**Usage:**
+
+```js
+const appConfig = useAppConfig() // { foo: 'bar' }
+
+const newAppConfig = { foo: 'baz' }
+
+updateAppConfig(newAppConfig)
+
+console.log(appConfig) // { foo: 'baz' }
+```
+
+::ReadMore{link="/guide/features/app-config"}
--- a/docs/content/3.api/1.composables/update-app-config.md
+++ b/docs/content/3.api/1.composables/update-app-config.md
@@ -3,7 +3,7 @@
 ::StabilityEdge
 ::
 
-Updates [app config](/guide/features/app-config):
+Updates [app config](/guide/features/app-config) using deep assignment. Existing (nested) properties will be preserved. 
 
 **Usage:**
 
--- a/docs/content/3.api/1.composables/update-app-config.md
+++ b/docs/content/3.api/1.composables/update-app-config.md
@@ -3,7 +3,7 @@
 ::StabilityEdge
 ::
 
-Updates [app config](/guide/features/app-config) using deep assignment. Existing (nested) properties will be preserved. 
+Updates [app config](/guide/features/app-config) using deep assignment. Existing (nested) properties will be preserved.
 
 **Usage:**
 
--- a/packages/nuxt/src/app/config.ts
+++ b/packages/nuxt/src/app/config.ts
@@ -21,10 +21,18 @@ if (process.dev) {
     const appConfig = useAppConfig()
     if (newConfig && appConfig) {
       deepAssign(appConfig, newConfig)
-      for (const key in appConfig) {
-        if (!(key in newConfig)) {
-          delete (appConfig as any)[key]
-        }
+      deepDelete(appConfig, newConfig)
+    }
+  }
+
+  function deepDelete (obj: any, newObj: any) {
+    for (const key in obj) {
+      if (!(key in newObj)) {
+        delete (obj as any)[key]
+      }
+
+      if (typeof obj[key] === 'object' && newObj[key]) {
+        deepDelete(obj[key], newObj[key])
       }
     }
   }
--- a/packages/nuxt/src/app/config.ts
+++ b/packages/nuxt/src/app/config.ts
@@ -18,6 +18,7 @@ function deepDelete (obj: any, newObj: any) {
     }
   }
 }
+
 function deepAssign (obj: any, newObj: any) {
   for (const key in newObj) {
     const val = newObj[key]
--- a/packages/nuxt/src/app/config.ts
+++ b/packages/nuxt/src/app/config.ts
@@ -7,12 +7,55 @@ import __appConfig from '#build/app.config.mjs'
 // Workaround for vite HMR with virtual modules
 export const _getAppConfig = () => __appConfig as AppConfig
 
+function deepDelete (obj: any, newObj: any) {
+  for (const key in obj) {
+    if (!(key in newObj)) {
+      delete (obj as any)[key]
+    }
+
+    if (typeof obj[key] === 'object' && newObj[key]) {
+      deepDelete(obj[key], newObj[key])
+    }
+  }
+}
+function deepAssign (obj: any, newObj: any) {
+  for (const key in newObj) {
+    const val = newObj[key]
+    if (val !== null && typeof val === 'object') {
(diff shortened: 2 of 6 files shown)


packages/nuxt/src/app/config.ts
```diff
@@ -15,28 +38,34 @@ export function useAppConfig (): AppConfig {
...
+ * Deep assign the current appConfig with the new one.
+ *
+ * Will preserve existing properties.
+ */
+export function updateAppConfig (appConfig: Partial<AppConfig>) {
+  const nuxtApp = useNuxtApp()
+  if (!nuxtApp._appConfig) {
+    nuxtApp._appConfig = reactive(__appConfig) as AppConfig
+  }
+  deepAssign(nuxtApp._appConfig, appConfig)
```
pi0, line 51: We could use `useAppConfig()`. It does same logic.

packages/nuxt/src/app/config.ts
```diff
@@ -15,28 +38,34 @@ export function useAppConfig (): AppConfig {
...
+}
+
+/**
+ * Overwrites the current app config with the new one.
+ */
+export function setAppConfig (appConfig: AppConfig) {
```
pi0, line 57: This composable doesn't makes sense to replace whole app config. Please remove.

packages/nuxt/src/imports/presets.ts
```diff
@@ -51,6 +51,8 @@ const appPreset = defineUnimportPreset({
     'createError',
     'defineNuxtLink',
     'useAppConfig',
+    'setAppConfig',
+    'updateAppConfig',
```
pi0, line 55: Do you mind to add api docs page for it (empty or minimum description is enough)

pi0:
Thanks for PR @Tahul! I think after resolving comments it is good to go 👍🏼 

Please for next ones keep them small to one change.

packages/nuxt/src/core/templates.ts
```diff
@@ -207,12 +207,13 @@ export const appConfigTemplate: NuxtTemplate = {
   write: true,
   getContents: ({ app, nuxt }) => {
     return `
-import defu from 'defu'
+import { defuArrayFn } from 'defu'
```
pi0, line 210: (please locally confirm) But we can use `defu.defuFn` to allow overriding all values with custom fn.

Tahul: Yes, it seem to work and overwrites other layers completely.

Tahul:
https://github.com/nuxt/framework/pull/6905#issuecomment-1225917346

I mean from inside Nuxt, this seem needed to generate typings and support `_schema` key as mentioned in the PR description.

packages/nuxt/src/app/config.ts
```diff
@@ -7,6 +7,29 @@ import __appConfig from '#build/app.config.mjs'
 // Workaround for vite HMR with virtual modules
 export const _getAppConfig = () => __appConfig as AppConfig
 
+function deepDelete (obj: any, newObj: any) {
+  for (const key in obj) {
+    if (!(key in newObj)) {
+      delete (obj as any)[key]
+    }
+
+    if (typeof obj[key] === 'object' && newObj[key]) {
```
pi0, line 16: Any reason condition is not same as `deepAssign`? (for null exception)

packages/nuxt/src/app/config.ts
```diff
@@ -15,28 +38,23 @@ export function useAppConfig (): AppConfig {
   return nuxtApp._appConfig
 }
 
+/**
+ * Deep assign the current appConfig with the new one.
+ *
+ * Will preserve existing properties.
+ */
+export function updateAppConfig (appConfig: Partial<AppConfig>) {
```
pi0, line 46: Correct typing should be DeepPartial IMO. But we can do refinement later...

pi0:
> I mean from inside Nuxt, this seem needed to generate typings and support _schema key as mentioned in the PR description.

We have access to `nuxt.options.appConfig`... But more thinking, a generic schema support for Nuxt makes more sense (it is not limited to app config feature) but can type `appConfig` too.

packages/nuxt/src/app/config.ts
Tahul: I used DeepPartial that was in some other places of the codebase locally to replace it.

packages/nuxt/src/app/config.ts
Tahul: I changed it to keep styling from other function

Tahul:
> > I mean from inside Nuxt, this seem needed to generate typings and support _schema key as mentioned in the PR description.
> 
> We have access to `nuxt.options.appConfig`... But more thinking, a generic schema support for Nuxt makes more sense (it is not limited to app config feature) but can type `appConfig` too.

CleanShot 2022-08-24 at 18 20 12@2x

I understand we have access to nuxt.options.appConfig, but does that mean people can only write their schema from `nuxt.config.ts` file then?

docs/content/3.api/1.composables/update-app-config.md
```diff
@@ -0,0 +1,20 @@
+# `updateAppConfig`
+
+::StabilityEdge
+::
+
+Updates [app config](/guide/features/app-config):
```
pi0, line 6:
```suggestion
Updates [app config](/guide/features/app-config) using deep assignment. Existing (nested) properties will be preserved. 
```

pi0:
> I understand we have access to nuxt.options.appConfig, but does that mean people can only write their schema from nuxt.config.ts file then?

I don't expect ordinary people to use (advanced) _schema feature we are planning to add. But yes it is a built-time thing for schema not runtime. Defined either using nuxt.config or hooks.

pi0 (approved):
Thanks

PR #6911, opened 2022-08-24
perf(nuxt): copy assets and prerender in parallel

sanjaiyan-dev:
Improve the performance by using [`Promise.all`](https://www.geeksforgeeks.org/javascript-promise-all-method/) syntax :)

_Sorry if I made any mistakes :(_

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -10,7 +10,7 @@ import { toEventHandler, dynamicEventHandler } from 'h3'
 import { distDir } from '../dirs'
 import { ImportProtectionPlugin } from './plugins/import-protection'
 
-export async function initNitro(nuxt: Nuxt) {
+export async function initNitro (nuxt: Nuxt) {
   // Resolve handlers
   const { handlers, devHandlers } = await resolveHandlers(nuxt)
 
@@ -64,10 +64,10 @@ export async function initNitro(nuxt: Nuxt) {
         ...(nuxt.options.dev
           ? []
           : [
-            ...nuxt.options.experimental.externalVue ? [] : ['vue', '@vue/'],
-            '@nuxt/',
-            nuxt.options.buildDir
-          ]),
+              ...nuxt.options.experimental.externalVue ? [] : ['vue', '@vue/'],
+              '@nuxt/',
+              nuxt.options.buildDir
+            ]),
         'nuxt/dist',
         'nuxt3/dist'
       ]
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for the PR. Do you mind to revert formatting changes

packages/nuxt/src/core/nitro.ts
```diff
@@ -160,19 +160,17 @@ export async function initNitro (nuxt: Nuxt) {
     if (nuxt.options.dev) {
       await build(nitro)
     } else {
-      await prepare(nitro)
```
pi0, line 163: Prepare needs to be done before both not in parallel.

packages/nuxt/src/core/nitro.ts
```diff
@@ -160,19 +160,17 @@ export async function initNitro (nuxt: Nuxt) {
...
-      await copyPublicAssets(nitro)
-      await prerender(nitro)
+      await Promise.allSettled([prepare(nitro), copyPublicAssets(nitro), prerender(nitro)])
```
DamianGlowala, line 163: Also, should it be `Promise.all()` instead of `Promise.allSettled()`, as the latter will never reject?

sanjaiyan-dev:
Done @DamianGlowala  & @pi0  👍🏽

packages/nuxt/src/core/nitro.ts
```diff
@@ -161,8 +161,7 @@ export async function initNitro (nuxt: Nuxt) {
       await build(nitro)
     } else {
       await prepare(nitro)
-      await copyPublicAssets(nitro)
-      await prerender(nitro)
+      await Promise.all([copyPublicAssets(nitro), prerender(nitro)])
```
pi0, line 164: @danielroe Would it break critters? (i assume you somehow depend on build assets being available during prerendering?)

danielroe: Nice catch; you are right.

pi0:
Hi @sanjaiyan-dev. I really appreciate time you put on this PR please do it again 💚 (you can make an issue to confirm before making PRs)

Considering build steps, if we do prerendering and copying of assets in parallel, it can cause some issues when prerendering depends on public assets in `/_nuxt/`. We need to read them to inline or optimize some assets. Also it shouldn't be a major bottleneck for performance we could instead focus on increasing prerender parallel performance and prerender build time.

Cheers.

sanjaiyan-dev:
> Hi @sanjaiyan-dev. I really appreciate time you put on this PR please do it again 💚 (you can make an issue to confirm before making PRs)
> 
> Considering build steps, if we do prerendering and copying of assets in parallel, it can cause some issues when prerendering depends on public assets in `/_nuxt/`. We need to read them to inline or optimize some assets. Also it shouldn't be a major bottleneck for performance we could instead focus on increasing prerender parallel performance and prerender build time.
> 
> Cheers.

Ok, tks 💪🙌

Issue #6920, opened 2022-08-25
ignore `.ts` files inside the `pages/`


warmthsea:
### Describe the feature

![B2C0E6E5107E45F1BBC106BC8C25C375](https://user-images.githubusercontent.com/45450994/186596417-c526c842-8e1d-47df-8653-d3827619e705.png)
I want to write the code in an external Ts file (it is only related to the adjacent vue file), and then import it in the Vue file, but it is registered as a route

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

516310460:
Yes, need to exclude `.ts`

warmthsea:
> Yes, need to exclude `.ts`

Is there any specific documentation?

warmthsea:
> Yes, need to exclude `.ts`

Is there any docs to configure it?

516310460:
See: https://v3.nuxtjs.org/api/configuration/nuxt.config#extensions

> > Yes, need to exclude `.ts`
> 
> Is there docs to configure it?

516310460:
`extensions` have bugs

Need to be resolved

warmthsea:
> `extensions`有错误
> 
> 需要解决

好吧

516310460:
> > `extensions`有错误
> > 需要解决
> 
> 好吧

small issue

516310460:
When customizing `extensions`, it causes duplication, and the data is as follows:
```ts
import { defineNuxtConfig } from 'nuxt'

export default defineNuxtConfig({
  extensions: [
    ".js",
    ".jsx",
    ".mjs",
    // ".ts",
    ".tsx",
    ".vue"
  ]
})

```
nuxtConfig.extensions: [
   '.js',
   '.jsx',
   '.mjs',
   '.ts',
   '.tsx',
   '.vue',
   '.js',
   '.jsx',
   '.mjs',
   '.tsx',
   '.vue'
]

516310460:
PR: https://github.com/nuxt/framework/pull/6930

pi0:
You can use `ignore` option in `nuxt.config` to prevent `.ts` files inside pages being scanned.

See example:

```js
[...]

PR #6930, opened 2022-08-25
fix(schema)!: override `extensions` instead of extending

516310460:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/_common.ts
+++ b/packages/schema/src/config/_common.ts
@@ -496,7 +496,7 @@ export default {
    */
   extensions: {
     // https://github.com/nuxt/framework/issues/6920
-    $resolve: val => val ? resolve(val) : ['.js', '.jsx', '.mjs', '.ts', '.tsx', '.vue']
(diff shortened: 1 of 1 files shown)


516310460:
https://github.com/nuxt/framework/issues/6920

pi0:
Thanks for this PR.

About linked issue, I've explained in https://github.com/nuxt/framework/issues/6920#issuecomment-1228541492 to use `ignore` (or `.nuxtignore`) we can avoid scanning specific patterns. This is best way to ensure a specific pattern or extension is not being resolved.

In general for array options sometimes it makes sense to allow overriding them instead of extending. In nuxt 2, we were doing [this](https://github.com/nuxt/nuxt.js/blob/dev/packages/config/src/options.js#L203) to add only mandatory extensions during normalization. (In edge-cases, a module can still do this to modify `nuxt.options`)

I think it doesn't worth introducing this breaking change finally because all of the default extensions are built-in and overriding can produce more unpredictable cases to resolve functions based on user config.

516310460:
> 

:)

PR #6966, opened 2022-08-26
docs(deploy): update command for csr prerendering

oumarbarry:
### ❓ Type of change

### 📚 Description

As tried, when you want to generate client-side only output, even with "ssr : false" in nuxt.config.ts,
"nuxi build" doesn't generate the correct output, it still generates node ssr output,
but nuxi generate does the job.

### DIFF
--- a/docs/content/2.guide/5.deploy/2.static-hosting.md
+++ b/docs/content/2.guide/5.deploy/2.static-hosting.md
@@ -19,7 +19,7 @@ npx nuxi generate
 
 ## Client-side Only Rendering
 
-If you don't want to prerender your routes, another way of using static hosting is to set the `ssr` property to `false` in the `nuxt.config` file. The `nuxi build` command will then output an `index.html` entrypoint like a classic client-side Vue.js application.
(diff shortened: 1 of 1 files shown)


docs/content/2.guide/5.deploy/2.static-hosting.md
```diff
@@ -19,7 +19,7 @@ npx nuxi generate
 
 ## Client-side Only Rendering
 
-If you don't want to prerender your routes, another way of using static hosting is to set the `ssr` property to `false` in the `nuxt.config` file. The `nuxi build` command will then output an `index.html` entrypoint like a classic client-side Vue.js application.
+If you don't want to prerender your routes, another way of using static hosting is to set the `ssr` property to `false` in the `nuxt.config` file. The `nuxi generate` command will then output an `index.html` entrypoint like a classic client-side Vue.js application.
```
pi0, line 22: Note: Actually `nuxi build` does the same to add pretender SPA routes. `generate` enables crawler only. But generate seem better for usage example here 👍🏼

Issue #4579, opened 2022-04-25
Support router in hash mode (spa)


1isten:
### Environment

------------------------------
- Operating System: `Darwin`
- Node Version:     `v14.19.1`
- Nuxt Version:     `3.0.0-rc.1`
- Package Manager:  `yarn@1.22.18`
- Builder:          `vite`
- User Config:      `ssr`, `app`, `typescript`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

-

### Describe the bug

How to config nuxt 3 router to use [hash-mode](https://router.vuejs.org/guide/essentials/history-mode.html#hash-mode)?

I know the docs is not finished yet, so I searched the source code, but it seems there is no way to use the `createWebHashHistory`

https://github.com/nuxt/framework/blob/0eebb0a390c53c05c5bcbcb4fbda64d59a8945b6/packages/nuxt/src/pages/runtime/router.ts#L64

Aareksio:
Hash mode would not work with SSR, as [clever people on StackOverflow explained](https://stackoverflow.com/questions/3664257/why-is-the-hash-part-of-the-url-not-available-on-the-server-side). Might be useful only in SPA mode - making this a feature request, I guess.

1isten:
I want to use Nuxt 3 to build an SPA and pack it into a desktop app using [Electron](https://electronjs.org/). According to the vue-router [docs](https://router.vuejs.org/api/#createwebhashhistory), hash mode is

> Useful for web applications with no host (e.g. file://) or when configuring a server to handle any URL isn't an option.

dominixz-dzvz:
I think I need this too. I want to do website that deploy on IPFS and I can't enter page by directly from Browser URL

SvetlozarValchev:
I didn't even know this wasn't part of 'core' features, I'm using Nuxt for purely static client-side web apps and always have set the router mode to 'hash' in Nuxt 2, but the lack of this option seems to be a breaking change if I want to migrate my projects to 3.0

o1lab:
Ahh we need this badly - we're almost through with migrating to nuxt3 and are struggling with this :sob: 

@danielroe : can you please help ?

danielroe:
This was implemented in https://github.com/nuxt/framework/pull/6980.

PR #6980, opened 2022-08-26
feat(nuxt): router with hash mode

pi0:
### 🔗 Linked issue

#4579

### ❓ Type of change

### 📚 Description

This PR adds support for hash history mode for router (client only) with a new option. In order to enable:

```js
import { defineNuxtConfig } from 'nuxt'

  ssr: false,
  router: {
    options: {
      hashMode: true
    }
  }
})
```

Remarks:
- This option only works for spa mode
- Implementation of `createWebHashHistory` is super simple. It basically [appends `#` to the base with location.host](https://github.com/vuejs/router/blob/main/packages/router/src/history/hash.ts#L35) and reuse  `createWebHistory`. So there is almost no bundle overhead for this option
- I know, API is not beautiful but this is also an edge option! Later, we might move `router.options` to `app.router` this way option would be simpler `app.router.hashMode: false` instead of `router.options.hashMode: false`. `router.mode` from Nuxt 2 is also deprecated. We could portal it but it makes it harder to merge with one options object.

TODO: 
- [ ] Add example to the docs
- [ ] Investigate issue with initial route pushing additional `/#` each time
- [ ] Somehow warn about not using `ssr: false` with `hashMode: true`


pranavxc:
Thanks for the PR, we have been struggling due to the issue since last week. As a workaround we have updated the build files under node_modules directly ( `node_modules/nuxt/dist/pages/runtime/router.mjs` ).

For enabling hash-based routing we had modified

```js
import {
  createRouter,
  createWebHistory,
  createMemoryHistory
} from "vue-router";
```

to 


```js
import {
  createRouter,
  createWebHashHistory as createWebHistory,
  createMemoryHistory
} from "vue-router";
```

And  fixed following issue mentioned in PR 

> Investigate issue with initial route pushing additional /# each time

by changing following 

```js
function createCurrentLocation(base, location) {
  const { pathname, search, hash } = location;
```

to 


```js
function createCurrentLocation(base, location) {
  const { pathname, search, hash } = new URL((location.hash || '').replace(/^#/, ''), location.origin);
```
---
As a temporary workaround it's working  :+1: waiting for your PR to merge.

bcakmakoglu:
@pi0 is there something we could potentially help with to get this PR merged quicker? 😄

pi0:
WoW I didn't expect this amount of welcoming for this particular feature <3 Surely we can move it to the edge. Only there is an issue that redirects to `#/#/` everytime page is refreshed. I'm worried we are missing something. When fixed we can move to the edge.

PR #7021, opened 2022-08-29
fix(vite): remove client manifest.json from public dir

### LINKED ISSUE #7020: client manifest is included in `public/`
I could not run PWA on production mode because PWA manifest get replaced by Nuxt manifest. 
is there any way to change the name of Nuxt manifest?

_Originally posted by @mjzarrin in https://github.com/nuxt/framework/discussions/7017_

danielroe:
### 🔗 Linked issue

resolves  #7020

### ❓ Type of change

### 📚 Description

Currently, vite manifest gets accidentally included in public assets dir. It is not used and the data within it is incorrect (i.e. non-normalised)

### 📝 Checklist

### DIFF
--- a/packages/vite/src/manifest.ts
+++ b/packages/vite/src/manifest.ts
@@ -52,4 +52,8 @@ export async function writeManifest (ctx: ViteBuildContext, css: string[] = []) 
 
   await fse.writeFile(resolve(serverDist, 'client.manifest.json'), JSON.stringify(manifest, null, 2), 'utf8')
   await fse.writeFile(resolve(serverDist, 'client.manifest.mjs'), 'export default ' + JSON.stringify(manifest, null, 2), 'utf8')
+
+  if (!ctx.nuxt.options.dev) {
+    fse.removeSync(resolve(clientDist, 'manifest.json'))
+  }
 }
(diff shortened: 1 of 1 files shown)


packages/vite/src/manifest.ts
```diff
@@ -52,4 +52,8 @@ export async function writeManifest (ctx: ViteBuildContext, css: string[] = [])
 
   await fse.writeFile(resolve(serverDist, 'client.manifest.json'), JSON.stringify(manifest, null, 2), 'utf8')
   await fse.writeFile(resolve(serverDist, 'client.manifest.mjs'), 'export default ' + JSON.stringify(manifest, null, 2), 'utf8')
+
+  if (!ctx.nuxt.options.dev) {
+    fse.removeSync(resolve(clientDist, 'manifest.json'))
```
pi0, line 57: Any reason not using async version to remove? Also maybe ensuring for existence of file before remove? (I think there is a `force` option too not sure works per-file)

danielroe: we should know it exists because we have just read the contents a few lines before - unless something deletes it in the mean time. are you worried about a race condition?

pi0 (approved):

Issue #7026, opened 2022-08-29
Factory function to create `useFetch` and `useAsyncData` with defaults


pi0:
### Describe the feature

It is common that a (bigger) project and library/module authors want to create a composable that is _based_ on either `useFetch` or `useAsyncData` and provides some defaults.

There is a [nice example](https://github.com/nuxt/framework/blob/712fdae8a580d1c514bf9ca852d2f2965e00b618/examples/advanced/encapsulate-use-fetch/api/utils/http/request.ts#L39) by @516310460 that demonstrates it. 

The basic idea is to create factory composables that wrap `useFetch` and `useAsyncData` with defaults while preserving the behavior:

- `export const useMyAPI = createUseFetch(/* defaults */)`
- `export const useMyAPI = createUseAsyncData(/* defaults */)`

Currently it is possible with [`$fetch.create`](https://github.com/unjs/ohmyfetch#%EF%B8%8F-create-fetch-with-default-options) to make `const useMyAPI = $fetch.create(/* defaults */)` but it is just plain fetch without benefits of `useFetch` with reactivity.

With this new factory function, we can also find an opportunity to auto-generate keys for these custom instances too based on eager scanning with AST or fast regex. Currently, if a wrapper is custom-made, the key and state of any usage are shared and the same.

### Additional information

- [ ] Would you be willing to help implement this feature?
- [X] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

516310460:
I think the benefit of using `useFetch` is reactivity
The possible focus is how to constrain your own parameters and response data, and the code structure is clear and easy to use

edgarsn:
In context of this feature, I would like to point out suggestion. We could allow defaults not only as an object, but as a function as well. 

My current implementation is something like this:

```typescript
export function useCustomFetch<ResT = void,
    ErrorT = Error,
    ReqT extends NitroFetchRequest = NitroFetchRequest,
    _ResT = ResT extends void ? FetchResult<ReqT> : ResT,
    Transform extends (res: _ResT) => any = (res: _ResT) => _ResT,
    PickKeys extends KeyOfRes<Transform> = KeyOfRes<Transform>>(
    request: Ref<ReqT> | ReqT | (() => ReqT),
[...]

PR #7055, opened 2022-08-30
fix(nuxt): use shared state for asyncData

pi0:
### 🔗 Linked issue

#4758 #5078 #5738

### ❓ Type of change

### 📚 Description

With the initial implementation of `useAyncData`, we were using a shared promise to avoid running two asyncData promises with the same key at the same time but the fetching state and data was not shared. It makes problems if someone tries to make a higher-order composable that **returns a new instance of asyncData with the same key**.

This PR resolves it using a shared `_asyncData` map in nuxtApp interface also adding tests from #5738 by @OhB00.

**Remarks:**

We probably can iterate over this. Normally each useAsyncData should be **unique** based on its key and a factory pattern shall be introduced for `createAsyncData` and `createFetch` (discussion: #7026)

I will follow up with more PRs to have proper changelog.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -114,11 +114,16 @@ export function useAsyncData<
 
   const useInitialCache = () => (nuxt.isHydrating || options.initialCache) && nuxt.payload.data[key] !== undefined
 
-  const asyncData = {
-    data: ref(useInitialCache() ? nuxt.payload.data[key] : options.default?.() ?? null),
-    pending: ref(!useInitialCache()),
-    error: ref(nuxt.payload._errors[key] ?? null)
-  } as AsyncData<DataT, DataE>
+  // Create or use a shared asyncData entity
+  if (!nuxt._asyncData[key]) {
+    nuxt._asyncData[key] = {
+      data: ref(useInitialCache() ? nuxt.payload.data[key] : options.default?.() ?? null),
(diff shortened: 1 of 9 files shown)


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -114,11 +114,16 @@ export function useAsyncData<
...
   const useInitialCache = () => (nuxt.isHydrating || options.initialCache) && nuxt.payload.data[key] !== undefined
 
-  const asyncData = {
-    data: ref(useInitialCache() ? nuxt.payload.data[key] : options.default?.() ?? null),
-    pending: ref(!useInitialCache()),
-    error: ref(nuxt.payload._errors[key] ?? null)
-  } as AsyncData<DataT, DataE>
+  // Create or use a shared asyncData entity
+  if (!nuxt._asyncData[key]) {
+    nuxt._asyncData[key] = {
```
danielroe, line 119: @pi0 Have you checked whether this retains reactivity if this is created outside a component setup (eg in plugin), or when it is used after the original component that created it has unmounted?

pi0: You can try it (as mentioned in the description we need to iterate over to fix the related issue).

BTW really the use case of useAsyncData is to be used within the component lifecycle, not the plugins even if partially possible. 


Good point about reactivity after unmounting. Will test this.

PR #6826, opened 2022-08-22
feat(nuxt): add `setLayout` utility

516310460:
### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #7075, opened 2022-08-30
feat(nuxt): add `setLayout` utility (#6826)

### LINKED ISSUE #6977: `setLayout`/`setPageLayout` utility
Originally made by @516310460 in #6826 moved to [`feat/set-page-layout`](https://github.com/nuxt/framework/compare/feat/set-page-layout?expand=1)

Todo: (@danielroe)

- [ ] rename to `setPageLayout` to make it clearer how this composable operates
- [ ] update more docs to refer to this composable
- [ ] fix behaviour in plugins to use router hook as well

danielroe:
### 🔗 Linked issue

resolves #6977

### ❓ Type of change

### 📚 Description

This adds a `setPageLayout` utility to handle complexities associated with changing layout keyed off of `route.meta`. Namely, it's complicated to persist the change from server -> client. This persists the server-set layout via a custom `useState` which is only used on the initial hydration. Subsequent layout changes are reactive and don't need to be hydrated.

> Originally implemented by @516310460 ❤️

### 📝 Checklist


pi0:
Following up https://github.com/nuxt/framework/issues/6977#issuecomment-1231868127, i understand there is a way to use prop binding for global usage of `NuxtLayout` but I'm thinking we could easily (what I suppose) use `_layout` state for **default value** of `:name` for `<NuxtLayout>`. This way utility can be already universal as `setLayout`.

danielroe:
I did consider updating the default value but feel that it's not optimal. Consider this use case:

1. user visits a blog page (`~/pages/blog/[slug].vue`). This blog entry is full-width, so we set the layout to `full-width` using `setPageLayout` (so that it can be hydrated correctly on client-side).
2. user changes route to `~/pages/index.vue` or `~/pages/about.vue`. Now these are also displaying `full-width` layout unless there is a hard-coded `definePageMeta({ layout: 'default })` in the page.

If we want to update the default layout (and I'm not sure we should) then a `setDefaultLayout` utility or similar might be a better approach.

516310460:
Hi, brothers, do you remember what I said at the beginning. `setLayout` extends the configuration. The previous default value is name. Now it is completely possible to extend the global setting. The user can set it according to different business needs

pi0 (approved):

pi0:
Makes sense to introduce setDefaultLayout utility for universal usage 👍🏼

PR #7089, opened 2022-08-30
docs(deploy): add node cluster mode

Mastercuber:
### ❓ Type of change

### 📚 Description

This PR adds a section to the docs for starting load balanced network processes with a Node.js internal module.

### 📝 Checklist


pi0:
Hi @Mastercuber and thanks for nice PR. I've made https://github.com/unjs/nitro/issues/451 to introduce cluser as a nitro preset. Would you like to help implementing that?

Mastercuber:
Yes I would like to.
I made me some thoughts about how could a cluster module looks like, but maybe it's better to switch to the issue in [unjs](https://github.com/unjs/nitro/issues/451).

pi0:
Surely. Feel free to share your thoughts in https://github.com/unjs/nitro/issues/451 but already a similar usage of current example to implement preset should be amazing start.

docs/content/2.guide/5.deploy/1.node-server.md
```diff
@@ -49,6 +49,32 @@ module.exports = {
...
 ```
 
+### Using Node's Cluster Module
+
+Another way of starting a multi-process server with a Node.js internal module is to use the [cluster](https://nodejs.org/dist/latest/docs/api/cluster.html):
+
+```js
+import cluster from 'cluster'
+import os from 'os'
+import process from 'process'
```
stafyniaksacha, line 59: I suggest using `node:` protocol for built-in modules

```js
import cluster from 'node:cluster'
import os from 'node:os'
import process from 'node:process'
```js

Mastercuber: I will add this to the example.

Mastercuber: Removed it because it reads a bit better.

pi0:
@Mastercuber I think it would be better if you put changes to nitro PR. (I can help on this if you like). This way we can include preset in next release :)

Mastercuber:
I'm still writing at a comment. The nitro repo is forked and cloned, but I have trouble with `pnpm` or the registry I use. I will investigate this after finishing the comment. Afterwards I can create a PR with the first example code

pi0:
New preset is avialable since nitro 0.5.1 (updating lockfile). Thanks for helping on this 🚀

PR #7104, opened 2022-08-31
fix(schema): sync types of vite v3.x

brownsugar:
### ❓ Type of change

### 📚 Description

1. `globEager` is [deprecated](https://vitejs.dev/guide/migration.html#general-changes)
2. add some types according to [latest types](https://github.com/vitejs/vite/blob/main/packages/vite/types/importGlob.d.ts)
3. `glob` will always exists in `import.meta`, I think there is no need to set it as an optional field here?
If we try to use optional chain in the code, it will break this feature in Vite 3:
```ts
// Works in Vite 2 but not in Vite 3
const result = import.meta.glob?.('assets/images/svg/**/*.svg')
console.log(result) // `undefined` in Vite 3

// Works fine
const result = import.meta.glob('assets/images/svg/**/*.svg')
console.log(result) // { '.../...svg': () => import(...) }
```

### 📝 Checklist

### DIFF
--- a/packages/schema/src/types/global/vite.ts
+++ b/packages/schema/src/types/global/vite.ts
@@ -20,15 +20,18 @@ export interface ViteHot {
 
 export interface ViteGlobOptions {
   as?: string
+  eager?: boolean
+  import?: string
+  query?: string | Record<string, string | number | boolean>
+  exhaustive?: boolean
 }
 
 export interface ViteImportMeta {
+  url: string
+
   /** Vite client HMR API - see https://vitejs.dev/guide/api-hmr.html */
   readonly hot?: ViteHot
 
   /** vite glob import utility - https://vitejs.dev/guide/features.html#glob-import */
-  glob?(pattern: string, options?: ViteGlobOptions): Record<string, () => Promise<Record<string, any>>>
-
-  /** vite glob import utility - https://vitejs.dev/guide/features.html#glob-import */
-  globEager?(pattern: string, options?: ViteGlobOptions): Record<string, Record<string, any>>
+  glob (glob: string | string[], options?: ViteGlobOptions): Record<string, () => Promise<Record<string, any>>>
 }
--- a/packages/schema/src/types/global/vite.ts
+++ b/packages/schema/src/types/global/vite.ts
@@ -27,8 +27,6 @@ export interface ViteGlobOptions {
 }
 
 export interface ViteImportMeta {
-  url: string
-
   /** Vite client HMR API - see https://vitejs.dev/guide/api-hmr.html */
   readonly hot?: ViteHot
 


packages/schema/src/types/global/vite.ts
```diff
@@ -20,15 +20,18 @@ export interface ViteHot {
...
   as?: string
+  eager?: boolean
+  import?: string
+  query?: string | Record<string, string | number | boolean>
+  exhaustive?: boolean
 }
 
 export interface ViteImportMeta {
+  url: string
+
```
danielroe, line 31:
```suggestion
```

This is present in https://github.com/nuxt/framework/blob/d68318f9d5ee25f6377edb866cd1b5846e8bfdce/packages/schema/src/types/global/index.ts#L6-L13.

packages/schema/src/types/global/vite.ts
```diff
@@ -20,15 +20,16 @@ export interface ViteHot {
...
 export interface ViteImportMeta {
   /** Vite client HMR API - see https://vitejs.dev/guide/api-hmr.html */
   readonly hot?: ViteHot
 
   /** vite glob import utility - https://vitejs.dev/guide/features.html#glob-import */
-  glob?(pattern: string, options?: ViteGlobOptions): Record<string, () => Promise<Record<string, any>>>
-
-  /** vite glob import utility - https://vitejs.dev/guide/features.html#glob-import */
-  globEager?(pattern: string, options?: ViteGlobOptions): Record<string, Record<string, any>>
+  glob (glob: string | string[], options?: ViteGlobOptions): Record<string, () => Promise<Record<string, any>>>
```
danielroe, line 34: This is meant to be optional as the declaration is also present in webpack builder.

```suggestion
  glob?(glob: string | string[], options?: ViteGlobOptions): Record<string, () => Promise<Record<string, any>>>
```

danielroe, line 34: Tricky. This is meant to be optional as the declaration is also present in webpack builder.

It might be worth reconsidering whether we add these types within `vite-builder` rather than schema, so it can be more precisely targeted to both the vite version and the builder that is being used. wdyt @pi0?

pi0: I like the idea of moving types but that also makes a tricky question of how we tie vite versions to Nuxt releases (it uses caret). As you also mentioned, reason to have types here was that they are shared between different bundlers. I think we can revise this refactor later but you are right about being optional.

pi0:
```suggestion
  glob?(glob: string | string[], options?: ViteGlobOptions): Record<string, () => Promise<Record<string, any>>>
```

danielroe: @brownsugar seemed to say that the optical chaining

pi0: It really seems a bug in vite to me with static replacement (import.meta by spec is an object while static replacement is not covering another valid syntax of optional chaining...)

Okay then let's keep it non optional until issue is fixed. Webpack is not that important.

pi0 (approved):

PR #7118, opened 2022-08-31
feat(nuxi): display nuxt and nitro versions for dev and build commands

### LINKED ISSUE #6935: Show nitro version in `nuxi info` and cli banner
### Describe the feature

It should help to diagnose issues with the known nitro version. It is also kinda important before having #6937.

### Additional information

### Final checks

pi0:
### 🔗 Linked issue

Resolves #6935

### ❓ Type of change

### 📚 Description

With this PR, we show (installed) versions of Nuxt (it can be different than Nuxi!) and Nitro.

image

Remarks: I used `_require` instead of getPkg or inference from `package.json` to ensure we show actually installed versions and use native resolver. This also reduces CLI entry dependencies. Later we should refactor this.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/utils/banner.ts
+++ b/packages/nuxi/src/utils/banner.ts
@@ -19,7 +19,7 @@ export function showVersions (cwd: string) {
     } catch { /* not found */ }
     return ''
   }
-  const pkgs = ['nitropacl', 'vite', 'webpack', 'vue']
+  const pkgs = ['nitropack', 'vite', 'webpack', 'vue']
   console.log(
     green(getPkgWithVersion('nuxt') || getPkgWithVersion('nuxt-edge') /* bridge */ || 'nuxt'),
     gray('running with'),
--- a/packages/nuxi/src/utils/banner.ts
+++ b/packages/nuxi/src/utils/banner.ts
@@ -5,23 +5,23 @@ import { version } from '../../package.json'
 
 export function showBanner (_clear?: boolean) {
   if (_clear) { clear() }
-  console.log(gray(`nuxi@${version}`))
+  console.log(gray(`Nuxi ${version}`))
 }
 
 export function showVersions (cwd: string) {
   const _require = createRequire(cwd)
-  const getPkgWithVersion = (pkg: string) => {
+  const getPkgWithVersion = (pkg: string, name: string) => {
     try {
       const { version } = _require(`${pkg}/package.json`)
       if (version) {
-        return pkg + '@' + version
+        return name + ' ' + version
       }
     } catch { /* not found */ }
     return pkg
   }
   console.log(
-    green(getPkgWithVersion('nuxt')),
+    green(getPkgWithVersion('nuxt', 'Nuxt')),
     gray('running with'),
-    gray(getPkgWithVersion('nitropack'))
+    gray(getPkgWithVersion('nitropack', 'Nitro'))
   )
 }
--- a/packages/nuxi/src/utils/banner.ts
+++ b/packages/nuxi/src/utils/banner.ts
@@ -5,7 +5,7 @@ import { version } from '../../package.json'
 
 export function showBanner (_clear?: boolean) {
   if (_clear) { clear() }
-  console.log(green(`Nuxi ${version}`))
+  console.log(green(`nuxi ${version}`))
 }
 
 export function showVersions (cwd: string) {
--- a/packages/nuxi/src/utils/banner.ts
+++ b/packages/nuxi/src/utils/banner.ts
@@ -17,12 +17,11 @@ export function showVersions (cwd: string) {
         return pkg + '@' + version
       }
     } catch { /* not found */ }
-    return ''
+    return pkg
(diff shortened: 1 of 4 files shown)


packages/nuxi/src/utils/banner.ts
```diff
@@ -1,8 +1,24 @@
...
+  console.log(gray(`nuxi ${version}`))
+}
+
+export function showVersions (cwd: string) {
+  const _require = createRequire(cwd)
+  const versions = []
+  for (const pkg of ['nuxt', 'nitropack', 'vite', 'webpack']) {
+    try {
+      const { version } = _require(`${pkg}/package.json`)
+      const fullName = pkg + '@' + version
```
danielroe, line 17: Seems slightly inconsistent:

```
nuxi 3.0.0-rc.8
nuxt@3.0.0-rc.8 vite@3.0.9 webpack@5.74.0
```

Maybe drop the `@` in the second line or make versions gray and differentiate nuxt differently - something like

```
nuxt 3.0.0-rc.8 running with vite 3.0.9 and webpack 5.74.0
```

pi0: Thinking to choose last one. If we don't have `@` seperator, need to use `and` or `, ` to split each which makes longer if we want to add more items. Also is more like actual dependency@version representation for npm.

pi0:
Update: Simplified to only show Nuxt and Nitro versions as original proposal. It was leading to verbosity of showing all sub deps and more complex code. We can always revise to add more info when needed.

packages/nuxi/src/utils/banner.ts
pi0: (check updated. finally simplified to two with alternative and shorter display name)

danielroe: Sorry - missed this! Looks great.

packages/nuxi/src/utils/banner.ts
```diff
@@ -1,8 +1,27 @@
...
+export function showVersions (cwd: string) {
+  const _require = createRequire(cwd)
+  const getPkgWithVersion = (pkg: string, name: string) => {
+    try {
+      const { version } = _require(`${pkg}/package.json`)
+      if (version) {
+        return name + ' ' + version
+      }
+    } catch { /* not found */ }
+    return pkg
```
danielroe, line 20: Note, this prints

```
Nuxi 3.0.0-rc.8
Nuxt 3.0.0-rc.8 running with nitropack
```

if the version cannot be determined. That's absolutely fine here (also because _surely_ there must be nitropack in the project) but if we add more options later we may need to revisit approach.

pi0: Good point! Actually, it probably prints `Nuxt running with Nitro` in case deps are not found..

packages/nuxi/src/utils/banner.ts
danielroe: One more suggestion, though happy with what you have too....

CleanShot 2022-09-01 at 10 02 05

packages/nuxi/src/utils/banner.ts
pi0: Made refactor to add back protection.

packages/nuxi/src/utils/banner.ts
pi0: I like removing extra `with` to simplify. Importance of both label (nitro, vite webpack) and version are the same in regard of display. If we gray out I think is less verbose coloring to keep tune same.

danielroe: (This is of course for another PR, right? Not trying to delay this - using nitro version alone looks good to me.) In that case, I think we could make all of them greyed out except nuxt version for ultra-minimal look.

pi0: You are not delaying anything :) Are you thinking to use white instead of green? One other option is also bolding nuxt version.

pi0: image

pi0: wyt?

danielroe: Yes, I was thinking white instead of green. Bolding version looks nice 👍

pi0: It is how it likes with white (default term color)

image

danielroe: Thinking about it, I would make Nuxt color same as the Nuxt version personally. And with bold versions, I think it could all be grey. The Nuxi/Nuxt similarity also is a bit confusing as it may look like duplication.

pi0: Green is good?

pi0: image

danielroe: perfect

manniL: Looking sharp 🔥

PR #7120, opened 2022-08-31
docs: add note about fetching data on initial load

### LINKED ISSUE #6921: useAsyncData's data._rawValue is null
### Environment

3.0.0-rc.8

### Reproduction

https://stackblitz.com/edit/nuxt-starter-qfirv9?file=app.vue

### Describe the bug

When I send the first request using useAsyncData, and then send the second request when the first request is not finished, the data._rawValue of the first request will be null

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #6921

### ❓ Type of change

### 📚 Description

This adds a note about a limitation of `server: false` data fetching and updates a couple of other minor inaccuracies..

### 📝 Checklist


pi0 (approved):
Thanks for pointing this out. It seems a buggy behavior to me since we could make data available after awaiting for client-side only. But let's document warning until fixing

danielroe:
I don't think we can wait in setup, or we will cause a hydration error.

Issue #7122, opened 2022-08-31
Support memory history from Vue Router in client side (createMemoryHistory)


jonyw4:
### Describe the feature

I want to choose the history method of Vue Router in nuxt config file (in my case, to use specifically the `createMemoryHistory`). 

In Nuxt 2 we had that available via [router mode](https://nuxtjs.org/docs/configuration-glossary/configuration-router#mode). Although [we can config vue router](https://github.com/nuxt/framework/pull/3485), we can't change the history method [because is based on `process.client` boolean](https://github.com/nuxt/framework/blob/main/packages/nuxt/src/pages/runtime/router.ts#L57-L60).

### Proposal

Accept a new `router.history` prop in `nuxt.config.ts`
```ts
defineNuxtConfig({
  router: {
    history: 'memory'
  }
})
```

That can accept both strings or a router history implementation, using the `RouterHistory` interface from Vue Router
```ts
'memory' | 'web-hash' | 'web-html5' |  RouterHistory
```

This can override the default ternary in the current router runtime

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

516310460:
See: https://v3.nuxtjs.org/guide/directory-structure/pages#router-options

jonyw4:
That looks great. Thank you @danielroe 🙇🏻‍♂️

PR #7129, opened 2022-09-01
feat(nuxt): custom `history` and `routes` for `app/router.options.ts`

danielroe:
### 🔗 Linked issue

#7122

### ❓ Type of change

### 📚 Description

Alongside first-class 'presets' like #7122 and https://github.com/nuxt/framework/issues/4579 which can be configured with a string, it might be useful to allow fully overriding router history + routes for advanced use cases (= escape hatch).

This allows a router options something like this:

```js
import type { RouterConfig } from '@nuxt/schema'
import { createMemoryHistory } from 'vue-router'

export default <RouterConfig> {
  history: process.client ? createMemoryHistory : undefined
}
```

Note that this does not replace https://github.com/nuxt/framework/pull/6980 as hash history has its own challenges with an SSR framework and needs its own implementation.

### 📝 Checklist


pi0:
I like the idea of being able to override history and routes via `app/router.options` file.

The factory function for `history` seems good idea too.

Maybe we can also pass default routes to a factory to allow overriding+modifiying routes from config?

(in both regards delaying this PR as well as history for after RC.9 to test better)

danielroe:
Great. Can look into factory routes option as well 👍

Big question in my mind (regardless of the factory function point) is tree-shaking - would be nice to support tree-shaking out auto-generated routes in the event that user fully overrides them, and I'm not sure that rollup will remove them from the build as is. (But then again, nothing stopping user from removing them entirely in the `pages:extend` hook so maybe I'm overthinking this.)

pi0:
Agreed that hook is probably best way to override routes. Do you mind adding function support for routes() in order to merge this?

packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -55,15 +55,16 @@ export default defineNuxtPlugin(async (nuxtApp) => {
   nuxtApp.vueApp.component('NuxtChild', NuxtPage)
 
   const baseURL = useRuntimeConfig().app.baseURL
-  const routerHistory = process.client
-    ? createWebHistory(baseURL)
-    : createMemoryHistory(baseURL)
+  const routerHistory = routerOptions.history?.(baseURL) ??
+    (process.client ? createWebHistory(baseURL) : createMemoryHistory(baseURL))
+
+  const routes = !routerOptions.routes ? _routes : Array.isArray(routerOptions.routes) ? routerOptions.routes : routerOptions.routes(_routes)
```
pi0, line 61: I vote to accept function for now. We might alternatively use a _merging strategy_ when `routes` as the array is provided...

```suggestion
  const routes = !routerOptions.routes ? _routes : routerOptions.routes(_routes)
```

danielroe: oh, I see - you mean support function only 👍

packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -55,15 +55,16 @@ export default defineNuxtPlugin(async (nuxtApp) => {
...
+    (process.client ? createWebHistory(baseURL) : createMemoryHistory(baseURL))
+
+  const routes = !routerOptions.routes ? _routes : routerOptions.routes(_routes)
 
   const initialURL = process.server ? nuxtApp.ssrContext!.url : createCurrentLocation(baseURL, window.location)
   const router = createRouter({
     ...routerOptions,
     history: routerHistory,
-    routes
+    routes: routerOptions.routes || routes
```
pi0, line 67: Shouldn't be resolved routes from L61?

pi0 (approved):

PR #7190, opened 2022-09-03
fix(head): case `http-equiv` correctly

harlan-zw:
### ❓ Type of change

### 📚 Description

Currently, when using the `Meta` component, it's not possible to provide a value for `http-equiv`. This is because `http-equiv` gets turned to  `httpEquiv`, from Vue prop passing. 

`httpequiv` is the final DOM result and it's not honoured by browsers

This PR simply swaps out the key for the correct one, I beleive this will be a temporary fix if unjs is working on a dom handling package.

#### Reproduction

https://stackblitz.com/edit/nuxt-starter-jdegdy?file=app.vue

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/head/runtime/components.ts
+++ b/packages/nuxt/src/head/runtime/components.ts
@@ -197,9 +197,17 @@ export const Meta = defineComponent({
     httpEquiv: String as PropType<HTTPEquiv>,
     name: String
   },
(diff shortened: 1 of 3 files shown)


danielroe (approved):
Thank you ❤️

pi0:
Thanks!

PR #7243, opened 2022-09-05
perf(nuxt): use a cache for createClientOnly components

### LINKED ISSUE #7226: Client only components (`x.client.vue`) force remount on prop change
### Environment

- Operating System: `Linux`
- Nitro Version:    `0.5.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-akmejo?file=app.vue,components%2FChild.client.vue,components%2FParent.vue

### Describe the bug

If props passed to a client only component (https://v3.nuxtjs.org/guide/directory-structure/components/#client-components)
change it forces a remount/rerender of the entire component, which really kills performance in many cases. If you remove the .client from the child component and refresh it works as expected.
Happy I found figured it out now but man that one was painful to find😭

### Additional context

_No response_

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve #7226

### ❓ Type of change

### 📚 Description

Hi :wave: , this PR improve (or fix) `createClientOnly()` performance by adding a cache. Before this, vue always re-render and remount completely the client component when the parent triggers a re-render, even if there's no changes in props or anything else. I think this might caused by the fact that `createClientOnly` return each time a different component definition (object reference).

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/client-only.mjs
+++ b/packages/nuxt/src/app/components/client-only.mjs
@@ -1,4 +1,4 @@
-import { ref, onMounted, defineComponent, createElementBlock, h, Fragment } from 'vue'
+import { ref, onMounted, defineComponent, createElementBlock, h } from 'vue'
 
 export default defineComponent({
   name: 'ClientOnly',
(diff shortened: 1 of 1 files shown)


huang-julien:
i'll test this in a few hours but this pr work on the referenced issue

antfu:
Thank you! But I think https://github.com/nuxt/framework/pull/7245 would be a more correct fix.

Caching could be a nice addition, but I would suggest using `WeakMap` to map the component reference directly.

huang-julien:
I've removed the comment on the first commit but had to revert this change on 6e7c5ea9 due to test fixtures being broken. Setting back Fragment on 6e7c5ea9 wasn't a bug fix but more of a refactor, However it still throws warnings in dev environment.

I'll close this PR in favor of #7245 and open another PR (after #7245 being merged) adding a cache based on @antfu suggestion

PR #7245, opened 2022-09-05
fix(nuxt): import and wrap client-only components once

antfu:
### 🔗 Linked issue

fix https://github.com/nuxt/framework/issues/7226

### ❓ Type of change

### 📚 Description

Previously `createClientOnly` was called inside the render function, which unexpectedly created new components on each render.

Before:
Screen Shot 2022-09-05 at 20 30 12

After:
Screen Shot 2022-09-05 at 20 48 00

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/client-only.mjs
+++ b/packages/nuxt/src/app/components/client-only.mjs
@@ -24,7 +24,7 @@ export function createClientOnly (component) {
     // override the component render (non script setup component)
     component.render = (ctx, ...args) => {
       return ctx.mounted$
-        ? h(Fragment, null, [h(_render(ctx, ...args), ctx.$attrs ?? ctx._.attrs)])
+        ? h(Fragment, ctx.$attrs ?? ctx._.attrs, _render(ctx, ...args))
         : h('div', ctx.$attrs ?? ctx._.attrs)
     }
   } else if (_template) {
@@ -34,6 +34,7 @@ export function createClientOnly (component) {
       <template v-else><div></div></template>
     `
   }
+
   return defineComponent({
     ...component,
     setup (props, ctx) {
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/components/client-only.mjs
```diff
@@ -19,38 +19,39 @@ export default defineComponent({
...
 
 export function createClientOnly (component) {
-  const { setup, render: _render, template: _template } = component
-  if (_render) {
+  const clone = { ...component }
+
+  if (clone.render) {
     // override the component render (non script setup component)
-    component.render = (ctx, ...args) => {
+    clone.render = (ctx, ...args) => {
```
antfu, line 26: Doing this because I think we should not mutate the input component. Not destructuring it would also make HMR works better as the HMR might replace the `render` function at runtime.

packages/nuxt/src/components/loader.ts
```diff
@@ -68,20 +68,26 @@ export const loaderPlugin = createUnplugin((options: LoaderOptions) => {
...
           const isClientOnly = component.mode === 'client'
           if (isClientOnly) {
             imports.add(genImport('#app/components/client-only', [{ name: 'createClientOnly' }]))
           }
           if (lazy) {
             imports.add(genImport('vue', [{ name: 'defineAsyncComponent', as: '__defineAsyncComponent' }]))
-            imports.add(`const ${identifier}_lazy = /*#__PURE__*/ __defineAsyncComponent(${genDynamicImport(component.filePath)})`)
+            identifier += '_lazy'
+            imports.add(`const ${identifier} = /*#__PURE__*/ __defineAsyncComponent(${genDynamicImport(component.filePath)})`)
             return isClientOnly ? `/*#__PURE__*/ createClientOnly(${identifier}_lazy)` : `${identifier}_lazy`
```
danielroe, line 81: should we also update this?

danielroe, line 81: should we also update this, something like:

```suggestion
            if (!isClientOnly) {
              return identifier
            }
```

danielroe (approved):
This looks good to me. 👍 Thanks @antfu. ❤️

pi0 (approved):
Nice changes 💯

Issue #7156, opened 2022-09-02
allow disabling client-side JS entirely


danielroe:
For some use cases, users may wish to have a purely HTML + CSS website where interactivity is not required. Shipping JS to hydrate such a site is unneeded.

For these use cases, we could consider support a top level option (or a per-path route rule) to disable rendering JS preload hints + script tags within the renderer.

Some initial steps:

- [x] create module: [nuxt-zero-js](https://github.com/danielroe/nuxt-zero-js)

**Note**: We may also need to update https://github.com/nuxtlabs/vue-telescope-api to accurately detect Nuxt on such sites. At the moment, it is unaware that https://nuxt.new/ is built with Nuxt 😆

pi0:
Pure HTML is not a normal output mode. We were considering it with @Atinux since Nuxt 2 but finally decided to not introduce but i really love this idea specially combined with reactive component islands. We can probably have a module to try this mode ahead of time. Later can be joined with route rules to set per page (or group of pages) bases.


Also good point about detection! What about a magical `<-- nuxt -->` comment replacing strings. (There is also `X-Powered-By` header but not working for static builds)

516310460:
This is great, just like building islands, non-interactive and interactive loading of js scripts

harlan-zw:
It'd be great to see it in the core with some sort of support for island hydration

516310460:
> It'd be great to see it in the core with some sort of support for island hydration

Yes, this is the beginning of a revolution

HelloAlexPan:
This would be the key feature for us not to switch to Astro

miclgael:
I have written post-build hacks to do this in nuxt 2 and it is a real pain point. Kudos @danielroe

516310460:
> 

Support nuxt2?

danielroe:
You can do this in Nuxt 2 by setting `render.injectScripts` to false.

kstraszewski:
The feature looks great, it would be amazing to use it per route!

ManasMadrecha:
Having absolutely no JS is not useful; we need some JS in header or in button to toggle dark mode.

It would be better to have this feature per component, like in Astro.

PR #7248, opened 2022-09-05
feat(nuxt): support experimental flag to render no client-side js

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/7156

### ❓ Type of change

### 📚 Description

This allows turning off all Nuxt-injected JS (preload/prefetch `<link>` and `<script>`). It does not affect user code, which can still inject scripts. It does remove all scripts from buildDir, though perhaps this could be reconsidered in future? (wdyt @pi0?)

> **Warning**
> This will only be advisable for the very small minority of sites that would not benefit from client-side JS.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -82,6 +82,7 @@ async function initNuxt (nuxt: Nuxt) {
     addWebpackPlugin(TreeShakePlugin.webpack({ sourcemap: nuxt.options.sourcemap, treeShake: removeFromClient }), { server: false })
   }
 
+  // TODO: [Experimental] Avoid emitting assets when flag is enabled
   if (nuxt.options.experimental.noScripts) {
     nuxt.hook('build:manifest', async (manifest) => {
       for (const file in manifest) {
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -5,6 +5,8 @@ import { loadNuxtConfig, LoadNuxtOptions, nuxtCtx, installModule, addComponent, 
 // Temporary until finding better placement
 /* eslint-disable import/no-restricted-paths */
 import escapeRE from 'escape-string-regexp'
+import fse from 'fs-extra'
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/core/nuxt.ts
```diff
@@ -80,6 +82,17 @@ async function initNuxt (nuxt: Nuxt) {
     addWebpackPlugin(TreeShakePlugin.webpack({ sourcemap: nuxt.options.sourcemap, treeShake: removeFromClient }), { server: false })
   }
 
+  if (nuxt.options.experimental.noScripts) {
```
pi0, line 85:
```suggestion
  // TODO: [Experimental] Avoid emitting assets when flag is enabled
  if (nuxt.options.experimental.noScripts) {
```

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -40,6 +40,8 @@ const getServerEntry = () => import('#build/dist/server/server.mjs').then(r => r
 // @ts-ignore
 const getSSRStyles = (): Promise<Record<string, () => Promise<string[]>>> => import('#build/dist/server/styles.mjs').then(r => r.default || r)
 
+const JS_HINT_RE = /<[^>]+(\.m?js"|as="script")[^>]*>/g
```
pi0, line 43: Seems unused :)

packages/nuxt/src/core/nuxt.ts
danielroe: What are you suggesting - not emitting them in build in the first place?

pi0 (approved):
LGTM to try!

packages/nuxt/src/core/nuxt.ts
pi0: Yes.

pi0: Indeed probably tricky but emiting assets is last step in rollup. I'm positive we can do this if finally it was a viable option to keep :)

Issue #7195, opened 2022-09-03
avoid double-downloading of inlined css


danielroe:
Possibly we could move CSS resource hints from preload -> prefetch, and/or change stylesheet to load after site is interactive, maybe following one of [these strategies](https://github.com/GoogleChromeLabs/critters#preloadstrategy).

pi0:
Maybe we can find a way to map client styles with server ids in order to remove them from manifest when inlining is enabled for them? Currently, on edge and after (#7160), there are duplicated styles (both inline `<style>` and external `<link>`)

PR #7264, opened 2022-09-05
perf(vite): remove duplicate css links from rendered page when inlined

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/7195

### ❓ Type of change

### 📚 Description

This improves the performance of inlining styles in SSR. We can complete remove styles from the manifest for individual/non-entry chunks, as we always know that these will be inlined. For entry CSS, it's a little more complex. Client-only components may have their CSS in the entry file, so we can't entirely exclude entry CSS. (And we deliberately tree shake client-only components from the server build.) I have lowered entry CSS priority via prefetching it; vite's preload helper then loads it.

**Note**: I think this is soluble, but it may require a secondary PR, which will also need to stop vite from preloading the entry css...

### 📝 Checklist


packages/vite/src/server.ts
```diff
@@ -113,12 +114,35 @@ export async function buildServer (ctx: ViteBuildContext) {
...
       shouldInline: typeof ctx.nuxt.options.experimental.inlineSSRStyles === 'function'
         ? ctx.nuxt.options.experimental.inlineSSRStyles
         : undefined
     }))
+
+    // Remove CSS entries for files that will have inlined styles
+    ctx.nuxt.hook('build:manifest', (manifest) => {
+      for (const key in manifest) {
+        const entry = manifest[key]
+        const shouldRemoveCSS = chunksWithInlinedCSS.has(key)
```
pi0, line 130: What will happen for a component with ssr inline syle enabled when rendering in hybrid/spa mode?

danielroe: Rendering in hybrid/spa mode would not be adding a `<link>` for the CSS that is specific to the page anyway (as `ssrContext.modules` would be empty). So there would be no change from these lines (L127-133).

However, there would be a change to the entry CSS, which would be prefetched and then loaded on client-side rather than preloaded. (This would have a negative performance impact.) 

```diff
-   <link rel="preload" as="style" href="/_nuxt/entry.6e3e3577.css">
-   <link rel="stylesheet" href="/_nuxt/entry.6e3e3577.css">
+   <link rel="prefetch stylesheet" href="/_nuxt/entry.6e3e3577.css">
```

Note that this also has a negative impact for client-only components for SSR pages. We could disable L135-R143, and though we would have double-loaded entry CSS values, we would not have any negative perf impact for client-side/or hybrid spa:

https://github.com/nuxt/framework/pull/7264/files#diff-9659ec6cc10298d2e39de216bea87573d29efdb58b9fd2e8fc9fc4404719bcaaR135-R143

pi0 (approved):

DamianGlowala:
I've tracked the PR which breaks my existing project and this one does it. There's an infinite refresh of the page with 'Reloading server...' message and a console log (`undefined`) in the terminal. Will provide any additional info if needed.

image
image

harlan-zw:
I haven't looked into it too much but something in RC.10 is causing this issue https://github.com/windicss/nuxt-windicss/issues/180, this PR seems like a likely candidate. 

Any ideas before I dig into it?

Issue #7251, opened 2022-09-05
useLazyAsyncData with initialCache: false is still using cache


StefanCardnell:
### Environment

- Operating System: `Linux`
- Node Version:     `v18.5.0`
- Nuxt Version:     `3.0.0-rc.9`
- Nitro Version:    `0.5.0`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Example reproduction without setting up APIs:

1. `ssr: false` in `nuxt.config.ts`
2. Put this in a `pages` file:

```
const { data } = useLazyAsyncData(
  () => {
    console.log("fetching!");
    return fetch('/api/users')
  },
  { initialCache: false }
);

console.log(data.value);
```

3. Enter the page, open the console log and see that `data.value` is null because nothing has been fetched yet.
4. Leave the page, come back. See the the logged `data.value` is not-null but also that the console logged `fetching` indicating it is at least re-fetching the `api/users` URL.

### Describe the bug

When returning to the page the second time, `data` should be set to null initially and shouldn't use any cached value as this is what `initialCache` is trying to avoid.

Although it is eventually replaced by the refreshed data, its value should be null initially.

Note that I am not using a `key` for `useLazyAsyncData`, so the cache is based on the file+line number.

516310460:
Thanks

PR #7265, opened 2022-09-06
fix(nuxt): `initialCache: false` cache not deleted

516310460:
### ❓ Type of change

### 📝 Checklist


516310460:
https://github.com/nuxt/framework/issues/7251

516310460:
Solve the problem that cache data is not deleted when `initialCache` is false, and there is a large amount of invalid data, which will affect performance. Data should be deleted when `onUnmounted` to reduce stack and improve performance

pi0:
Hi, @516310460 sorry took long to check on this. The initial cache is disabled and removed (#8885). 

Releasing data from memory is a good idea. Both from `_asyncData` and `payload.data` without cache. At this point, it might introduce unwanted regressions and we are close to the 3.0.0 release. 

I have created #8894 to not forget about this enhancement and will ping you after v3 release if interested to help on it.

PR #7268, opened 2022-09-06
refactor(vite): reuse resolved server entry from context

danielroe:
### ❓ Type of change

### 📚 Description

We are [already resolving the entrypoint for the app](https://github.com/nuxt/framework/blob/de6252099052b5eae6af7cb187e9549dba79a412/packages/vite/src/server.ts#L15-L17), so I think we can remove this extra code. I see that the `.mjs` extension by default in vite node creation was added to fix a bug, so instead I'm moving that logic to the vite server creation so we can share that path (for example) in warming up the app. I'm not sure this makes a difference, but thought it would be better to keep logic central.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/server.ts
+++ b/packages/vite/src/server.ts
@@ -3,7 +3,7 @@ import { resolve } from 'pathe'
 import * as vite from 'vite'
 import vuePlugin from '@vitejs/plugin-vue'
 import viteJsxPlugin from '@vitejs/plugin-vue-jsx'
-import { logger, resolveModule } from '@nuxt/kit'
+import { logger, resolveModule, resolvePath } from '@nuxt/kit'
 import { joinURL, withoutLeadingSlash, withTrailingSlash } from 'ufo'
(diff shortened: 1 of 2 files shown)


packages/vite/src/server.ts
```diff
@@ -14,7 +14,7 @@ import { ssrStylesPlugin } from './plugins/ssr-styles'
 export async function buildServer (ctx: ViteBuildContext) {
   const useAsyncEntry = ctx.nuxt.options.experimental.asyncEntry ||
     (ctx.nuxt.options.vite.devBundler === 'vite-node' && ctx.nuxt.options.dev)
-  ctx.entry = resolve(ctx.nuxt.options.appDir, useAsyncEntry ? 'entry.async' : 'entry')
+  ctx.entry = await resolvePath(resolve(ctx.nuxt.options.appDir, useAsyncEntry ? 'entry.async' : 'entry'))
```
pi0, line 17: Can we use `resolve` (sync)?

pi0 (approved):

PR #7288, opened 2022-09-06
fix(nuxt): don't try to set cookie after redirect

### LINKED ISSUE #5183: [unhandledRejection] Cannot set headers after they are sent to the client
[unhandledRejection] Cannot set headers after they are sent to the client                                                                                                                                                   

  at new NodeError (node:internal/errors:372:5)
  at ServerResponse.setHeader (node:_http_outgoing:576:11)
  at sendError (/H:/nuxt%20projects/trademark/node_modules/h3/dist/index.mjs:216:13)
  at nodeHandler (/H:/nuxt%20projects/trademark/node_modules/h3/dist/index.mjs:375:15)

Due to a problem with dynamic imports, I moved all the resources of the application to the /public folder and now I constantly get this error, what could be the problem?

danielroe:
### 🔗 Linked issue

resolves #5183

### ❓ Type of change

### 📚 Description

We were trying to set headers after redirect when cookies were used. Alternatively we could check writableEnded but I think it's better to have warning surface in this kind of way.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/cookie.ts
+++ b/packages/nuxt/src/app/composables/cookie.ts
@@ -9,9 +9,9 @@ import { useNuxtApp } from '#app'
 
 type _CookieOptions = Omit<CookieSerializeOptions & CookieParseOptions, 'decode' | 'encode'>
 
-export interface CookieOptions<T=any> extends _CookieOptions {
+export interface CookieOptions<T = any> extends _CookieOptions {
   decode?(value: string): T
-  encode?(value: T): string;
+  encode?(value: T): string
   default?: () => T | Ref<T>
 }
 
@@ -38,8 +38,12 @@ export function useCookie <T = string> (name: string, _opts?: CookieOptions<T>):
         writeServerCookie(useRequestEvent(nuxtApp), name, cookie.value, opts)
       }
     }
-    nuxtApp.hooks.hookOnce('app:rendered', writeFinalCookieValue)
-    nuxtApp.hooks.hookOnce('app:redirected', writeFinalCookieValue)
+    const unhook = nuxtApp.hooks.hookOnce('app:rendered', writeFinalCookieValue)
+    nuxtApp.hooks.hookOnce('app:redirected', (...args) => {
+      // don't write cookie subsequently when app:rendered is called
+      unhook()
+      return writeFinalCookieValue(...args)
+    })
   }
 
   return cookie as CookieRef<T>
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -38,8 +38,12 @@ export function useCookie <T = string> (name: string, _opts?: CookieOptions<T>):
         writeServerCookie(useRequestEvent(nuxtApp), name, cookie.value, opts)
       }
     }
-    nuxtApp.hooks.hookOnce('app:rendered', writeFinalCookieValue)
-    nuxtApp.hooks.hookOnce('app:redirected', writeFinalCookieValue)
+    const unhook = nuxtApp.hooks.hookOnce('app:rendered', writeFinalCookieValue)
+    nuxtApp.hooks.hookOnce('app:redirected', (...args) => {
+      // don't write cookie subsequently when app:rendered is called
+      unhook()
+      return writeFinalCookieValue(...args)
```
pi0, line 45: We are not consuming args right?

danielroe: No. Just wrote it this way for forward compatibility.

pi0: But it is confusing. Can we please pass them (by name) when actually needed?

danielroe, line 45:
```suggestion
      return writeFinalCookieValue()
```

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -38,8 +38,12 @@ export function useCookie <T = string> (name: string, _opts?: CookieOptions<T>):
         writeServerCookie(useRequestEvent(nuxtApp), name, cookie.value, opts)
       }
     }
-    nuxtApp.hooks.hookOnce('app:rendered', writeFinalCookieValue)
-    nuxtApp.hooks.hookOnce('app:redirected', writeFinalCookieValue)
+    const unhook = nuxtApp.hooks.hookOnce('app:rendered', writeFinalCookieValue)
+    nuxtApp.hooks.hookOnce('app:redirected', (...args) => {
```
danielroe, line 42:
```suggestion
    nuxtApp.hooks.hookOnce('app:redirected', () => {
```

packages/nuxt/src/app/composables/cookie.ts
danielroe: Oh, agreed.

PR #7290, opened 2022-09-06
fix(nuxt): don't override payload error if it is present

### LINKED ISSUE #6918: Error 500 (from template) rendered instead of 404 (from createErrror)
### Environment

- Operating System: `Linux`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-8gwl3q?file=app.vue

### Describe the bug

If `useAsyncData` returns an error, `throw createError(...)` doesn't work properly. Template errors are being caused and error 500 is rendered.

Actual:
- Error 500 (`Cannot read properties of undefined (reading 'attr')`) is rendered if template uses `state.attr`
- Error 500 (`Cannot read properties of undefined (reading 'something')`) is rendered if template uses `data.something`
- Error 404 is rendered if template doesn't access `state.attr` nor `data.something`

Expected:
- Error 404 should be rendered in all cases

### Additional context

https://github.com/nuxt/framework/issues/4520

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves  #6918

### ❓ Type of change

### 📚 Description

Currently if a renderering error occurs after another error the rendering error will take precedence. (e.g. if an error is thrown in `<script setup>`). This PR simply ensures that the original error is the one we handle.

I've moved the test case to the end of the file as https://github.com/vuejs/core/issues/6110 means that the dev test fixture is broken after a rendering error. We either need to work around it in a separate PR or 🤞 that it is resolved upstream. (But either way, the issue already exists apart from this PR.)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -124,7 +124,8 @@ export default defineRenderHandler(async (event) => {
   const renderer = (process.env.NUXT_NO_SSR || ssrContext.noSSR) ? await getSPARenderer() : await getSSRRenderer()
   const _rendered = await renderer.renderToString(ssrContext).catch((err) => {
     if (!ssrError) {
-      throw err
+      // Use explicitly thrown error in preference to subsequent rendering errors
+      throw ssrContext.payload?.error || err
     }
   })
   await ssrContext.nuxt?.hooks.callHook('app:rendered', { ssrContext })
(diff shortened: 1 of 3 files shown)


pi0 (approved):

danielroe:
The upstream bug has been fixed in Vue 3.2.29, for anyone tracking it. 🥳

confirmed in https://github.com/nuxt/framework/pull/7321.

PR #7313, opened 2022-09-07
feat(vite, webpack): allow client-side sourcemaps in production

### LINKED ISSUE #7294: `nuxt generate` won't output javascript source maps
### Environment

- Operating System: `Linux` (WSL)
- Node Version:     `v16.16.0`
- Nitro Version:    `0.5.1`
- Package Manager:  `npm@8.11.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Generate a starter project. That's it.

### Describe the bug

Running the command `npm run generate` to statically generate the website won't gerenate source maps along the javascript files.

There's an `sourcemap`  property in the nuxt config file, but it default to true, so it should generate the files even if this property isn't specificed. Explicitly setting it to `true` also won't generate them.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #7294

### ❓ Type of change

### 📚 Description

This PR bundles a couple of changes to sourcemap handling:

1. fixes a bug where nitro sourcemaps were not respecting `sourcemap` config (a typo)
2. generates client-side sourcemaps for webpack/vite in production if sourcemap is set to true (or allows either falling back to previous behaviour with new option 'server' or generating _only_ client-side sourcemaps with 'client')
3. defaults to previous behaviour (generates all sourcemaps in dev, and only server-side sourcemaps in production)
4. fixes (unrelated) issue with https://github.com/nuxt/framework/pull/5750 where client-only components were being tree-shaken out of client build as well 🙈 

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/imports/module.ts
+++ b/packages/nuxt/src/imports/module.ts
@@ -91,8 +91,8 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
       })
     } else {
       // Transform to inject imports in production mode
-      addVitePlugin(TransformPlugin.vite({ ctx, options, sourcemap: Object.values(nuxt.options.sourcemap).some(Boolean) }))
-      addWebpackPlugin(TransformPlugin.webpack({ ctx, options, sourcemap: Object.values(nuxt.options.sourcemap).some(Boolean) }))
+      addVitePlugin(TransformPlugin.vite({ ctx, options, sourcemap: nuxt.options.sourcemap.server || nuxt.options.sourcemap.client }))
+      addWebpackPlugin(TransformPlugin.webpack({ ctx, options, sourcemap: nuxt.options.sourcemap.server || nuxt.options.sourcemap.client }))
     }
 
     const regenerateImports = async () => {
--- a/packages/nuxt/src/imports/module.ts
+++ b/packages/nuxt/src/imports/module.ts
@@ -91,8 +91,8 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
       })
     } else {
       // Transform to inject imports in production mode
-      addVitePlugin(TransformPlugin.vite({ ctx, options, sourcemap: nuxt.options.sourcemap }))
-      addWebpackPlugin(TransformPlugin.webpack({ ctx, options, sourcemap: nuxt.options.sourcemap }))
+      addVitePlugin(TransformPlugin.vite({ ctx, options, sourcemap: !!nuxt.options.sourcemap }))
+      addWebpackPlugin(TransformPlugin.webpack({ ctx, options, sourcemap: !!nuxt.options.sourcemap }))
     }
 
     const regenerateImports = async () => {
--- a/packages/nuxt/src/imports/module.ts
+++ b/packages/nuxt/src/imports/module.ts
@@ -91,8 +91,8 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
       })
     } else {
       // Transform to inject imports in production mode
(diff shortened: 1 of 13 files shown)


packages/schema/src/config/build.ts
```diff
@@ -26,9 +26,12 @@ export default defineUntypedSchema({
   /**
    * Whether to generate sourcemaps.
    *
+   * @type {boolean | 'server' | 'client'}
```
pi0, line 29: It doesn't makes sense to me to use single value. We could normalize boolean to `{ client: boolean, server: boolean }`. There are two different source maps.

danielroe: currently:
 - `false`: server: false, client: false
 - `true`: server: true, client: true
 - `client`: server: false, client: true
 - `server`: server: true, client: false

If you feel an object is more semantic then I can refactor.

pi0: Thanks. Yes object would be much more clear also in internal checks.

danielroe: Hm. Normalisation is tricky with types. For example, we can normalise but module may override later and think that 'true' is a valid option. Temporarily will need to check each time we access value, but mental note (as we've discussed before) to refer to the normalized nuxt config type & exclude types that are normalized at build time...

pi0: Agree. We can narrow types for `NuxtOptions` interface from boolean|obj to obj.

packages/vite/src/server.ts
```diff
@@ -81,6 +81,7 @@ export async function buildServer (ctx: ViteBuildContext) {
       ]
     },
     build: {
+      sourcemap: typeof ctx.nuxt.options.sourcemap === 'boolean' ? ctx.nuxt.options.sourcemap : ctx.nuxt.options.sourcemap.server,
```
pi0, line 84: I think we don't need typeof check here anymore

packages/nuxt/src/imports/module.ts
```diff
@@ -91,8 +91,8 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
       })
     } else {
       // Transform to inject imports in production mode
-      addVitePlugin(TransformPlugin.vite({ ctx, options, sourcemap: nuxt.options.sourcemap }))
-      addWebpackPlugin(TransformPlugin.webpack({ ctx, options, sourcemap: nuxt.options.sourcemap }))
+      addVitePlugin(TransformPlugin.vite({ ctx, options, sourcemap: Object.values(nuxt.options.sourcemap).some(Boolean) }))
+      addWebpackPlugin(TransformPlugin.webpack({ ctx, options, sourcemap: Object.values(nuxt.options.sourcemap).some(Boolean) }))
```
pi0, line 95: Why not making normalized type `false | { client, server }`. This way we can can simply use it as a boolean check. (when both disabled, normalize to false)

pi0, line 95: Wouldn't be more readable using `sourcemap.client || sourcemap.server` ?

packages/vite/src/server.ts
danielroe: nice catch - thanks

pi0 (approved):
LGTM. refactor note: i was thinking `false | { client, server }` from resolved options is more convenient but we can do it later if there are more `client || server` conditions needed to enable sourcemap for plugins...

DamianGlowala:
I've tracked the PR which breaks my existing project and this one does it. There's an infinite refresh of the page with 'Reloading server...' message and a console log (`undefined`) in the terminal. Will provide any additional info if needed.

image
image

Issue #7316, opened 2022-09-07
change vite.server.port not work


whichdew:
### Environment

Nuxt project info:                                                                                                  18:48:43

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.17.0`
- Nuxt Version:     `3.0.0-rc.9`
- Nitro Version:    `0.5.1`
- Package Manager:  `npm@8.15.0`
- Builder:          `vite`
- User Config:      `alias`, `ssr`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

1. git clone https://github.com/4tst/nuxt3-starter.git
2. change nuxt.config.ts, set vite.server.port=3001
3. npm run dev

### Describe the bug

always use port 3000 & _nuxt (hmr.path)

pi0:
Hi @whichdew. Yes this is expected. Nuxt lunches it's own server and this type is not picked.

whichdew:
> fix(schema): disallow setting vite server properties #7317

then, what's the differences between nuxt.port and vite.server.port, and how to change nuxt's host&port?

pi0:
Currently, it is possible to use `PORT=3001 nuxt dev` or `nuxt dev --port 3001` to set them.

NBaron:
> Currently, it is possible to use `PORT=3001 nuxt dev` or `nuxt dev --port 3001` to set them.

Thanks for your reply.
**Is there a difference between nuxt's port and vite.server.port?**

If yes, then how to set the later?

PR #7317, opened 2022-09-07
fix(schema): disallow setting vite server properties

manniL:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/7316#issuecomment-1239235095

### ❓ Type of change

### 📚 Description

Disallows setting the port via `vite.server.port` in the `nuxt.config.ts`.
@pi0 Any other options we should disallow? Or `vite.server` in general?

### 📝 Checklist


pi0:
> @pi0 Any other options we should disallow? Or vite.server in general?

I think makes sense to server `vite.server` to never too 👍🏼 (quickly) checking code we are not using it at all.

pi0 (approved):

tobiasdiez:
Setting `server` no `never` now leads to ts errors for hmr configs that are sometimes necessary (e.g. for gitpod):
```ts
vite: {
    server: {
      hmr: ...
    }
}
```
But `hmr` is not exposed by the top-level `server` config. What is the correct config with rc.10?

pi0:
@tobiasdiez Can you please open a PR to limit never to only port/host options? Configuring vite.server.hmr is valid.

manniL:
> @tobiasdiez Can you please open a PR to limit never to only port/host options? Configuring vite.server.hmr is valid.

Will fix this ☺️

PR #7323, opened 2022-09-07
feat(nuxt): filter support for `clearNuxtData`

cawa-93:
### 🔗 Linked issue

Continuation of  https://github.com/nuxt/framework/pull/5227

### ❓ Type of change

### 📚 Description

Uncovered use case: 
Remove cache for several unknown keys. Example: you have app with posts and filter for posts. Each query for posts saved in cache in uniq key, like `posts.list-{param1:1}`, `posts.list-{param2:2}`, `posts.list-{param1:1,param2:3}` (nuxt-trpc works llike that). In this case you may want clear data for all keys `posts.list-*`. 
May be `clearNuxtData` should also accept function to filter keys?
```ts
clearNuxtData(key => key.startsWith('posts.list'))
```
Or provide another method to get all keys in cache and filter it manually:
```ts
clearNuxtData(
  getNuxtDataKeys().filter(key => key.startsWith('posts.list'))
)
```

### 📝 Checklist

### DIFF
--- a/docs/content/3.api/3.utils/clear-nuxt-data.md
+++ b/docs/content/3.api/3.utils/clear-nuxt-data.md
@@ -10,7 +10,7 @@ This method is useful if you want to invalidate the data fetching for another pa
 ## Type
 
 ```ts
-clearNuxtData (keys?: string | string[]): void
+clearNuxtData (keys?: string | string[] | ((key: string) => boolean)): void
 ```
 
 ## Parameters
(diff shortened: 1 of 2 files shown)


Atinux:
I do like the idea of a filter method ☺️

Atinux:
Wondering if this could be applied also to refreshNuxtData(): https://v3.nuxtjs.org/api/utils/refresh-nuxt-data

cawa-93:
I tried, but `refreshNuxtData` doesn't have keys for filtering

https://github.com/cawa-93/framework/blob/ba4828c1dcb4ef7b24e3a3cb0294b754c4f2b811/packages/nuxt/src/app/composables/asyncData.ts#L249-L255

pi0:
LGTM. Supporting for `refreshNuxtData` is also good idea. We can implement similar for fallback when no keys are set.

PR #7344, opened 2022-09-08
fix(nuxt): do not apply import protection to top-level resolution

### LINKED ISSUE #2886: tailwind triggers `nuxt.config` import protection
### Environment

RootDir: /home/supp/code/test 
Nuxt project info:  

- Operating System: `Linux`
- Node Version:     `v17.4.0`
- Nuxt Version:     `3.0.0-27383858.03cc191`
- Bundler:          `Vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-vaufyy?devtoolsheight=33&file=package.json

### Describe the bug

I just create a new clear project with `npx nuxi create` and tried to run `yarn dev`

Errors compilng:
`nuxt3/nuxt cannot be imported directly. Instead, import runtime Nuxt composables from #app or #imports. [importing /home/supp/code/test/node_modules/nuxt3/dist/meta/runtime/plugin from virtual:/home/supp/code/test/.nuxt/plugins/server.mjs]`

` This module cannot be imported in the Vue part of your app. [importing /home/supp/code/test/node_modules/@nuxt/nitro/dist/runtime/app/nitro.client.mjs from virtual:/home/supp/code/test/.nuxt/plugins/server.mjs]`

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #2886

### ❓ Type of change

### 📚 Description

We can avoid triggering import protection from a top-level `this.resolve` call (i.e. within another vite plugin), as there may be valid use cases for this, like tailwind resolving the nuxt.config file.

**Relevant vite section**:

https://github.com/vitejs/vite/blob/757a92f1c7c4fa961ed963edd245df77382dfde6/packages/vite/src/node/server/pluginContainer.ts#L562-L566

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -1,4 +1,4 @@
-import { normalize, resolve } from 'pathe'
+import { join, normalize, resolve } from 'pathe'
 import { createHooks } from 'hookable'
 import type { Nuxt, NuxtOptions, NuxtConfig, ModuleContainer, NuxtHooks } from '@nuxt/schema'
 import { loadNuxtConfig, LoadNuxtOptions, nuxtCtx, installModule, addComponent, addVitePlugin, addWebpackPlugin, tryResolveModule } from '@nuxt/kit'
@@ -62,6 +62,8 @@ async function initNuxt (nuxt: Nuxt) {
   // Add import protection
   const config = {
     rootDir: nuxt.options.rootDir,
+    // Exclude top-level resolutions by plugins
+    exclude: [join(nuxt.options.rootDir, 'index.html')],
     patterns: vueAppPatterns(nuxt)
   }
   addVitePlugin(ImportProtectionPlugin.vite(config))
--- a/packages/nuxt/src/core/plugins/import-protection.ts
+++ b/packages/nuxt/src/core/plugins/import-protection.ts
@@ -32,6 +32,7 @@ export const ImportProtectionPlugin = createUnplugin(function (options: ImportPr
     enforce: 'pre',
     resolveId (id, importer) {
       if (!importer) { return }
+      if (isAbsolute(id)) { id = relative(options.rootDir, id) }
       if (importersToExclude.some(p => typeof p === 'string' ? importer === p : p.test(importer))) { return }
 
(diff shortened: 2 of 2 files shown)


pi0:
Thanks! This defenitly reduces some frustrations but why should a plugin attempt to import from `nuxt.config` or other potentially build-only files that we need to disable protection against? (i think tailwind for runtime config generates a virtual file)

danielroe:
I am not sure exactly how to answer that. Tailwind on its own is triggering this import protection rule. (ie. I'm reproducing it with pure tailwind, not the nuxt integration).

pi0:
I understand issue with our impl and tailwind but the change of this PR is also defusing the import protection for `nuxt.config` at all which was the initial point! e.g: Users now can directly import something from `nuxt.config` in `app.vue`.

pi0:
> note: the import protection rule triggers on resolving an id, not on importing from it.

What about marking triggered resolutions (`resolveId`) and then show error in `transform` step if id actually resolve?

danielroe:
What makes you say they can import in `app.vue`?

pi0:
It won't work because there is another import protection rule for `defineNuxtConfig` from `nuxt`. Directly returning config or import from any top level is broken.

danielroe:
Just did a quick test to confirm (thought I had done it when creating PR but always worth checking!) and indeed import protection works correctly in `app.vue`. The plugin gets called with an importer of `app.vue`. The change in this PR only applies to _top-level_ imports (from `index.html`).

pi0:
nuxt.config:

```ts
export const a = 123

export default {}
```

app.vue:

```vue
<script setup lang="ts">
import { a } from '~/nuxt.config'
console.log(a)
</script>
```

danielroe:
That's not caused by this PR - it's a bug with the import protection rules. Pushing fix.

pi0:
Before your last commit, protection works without change to exclude top level `index.html` and was not working after adding exclude. Now `~/nuxt.config` is working (thanks) but `import { a } from './nuxt.config'` not protected but this one was happening before too.

danielroe:
I think so. 👍 Will have a look.

pi0 (approved):

PR #7361, opened 2022-09-08
feat(nuxi): switch to `unjs/giget` for `nuxi init`

pi0:
### ❓ Type of change

### 📚 Description

Previously we were using degit and then tiged for cloning repository. They were not quite stable and had issues with subdir handling and also because of depending on local `git` command, fail with strange issues. This PR migrates to use [unjs/gitget](https://github.com/unjs/giget) that is designed for best performance and not depending on local git. It also adds `--force`, `--offline`, `--prefer-offline` flags and (experimental) `--shell` that replaces working dir.

Note: `-v|--verbose` previously passing to tiged is deprecated. Using `DEBUG=1` environment variable to enable global verbose logs is possible.

### 📝 Checklist


pi0:
(merging to try and fix early bugs on edge)

pi0:
Getting "global fetch is not available!". Seems an unbuild resolution issue with node-fetch-native... Tracking...

Issue #6592, opened 2022-08-13
Page component is created twice when navigating to it from page which uses different layout


samuelbednarcik:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.15.0`
- Nuxt Version:     `3.0.0-rc.8`
- Package Manager:  `npm@8.14.0`
- Builder:          `vite`
- User Config:      `modules`, `css`, `vite`, `runtimeConfig`, `build`
- Runtime Modules:  `@formkit/nuxt@1.0.0-beta.10`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-nodp3z?file=pages/page.vue

There are two pages and each of them uses a different layout. Just click on link to another page, then click on link to index page. You can see in console that index page component is created and mounted twice.

### Describe the bug

If you are navigating using NuxtLink to another page which uses different layout, that page component is created twice.

### Additional context

It started happening after upgrade from rc6 to rc7. Still happening on rc8.

I also noticed that await useAsyncData on rc6 does not block navigation when navigating to page with another layout. Layout shows without page, then after page data is loaded, page is mounted. This does not happen on rc7 and rc8, navigation is correctly blocked until data is fetched. Maybe it has something to do with this.

Hui-quantblock:
In this case, the <NuxtLoadingIndicator /> does not work properly

deanmikan:
This is a huge issue if you have logic running the created/mounted fields as they get run twice. In our case this is fairly destructive.

We have a default layout that is used across almost all pages. However within a single page we change the layout using:

```javascript
<script>
export default {
    setup() {
        definePageMeta({
            layout: 'alternate-layout'
       });
    }
}
</script>
```

This causes the page to be loaded first with the `default` layout, then again the `alternate-layout` hence running the `created()` and `mounted()` twice. Ouch.

MrWook:
@deanmikan We got around this with using the `NuxtLayout` component on every page and disable the layout in `definePageMeta`.
For Example:
```vue
<template>
  <NuxtLayout name="default">
    <div>
      someCOntent
    </div>
  </NuxtLayout>
</template>
<script setup lang="ts">
definePageMeta({
  layout: false,
})
</script>
```

Its really painfull but at least the pages are working again on our site

deanmikan:
Ouch. Are you running this on `rc-8` @MrWook?

kitae-yun:
!Same thing. I'm so painful

MrWook:
[...]

PR #7400, opened 2022-09-09
fix(nuxt): track suspense status so we can share single state

### LINKED ISSUE #7337: nuxtApp.isHydrating is always false in page setup script when using async page
### Environment

Nuxi 3.0.0-rc.9
RootDir: /home/projects/github-gagmal-zapxnv
Nuxt project info: 

- Operating System: `Linux`
- Nitro Version:    `0.5.1`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-gagmal-zapxnv?file=pages/index.vue

Open `/sync`, `/a`, `/b/b1`, `/c/c1`
Watch the client console to see if the app is hydrated or not.

### Describe the bug

The nuxtApp.isHydrating becomes false before async page component is resolved.
And it also affect all nested page of the async page (no matter it is async or not).

### Additional context

I am writing an app. And I want to get fresh data without the initial request.

To achieve that. I am writing it like

```js
const a = await useFetch('url', () => {
  initialCache: !!process.server || nuxtApp.isHydrating
})
```

so it will only try to actually fetch the data if it isn't initial load.
But it turns out it always say the page is hydrated if I am using async page.
Which is completely unexpected.

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #7337

### ❓ Type of change

### 📚 Description

This PR:
1. resolves a bug where we generated 'too many' Suspense components, meaning that `isHydrating` was `false` in nested child pages on initial load (which has bad consequences for data fetching, as we make decisions based on whether it's the first load or not).
2. works around a vue core issue where _nothing_ under an async component is correctly wrapped into a parent suspense (so this would include all pages with a `<NuxtLayout>`) meaning that `isHydrating` is incorrectly false for these entire trees.

### 📝 Checklist


pi0:
(added pending to land after RC.10)

pi0:
@danielroe Would you please help to resolve conflicts? Will do a quick test afterwards to merge 🚀

danielroe:
Sure 👍 I am also working on resolving https://github.com/nuxt/framework/issues/6592, which may touch some of these files too, but I think this is safe to merge and we can iterate.

mmis1000:
https://github.com/nuxt/framework/pull/7400/files#diff-38b334f71e61b2fb35956472399fb0c87ca4a9d84ef6f99c570c973115f20f0fR49

Isn't change of suspense wrapper one of the reason that page loaded twice?

https://stackblitz.com/edit/github-fktkzl-zxyvww?file=src%2FApp.vue

I don't think vue can keep the page instance if the wrapper changed.

danielroe:
@mmis1000 Yes, it is, and https://github.com/nuxt/framework/issues/6592 exists to track it. I don't believe it's affected by this PR; please let me know if you think otherwise.

mmis1000:
> @mmis1000 Yes, it is, and #6592 exists to track it. I don't believe it's affected by this PR; please let me know if you think otherwise.

I think use of any additional reactive property that does not change at the same time of route in the render function of NuxtPage component is a entirely bad idea. (Just like in this PR, which introduced more reactive state.). Because any structure change of parent of page component WILL RELOAD the whole page.

danielroe:
This PR deliberately doesn't make it reactive. So it should only change the suspense structure if there is _already_ a rerender happening.

mmis1000:
> This PR deliberately doesn't make it reactive. So it should only change the suspense structure if there is _already_ a rerender happening.

https://stackblitz.com/edit/github-gagmal-vjqqo3?file=pages%2Fb%2F[id].vue

That line alone is already enough to cause bug.

1. Navigate to /b/b1
2. reload page to get a SSR's initial page
3. click `b2` to go to /b/b2
4. page is reloaded even it shouldn't (The key is already specified as b via `definePageMeta`)
5. navigate to b1 again
6. it works correctly this time, page is not reloaded

danielroe:
You're right, it does cause an unnecessary rerender. We may pend this PR as I have local work-in-progress that doesn't cause it, and it makes sense to resolve #6592 at the same time.

mmis1000:
https://github.com/mmis1000/nuxt-framework/commit/6bb08590805a16d514dff5f63262af9e1acae796

I think one of potential fix would be using a counter to count how many `<Suspense>` we created, and delay the resolve until every one resolved?

By the way, is the `isNested` check there to workaround some issue? It seems that stop all redirection between sibling of nested routes (`/a/a1 -> /a/a2`) to emit `page:start` and `page:finish` even there is a page load.

PR #7404, opened 2022-09-10
feat(cli)!: use giget 0.1x with template registry for `nuxi init`

pi0:
### 🔗 Linked issue

#7361

### ❓ Type of change

### 📚 Description

This PR upgrades to [unjs/giget](https://github.com/unjs/giget) 0.1.x with custom template registry support. This way we can have beautiful aliases for template names without rebuilding and releasing nuxi CLI but directly from [nuxt/starter](https://github.com/nuxt/starter/tree/templates/templates). A custom environment variable `NUXI_INIT_REGISTRY` can be used to use an alternative registry.

**Breaking change:** Usage of `-t repo/name` is changed to `-t gh:repo/name`.

DX change: Using more minimal output now:

image
 

### 📝 Checklist


pi0:
(merging to try on edge and also unblock https://github.com/nuxt/framework/pull/7383)

danielroe:
@pi0 Could we avoid a breaking change? Often `npx nuxi` does not run the latest version of nuxi owing to npm caching so we will likely need to update documentation to `npx nuxi@latest` if we are providing a template option, which seems a bit verbose...

pi0:
Indeed actually _we could_ also not break the `org/name` usage and allow without `gh:` prefix but i thought it would be too implicit and making it fragile for future development with registry (example: supporting registry templates with slash in their name like. we don't know it is registry id or gh id)

Issue #5411, opened 2022-06-10
pnpm support without `--shamefully-hoist`


pi0:
History: #581

We have progressively improved plug-and-play support over time and almost all underlying packages including nitro support and are tested against pnpm. 

Last step is making nuxt 3 working with pnpm out of the box before migrating modules and templates from first-class yarn to pnpm.

Using pnpm without `--shamefully-hoist` is possible today 🥳 but requires explicitly installing `vue` and `ufo` dependencies and we don't want this since Nuxt was always a zero-config package and we need to add deps of all nuxt modules as well this way.

One obvious solution is that same as Nitro, we internally (also for `unimport` ~> https://github.com/unjs/unimport/issues/69) resolve them to their full path but we also need to support non-resolved implicit imports in templates as modules might have similar implicit dependencies in their scope and they need to be resolved as well. I think this can be done using a custom resolver for vite (and webpack).

jpsc:
It would be nice if fixing this also fixes for `pnpm workspaces` which it doesn't, even with `--shamefully-hoist` .
We have another [issue for workspaces](https://github.com/nuxt/framework/issues/158) that is probably closer to the issue.

asonnleitner:
> It would be nice if fixing this also fixes for `pnpm workspaces` which it doesn't, even with `--shamefully-hoist` .
> We have another [issue for workspaces](https://github.com/nuxt/framework/issues/158) that is probably closer to the issue.

I have a fully functional monorepo using pnpm workspaces, including cached docker build layers and deployment into k8s. Just by using --shamefully-hoist

jpsc:
@asonnleitner I'm assuming that's not anywhere public? I've tried multiple times when on rc.1. I would love to check and compare it.

asonnleitner:
@jpsc yeah, it's private, but if you could make a simple example monorepo, I could take a look at it over the next week and check what's causing the troubles

YannicEl:
@jpsc there are two ways i have found to use pnpm workspaces with nuxt3

1. Have a `.npmrc` file with `shared-workspace-lockfile=false` in your root folder and a `.npmrc` file with `shamefully-hoist=true` in your nuxt subfolder.  I use that approche here: https://github.com/YannicEl/try-finger/tree/dev

2. Have a `.npmrc` file with `shamefully-hoist=true` in your root folder. I use that approche here: https://github.com/YannicEl/hans

amery:
[...]

PR #7407, opened 2022-09-11
feat(nuxi): auto-generate `.npmrc` and setting for pnpm

kecrily:
### 🔗 Linked issue

ref https://github.com/nuxt/framework/issues/5411

### ❓ Type of change

### 📚 Description

Before we support using pnpm without --shamefully-hoist, let's automatically generate `.npmrc` for an out-of-the-box experience. Remove it after support.

### 📝 Checklist


packages/nuxi/src/commands/init.ts
```diff
@@ -25,11 +27,18 @@ export default defineNuxtCommand({
       registry: process.env.NUXI_INIT_REGISTRY || DEFAULT_REGISTRY
     })
 
+    const pkgManager = detectPackageManager()?.name || 'npm'
```
pi0, line 30: I'm not sure this works after init since we don't a lock file generated yet...

pi0:
I really love the idea of this PR! But I think [which-pm-runs](https://github.com/zkochan/packages/tree/master/which-pm-runs) only works when running `nuxi init` with pnpm and in docs we also use `npx` which could be misleading if someone uses `npx` + `pnpm install`.

pi0 (approved):

Issue #7085, opened 2022-08-30
Explicit import from #component of .client components throws an error server side


huang-julien:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.15.0`
- Nuxt Version:     `3.0.0-rc.8`
- Package Manager:  `yarn@3.2.3`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://stackblitz.com/edit/github-ewymne?file=app.vue,components%2FHello.client.vue

### Describe the bug

Hi 👋 I ran into this issue when trying to import a client only component with an explicit import 

````ts
import { Hello } from "#components"
````

However it seems that a .server component fallback is mandatory for Nuxt server-side, I can workaround this by adding a Hello.server.vue but i don't think that this is an expected behavior.

 Calling the component without specifying the import works.

huang-julien:
Not sure if this should be in a separate issue
https://stackblitz.com/edit/github-ewymne-tuweve?file=app.vue,components%2FHello.server.vue
adding a .server fallback and explicitly importing the component resultin a hydration mismatch

Issue #7409, opened 2022-09-11
Lazy .client components are not mounted or cause hydration issues


huang-julien:
### Environment

------------------------------
- Operating System: `Windows_NT`
- Node Version:     `v16.11.1`
- Nuxt Version:     `3.0.0-rc.9`
- Package Manager:  `yarn@3.2.3`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://stackblitz.com/edit/nuxt-starter-fmrgzt?file=app.vue,components%2FWorld.client.vue,components%2FFoo.client.vue,components%2FHello.vue

### Describe the bug

Hi! lazy importing component does not work with .client components. Using direct import from "#components" cause hydration issue and calling the component using auto-import does not mount the component even if `v-if` is set to `true`. 
On the stackblitz, the component  `World`  is being mounted when changing `count` due to  `App` being re-rendered client-side (https://github.com/nuxt/framework/issues/7226).
But after the PR https://github.com/nuxt/framework/pull/7245 that fixed the issue (i'm currently testing this on https://github.com/nuxt/framework/commit/b68bfb0a24c6ae6eaf9b127d56ba4a04672db692 ), the component `World` is not being mounted.

TheDutchCoder:
There's a warning in the docs saying this won't work: https://v3.nuxtjs.org/guide/directory-structure/components#client-components

huang-julien:
yes about the explicit import there's currently a PR, the linked issue is https://github.com/nuxt/framework/issues/7085

PR #7412, opened 2022-09-11
feat(nuxt): wrap `#components` client exports with createClientOnly

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/framework/issues/7085 , related https://github.com/nuxt/framework/issues/7409

### ❓ Type of change

### 📚 Description

hi ! i'm not sure if this should be considered as  bug fix or feature. This PR allow importing `.client` component as client only components when using explicit imports from `#components`. It also adds `ServerPlaceholder` as the `.client` server-side fallback   if there's no `.server` component to avoid hydration issues.

### 📝 Checklist


huang-julien:
convert to draft due to lazy loaded client components not being lazy loaded

huang-julien:
@danielroe
Importing from "#components" automatically break the tests due to `BreakServer` being exported in `components.server.mjs`.
It happens also on the edge channel. 
Do you have any tips or solution to test this ? Maybe renaming `BreakServer` to `BreakServer.client` ? this shouldn't break the test of `/client` since the import is done with it's path.

danielroe:
I'd be happy with renaming `BreaksServer` :+1: Would you confirm that this still preserve the integrity of the test? (That is, that if you set `experimental.treeshakeClientOnly` to false, the test breaks.)

huang-julien:
setting treeshakeClientOnly to false does not break the tests :+1:

packages/nuxt/src/components/scan.ts
```diff
@@ -130,5 +131,17 @@ export async function scanComponents (dirs: ComponentsDir[], srcDir: string): Pr
...
+  components.forEach((component) => {
+    if (component.mode === 'client' && !components.some(c => c.pascalName === component.pascalName && c.mode === 'server')) {
+      components.push({
+        ...component,
+        mode: 'server',
+        filePath: resolve(distDir, 'app/components/server-placeholder'),
+        chunkName: 'components/' + component.kebabName
+      })
+    }
+  })
```
danielroe, line 144: this should probably be located after the `components:extend` hook instead:

https://github.com/nuxt/framework/blob/eab4706614199594006b2132f57e838718ac3e73/packages/nuxt/src/components/module.ts#L146-L150

danielroe (approved):

PR #7435, opened 2022-09-12
fix(nuxt): use more specific `FetchError` for `useFetch` errors

### LINKED ISSUE #7429: `useFetch()` incorrectly typed error
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.20.0`
- Nitro Version:    `0.5.1`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `modules`, `css`, `typescript`, `vite`, `build`
- Runtime Modules:  `@vueuse/nuxt@9.2.0`, `@nuxtjs/supabase@0.1.25`, `@pinia/nuxt@0.4.2`
- Build Modules:    `-`

### Reproduction

Not required

### Describe the bug

The typing of the Error that is returned from the `useFetch` composable is as as follow `Ref<true | Error | null>` however the actually error returned is not the default Error type and is actually of type `FetchError` ([source](https://github.com/unjs/ohmyfetch/blob/a5ebc970414a46f403f552812a46850adcbc71f7/src/error.ts))

So in order to actually get to the extra information of the error casting is required.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves  #7429

### ❓ Type of change

### 📚 Description

In the case of `useFetch`, errors will almost certainly be of the `FetchError` type so we can provide a bit more type hinting to the user to avoid the requirement of casting...

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/fetch.ts
+++ b/packages/nuxt/src/app/composables/fetch.ts
@@ -1,4 +1,4 @@
-import type { FetchOptions } from 'ohmyfetch'
+import type { FetchError, FetchOptions } from 'ohmyfetch'
 import type { TypedInternalResponse, NitroFetchRequest } from 'nitropack'
 import { computed, isRef, Ref } from 'vue'
 import type { AsyncDataOptions, _Transform, KeyOfRes, AsyncData, PickFrom } from './asyncData'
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Looks good to me. However, currently, if interceptors fail for example we don't throw a network error. But is fixable for most cases.

PR #7439, opened 2022-09-12
feat(nuxt): add `workspaceDir` option and add it to `modulesDir`

danielroe:
### 🔗 Linked issue

resolves  #158

### ❓ Type of change

### 📚 Description

This adds `workspaceDir` as an option in nuxt config. By default it uses `pkg-types` to locate this directory.

### 📝 Checklist


packages/schema/src/config/_common.ts
```diff
@@ -51,6 +51,19 @@ export default defineUntypedSchema({
...
+   * Define the workspace directory of your application.
+   *
+   * Often this is used when in a monorepo setup. Nuxt will attempt to detect
+   * your workspace directory automatically, but you can override it here.
+   *
+   * It is normally not needed to configure this option.
+   * @version 3
+   */
+  workspaceDir: {
+    $resolve: (val, get) => val ? resolve(get('rootDir'), val) : ''
```
pi0, line 64: It would be nice if we directly use `fineWorkspaceDir` here for fallback value instead of core

danielroe: Ah, but it's async.

pi0: On it!

pi0: Should be possible now

pi0:
Would be nice if in same PR, we can sync workpaceDir to the nitro config as well (releasing next version shortly)

danielroe:
Great! Will update shortly. I think I already added the option in Nitro config in this PR (though it won't be effective until next Nitro version).

PR #7440, opened 2022-09-12
fix(vite): don't fail builds for virtual modules that don't support inlining

### LINKED ISSUE #7372: virtual css modules not working with inlineSSRStyles (on edge)
Currently if you are using a framework that injects a virtual CSS file it is not working

* https://github.com/windicss/nuxt-windicss/issues/180
* https://github.com/unocss/unocss/issues/1536

A temporary workaround is to exclude these files from processing:

```js
import { defineNuxtConfig } from 'nuxt'

  experimental: {
    inlineSSRStyles: id => !id.includes('windi')
  }
})
```

danielroe:
### 🔗 Linked issue

resolves #7372

### ❓ Type of change

### 📚 Description

Virtual CSS files that want Nuxt inline styles need to respond to the resolved ID with the correct query flags (`?inline&used`). I've updated three key virtual CSS modules (see linked issue). But we shouldn't fail builds if user is using some other virtual CSS module; this PR simply makes sure we skip those virtual modules. Any CSS they inject will not be inlined.)

We could also display a warning message so users at least know this is happening and know where to raise an issue. wdyt @pi0?

### 📝 Checklist

### DIFF
--- a/packages/vite/src/plugins/ssr-styles.ts
+++ b/packages/vite/src/plugins/ssr-styles.ts
@@ -104,6 +104,7 @@ export function ssrStylesPlugin (options: SSRStylePluginOptions): Plugin {
 
         const resolved = await this.resolve(i.specifier, id)
         if (!resolved) { continue }
+        if (!(await this.resolve(resolved.id + '?inline&used'))) { continue }
 
         const ref = this.emitFile({
           type: 'chunk',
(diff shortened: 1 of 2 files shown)


pi0:
A warning when bailing inlining because of unsupported query seems nice idea 👍🏼 This way we can signal plugin authors to add support.

pi0 (approved):

PR #7487, opened 2022-09-13
refactor(nuxt): use relative imports into composables

danielroe:
### ❓ Type of change

### 📚 Description

I came a significant performance hit (in Nuxt Bridge) from importing from `#app` alias with vite, because `#app` re-exports lots of files which themselves import from `#app`.

We can work around that by using relative imports, which I think is also slightly nicer. Even though I haven't measured a slow-down in this repo, it can't hurt and should be a more intelligible approach for vite. I've left a number of imports from `#app` untouched within plugins/components as these are not re-exported from `#app`. (It might also be possible to do the same with `head`, but it's still in a quasi-independent state and this can wait for improvements to its implementation.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -2,7 +2,8 @@ import { defineComponent, h, ref, resolveComponent, PropType, computed, DefineCo
 import { RouteLocationRaw, Router } from 'vue-router'
 import { hasProtocol } from 'ufo'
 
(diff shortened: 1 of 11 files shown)


pi0 (approved):

pi0:
I don't think it should significantly affect performance unless something is wrong with the intermediate resolver cache. Also entrypoint of app is anyway finally imported. But using relative import paths when possible seems more readable and if it is helping performance, why not :)

PR #7527, opened 2022-09-14
feat(nuxt): parse html to treeshake client-only components

### LINKED ISSUE #7517: `<ClientOnly>` with `<template #fallback>` always results in hydration error in RC10
### Environment

- Operating System: `Darwin`
- Nitro Version:    `0.5.3`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-kmye2r?file=components%2FTestThing.vue,components%2FTestThing2.vue,app.vue

### Describe the bug

When using a ClientOnly component with a fallback templates, as described in the docs, you will always get hydration errors.

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #7517

### ❓ Type of change

### 📚 Description

This switches from regexp -> parse5 (fast HTML parsing) to remove client only component nodes. It's more stable and  also allows us to preserve fallback slots, for example.

The blocker for this PR is currently https://github.com/unjs/jiti/issues/84; jiti is having trouble transforming `parse5`, blocking testing.

### 📝 Checklist


pi0:
I will look into jiti issue. Isn't there an easy way to fix the regex issue? Parse5, even if super fast is stil parsing the strings. We might use a hybrid method (like mlly) to improve precision.

danielroe:
Indeed I could fix that particular issue with regexp but I feel parsing the HTML is more reliable. For example, there was always the caveat of nested client-only components.

At some point you reinvent the parser. And I believe parse5 will soon be exposed to us by vite, so we can even benefit from vite caching ast (iirc): https://github.com/vitejs/vite/pull/9678

pi0:
Yes, I'm aware that vite itself is using parse5 and it is a really good decision.

But here we are (re) parsing the same template code into AST in a plugin for a particular optimization that benefits only some components. Full parsing makes the built-time overall slower.

If there are finally more edge cases that we cannot handle with regex only, same as mlly we can use a tho pahse detection with quick regex matcher and precision AST (or Tokenized) verifier when we really are going to make a change to the code.


It would be nice if first could resolve the linked issue via regex and keep working on this PR as a general enhancenment.

pi0:
@danielroe Anyway we can workaround https://github.com/nuxt/framework/issues/7517 regex issue?

danielroe:
~> https://github.com/nuxt/framework/pull/7659

danielroe:
Updating this PR just to keep it in sync. You were right to be careful; `parse5` doesn't handle arbitrary self-closing elements. I suppose we could use an xml parser, but a big appeal of it for me was aligning with vite.

danielroe:
🚀 Switched to **[ultrahtml](https://github.com/natemoo-re/ultrahtml)**, a 1.75kB library with zero dependencies and deliberate support for parsing Vue.

Atinux (approved):

PR #7535, opened 2022-09-15
fix(nuxt): disable payload extraction for spa generated pages

pi0:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/7513#discussioncomment-3649155

### ❓ Type of change

### 📚 Description

SPA pages have no actual payload data (other than initial state). 

This also fixes issue for special spa (`index.html`, `200.html` and `404.html`) routes.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -139,7 +139,7 @@ export default defineRenderHandler(async (event) => {
   }
 
   // Whether we are prerendering route
-  const payloadURL = process.env.prerender ? joinURL(url, '_payload.js') : undefined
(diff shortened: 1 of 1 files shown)


danielroe (commented):
Looks good to me. Unless we can imagine a nitro plugin injecting initial info into a payload?

What about checking whether there is content in the payload?

pi0:
> Looks good to me. Unless we can imagine a nitro plugin injecting initial info into a payload?

That is still considered the initial state of the page. What we **extract** to the `_payload.js` are `data` and `state` keys which are nuxt-app state during SSR. 

If Nitro plugins extend anything in the payload, it will be ignored by SPA renderer (following up with another PR to change it). But this is a rare and new usecase for prerendering spa with payload support.

PR #7541, opened 2022-09-15
fix(schema): update resolver for `cssSourceMap` to new `sourcemap`

danielroe:
### ❓ Type of change

### 📚 Description

This should only affect typings for Bridge, but we hadn't updated the `get('sourcemap')` call.

### 📝 Checklist


pi0 (approved):
LGTM. We need to split v2 schema resolution in next versions and move it to bridge directly...

PR #7553, opened 2022-09-15
fix(nuxt): remove modulepreload for spa fallback routes with `ssr:true`

### LINKED ISSUE #7548: generated 200.html references /200.html/_payload.js
### Environment

- Operating System: `Linux`
- Nitro Version:    `0.5.3`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

- Run `nuxt generate` on nuxt3 project.

### Describe the bug

It's great to have the fallback file `200.html` now! 💖

However, the fallback file `200.html` references a `/200.html/_payload.js` which does not exist.

```html
<!DOCTYPE html>
<html >
<head><link rel="modulepreload" href="/200.html/_payload.js"><link rel="modulepreload" as="script" crossorigin...
```

This causes this error in the console as the `200.html` is served as a fallback 😁
```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/html". Strict MIME type checking is enforced for module scripts per HTML spec.
```

Not sure if this is the reason, the fallback is not working properly, but let's take one step at a time 😉

### Additional context

_No response_

### Logs

_No response_

pi0:
### 🔗 Linked issue

closes #7548

### ❓ Type of change

### 📚 Description

[7548](https://github.com/nuxt/framework/issues/7548#issuecomment-1248003267)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -194,7 +194,7 @@ export default defineRenderHandler(async (event) => {
     htmlAttrs: normalizeChunks([renderedMeta.htmlAttrs]),
     head: normalizeChunks([
       renderedMeta.headTags,
-      (!ssrContext.noSSR && process.env.prerender) ? `<link rel="modulepreload" href="${payloadURL}">` : null,
+      (process.env.prerender && !ssrContext.noSSR) ? `<link rel="modulepreload" href="${payloadURL}">` : null,
       _rendered.renderResourceHints(),
       _rendered.renderStyles(),
       inlinedStyles,
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -194,7 +194,7 @@ export default defineRenderHandler(async (event) => {
     htmlAttrs: normalizeChunks([renderedMeta.htmlAttrs]),
     head: normalizeChunks([
       renderedMeta.headTags,
-      !process.env.NUXT_NO_SCRIPTS && process.env.prerender ? `<link rel="modulepreload" href="${payloadURL}">` : null,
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -194,7 +194,7 @@ export default defineRenderHandler(async (event) => {
     htmlAttrs: normalizeChunks([renderedMeta.htmlAttrs]),
     head: normalizeChunks([
       renderedMeta.headTags,
-      !process.env.NUXT_NO_SCRIPTS && process.env.prerender ? `<link rel="modulepreload" href="${payloadURL}">` : null,
+      (process.env.prerender && !ssrContext.noSSR) ? `<link rel="modulepreload" href="${payloadURL}">` : null,
```
danielroe, line 197:
```suggestion
      (process.env.prerender && !ssrContext.noSSR && !process.env.NUXT_NO_SCRIPTS) ? `<link rel="modulepreload" href="${payloadURL}">` : null,
```

danielroe: But this way we can treeshake it.

danielroe: If we want to have the ability to switch it on and off via ssrContext.noSSR then we should also update in L213-218.

pi0: There is no tree-shaking benefit. `env.prerender` is already causing it.

danielroe: There is a tree shaking benefit if we are not prerendering.

pi0: Have to tried building to confirm?

image

PR #7567, opened 2022-09-15
fix(nuxt): keep `state` in initial state instead of extracting it

### LINKED ISSUE #7562: [rc.10] static payload replaces current state
### Environment

- Operating System: `Linux`
- Nitro Version:    `0.5.3`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

Minimal: https://stackblitz.com/edit/github-vz4xcz-wjzklq?file=app.vue,nuxt.config.ts (updated v3)

Original reproduction by @Atinux with content module: https://github.com/Atinux/nuxt-dd-bug

### Describe the bug

With https://github.com/nuxt/framework/pull/6455, we extract page payloads and assign new state on navigation. This overrides any previous `useState` value with new one. We could instead support a merging strategy or by default avoid **overriding** any initialized state. or introduce an explicit composable for it https://github.com/nuxt/framework/issues/7568 (useData)

### Additional context

_No response_

### Logs

_No response_

pi0:
### 🔗 Linked issue

Resolves #7562

### ❓ Type of change

### 📚 Description

#6455 adds `payload.state` (set by `useState`) to the extracted payload. This will cause hydration and unwanted behavior on navigation by orriding already initialized state. 

This PR, makes behavior closer to Nuxt 2 by keeping state to the initial state inlined in the pages. Only payload data is now extracted.

Alternatives: We could instead keep extracting state and only try to set keys that are not assigned yet but it seems wrong performance wise if state is _initial state_ it should be set once and if is same across pages, loading it adds to network overhead. We can revise it later with some better approach to only record state change diffs and include that in `_payload.js` somehow.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/plugins/payload.client.ts
+++ b/packages/nuxt/src/app/plugins/payload.client.ts
@@ -10,7 +10,6 @@ export default defineNuxtPlugin((nuxtApp) => {
     const payload = await loadPayload(url)
     if (!payload) { return }
     Object.assign(nuxtApp.payload.data, payload.data)
-    Object.assign(nuxtApp.payload.state, payload.state)
   }
   nuxtApp.hooks.hook('link:prefetch', async (to) => {
     await prefetchPayload(to)
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -296,9 +296,9 @@ function renderPayloadResponse (ssrContext: NuxtSSRContext) {
 }
 
 function splitPayload (ssrContext: NuxtSSRContext) {
-  const { data, state, prerenderedAt, ...initial } = ssrContext.payload
+  const { data, prerenderedAt, ...initial } = ssrContext.payload
   return {
     initial: { ...initial, prerenderedAt },
-    payload: { data, state, prerenderedAt }
+    payload: { data, prerenderedAt }
   }
 }
--- a/test/basic.test.ts
+++ b/test/basic.test.ts
@@ -590,7 +590,7 @@ describe.skipIf(process.env.NUXT_TEST_DEV || isWindows)('payload rendering', () 
   it('renders a payload', async () => {
     const payload = await $fetch('/random/a/_payload.js', { responseType: 'text' })
     expect(payload).toMatch(
-      /export default \{data:\{\$frand_a:\[[^\]]*\]\},state:\{"\$srandom:rand_a":\d*,"\$srandom:default":\d*\},prerenderedAt:\d*\}/
+      /export default \{data:\{\$frand_a:\[[^\]]*\]\},prerenderedAt:\d*\}/
     )
   })
 


danielroe:
I would suggest not treating data differently from state. `useState` is pretty much identical to `useAsyncData` as far as data initialisation. Unlike Nuxt 2, where the store was a global object that was initialised no matter what page you started on, in Nuxt 3, state may only be initialised on a subset of pages.

I agree we shouldn't override keys that are already defined.

danielroe:
I would suggest not treating data differently from state. `useState` is pretty much identical to `useAsyncData` as far as data initialisation. Unlike Nuxt 2, where the store was a global object that was initialised no matter what page you started on, in Nuxt 3, state may only be initialised on a subset of pages.

It's also equally likely that users who have deliberately duplicated an asyncData key, such as (for example) an auto key defined in `[slug].vue`, will have issues with the current behaviour.

I agree we shouldn't override keys that are already defined.

pi0:
`useState` has two use cases. Main is to hydrate (sync) state from server to the client and one new use case of content module for custom data fetching. I agree second is valid but combining it with useState, means we should sacrifice one usage performance to each other (loading full state to the payload is duplicate). 

We might instead introduce new `useData` (https://github.com/nuxt/framework/pull/6455) that is purposed for this. It can reuse same `data` namespace of `useAsyncData`.

danielroe:
I've raised a separate PR partially dealing with the issue of conflicting state/data keys: https://github.com/nuxt/framework/pull/7574.

If we really need to exclude data from payload, then what about, instead, only excluding data that has been set outside of a vue component instance? E.g. in a plugin. We could define on `ssrContext` which keys are 'initial' and which represent the payload?

pi0:
I thought about making differentiated behavior too but honestly, it makes composable behavior even more unpredictable if it behaves different when used in plugin or component contexts.

Actually, I made `useState` as a global state replacement initially. It was (and is its main purpose) `useAsyncData` for data fetching per page.

New `useData` can be exactly purposed for what we want.

danielroe:
I would really appreciate it if you would wait an hour or two for me to write down my comments on this and https://github.com/nuxt/framework/pull/7567.

danielroe:
**tldr;** I think it's safe to disable the inclusion of `useState` data in the payload because we only just released full static mode - no objections to merging this PR as-is. (I think we may need to reconsider this approach in future, but that would be in the nature of an enhancement rather than a breaking change...)

---

I think there are several use-cases for it at the moment.

1. It is a replacement for any kind of ref meant to be SSR-safe. This is typically write-once, and it may be updated in each successive page - in other words, it might be unique to each page. This would make sense to extract into a payload.
2. It is a replacement for global state (e.g. nav menu display or not). This should _not_ be extracted (more specifically, it shouldn't be _updated_ on navigation).
3. It is a replacement for an initial state. E.g. a list of navigation items or settings from a CMS. This only needs to be loaded once and shouldn't be extracted into a payload owing to the increased network payload.

With this change, `useState` becomes usable for #2 & #3. And I think that `useData`/`usePageData` would be a sufficient replacement for #1, with my suggested change. (This would be somewhat breaking as users would not have a way of extracting data into a payload which is in the global scope, e.g. shared across pages. But I think that is a necessary evil to prevent the Suspense issue I referred to - which would have been a problem anyway, independent of payload extraction.)

pi0:
Thanks for the nice summary and input @danielroe ❤️  Indeed this revert a short-lived behavior change we introduced in RC.10 for payload extraction making `useState` primarily usable for global/initial state such as navigation. And that we might include **state changes** as a future and non-breaking change enhancement to the payload extraction in the future.

danielroe (approved):

Issue #7534, opened 2022-09-15
[vite-node] [ERR_MODULE_NOT_FOUND]


aslero:
### Environment

[vite-node] [ERR_MODULE_NOT_FOUND]

### Reproduction

[vite-node] [ERR_MODULE_NOT_FOUND]

### Describe the bug

[vite-node] [ERR_MODULE_NOT_FOUND]

marcelxpfeifer:
Are you using pnpm?

danielroe:
Would you provide a reproduction? 🙏

aslero:
> Are you using pnpm? If so you probably just need to run _pnpm install --shamefully-hoist_

no, use npm

hbilal9:
same error is coming with me:

i Vite client warmed up in 2963ms
√ Nitro built in 1292 ms
[Vue warn]: inject() can only be used inside setup() or functional components.
[Vue warn]: Server rendering context not provided. Make sure to only call useSSRContext() conditionally in the server build.
[Vue warn]: Unhandled error during execution of setup function
  at <NuxtRoot>
[nuxt] [request error] [unhandled] [500] Cannot read properties of undefined (reading 'modules')
  at _sfc_main.setup (H:/Workspace/Web Workspace/Productions/axlic-nuxt/node_modules/nuxt/dist/app/components/nuxt-root.vue:67:16)
  at callWithErrorHandling (H:\Workspace\Web Workspace\Productions\axlic-nuxt\node_modules\@vue\runtime-core\dist\runtime-core.cjs.js:157:22)
  at setupStatefulComponent (H:\Workspace\Web Workspace\Productions\axlic-nuxt\node_modules\@vue\runtime-core\dist\runtime-core.cjs.js:7109:29)
  at setupComponent (H:\Workspace\Web Workspace\Productions\axlic-nuxt\node_modules\@vue\runtime-core\dist\runtime-core.cjs.js:7064:11)
  at renderComponentVNode (H:\Workspace\Web Workspace\Productions\axlic-nuxt\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:172:17)
  at renderToString (H:\Workspace\Web Workspace\Productions\axlic-nuxt\node_modules\@vue\server-renderer\dist\server-renderer.cjs.js:450:26)
  at Object.renderToString$1 [as renderToString] (/H:/Workspace/Web%20Workspace/Productions/axlic-nuxt/.nuxt/dev/index.mjs:472:24)
[...]

PR #7575, opened 2022-09-15
fix(vite): provide fallback reason

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/7534

### ❓ Type of change

### 📚 Description

This adds more info in the event of minimal error data. In the linked issue, error data is:

```js
const errorData = {
  code: "ERR_MODULE_NOT_FOUND",
}
```

That's enough to thwart our formatter.

CleanShot 2022-09-15 at 23 09 56@2x

This PR adds stack + reason fallbacks, transforming it to:

CleanShot 2022-09-15 at 23 09 19@2x

Admittedly probably the stack isn't helpful in this case, but at least the error message tells us the module that couldn't be fetched. Alternatively, maybe we could directly throw the error if we only have a `code` in errorData?

### 📝 Checklist

### DIFF
--- a/packages/vite/src/runtime/vite-node.mjs
+++ b/packages/vite/src/runtime/vite-node.mjs
@@ -40,6 +40,8 @@ function createRunner () {
         if (!errorData) {
           throw err
         }
+        errorData.stack = errorData.stack || err.stack
+        errorData.reason = errorData.reason || err.message
         let _err
         try {
           const { message, stack } = formatViteError(errorData)
--- a/packages/vite/src/runtime/vite-node.mjs
+++ b/packages/vite/src/runtime/vite-node.mjs
@@ -40,7 +40,6 @@ function createRunner () {
         if (!errorData) {
           throw err
         }
-        errorData.stack = errorData.stack || err.stack
         errorData.reason = errorData.reason || err.message
         let _err
         try {


packages/vite/src/runtime/vite-node.mjs
```diff
@@ -40,6 +40,8 @@ function createRunner () {
         if (!errorData) {
           throw err
         }
+        errorData.stack = errorData.stack || err.stack
```
pi0, line 43: This stack is really unrelated (other than URL that can hint about source). I think we shall use normalization directly from vite-node server.

danielroe, line 43:
```suggestion
```

pi0:
I've made #7589 for a better normalization approach. I think we can change this PR to make fatal errors (unhandled while normalizing) to include module id in the message. But `reason` is internal preserving it will be confusing.

danielroe:
Sure 👍 Also happy to close this and we can do it all in your linked PR.

pi0:
I've double-checked now. Unless `formatViteError` fails (which is solved with other PR), we show 500 error (network error actually) with path to module id. That should only happen in rare cases of internal error.

PR #7577, opened 2022-09-15
fix(nuxi): print resolved public directory after generate

### LINKED ISSUE #7511: Output directory message mismatch confi
### Environment

Nuxt 3.0.0.rc9

### Reproduction

Example: https://stackblitz.com/edit/github-eeaz87?file=nuxt.config.ts
(ctrl+c and run `npm install && npm run generate`)

Setup in `nuxt.config.ts`:
```js
  target: 'static',
  ssr: true,
  nitro: {
    output: {
      publicDir: 'SHOW_THIS',
    },
  },
  dir: {
    static: 'public',
    public: 'SHOW_THIS',
  },
});
```
See after running command `npm install && npm run generate`:
https://github.com/nuxt/framework/blob/d567a91d7d63c9928d3eb0f2da8ccf76b336a971/packages/nuxi/src/commands/generate.ts#L14

### Describe the bug

Message have hard coded value. Changing setting need change message

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #7511

### ❓ Type of change

### 📚 Description

Using new useNitro utility (https://github.com/nuxt/framework/pull/7557) we can access the resolved publicDir for nitro and print the relative path to it, rather than hard coding `.output/public`.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/build.ts
+++ b/packages/nuxi/src/commands/build.ts
@@ -42,8 +42,9 @@ export default defineNuxtCommand({
     await buildNuxt(nuxt)
 
     if (args.prerender) {
+      // TODO: revisit later if/when nuxt build --prerender will output hybrid
       const dir = nitro?.options.output.publicDir
-      const publicDir = dir ? relative(rootDir, dir) : '.output/public'
+      const publicDir = dir ? relative(process.cwd(), dir) : '.output/public'
       consola.success(`You can now deploy \`${publicDir}\` to any static hosting!`)
     }
   }
--- a/packages/nuxi/src/commands/build.ts
+++ b/packages/nuxi/src/commands/build.ts
@@ -1,4 +1,4 @@
-import { resolve } from 'pathe'
+import { relative, resolve } from 'pathe'
 import consola from 'consola'
 import { writeTypes } from '../utils/prepare'
 import { loadKit } from '../utils/kit'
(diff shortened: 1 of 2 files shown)


packages/nuxi/src/commands/build.ts
```diff
@@ -38,5 +40,11 @@ export default defineNuxtCommand({
     })
 
     await buildNuxt(nuxt)
+
+    if (args.prerender) {
+      const dir = nitro?.options.output.publicDir
+      const publicDir = dir ? relative(rootDir, dir) : '.output/public'
```
pi0, line 46: We print CLI logs relative to cwd() so that it is clickable from IDE (at least on best hope).

packages/nuxi/src/commands/build.ts
```diff
@@ -38,5 +40,11 @@ export default defineNuxtCommand({
...
+      consola.success(`You can now deploy \`${publicDir}\` to any static hosting!`)
```
pi0, line 47: Note: This is accurate currently because `--prerender` disables server build but it might be inaccurate in the future if `nuxt build --prerender` outputs hybrid. Would you please add a comment to not forget about this? (or simply i would suggest to keep it in `generate.ts` we can refactor later for different messages

danielroe: Unfortunately we have to call it here as in generate (after build) `useNitro()` returns null.

pi0: I see! Okay then let's just keep it in mind for future.

pi0 (approved):

PR #7593, opened 2022-09-16
fix(nuxi): backward compatibility for kit <= RC.10 without `useNitro`

danielroe:
### ❓ Type of change

### 📚 Description

If used with an older version of kit without the feature it will error out on:

```
 ERROR  useNitro is not a function
```

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/build.ts
+++ b/packages/nuxi/src/commands/build.ts
@@ -28,6 +28,7 @@ export default defineNuxtCommand({
       }
     })
 
+    // Use ? for backward compatibility for Nuxt <= RC.10
     const nitro = useNitro?.()
 
     await clearDir(nuxt.options.buildDir)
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/commands/build.ts
```diff
@@ -28,7 +28,7 @@ export default defineNuxtCommand({
       }
     })
 
-    const nitro = useNitro()
+    const nitro = useNitro?.()
```
pi0, line 31: Refactor: Do you mind to move this line to L45?

danielroe: It doesn't work there. The Nuxt instance is unset.

pi0: I see. This is a strange issue `build()` function is directly closing instance. will fix in subsequent PR.

PR #7606, opened 2022-09-16
feat(nuxi): move `module-builder` to unified `nuxi` cli

cpreston321:
### ❓ Type of change

- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description
Move core functionality from [nuxt-module-builder](https://github.com/nuxt/module-builder) -> `nuxi`

**CLI Options**

Command: `npx nuxi build-module [--stub] [--outDir] [rootDir]`

- [Added]: new command `build-module`
   - [rootDir]: The rootDir
   - [--stub]: stub directory for module development.
   - [--outDir] (optional): change the outDir name  default `dist` 

###  Why ? 

Why you ask? The goal was to make nuxi cli feel like it has all the tool sets needed for nuxt 💚 . I do feel like it would be nice to see even the `nuxt-module-builder` inside nuxi so it's easier to maintain code. 

Maybe there are few improvements to make but I can leave this as a draft PR to see what you say.

Thanks!

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/commands/build-module.ts
+++ b/packages/nuxi/src/commands/build-module.ts
@@ -1,0 +1,82 @@
+import { promises as fsp } from 'node:fs'
+import { pathToFileURL } from 'node:url'
+import { resolve } from 'pathe'
+import consola from 'consola'
+
+import { writeModuleTypes, writeModuleCJSStub, loadUnbuild } from '../utils/build-module'
+import { defineNuxtCommand } from './index'
+
+import type { NuxtModule } from '@nuxt/schema'
+
+export default defineNuxtCommand({
+  meta: {
+    name: 'build-module',
+    usage: 'npx nuxi build-module [--stub] [--outDir] [rootDir]',
+    description: 'Build a nuxt module for development & production'
+  },
+  async invoke (args) {
+    const rootDir = resolve(args._[0] || '.')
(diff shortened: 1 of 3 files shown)


pi0:
Hi @cpreston321 and thanks for the PR. I do share the same idea that `nuxi` CLI should magically support building modules.

Since we have external dependencies (unbuild), it would add to boilerplate of modules and we have to make sure modules use a supported version of this dependency. We might also add more dependencies over time to the module-builder stack that makes it beyond unbuild. This is the primary reason I decided to keep it separated.

We do however plan to support `nuxi-` subcommand. (https://github.com/nuxt/nuxt.js/issues/13634 and nuxt/module-builder#32). This way when module builder dependency is added, `nuxi build-module` will work out of the box. 

Something we might do before introducing subcommand support, is to have a stub `nuxi build-module` command in nuxi that executes `nuxt-build-module` internally and if not found, makes an error to install the dependency of `@nuxt/module-builder`. Would you like to help on this? 🙏🏼 

(I'm closing this PR because moving module-builder to the core is not planned at this point because of above reasons)

cpreston321:
@pi0 I would love to help! I appreciate the in detail response. I can work the stub in the mean-time. 

Thanks 💚

PR #7607, opened 2022-09-16
fix(vite): include importer in `vite-node` error stack

pi0:
### 🔗 Linked issue

Ref #7534

### ❓ Type of change

### 📚 Description

This PR tries to include (last) importer in stack traces of vite-node formatted errors.

image

### 📝 Checklist

### DIFF
--- a/packages/vite/src/runtime/vite-node.mjs
+++ b/packages/vite/src/runtime/vite-node.mjs
@@ -29,12 +29,15 @@ export default async (ssrContext) => {
 }
 
 function createRunner () {
+  const _importers = new Map()
   return new ViteNodeRunner({
     root: viteNodeOptions.root, // Equals to Nuxt `srcDir`
     base: viteNodeOptions.base,
+    resolveId (id, importer) { _importers.set(id, importer) },
     async fetchModule (id) {
-      // TODO: fix in vite-node
-      id = id.replace(/\/\//g, '/')
+      const importer = _importers.get(id)
+      _importers.delete(id)
+      id = id.replace(/\/\//g, '/') // TODO: fix in vite-node
       return await viteNodeFetch('/module/' + encodeURI(id)).catch((err) => {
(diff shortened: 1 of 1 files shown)


packages/vite/src/runtime/vite-node.mjs
```diff
@@ -29,20 +29,23 @@ export default async (ssrContext) => {
 }
 
 function createRunner () {
+  const _importers = new Map()
   return new ViteNodeRunner({
     root: viteNodeOptions.root, // Equals to Nuxt `srcDir`
     base: viteNodeOptions.base,
+    resolveId (id, importer) { _importers.set(id, importer) },
     async fetchModule (id) {
```
pi0, line 37: @antfu is it possible for vite-node to directly give importer to the `fetchModule`? It could be useful to format errors better.

antfu: The problem is that a module could be imported by multiple importers and `vite-node` caches them to reuse the existing one. That means if we could have the importer it could only be the first one hit the module and can be changed in different requests.

pi0: I see. Indeed makes sense. But how do you think about this solution for Nuxt to hint about error? I suppose _first_ importer triggers the error this way.

danielroe: Maybe we could update the logging message to be something like `(first imported by x)` to keep it accurate?

pi0: having imported by is also accurate (it _could be_, _also_ imported by y but as result of being imported by x is now being failed). Similar to other stack traces if a util is used by two functions, first function calling it shows in stack trace.

antfu: I will see what I can do on the vite-node side

PR #6153, opened 2022-07-26
fix(vite): use `externality` to handle vite-node externals

antfu:
### 🔗 Linked issue

#5018

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This reuses the externalization options from `dev-bundler` to make the behaviour more consistent.

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #7612, opened 2022-09-16
fix(vite): respect `ctx.nuxt.options.modulesDir` for resolving externals with `vite-node`

### LINKED ISSUE #7609: [RC 9] [RC 10] `Cannot read properties of undefined (reading 'modules')` when `vue` is installed in parent dir
### Related issues

- #7532 
- #7220
- #7573
- sidestream-tech/sidebase#16

### Reproduction

Minimal reproduction: https://stackblitz.com/edit/github-2yn1ze?file=app%2Fpackage.json,app%2Fnuxt.config.ts

### Describe the bug

How it happens: If vue dependency is installed outside of project directory in an adjutant directory (usually home directory with `node_modules`), `vite-node` that is enabled by default for RC.9+ makes issues. (issue is with nuxt custom external options)

## Workaround 1 

Check you home directory. If there is any `node_modules` dir, remove it or rename it. It is probably created wrongly in the past.

## Workaround 2

It might be also because of a monorepo setup that has `../node_modules/vue` dependency. In this case, switch to the legacy vite bundler:

```ts
  vite: {
    devBundler: 'legacy',
  },
});
```

### Logs

_No response_

pi0:
### 🔗 Linked issue

resolves #7609

### ❓ Type of change

### 📚 Description

We are overriding `shouldExternalize` (PR: #6153) of vite-node, and use mlly resolver to directly resolve paths. Issue with #7609 happens because directly setting URL to the directory of `rootDir`, causes resolving from parent `node_modules` and finally different resolved `vue` version. (Correct way is setting `rootDir + '/_node_modules'` or `rootDir + /index.mjs`)

This PR resolves issue by respecting `modulesDir` for resolving `vue` and other dependencies. In the future we might come of with better implementation to actually reuse same vite resolution. I didn't use `return id` on external branch (while it works!) until getting confirmation from @antfu. Vite-node's default external handler however does this. [src](https://github.com/vitest-dev/vitest/blob/95da4d66933db46da368d5d66f422afd26fba4c1/packages/vite-node/src/externalize.ts#L85)

Also worth to mention that `isExternal('vue')` resolves to `null` as it is unable to determine external status.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/vite-node.ts
+++ b/packages/vite/src/vite-node.ts
@@ -102,7 +102,7 @@ function createViteNodeMiddleware (ctx: ViteBuildContext, invalidates: Set<strin
     node.shouldExternalize = async (id: string) => {
       const result = await isExternal(id)
       if (result?.external) {
-        return resolveModule(result.id, { url: ctx.nuxt.options.rootDir })
+        return resolveModule(result.id, { url: ctx.nuxt.options.modulesDir })
       }
       return false
     }


packages/vite/src/vite-node.ts
```diff
@@ -102,7 +102,7 @@ function createViteNodeMiddleware (ctx: ViteBuildContext, invalidates: Set<strin
     node.shouldExternalize = async (id: string) => {
       const result = await isExternal(id)
       if (result?.external) {
-        return resolveModule(result.id, { url: ctx.nuxt.options.rootDir })
+        return resolveModule(result.id, { url: ctx.nuxt.options.modulesDir })
```
pi0, line 105: @antfu Shall we do this to be same as [vite-node](https://github.com/vitest-dev/vitest/blob/95da4d66933db46da368d5d66f422afd26fba4c1/packages/vite-node/src/externalize.ts#L85)?

```suggestion
  return id
```

pi0:
Merging as hotfix.

packages/vite/src/vite-node.ts
antfu: The id passed here should already be resolved by Vite. So yes I think we could directly do this.

pi0: Thanks for checking. FYI, it is not always resolved. You can reproduce it only with an external repository by putting a debugger. `vue` for instance is directly passed to `shouldExternalize`.

PR #7622, opened 2022-09-17
fix(vite, webpack): avoid generating keys where a key is already provided

### LINKED ISSUE #7578: [nitro] [dev] [unhandledRejection] TypeError: Cannot create property 'server' on string '$KxdiLPLTek'
### Environment

- Node Version:     `v14.19.0`
- Nuxt Version:     `3.0.0-rc.10-27709424.bb8942a`
- Nitro Version:    `0.5.2-27709289.11cc5cd`
- Package Manager:  `yarn@1.22.10`
- Builder:          `webpack`
- User Config:      `builder`, `modules`, `proxy`, `hooks`, `postcss`, `build`, `webpack`, `css`
- Runtime Modules:  `nuxt-proxy@0.2.4`
- Build Modules:    `-`

### Reproduction

repository: 
https://gitlab.com/zengshine/nuxt-web-app.git

yarn dev and open the test page: 
http://localhost.qizhidao.com:3000/useAsyncTest

### Describe the bug

[nitro] [dev] [unhandledRejection] TypeError: Cannot create property 'server' on string '$KxdiLPLTek'

image

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #7578

### ❓ Type of change

### 📚 Description

With webpack, we were processing code _twice_ and therefore injecting several auto-generated keys, one after the other. This PR prevents double-insertion of keys, and also prevents insertion when there is an obvious string or template literal key provided already.

It also excludes nuxt sourcefiles (e.g. `#app`) from having keys inserted into them.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/plugins/composable-keys.ts
+++ b/packages/vite/src/plugins/composable-keys.ts
@@ -24,7 +24,7 @@ export const composableKeysPlugin = createUnplugin((options: ComposableKeysOptio
     enforce: 'post',
     transformInclude (id) {
       const { pathname, search } = parseURL(decodeURIComponent(pathToFileURL(id).href))
-      return pathname.match(/node_modules\/nuxt3?\//) && pathname.match(/\.(m?[jt]sx?|vue)/) && parseQuery(search).type !== 'style'
+      return !pathname.match(/node_modules\/nuxt3?\//) && pathname.match(/\.(m?[jt]sx?|vue)/) && parseQuery(search).type !== 'style'
     },
     transform (code, id) {
       if (!KEYED_FUNCTIONS_RE.test(code)) { return }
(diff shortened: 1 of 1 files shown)


pi0:
I love this as workaround but also it can be even trickier in the future when we support more arguments (like 3rd options to useState) and if first arg is not a string but variable/reference. Thinking if we could change auto key injection method to be safe out of the box.

danielroe:
You're right that we would need to update if we add more args, but the main thing it is protecting against is duplicate string keys at the end of the argument chain, which is what causes the issue.

Detecting explicit keys at the beginning is only an added benefit - not a bug fix.

pi0 (approved):

Issue #3252, opened 2022-02-16
Vitest was initialized with native Node instead of Vite Node


danielroe:
The async import doesn't seem to respect the environment of the test runner.

**Reproduction**: https://github.com/danielroe/test-utils-test

cc: @antfu

schickling:
Are you sure this is a problem with Nuxt? I was running into the same problem when running Vitest on pre-compiled `.js` test files with the following config:

```ts
import { defineConfig } from 'vitest/config'

export default defineConfig({
  test: {
    include: ['dist/__test__/**/mapping.spec.js'],
    // include: ['src/__test__/**/mapping.spec.ts'],
    exclude: [],
    watchIgnore: [],
  },
})
```

Note: Running the tests on the `src/**/*.spec.ts` files works as expected.

antfu:
Vitest v0.5.1 should fix it.

schickling:
Still doesn't seem to work with `0.5.1`. Here's a repro: https://github.com/schickling-test/2022-feb-vitest-effect/tree/repro/dist-bug

Please run:

- `yarn`
- `yarn dev:ts` (compiles to `./dist`)
- `yarn test`

You should see: 

<img width="627" alt="image" src="https://user-images.githubusercontent.com/1567498/155089489-3501641a-5899-4a15-a006-1965e122cbb9.png">

danielroe:
I can confirm that my issue is no longer reproducible. @schickling, your issue might be a different one that simply has the same symptom - possibly worth raising at the vitest repo?

jorge-obsidian:
> I can confirm that my issue is no longer reproducible. @schickling, your issue might be a different one that simply has the same symptom - possibly worth raising at the vitest repo?

I've just cloned that repo and i'm currently getting the `Error: Vitest was initialized with native Node instead of Vite Node` when launching test.

CodyBontecou:
I'm also running into this. Here's a reproduction [repo](https://github.com/CodyBontecou/nuxt3-vitest-native-node-error).

ytr0903:
In my case, creating `vitest.config.ts` with `test.deps.inline` resolves this error.
```ts:vitest.config.ts
import { defineConfig } from "vitest/config";
export default defineConfig({
  test: {
    environment: "jsdom",
    deps: {
      inline: [/@nuxt\/test-utils-edge/],
    },
  },
})
```

Ref: https://github.com/vitest-dev/vitest/issues/843

asmerkin:
[...]

PR #7712, opened 2022-09-21
fix(test-utils): support vitest v0.20.x

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/3252, resolves https://github.com/nuxt/framework/issues/7709

### ❓ Type of change

### 📚 Description

Vitest v0.20 added an additional parameter, which we weren't respecting with test-utils. This PR also adds nuxt/test-utils to `deps.inline` which is required to avoid `Vitest was initialized with native Node instead of Vite Node` error.

We can drop the workaround if we want to have a hard dependency on vitest 0.20+

### 📝 Checklist

### DIFF
--- a/packages/test-utils/src/run.ts
+++ b/packages/test-utils/src/run.ts
@@ -21,14 +21,20 @@ export async function runTests (opts: RunTestOptions) {
     process.env.NUXT_TEST_DEV = 'true'
   }
 
+  // TODO: add `as typeof import('vitest/dist/node')` and remove workaround
+  // when we upgrade vitest: see https://github.com/nuxt/framework/issues/6297
+
   // @ts-ignore missing types
   const { startVitest } = await import('vitest/dist/node.mjs')
-  const succeeded = await startVitest(
+  const args: any[] = [
     [] /* argv */,
     // Vitest options
     {
       root: opts.rootDir,
-      run: !opts.watch
+      run: !opts.watch,
+      deps: {
+        inline: [/@nuxt\/test-utils/]
+      }
     },
     // Vite options
     {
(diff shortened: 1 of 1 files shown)


packages/test-utils/src/run.ts
```diff
@@ -21,22 +21,30 @@ export async function runTests (opts: RunTestOptions) {
...
-  const succeeded = await startVitest(
+  const args: any[] = [
     [] /* argv */,
     // Vitest options
     {
       root: opts.rootDir,
-      run: !opts.watch
+      run: !opts.watch,
+      deps: {
+        inline: [/@nuxt\/test-utils/]
```
pi0, line 36: Is it for stub mode with .ts sources?

danielroe: It just doesn't work without it, printing the following:

```
ERROR  Error: Vitest was initialized with native Node instead of Vite Node.

One of the following is possible:
- "vitest" is imported outside of your tests (in that case, use "vitest/node" or import.meta.vitest)
- "vitest" is imported inside "globalSetup" (use "setupFiles", because "globalSetup" runs in a different context)
- Your dependency inside "node_modules" imports "vitest" directly (in that case, inline that dependency, using "deps.inline" config)
- Otherwise, it might be a Vitest bug. Please report it to https://github.com/vitest-dev/vitest/issues


 ❯ setupVitest node_modules/.pnpm/@nuxt+test-utils-edge@3.0.0-rc.12-27728205.6309d22/node_modules/@nuxt/test-utils-edge/dist/index.mjs:212:18
 ❯ Module.setup node_modules/.pnpm/@nuxt+test-utils-edge@3.0.0-rc.12-27728205.6309d22/node_modules/@nuxt/test-utils-edge/dist/index.mjs:273:3
 ❯ async /private/tmp/sourcemap/test/index.test.ts:6:3
```

I believe it's because of the dynamic import of vitest within `setup()`.

pi0: Okay lets investigate later for root cause.

pi0 (approved):

PR #7717, opened 2022-09-21
docs: recommend to not overwrite some keys in tsconfig

tobiasdiez:
### ❓ Type of change

### 📚 Description

I hit a strange error while upgrading from rc 9 to 11 that `import_meta.hot.accept` was not defined. After a bit of research, I think, I hit https://github.com/evanw/esbuild/issues/2294. At least after removing `"target": "ES2018",` from the tsconfig file everything is working smoothly now. So I added a small note to the docs that these keys shouldn't be overwritten.

### 📝 Checklist


danielroe (approved):

docs/content/2.guide/2.directory-structure/3.tsconfig.md
```diff
@@ -14,4 +14,4 @@ Nuxt [automatically generates](/guide/concepts/typescript) a `.nuxt/tsconfig.jso
 }
 ```
 
-As you need to, you can customize the contents of this file. However, note that if you need to customize your `paths`, this will override the auto-generated path aliases. Instead, we recommend that you add any path aliases you need to the `alias` property within your `nuxt.config`, where they will get picked up and added to the autogenerated `tsconfig`.
+As you need to, you can customize the contents of this file. However, it is recommended that you don't overwrite `target`, `module` and `moduleResolution`. Moreover, note that if you need to customize your `paths`, this will override the auto-generated path aliases. Instead, we recommend that you add any path aliases you need to the `alias` property within your `nuxt.config`, where they will get picked up and added to the autogenerated `tsconfig`.
```
DamianGlowala, line 17:
```suggestion
As you need to, you can customize the contents of this file. However, it is recommended that you don't overwrite `target`, `module` and `moduleResolution`. Moreover, note that if you need to customize your `paths`, this will override the auto-generated path aliases. Instead, we recommend that you add any path aliases you need to the `alias` property within your `nuxt.config`, where they will get picked up and added to the auto-generated `tsconfig`.
```

DamianGlowala (changes requested):

pi0 (approved):

pi0:
Thanks!

PR #7734, opened 2022-09-21
docs(navigate-to): use await instead of return

### LINKED ISSUE #7732: navigateTo examples throw errors
### Environment

- Operating System: `Darwin`
- Nitro Version:    `0.5.4`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

N/A

### Describe the bug

In the docs (https://v3.nuxtjs.org/api/utils/navigate-to/#navigating-within-a-vue-component) when using `navigateTo` in a component, it is said to use `return` in front of it, but this will throw a compiler error:

```
 ERROR  [@vue/compiler-sfc] 'return' outside of function. (2:0)                                 13:19:21

/Users/reinierkaper/projects/remx/apps/nuxt-app/pages/vision.vue
1  |  <script setup lang="ts">
   |                           ^
2  |  return navigateTo('/', { redirectCode: 301 })
   |  ^
3  |  </script>
```

Should it be awaited instead?

If so, I can PR a docs change if you like.

### Additional context

_No response_

### Logs

_No response_

TheDutchCoder:
### 🔗 Linked issue

resolves #7732 

### ❓ Type of change

### 📚 Description

1. Changes the `navigateTo` example to use `await` instead of return, as `return` will throw a compilation error.
2. Formats the block of code to be consistent with other blocks.

### 📝 Checklist

### DIFF
--- a/docs/content/3.api/3.utils/navigate-to.md
+++ b/docs/content/3.api/3.utils/navigate-to.md
@@ -70,20 +70,20 @@ An object accepting the following properties:
 
 ```vue
 <script setup>
-  // passing 'to' as a string
-  await navigateTo('/search')
+// passing 'to' as a string
+await navigateTo('/search')
 
-  // ... or as a route object
-  await navigateTo({ path: '/search' })
+// ... or as a route object
+await navigateTo({ path: '/search' })
 
-  // ... or as a route object with query parameters
-  await navigateTo({
-    path: '/search',
-    query: {
-      page: 1,
-      sort: 'asc'
-    }
-  })
+// ... or as a route object with query parameters
+await navigateTo({
+  path: '/search',
+  query: {
+    page: 1,
+    sort: 'asc'
+  }
+})
 </script>
 ```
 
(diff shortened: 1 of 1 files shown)


docs/content/3.api/3.utils/navigate-to.md
```diff
@@ -70,20 +70,20 @@ An object accepting the following properties:
...
+  await navigateTo({ path: '/search' })
+
+  // ... or as a route object with query parameters
+  await navigateTo({
+    path: '/search',
+    query: {
+      page: 1,
+      sort: 'asc'
+    }
+  })
```
DamianGlowala, line 86: Good catch! In terms of top-level indentation within the `<script setup>` block, there is currently a mix of both in the code snippets across the documentation, but I'd be in favor of no indentation. An example to back this up:

https://vuejs.org/api/sfc-script-setup.html#basic-syntax

```suggestion
// passing 'to' as a string
await navigateTo('/search')

// ... or as a route object
await navigateTo({ path: '/search' })

// ... or as a route object with query parameters
await navigateTo({
  path: '/search',
  query: {
    page: 1,
    sort: 'asc'
  }
})
```

DamianGlowala (changes requested):

docs/content/3.api/3.utils/navigate-to.md
TheDutchCoder: I personally also use unindented code, just wasn't sure which one was preferred in general and this page had both.
Made it all unindented, thanks!

danielroe (approved):
Thank you! ❤️

Issue #7755, opened 2022-09-22
client only components don't respect v-show


warflash:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-rc.11`
- Nitro Version:    `0.5.4`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://stackblitz.com/edit/github-zpzmng-lfbt8x?file=app.vue

### Describe the bug

Marking a component as client only using `xx.client.vue` stops v-show from working and instead displays the component all the time. v-if works fine.

PR #7774, opened 2022-09-22
fix(nuxt): remove Fragment from `createClientOnly`

huang-julien:
### 🔗 Linked issue

linked to https://github.com/nuxt/framework/pull/6165#issuecomment-1199255073
resolve https://github.com/nuxt/framework/issues/7755

### ❓ Type of change

### 📚 Description
Hi 👋 👋  This PR remove the Fragment workaround which prevent using v-show on .client components.

There is additionnal changes in the setup part where we return the render function. 

Previously in the linked PR, there's was a `Oldchildren is null` issue if we returned directly the setup result as a render function. 
Using the `h()` function to return the VNode was working only for some components (based on @danielroe 's  https://github.com/nuxt/framework/pull/6165#issuecomment-1197878279).

However the component fails to render or update its render silently if it only has a root tag with a string and a variable such as :

`````vue
<template>
  <div>
      I'm a string children {{ count }}. I will be mounted but will still render a simple div
  </div>
</template>
``````

The workaround was to use a Fragment but it leads to https://github.com/nuxt/framework/issues/7755 issue .

This is fixed by verifying the result of `setupState`.children which can be `string|array|null|object` and rendering the component using `createElementBlock` when res.children is null or a string
### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/client-only.mjs
+++ b/packages/nuxt/src/app/components/client-only.mjs
@@ -1,4 +1,4 @@
-import { ref, onMounted, defineComponent, createElementBlock, h, Fragment } from 'vue'
+import { ref, onMounted, defineComponent, createElementBlock, h } from 'vue'
 
 export default defineComponent({
   name: 'ClientOnly',
@@ -31,7 +31,7 @@ export function createClientOnly (component) {
     // override the component render (non script setup component)
     clone.render = (ctx, ...args) => {
       return ctx.mounted$
-        ? h(Fragment, ctx.$attrs ?? ctx._.attrs, component.render(ctx, ...args))
+        ? h(component.render(ctx, ...args))
         : h('div', ctx.$attrs ?? ctx._.attrs)
     }
   } else if (clone.template) {
@@ -52,8 +52,7 @@ export function createClientOnly (component) {
           ? { ...setupState, mounted$ }
           : (...args) => {
               return mounted$.value
-                // use Fragment to avoid oldChildren is null issue
-                ? h(Fragment, ctx.attrs, setupState(...args))
+                ? h(setupState(...args))
                 : h('div', ctx.attrs)
             }
       })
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by danielroe — COMMENTED** [review]
Thank you! ❤️ 

This seems like a good fix to me. Is there a way we can add a test for the different kinds of client-only components? Maybe a page in the fixture with a variety of types and we can use `createPage` to assert that all content is correctly displayed?

**Comment by huang-julien** [comment]
I think this should now cover most of the issue we can run into.
- [x] Multi root node component (script + script setup) -> Fragment https://github.com/nuxt/framework/issues/7755
- [x] Single Root with string children component (script + script setup) -> https://github.com/nuxt/framework/pull/6165#issuecomment-1199255073
- [x] component with slot  https://github.com/nuxt/framework/issues/5867
- [x] apply directive on a component (script + script setup) https://github.com/nuxt/framework/issues/7755
- [x] component with an array of children  (script + script setup)
- [x] component mounted but not rendered in HTML  (script + script setup) https://github.com/nuxt/framework/pull/6165#issuecomment-1199255073

File: test/basic.test.ts [hunk 1]
```diff
@@ -144,13 +144,68 @@ describe('pages', () => {
   })
 
   it('/client-only-components', async () => {
+    // ensure components are not rendered server-side
     const html = await $fetch('/client-only-components')
     expect(html).toContain('<div class="client-only-script" foo="bar">')
     expect(html).toContain('<div class="client-only-script-setup" foo="hello">')
     expect(html).toContain('<div>Fallback</div>')
     expect(html).not.toContain('Should not be server rendered')
```
danielroe, line 152:
```suggestion
    const html = await $fetch('/client-only-components')
    // ensure fallbacks with classes and arbitrary attributes are rendered
    expect(html).toContain('<div class="client-only-script" foo="bar">')
    expect(html).toContain('<div class="client-only-script-setup" foo="hello">')
    expect(html).toContain('<div>Fallback</div>')
    // ensure components are not rendered server-side
    expect(html).not.toContain('Should not be server rendered')
```

danielroe (approved):
This looks great, and I'm really happy with the test suite - thank you 😊

antfu (approved):

huang-julien:
i forgot to set .client on some test components

huang-julien:
updated https://github.com/nuxt/framework/pull/7774#issuecomment-1256336768
fixed missing .client in components name

danielroe (approved):

PR #7805, opened 2022-09-24
fix(nuxt): only fetch payloads for prerendered routes

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt.js/issues/14507

resolves https://github.com/nuxt/nuxt.js/issues/15024

### ❓ Type of change

### 📚 Description

This PR adds support for an app manifest (at `/manifest.json`) which contains, at the moment, simply a list of routes for which there are corresponding payloads. With the implementation of [route rules](https://github.com/nuxt/framework/discussions/560) this manifest can be extended/modified. In addition, it can be extended with a build hash to allow version change detection.

With this PR:
 * we **will not fetch** payloads for pages that were not prerendered (they would give a 404)
 * will **will fetch** all payloads if there is an SSR/hybrid app
 * we can fetch payloads in development mode if `experimental.payloadExtraction` is set by the user to `true` (there are more edge cases, such as CORS, with client-side fetching, so I felt it was better to set default to false in dev mode so user handles them properly, in case they are not prerendering all their pages).
 * support for _hybrid_ payload extraction (e.g. _some but not all_ routes have SSR payloads) is deferred to implementation of [route rules](https://github.com/nuxt/framework/discussions/560)

(**Note**, @pi0, I'm also happy to go ahead and implement route rules, but thought we could adopt this approach for now and modify it at that point.)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/payload.ts
+++ b/packages/nuxt/src/app/composables/payload.ts
@@ -9,6 +9,8 @@ interface LoadPayloadOptions {
 
 export function loadPayload (url: string, opts: LoadPayloadOptions = {}) {
   if (process.server) { return null }
+  if (!_hasPayload(url)) { return null }
+
   const payloadURL = _getPayloadURL(url, opts)
   const nuxtApp = useNuxtApp()
   const cache = nuxtApp._payloadCache = nuxtApp._payloadCache || {}
@@ -26,6 +28,8 @@ export function loadPayload (url: string, opts: LoadPayloadOptions = {}) {
 }
 
 export function preloadPayload (url: string, opts: LoadPayloadOptions = {}) {
+  if (!_hasPayload(url)) { return }
+
   const payloadURL = _getPayloadURL(url, opts)
   useHead({
     link: [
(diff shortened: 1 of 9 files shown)


danielroe:
@pi0 How can I help on this? What's the reason this is marked as draft?

ice-open-source:
🙏

danielroe:
@pi0 Would you let me know why you have marked this as a draft/pending?

danielroe:
On board with `nuxt build` 👍 

So, just to understand the situation with :**full static mode** - you're saying that from your point of view, 404s are an inevitable evil, with two possible mitigations: We can (1) prevent fetching them if there is a `prerender` route rule, and (2) prevent retrying a payload url that has already 404ed.

ConnorCairns:
Are there any updates on if/when this, or a solution to solve this problem will be completed?

Currently having this issue in production where payloads for non prerendered routes are being requested resulting in a 404 which is not ideal.

danielroe:
~> https://github.com/nuxt/nuxt/pull/18419

PR #7807, opened 2022-09-24
feat(nuxt): Feature/add `initialCache` option in `runtimeConfig.app`

TheDutchCoder:
Note: mainly looking for feedback on:
 1. Is this a welcome change?
 2. Did I do it right?
 3. Did I update the docs appropriately?
 4. Do I need to add more docs (probably)?

I don't want to waste your time, so feel free to close this if it's an unwanted addition!

### 🔗 Linked issue

Currently not an open issue, but a started idea/discussion. I can make a related issue, but first wanted to run this by you guys.

### ❓ Type of change

### 📚 Description

Currently, all data is fetched with `{ initialCache: true }` as the default setting. This setting can't be changed on a configuration level, which means people have to set `{ initialCache: false }` on every single query if they don't want caching by default, which is cumbersome.

This setting will allow people to specify a default caching strategy on a config level:
```ts
  runtimeConfig: {
    app: {
      initialCache: false
    }
  }
})
```

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/asyncData.ts
+++ b/packages/nuxt/src/app/composables/asyncData.ts
@@ -91,7 +91,6 @@ export function useAsyncData<
   options.default = options.default ?? getDefault
 
   const config = useRuntimeConfig()
-  console.log(config)
 
   // TODO: remove support for `defer` in Nuxt 3 RC
   if ((options as any).defer) {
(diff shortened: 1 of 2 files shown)


TheDutchCoder:
This obviously doesn't work yet, I'm probably approaching the config part wrong.
Will try to make it work so the tests pass, and add a new test case if I can!

TheDutchCoder:
Ok, it works properly now, but I'm list with the tests. Not really sure how they work under the hood.

pi0:
Thanks for PR. Planning to drop `initialCache` and disable it by default (#8885).

BTW supporting global options that change behavior might be tricky. If a Nuxt module or library depends on default behavior, it can be broken easily.

Issue #7806, opened 2022-09-24
Base URL doesn't get applied to /_payload.js when using Nuxi Generate


Redemption198:
### Environment

- Operating System: `Linux`
- Node Version:     `v16.15.0`
- Nuxt Version:     `3.0.0-rc.12-27730995.2894a75`
- Nitro Version:    `0.5.4-27726613.640c2b7`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `modules`, `components`, `runtimeConfig`, `experimental`, `app`, `pwa`
- Runtime Modules:  `@nuxtjs/tailwindcss@5.3.3`, `@vueuse/nuxt@9.2.0`, `@kevinmarrec/nuxt-pwa@0.4.2`
- Build Modules:    `-`

### Reproduction

1. Set `app.baseURL` to a custom base url in `nuxt.config.ts`
2. Run `npx nuxi generate`
3. Inspect `.output/public/index.html` and search for `/_payload.js` two instances should appear with no base url applied:
    1. `<link rel="modulepreload" href="/_payload.js">`
    2. `...import p from "/_payload.js";...`

### Describe the bug

Base url doesn't get applied when using `nuxi generate`. With `experimental.payloadExtraction` set to `false` it works fine.

PR #7809, opened 2022-09-24
fix(nuxt): respect `baseURL` when rendering payload path

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/7806

### ❓ Type of change

### 📚 Description

This respects the baseURL when rendering payload URL in HTML.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -143,7 +143,7 @@ export default defineRenderHandler(async (event) => {
 
   // Whether we are prerendering route
   const _PAYLOAD_EXTRACTION = process.env.prerender && process.env.NUXT_PAYLOAD_EXTRACTION && !ssrContext.noSSR
-  const payloadURL = _PAYLOAD_EXTRACTION ? joinURL(url, '_payload.js') : undefined
+  const payloadURL = _PAYLOAD_EXTRACTION ? joinURL(useRuntimeConfig().app.baseURL, url, '_payload.js') : undefined
   if (process.env.prerender) {
     ssrContext.payload.prerenderedAt = Date.now()
   }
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -177,7 +177,7 @@ export default defineRenderHandler(async (event) => {
 
   if (_PAYLOAD_EXTRACTION) {
     // Hint nitro to prerender payload for this route
-    appendHeader(event, 'x-nitro-prerender', payloadURL!)
+    appendHeader(event, 'x-nitro-prerender', joinURL(url, '_payload.js'))
```
pi0, line 180: I'm wondering if we should fix (or already supported by nitro) to support both with/without base UR prerendering.

pi0 (approved):

packages/nuxt/src/core/runtime/nitro/renderer.ts
woldtwerk: [  app/composables/payload.ts](https://github.com/nuxt/framework/blob/92ff066ab26c72e3488b16b8da391178647ada9d/packages/nuxt/src/app/composables/payload.ts#L40)

I think we need to inject the baseurl here aswell, don't we?

```ts
const parsed = parseURL(joinURL(useRuntimeConfig().app.baseURL, url));
```

```diff
diff --git a/dist/app/composables/payload.mjs b/dist/app/composables/payload.mjs
index 8f277d5d82c0d452b4e87bd91bd2c5029e02006c..88103dcad954855180737c1b064d4047deb5bbf1 100644
--- a/dist/app/composables/payload.mjs
+++ b/dist/app/composables/payload.mjs
@@ -1,6 +1,7 @@
 import { parseURL, joinURL } from "ufo";
 import { useNuxtApp } from "../nuxt.mjs";
 import { useHead } from "../index.mjs";
+import { useRuntimeConfig } from "#internal/nitro";
 export function loadPayload(url, opts = {}) {
   if (process.server) {
     return null;
@@ -29,7 +30,7 @@ export function preloadPayload(url, opts = {}) {
   });
 }
 function _getPayloadURL(url, opts = {}) {
-  const parsed = parseURL(url);
+  const parsed = parseURL(joinURL(useRuntimeConfig().app.baseURL, url));
   if (parsed.search) {
     throw new Error("Payload URL cannot contain search params: " + url);
   }
```

danielroe: Thanks for spotting it!

~> https://github.com/nuxt/framework/pull/7984

PR #7831, opened 2022-09-26
fix(nuxt): don't prerender index.html with a server

### LINKED ISSUE #7826: Public runtimeConfig not working on index page
### Environment

- Operating System: `Darwin`
- Nitro Version:    `0.5.4`
- Package Manager:  `npm@8.11.0`
- Builder:          `vite`
- User Config:      `runtimeConfig`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/JorgeVazquezS/nuxt-env-test

### Describe the bug

I detected this issue when upgrading from RC8 a simple app i'm working with. Bug is present from RC9 until RC11, i've tried them all.

The problem is present when `ssr` is set to false on config. `runtimeConfig` public  vars are not present on index page unless it was defined on the environment on build time. Every other page with acces to runtimeConfig works as expected.

_In RC8 this was working fine._ 

I think this is a very common use case since you can define some public vars on local dev environment, build the app on a different environment (like a pipeline) where the vars are not present, and finally run on a different environment like a container where you can set and change this env variables.

**_Steps to reproduce the issue with RC9:_**
- npm install
- npm run build
- Now we set an env variable: `export NUXT_PUBLIC_TEST_VAR="this is a test var"`
- Run `node .output/server/index.mjs`
- In the browser we access to `http://localhost:3000` Var is not shown. ❌
- In the browser we access to `http://localhost:3000/test` Var is present as expected. ✅

### Additional context

_No response_

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #7826

### ❓ Type of change

### 📚 Description

We should not be prerendering `index.html` if there is a dynamic server as runtime config (or nitro plugins) may change the response.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nitro.ts
+++ b/packages/nuxt/src/core/nitro.ts
@@ -57,7 +57,8 @@ export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
       crawlLinks: nuxt.options._generate ? nuxt.options.generate.crawler : false,
       routes: ([] as string[])
         .concat(nuxt.options._generate ? ['/', '/200.html', ...nuxt.options.generate.routes] : [])
-        .concat(nuxt.options.ssr === false ? ['/index.html', '/200.html', '/404.html'] : [])
+        .concat(nuxt.options.ssr === false ? ['/200.html', '/404.html'] : [])
+        .concat(nuxt.options.ssr === false && nuxt.options._generate ? ['/index.html'] : [])
     },
     sourceMap: nuxt.options.sourcemap.server,
     externals: {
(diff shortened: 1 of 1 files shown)


LastxTemplar:
I would just like to say that this is not a minor bug, it's an app breaking bug. We have the `ssr` flag set to false but it still generates the index page and our runtime env values are not available in there. This breaks the app especially if you have middleware using a service like auth0. Can this please be merge as soon as possible? or at least provide a workaround to forcefully not generate the index page.
Thank you!

manniL:
@LastxTemplar see https://github.com/nuxt/framework/issues/7826#issuecomment-1259126029 for a workaround.

LastxTemplar:
> @LastxTemplar see https://github.com/nuxt/framework/issues/7826#issuecomment-1259126029 for a workaround.

Thank you!

danielroe:
I would also add that when I mark something `p3` vs `p4` it doesn't mean it will get less attention. For me, it's:

```
(num people affected) x (severity of the issue) - (possibility of workaround)
```

Atinux (approved):

Atinux (approved):

PR #7870, opened 2022-09-27
feat(pages): add `validate` hook for `definePageMeta`

### LINKED ISSUE #5172: `validate` function in `definePageMeta`
In Nuxt v2: https://nuxtjs.org/docs/components-glossary/validate/

Using middleware could indeed achieve the same, but it would be a bit verbose to create and might lose the context since now they are in different files in different directories. I think it would be nice to support the `validate` function in `definePageMeta`. For example:

```vue
<script setup>
definePageMeta({
  validate({ params }) {
    return ['foo', 'bar'].includes(params.type)
  }
})
</script>
```

danielroe:
### 🔗 Linked issue

resolves #5172

### ❓ Type of change

### 📚 Description

This adds `validate()` function in `definePageMeta` for easily specifying what valid route options are for a given page. It supports returning an object error format that will then determine the error response on SSR.

### 📝 Checklist


docs/content/migration/7.component-options.md
```diff
@@ -117,7 +117,7 @@ See [layout migration](/migration/pages-and-layouts).
 
 ## `validate`
 
-There is no longer a validate hook in Nuxt 3. Instead, you can create a custom middleware function, or directly throw an error in the setup function of the page.
+The validate hook in Nuxt 3 only accepts a single argument, the `route`.
```
Atinux, line 120: Could be nice to also explain what values we can return in the validate hook.

docs/content/migration/7.component-options.md
```diff
@@ -128,14 +130,10 @@ There is no longer a validate hook in Nuxt 3. Instead, you can create a custom m
...
 + definePageMeta({
-+   middleware: [
-+     async function (to, from) {
-+       const nuxtApp = useNuxtApp()
-+       if (!valid) {
-+         return abortNavigation('Page not found')
-+       }
-+     }
-+   ]
++   validate: async (route) => {
```
Atinux, line 133: What about checking if the route.params.id is a number?

Atinux:
Shall we add also a section after https://v3.nuxtjs.org/getting-started/routing#route-middleware called "Route Validation" and taking the similar content as https://deploy-preview-7870--nuxt3-docs.netlify.app/migration/component-options#validate ?

danielroe:
Great idea!

docs/content/1.getting-started/5.routing.md
```diff
@@ -113,3 +113,25 @@ definePageMeta({
...
+
+If you have a more complex use case, then you can use anonymous route middleware instead.
+
+:StabilityEdge
+
+```vue [pages/post/[id].vue]
+<script setup>
+definePageMeta({
+  validate: async (route) => {
+    const nuxtApp = useNuxtApp()
```
Atinux, line 131: Is it necessary to have this line since it will provide a warning with TS since the variable is not used?

docs/content/migration/7.component-options.md
```diff
@@ -117,25 +117,23 @@ See [layout migration](/migration/pages-and-layouts).
...
++     const nuxtApp = useNuxtApp()
```
Atinux, line 134: Same here, can be removed

danielroe: It's here to show how to access the rest of the Nuxt context (which previously was directly provided to the validate hook).

antfu:
That's awesome, thank you!

Atinux (approved):

PR #7895, opened 2022-09-28
chore: use pnpm for framework monorepo

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/939

### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

We've been discussing migrating to pnpm in our [move towards full pnpm support](https://github.com/nuxt/framework/discussions/939). This PR implements it using `--shamefully-hoist ` and `--strict-peer-dependencies=false`; we can remove these flags when possible.

~~We also have the opportunity to lose a lot of weight in the monorepo if we remove `lerna` - not sure how possible that is @pi0 but worth mentioning; it's now only used for the actual release.~~ (lerna replaced with local scripts for bumping rc)

**Contributors note:** 

If you have already cleaned the repository locally, run this to migrate:

```sh
npm run cleanup && corepack enable && pnpm install && pnpm build:stub
```

### 📝 Checklist


Atinux:
This is really convincing as well as pnpm speed 🚀 

CleanShot 2022-10-03 at 15 39 42@2x

netlify.toml
```diff
@@ -0,0 +1,14 @@
+# Global settings applied to the whole site.
+#
+# “base” is the directory to change to before starting build. If you set base:
+#        that is where we will look for package.json/.nvmrc/etc, not repo root!
+# “command” is your build command.
+# “publish” is the directory to publish (relative to the root of your repo).
+
+[build]
+base = "."
+command = "npx pnpm@7.12.0 install && npx pnpm@7.12.0 stub && npx pnpm@7.12.0 --filter docs build"
```
pi0, line 10: Can't we use corepack?

pi0:
Netlify build seems going wrong (not related to pnpm but since deps are now shared it is happening)

> 3:12:12 PM: [error] [nuxt] [request error] [unhandled] [500] app.ssrContext.nuxt.vueApp.config.globalProperties.$pinceauSsr.get is not a function

pinceauSsr usage wrong?

> 3:12:12 PM: [error] [nuxt] [request error] [unhandled] [500] Cannot read properties of undefined (reading 'url')

Hint: `useQuery` is called with `req` instead of `event` somewhere (github module and another place)

netlify.toml
danielroe: Happy to try. I've found it very tricky getting netlify to work with `pnpm`.

pi0: I see. Since seems there are other issues, shall we keep using yarn for docs and fix it later? Rest of mono repo seems in good state with pnpm to merge this PR sooner

danielroe: Sure!

pi0 (approved):

PR #7906, opened 2022-09-29
feat(nuxt): add useDecodeURI method to decode url

yufanzheng0723:
### ❓ Type of change

### 📚 Description

Hi, sorry for my bad english, used translation
When I use nuxt3 to send a request like /api/search/[content], if I change the value of content through `<input type="text" v-model="text">`, the requested url will be encoded . I want to add a method to decode the url
### 📝 Checklist


danielroe:
Many thanks indeed!

We might need to fix a bug in h3 or Nuxt if you can explain more about the problem. Would you open an issue with a reproduction? 🙏

If it's not a bug to be fixed, then I still think we probably shouldn't add this utility to the Nuxt composables. (It's possible a utility like this might be useful in [**h3**](https://github.com/unjs/h3) - but do explain more about the use case first in an issue.)

I'm going to close this PR, but it's no reflection on your work, which is much appreciated.

PR #7929, opened 2022-09-30
Improve loading sequence of hydration `<script />` 🚀

sanjaiyan-dev:
### Improve loading sequence of hydration `<script />` 🚀

>This will help us to improve-: 
>- Improvements in FCP, LCP metric, faster time to visual-completeness.
>- Reduced developer cost for curating the loading sequence for FCP, LCP.
>- Additional optimization and predictability given the guaranteed order of execution of defer scripts.

References -:

1.  Next js current implementation -: https://github.com/vercel/next.js/discussions/24938
2. About script loading -: https://gist.github.com/jakub-g/385ee6b41085303a53ad92c7c8afd7a6

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -196,7 +196,7 @@ export default defineRenderHandler(async (event) => {
     head: normalizeChunks([
       renderedMeta.headTags,
       _PAYLOAD_EXTRACTION ? `<link rel="modulepreload" href="${payloadURL}">` : null,
-      _rendered.renderResourceHints(),
+      _rendered.renderScripts(),
       _rendered.renderStyles(),
       inlinedStyles,
       ssrContext.styles
@@ -217,7 +217,6 @@ export default defineRenderHandler(async (event) => {
             ? `<script type="module">import p from "${payloadURL}";window.__NUXT__={...p,...(${devalue(splitPayload(ssrContext).initial)})}</script>`
             : `<script>window.__NUXT__=${devalue(ssrContext.payload)}</script>`
           ),
-      _rendered.renderScripts(),
       // Note: bodyScripts may contain tags other than <script>
       renderedMeta.bodyScripts
     ])


danielroe:
At the moment we are preloading the script in the head, and then adding it at the end of the document. That is important for hydration, as we can't hydrate the DOM before the DOM is present.

Perhaps you could open an issue with some issues to address, and we could discuss it there?

sanjaiyan-dev:
> At the moment we are preloading the script in the head, and then adding it at the end of the document. That is important for hydration, as we can't hydrate the DOM before the DOM is present.
> 
> Perhaps you could open an issue with some issues to address, and we could discuss it there?

Sorry, actually here script is fetched parallel with html parsing (same as `<link rel="preload" />`) but for browser which doesn't support `<link rel="preload" />`.

Defer attribute and `type="module"` behave same as adding script to bottom but parallel to html parsing :)

sanjaiyan-dev:
> > At the moment we are preloading the script in the head, and then adding it at the end of the document. That is important for hydration, as we can't hydrate the DOM before the DOM is present.
> > Perhaps you could open an issue with some issues to address, and we could discuss it there?
> 
> Sorry, actually here script is fetched parallel with html parsing (same as `<link rel="preload" />`) but for browsers which doesn't support `<link rel="preload" />`.
> 
> Defer attribute and `type="module"` behave same as adding script to bottom but fetches scripts parallel to html parsing :)

For ref-: 

https://web.dev/efficiently-load-third-party-javascript/#defer

https://v8.dev/_img/modules/async-defer.svg

sanjaiyan-dev:
> > At the moment we are preloading the script in the head, and then adding it at the end of the document. That is important for hydration, as we can't hydrate the DOM before the DOM is present.
> > Perhaps you could open an issue with some issues to address, and we could discuss it there?
> 
> Sorry, actually here script is fetched parallel with html parsing (same as `<link rel="preload" />`) but for browsers which doesn't support `<link rel="preload" />`.
> 
> Defer attribute and `type="module"` behave same as adding script to bottom but fetches scripts parallel to html parsing :)

@danielroe For ref-: 

https://web.dev/efficiently-load-third-party-javascript/#defer

https://v8.dev/_img/modules/async-defer.svg

danielroe:
I'm familiar with what async/defer do.

We have a hard dependency in Nuxt 3 on [browsers that support Proxy](https://caniuse.com/proxy) as this is a requirement for Vue 3.

To implement something like this we would need to make some more changes to rendering, not completely omit all preload/prefetches, as this PR does.

I'm very receptive to anything that will improve Nuxt performance, and appreciate your time in raising this PR.

However, would you instead open a separate issue and explain a bit more about the benefit this will bring? (Ideally with benchmarks.) Is it mostly about increasing support for browsers that don't support preload?

We can then discuss how it might implemented.

sanjaiyan-dev:
> I'm familiar with what async/defer do.
> 
> We have a hard dependency in Nuxt 3 on [browsers that support Proxy](https://caniuse.com/proxy) as this is a requirement for Vue 3.
> 
> To implement something like this we would need to make some more changes to rendering, not completely omit all preload/prefetches, as this PR does.
> 
> I'm very receptive to anything that will improve Nuxt performance, and appreciate your time in raising this PR.
> 
> However, would you instead open a separate issue and explain a bit more about the benefit this will bring? (Ideally with benchmarks.) Is it mostly about increasing support for browsers that don't support preload?
> 
> We can then discuss how it might implemented.

https://github.com/nuxt/framework/issues/7939

PR #7940, opened 2022-10-01
fix(nuxt): page hydration and double load

### LINKED ISSUE #6592: Page component is created twice when navigating to it from page which uses different layout

### LINKED ISSUE #7337: nuxtApp.isHydrating is always false in page setup script when using async page

mmis1000:
### 🔗 Linked issue
Fix #7337, fixes #6592, closes https://github.com/nuxt/framework/pull/7400

### ❓ Type of change

### 📚 Description

This PR is a remake of #7400

Besides of fix and workaround in #7400.

1. This PR use a different approach to handle hydrating, so change of page wrapper isn't required.
2. This PR ensure that layout and any page wrapper never change outside of the window that route navigation happens.
    - so #6592 should never happen again
3. This PR adds proper test for linked issue
    - so regression after this should be noticed

A remaining issue is temporary blank page when switching between layout and navigate at the same time.
But that isn't fixable without vuejs/core#6736.
So it isn't handled here.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/layout.ts
+++ b/packages/nuxt/src/app/components/layout.ts
@@ -1,6 +1,6 @@
 import { computed, defineComponent, h, inject, nextTick, onMounted, Ref, Transition, unref, VNode } from 'vue'
-import { _wrapIf } from './utils'
 import { RouteLocationNormalizedLoaded, useRoute as useVueRouterRoute } from 'vue-router'
+import { _wrapIf } from './utils'
 import { useRoute } from '#app'
 // @ts-ignore
 import layouts from '#build/layouts'
--- a/packages/nuxt/src/app/components/layout.ts
+++ b/packages/nuxt/src/app/components/layout.ts
@@ -1,4 +1,4 @@
-import { computed, defineComponent, h, isRef, nextTick, onMounted, Ref, Transition, VNode } from 'vue'
+import { computed, defineComponent, h, nextTick, onMounted, Ref, Transition, unref, VNode } from 'vue'
 import { _wrapIf } from './utils'
 import { useRouter } from '#app'
 // @ts-ignore
@@ -47,7 +47,7 @@ export default defineComponent({
     // Use router.currentRoute.value instead because this must be changed synchronized with route
     const router = useRouter()
     const route = computed(() => router.currentRoute.value)
-    const layout = computed(() => (isRef(props.name) ? props.name.value : props.name) ?? route.value.meta.layout as string ?? 'default')
+    const layout = computed(() => unref(props.name) ?? route.value.meta.layout as string ?? 'default')
 
     let vnode: VNode
     let _layout: string | false
--- a/packages/nuxt/src/app/components/layout.ts
+++ b/packages/nuxt/src/app/components/layout.ts
@@ -1,6 +1,7 @@
-import { computed, defineComponent, h, nextTick, onMounted, Ref, Transition, unref, VNode } from 'vue'
+import { computed, defineComponent, h, inject, nextTick, onMounted, Ref, Transition, unref, VNode } from 'vue'
 import { _wrapIf } from './utils'
-import { useRouter } from '#app'
+import { RouteLocationNormalizedLoaded, useRoute as useVueRouterRoute } from 'vue-router'
+import { useRoute } from '#app'
 // @ts-ignore
 import layouts from '#build/layouts'
 // @ts-ignore
@@ -44,10 +45,10 @@ export default defineComponent({
     }
   },
   setup (props, context) {
-    // Use router.currentRoute.value instead because this must be changed synchronized with route
-    const router = useRouter()
-    const route = computed(() => router.currentRoute.value)
-    const layout = computed(() => unref(props.name) ?? route.value.meta.layout as string ?? 'default')
+    // Need to ensure (if we are not a child of `<NuxtPage>`) that we use synchronous route (not deferred)
+    const injectedRoute = inject('_route') as RouteLocationNormalizedLoaded
+    const route = injectedRoute === useRoute() ? useVueRouterRoute() : injectedRoute
+    const layout = computed(() => unref(props.name) ?? route.meta.layout as string ?? 'default')
 
     let vnode: VNode
     let _layout: string | false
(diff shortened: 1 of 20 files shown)


danielroe (approved):
This looks great to me - thank you so much 😊

I've made a couple of minor changes and one fix (to ensure this works for layouts that are children of `<NuxtPage>`); let me know if you spot any issues.

mmis1000:
> This looks great to me - thank you so much 😊
> 
> I've made a couple of minor changes and one fix (to ensure this works for layouts that are children of `<NuxtPage>`); let me know if you spot any issues.
> 
> P.S. really appreciate the work on the tests...

Looks fine to me except I am wondering what is this for https://github.com/nuxt/framework/pull/7940/commits/21d872ca3f1b51da7fa2787551db2d85acce24e6#diff-ef9ac7f796e2b5d9a4a232b098d7ce5dd92610191d35dfaf10612a76bfe532ddR50

Is this exist to ensure that layout used inside page changes when writing things like `useRoute().meta.layout = xxx` inside a page component? 

Also, this didn't seem to change any test results, is there a test for it?

danielroe:
Yes, so, for example of you are changing page which has a layout within it, it doesn't force rerender the existing page with the new layout before suspense resolves. And yes, we should add a test for that case. I'll do that later unless you get to it first.

mmis1000:
> Yes, so, for example if you are changing page which has a layout within it, it doesn't force rerender the existing page with the new layout before suspense resolves. And yes, we should add a test for that case. I'll do that later unless you get to it first.

Yea, I think it would makes more sense in anthoer PR instead of this due to following reason.

1. That isn't actually part of this change set
2. You can't test this without alternative app.vue. It need a another fixture of app because the default app.vue defaults to have a layout in it. 
    - To test this, the change set would be too big for this PR.

danielroe:
I have great sympathy for your desire to push it to another PR, but I think it's not as complex as you are worried it will be.

It's just a matter of creating two files:

`~/pages/internal-layout.vue` and `~/pages/internal-layout/[slug].vue`. We can turn off normal layouts in `internal-layout.vue` with `definePageMeta({ layout: false })` and take full control by adding our custom `<NuxtLayout>` inline in the page.

I do think it would be good to add tests for this in this PR as it would be good to check that https://github.com/nuxt/framework/pull/7940/commits/21d872ca3f1b51da7fa2787551db2d85acce24e6 is doing what it's meant to do, and that we don't accidentally regress the fix.

mmis1000:
> I have great sympathy for your desire to push it to another PR, but I think it's not as complex as you are worried it will be.
> 
> It's just a matter of creating two files:
> 
> `~/pages/internal-layout.vue` and `~/pages/internal-layout/[slug].vue`. We can turn off normal layouts in `internal-layout.vue` with `definePageMeta({ layout: false })` and take full control by adding our custom `<NuxtLayout>` inline in the page.
> 
> I do think it would be good to add tests for this in this PR as it would be good to check that [21d872c](https://github.com/nuxt/framework/commit/21d872ca3f1b51da7fa2787551db2d85acce24e6) is doing what it's meant to do, and that we don't accidentally regress the fix.

https://stackblitz.com/edit/github-42bhq4?file=layouts%2Ftest3.vue,pages%2Findex.vue

I think it will still also change the outer layout even you already set `layout: false`. So you need a App.vue without outer layout to test it.

tada5hi:
would love to see this pr merged :see_no_evil:

danielroe:
The only thing it needs is the test I mentioned previously. I'm on it.

tada5hi:
any news on this ?

danielroe (approved):
Looks good to me

Atinux (approved):

Atinux:
Thank you very much!

It would be nice to have a way to detect how much kb add pull requests to the Nuxt default bundle size.

loilo:
So, just out of curiosity (because I ran into #6592 as well) – when is the next RC scheduled to be available? 🥹

danielroe:
Until then, you can try this out in the [edge channel](https://v3.nuxtjs.org/guide/going-further/edge-channel/).

loilo:
Thank you, will take a look. 🥳

PR #7946, opened 2022-10-01
fix(nuxt): handle schema types for relative module paths

danielroe:
### ❓ Type of change

### 📚 Description

In the case where a module is specified with a _relative_ path, it is not properly typed in `schema.d.ts`. For example:

```js
  modules: [
    '../src/module'
  ],
})
```

results in the following type:

```ts
declare module '@nuxt/schema' {
  interface NuxtConfig {
    ["metaKey"]?: typeof import("../src/module").default extends NuxtModule<infer O> ? Partial<O> : Record<string, any>
  }
}
```

But this relative import path either needs to be absolute or resolved relative to the generated schema.d.ts.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/templates.ts
+++ b/packages/nuxt/src/core/templates.ts
@@ -1,7 +1,7 @@
 import type { Nuxt, NuxtApp, NuxtTemplate } from '@nuxt/schema'
 import { genArrayFromRaw, genDynamicImport, genExport, genImport, genObjectFromRawEntries, genString, genSafeVariableName } from 'knitwork'
 
-import { isAbsolute, join, relative } from 'pathe'
+import { isAbsolute, join, relative, resolve } from 'pathe'
 import { resolveSchema, generateTypes } from 'untyped'
 import escapeRE from 'escape-string-regexp'
 import { hash } from 'ohash'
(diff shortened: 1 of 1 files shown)


Atinux (approved):
LGTM

packages/nuxt/src/core/templates.ts
```diff
@@ -120,12 +120,14 @@ export const schemaTemplate: NuxtTemplate<TemplateContext> = {
...
 
+    const relativeRoot = relative(resolve(nuxt.options.buildDir, 'types'), nuxt.options.rootDir)
+
     return [
       "import { NuxtModule } from '@nuxt/schema'",
       "declare module '@nuxt/schema' {",
       '  interface NuxtConfig {',
       ...moduleInfo.filter(Boolean).map(meta =>
-        `    [${genString(meta.configKey)}]?: typeof ${genDynamicImport(meta.importName, { wrapper: false })}.default extends NuxtModule<infer O> ? Partial<O> : Record<string, any>`
+        `    [${genString(meta.configKey)}]?: typeof ${genDynamicImport(meta.importName.startsWith('.') ? './' + join(relativeRoot, meta.importName) : meta.importName, { wrapper: false })}.default extends NuxtModule<infer O> ? Partial<O> : Record<string, any>`
```
pi0, line 130: Have you tested if it also works with extends layers that have different relative root?

danielroe: Could you explain more? If an extends layer adds a module with a relative path I suspect it wouldn't work at all, quite apart from this PR.

PR #8106, opened 2022-10-11
fix(nuxt): don't inline styles for per-request `ssr: false`

### LINKED ISSUE #8093: x-nuxt-no-ssr header: globalThis.__buildAssetsURL is not a function
### Environment

- Operating System: `Linux`
- Node Version:     `v14.19.3`
- Nitro Version:    `0.5.4`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `ssr`, `css`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://codesandbox.io/s/distracted-cookies-x37yoi?file=/assets/css/main.css

(Let me know if that link does not work for you)

### Describe the bug

We have a server middleware to dynamically enable/disable SSR depending on the user agent.

In `nuxt.config.ts` we set `ssr: true`, our middleware then overrides this using
```
event.req.headers['x-nuxt-no-ssr'] = 'true';
```
if serving a user agent that should not get server side rendering.

This leads to problems when using assets in CSS files:

```
[nuxt] [request error] [unhandled] [500] globalThis.__buildAssetsURL is not a function
  at ./.output/server/chunks/app/styles.mjs:7:276
  at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
  at async Promise.all (index 0)
  at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
  at async renderInlineStyles (./.output/server/chunks/handlers/renderer.mjs:453:20)
  at async ./.output/server/chunks/handlers/renderer.mjs:386:25
  at async ./.output/server/chunks/handlers/renderer.mjs:30:22
  at async ./.output/server/node_modules/h3/dist/index.mjs:592:19
  at async Server.nodeHandler (./.output/server/node_modules/h3/dist/index.mjs:538:7)
```

Note that this only happens in production mode, dev mode is fine.

### Additional context

_No response_

### Logs

_No response_

Alex--C:
### 🔗 Linked issue

resolves #8093 
#7580 

### ❓ Type of change

### 📚 Description

Currently, if SSR is enabled during build, the renderer tries to inline styles if `process.env.NUXT_INLINE_STYLES` is true.
It does not take the current request's `ssrContext.noSSR` into account.

This leads to an error if SSR is disabled dynamically (e.g. using the `x-nuxt-no-ssr` header in a server middleware):

```
[nuxt] [request error] [unhandled] [500] globalThis.__buildAssetsURL is not a function
  at ./.output/server/chunks/app/styles.mjs:7:276
  at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
  at async Promise.all (index 0)
  at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
  at async renderInlineStyles (./.output/server/chunks/handlers/renderer.mjs:453:20)
  at async ./.output/server/chunks/handlers/renderer.mjs:386:25
  at async ./.output/server/chunks/handlers/renderer.mjs:30:22
  at async ./.output/server/node_modules/h3/dist/index.mjs:592:19
  at async Server.nodeHandler (./.output/server/node_modules/h3/dist/index.mjs:538:7)
```
 
My change here disables style inlining if `ssrContext.noSSR` is `true`.

Resolves #8093 and #7580 .

### 📝 Checklist


danielroe:
Ideally we would still render global inline styles with `ssr: false`. At the moment, the way we import asset URLs is with `globalThis.__buildAssetsURL` which is set here: https://github.com/nuxt/framework/blob/c404cb1be4bdd568cb77dcd3ae7628ec9068a730/packages/nuxt/src/core/templates.ts#L243-L244

We could either set this manually in `getSPARenderer` or (better) somehow import `paths.mjs` in our emitted `styles.mjs` file.

Thoughts welcome.

Alex--C:
> Ideally we would still render global inline styles with `ssr: false`. 

Wouldn't style inlining for SPA be a completely new feature? From what I understand, `inlineSSRStyles` is currently always false if the app is built without SSR enabled. This PR only addresses the issue that happens when the app is built with SSR enabled but then disabled during runtime.

danielroe:
Oh yes, you're quite right. I do think it would be a nice feature to inline default styles with a SPA, but no changes needed in this PR 😊

danielroe (approved):

PR #8131, opened 2022-10-12
docs: add app.config route to pre-render

Atinux:
### ❓ Type of change

### 📚 Description

Add missing url not detected by crawler 🤔 

### 📝 Checklist

### DIFF
--- a/docs/nuxt.config.ts
+++ b/docs/nuxt.config.ts
@@ -1,3 +1,8 @@
 export default defineNuxtConfig({
-  extends: '@nuxt-themes/website'
+  extends: '@nuxt-themes/website',
+  nitro: {
+    prerender: {
+      routes: ['/guide/directory-structure/app.config']
+    }
+  }
 })
--- a/docs/nuxt.config.ts
+++ b/docs/nuxt.config.ts
@@ -2,7 +2,7 @@ export default defineNuxtConfig({
   extends: '@nuxt-themes/website',
   nitro: {
     prerender: {
-      routes: ['/404.html', '/guide/directory-structure/app.config']
+      routes: ['/', '/404.html', '/guide/directory-structure/app.config']
     }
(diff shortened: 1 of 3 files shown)


danielroe:
Hm. We need to handle 404 page:

CleanShot 2022-10-12 at 12 14 37

Tahul:
https://github.com/nuxt/content/blob/main/src/runtime/pages/document-driven.vue

Docus has no "default" components for DocumentDrivenNotFound and DocumentDrivenEmpty components, but I guess this is the way to go for this kind of issues.

I remember that there was an "issue" (maybe this is expected behavior) with catch-all pages that was overriding the support for error pages in `pages` folder.

Maybe we could work around this by throwing an error in DocumentDriven page when the query fails, and then handling the 404 via Nuxt pages folder?

I also know that as `pages`aren't "extendable", the only way of allowing user to overwrite was to use replaceable components (DocumentDrivenNotFound...).

danielroe:
If we have a design for a custom error page, we can provide it in `~/error.vue` in this repo.

The 404 error is only being thrown on server side in docus, which means we're not redirected to it on a client-side 404 (in this case), and the console errors are coming from missing nav data (owing to not having a defined GITHUB_TOKEN) on client side. We should likely gracefully handle them as well.

Atinux:
It the comment blocking the current PR?

danielroe (approved):
No, this is good to go 👍

Tahul:
Just a little info, this can only be reproduced on built version, when using `dev`, landing on undefined link will display 404/error page

I just locally tried creating both:
```
docs/error.vue
// or
docs/pages/error.vue
```

It doesn't had any effect in built versions, I suspect this is upstream issue as mentioned, `error.vue` page won't be taken in account when catch-all page already setup.

PR #8133, opened 2022-10-12
test: add bundle size test

danielroe:
### ❓ Type of change

- [x] 🧹 Chore

### 📚 Description

This adds a test to constraint bundle size growth for an out-of-the-box install. This will prompt us to update snapshots when/if new packages are added to build.

Any ideas for improving assertion or test more than welcome.

I've also fixed an issue where `vue-router` and `@vue/devtools-api` were traced as dependencies even without router integration.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -1,11 +1,11 @@
 import { defineComponent, h, ref, resolveComponent, PropType, computed, DefineComponent, ComputedRef, onMounted, onBeforeUnmount } from 'vue'
-import { RouteLocationRaw, Router } from 'vue-router'
+import type { RouteLocationRaw, Router } from 'vue-router'
 import { hasProtocol } from 'ufo'
 
 import { navigateTo, useRouter } from '../composables/router'
 import { useNuxtApp } from '../nuxt'
 
-const firstNonUndefined = <T>(...args: (T | undefined)[]) => args.find(arg => arg !== undefined)
+const firstNonUndefined = <T> (...args: (T | undefined)[]) => args.find(arg => arg !== undefined)
 
 const DEFAULT_EXTERNAL_REL_ATTRIBUTE = 'noopener noreferrer'
 
--- a/packages/nuxt/src/app/composables/component.ts
+++ b/packages/nuxt/src/app/composables/component.ts
@@ -1,8 +1,8 @@
 import { defineComponent, getCurrentInstance, reactive, toRefs } from 'vue'
 import type { DefineComponent } from 'vue'
-import { useRoute } from 'vue-router'
 import { NuxtApp, useNuxtApp } from '../nuxt'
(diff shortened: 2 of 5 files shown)


pi0:
> I've also fixed an issue where vue-router and @vue/devtools-api were traced as dependencies even without router integration.

Nice find! Do you mind splitting this into a new fix/perf PR?


I have some concerns about how we do bundle testing. It made honestly lots of trouble over time in Nuxt 2 and made the CI of Nuxt 2 fragile as each package can actually affect bundle size easily and hardcoded size was not a good idea. BTW I'm fine with making test limited to testing used node_modules in snapshot as the list changes much less often.

danielroe:
Done. But for what it's worth, I think it is worth keeping a bundle size test. It takes 10s to update the snapshots for the bundle size, and at least that means that we are _deliberately_ choosing to increase server or client bundle size rather than doing it accidentally.

pi0:
True but it quite often happens because an (expected as we test) external library simply adds new functionality and it adds a few bytes to the server or app bundle. Over time all libs and our internal codebase increase and it is by itself a deliberately chosen of us or 3rd party library authors that make this choice. Bein required to manually update the test, is a duplicate effort unless it _often_ would cause us to revise even using a library or one of its upgrades.

I think the entire logic for size testing, could be an _informational_ CI action similar to code coverage integration signaling us about certain percentage increase. There might be even some ready to use integrations...

In the meantime i appreciate if we remove size cap logic from code completely and add it back as CI integration later.

danielroe:
The logic is already removed unless running manually locally for ease of testing. 😊

pi0:
Sizes are still hardcoded though... Can we write sizes to (a gitginored file) for local testing? Later CI can pick same file as well.

danielroe:
I agree, it would be nice to print results in CI.

pi0 (approved):

pi0:
> When originally pushed it was not.

(or it was but didn't reach the cap size. something else could added few bytes and reveal it.)

danielroe:
I saw that change. But the variability was not so high. The numbers were:

```
    expect(serverBundle.size).toBeLessThan(120000)
    // expect(serverBundle.size).toMatchInlineSnapshot('114018')

    expect(modules.size).toBeLessThan(2700000)
    // expect(modules.size).toMatchInlineSnapshot('2637251')
```

Whereas the difference in your last example is more like 250kB, which is a huge difference!

pi0:
Yeah that's strange! Specially that only windows is bigger. We could make different values between windows and linux for server bundle size....

danielroe:
Is it caused by reading byte size? Maybe this is the difference? Why should the size on disk have been the same, but byte size differ so much?

pi0:
I'm not suggesting it necessary was the reason. I think since then we also migrated to pnpm and upgraded couple of dependencies. 

Anyway since byte size difference now, i think something is really wrong probably with nitro externals (not tests) 😬

Issue #7431, opened 2022-09-12
components.d.ts file does not contain NuxtPage components


Ryongyon:
### Environment

- Operating System: `Windows_NT`
- Node Version:     `v18.7.0`
- Nuxt Version:     `3.0.0-rc.9`
- Nitro Version:    `0.5.1`
- Package Manager:  `npm@8.15.0`
- Builder:          `vite`
- User Config:      `srcDir`, `sourcemap`, `app`, `typescript`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

-

### Describe the bug

I need to import explicitly. 
When I try to `import { NuxtPage } from '#components'`
I found that the NuxtPage component does not exist

danielroe:
Do you have a `pages/` directory? NuxtPage is only added as a component if you do (or if the pages option in your Nuxt config is truthy).

Ryongyon:
@danielroe 
Yes, I have the `pages/` directory

`.nuxt/components.d.ts` :


```
// Generated by components discovery
declare module 'vue' {
  export interface GlobalComponents {
    'NuxtWelcome': typeof import("../node_modules/@nuxt/ui-templates/dist/templates/welcome.vue")['default']
    'NuxtLayout': typeof import("../node_modules/nuxt/dist/app/components/layout")['default']
    'NuxtErrorBoundary': typeof import("../node_modules/nuxt/dist/app/components/nuxt-error-boundary")['default']
    'ClientOnly': typeof import("../node_modules/nuxt/dist/app/components/client-only")['default']
    'ServerPlaceholder': typeof import("../node_modules/nuxt/dist/app/components/server-placeholder")['default']
    'NuxtLink': typeof import("../node_modules/nuxt/dist/app/components/nuxt-link")['default']
    'NuxtLoadingIndicator': typeof import("../node_modules/nuxt/dist/app/components/nuxt-loading-indicator")['default']
    'LazyNuxtWelcome': typeof import("../node_modules/@nuxt/ui-templates/dist/templates/welcome.vue")['default']
    'LazyNuxtLayout': typeof import("../node_modules/nuxt/dist/app/components/layout")['default']
    'LazyNuxtErrorBoundary': typeof import("../node_modules/nuxt/dist/app/components/nuxt-error-boundary")['default']
    'LazyClientOnly': typeof import("../node_modules/nuxt/dist/app/components/client-only")['default']
[...]

PR #8145, opened 2022-10-12
feat(nuxt)!: add NuxtPage to `#components`

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/framework/issues/7431

### ❓ Type of change

### 📚 Description
Hi 🙂  , this PR adds NuxtPage component to `#components`

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -157,12 +157,6 @@ async function initNuxt (nuxt: Nuxt) {
     filePath: resolve(nuxt.options.appDir, 'components/nuxt-loading-indicator')
   })
 
-  // Add <NuxtPage>
-  addComponent({
-    name: 'NuxtPage',
-    filePath: resolve(distDir, 'pages/runtime/page')
-  })
-
   // Deprecate hooks
   nuxt.hooks.deprecateHooks({
     'autoImports:sources': {
--- a/packages/nuxt/src/core/nuxt.ts
+++ b/packages/nuxt/src/core/nuxt.ts
@@ -157,6 +157,12 @@ async function initNuxt (nuxt: Nuxt) {
     filePath: resolve(nuxt.options.appDir, 'components/nuxt-loading-indicator')
   })
 
+  // Add <NuxtPage>
+  addComponent({
+    name: 'NuxtPage',
+    filePath: resolve(distDir, 'pages/runtime/page')
+  })
+
   // Deprecate hooks
   nuxt.hooks.deprecateHooks({
     'autoImports:sources': {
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/core/nuxt.ts
```diff
@@ -157,6 +157,12 @@ async function initNuxt (nuxt: Nuxt) {
     filePath: resolve(nuxt.options.appDir, 'components/nuxt-loading-indicator')
   })
 
+  // Add <NuxtPage>
+  addComponent({
+    name: 'NuxtPage',
+    filePath: resolve(distDir, 'pages/runtime/page')
+  })
```
danielroe, line 164: this should probably only be done in the pages module instead (`packages/nuxt/src/pages/module.ts`) - this should not be added if the pages module isn't enabled.

note that we manually register the component here:

https://github.com/nuxt/framework/blob/829a5505809748083feb06892a9895299af5e4a4/packages/nuxt/src/pages/runtime/router.ts#L53-L56

huang-julien: yes of course 😅 .  
With the components loader, is the manual register still needed ? 🤔

danielroe: Yes, let's try removing it 🤯  But let's manually import in the default `app.vue` file...

We've just done a very similar thing in https://github.com/nuxt/framework/pull/8167#pullrequestreview-1141016400.

huang-julien: Since the release should happens soon (i hope so 😄), what do you think about removing NuxtChild and NuxtChild too ?

Do you mean this one https://github.com/nuxt/framework/blob/829a5505809748083feb06892a9895299af5e4a4/packages/nuxt/src/pages/runtime/app.vue#L3 with a manual import from `#components` ?

pi0 (approved):

packages/nuxt/src/core/nuxt.ts
danielroe: Agreed. I would do a manual relative import.

packages/nuxt/src/pages/runtime/app.vue
```diff
@@ -3,3 +3,7 @@
     <NuxtPage />
   </NuxtLayout>
 </template>
+
+<script setup>
+import NuxtPage from './page'
+</script>
```
danielroe, line 9: Apologies, you were probably right the first time. We don't import `<NuxtLayout>` but rely on the loader here; we can do the same for NuxtPage.

```suggestion

<script setup>
import NuxtPage from './page'
</script>
```

huang-julien: nice !

danielroe (approved):

pi0:
Thanks!

PR #8172, opened 2022-10-13
types: include `nuxt-link` target types

vencho-mdp:
### 🔗 Linked issue

resolves #8171 

### ❓ Type of change

### 📚 Description
Added possible types for target value, in order to provide autocompletion.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -26,7 +26,7 @@ export type NuxtLinkProps = {
   custom?: boolean
 
   // Attributes
-  target?: '_blank' | '_parent' | '_self' | '_top' | string
+  target?: '_blank' | '_parent' | '_self' | '_top' | (string & {}) | null
   rel?: string | null
   noRel?: boolean
 
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -26,7 +26,7 @@ export type NuxtLinkProps = {
   custom?: boolean
 
   // Attributes
-  target?: string | null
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -26,7 +26,7 @@ export type NuxtLinkProps = {
   custom?: boolean
 
   // Attributes
-  target?: string | null
+  target?: '_blank' | '_parent' | '_self' | '_top' | string
```
danielroe, line 29:
```suggestion
  target?: '_blank' | '_parent' | '_self' | '_top' | (string & {}) | null
```

DamianGlowala: Knowing that `target` is optional, is `null` actually needed?

vencho-mdp: If null is a type option then it'd allow something like `:target="null"`. Not so sure if it has a use case.

DamianGlowala: Even if we wanted to use a ternary operator like the one below, I'd naturally go for `undefined` (knowing the property is optional).

```
:target="variable ? '_blank' : undefined"
```

@danielroe, shall all `null` values be removed for optional properties (line 14, 29 and 30)? There is some inconsistency at the moment.

danielroe: No, null values serve an important purpose of overwriting default values. See https://github.com/nuxt/framework/pull/3544#issuecomment-1062061430 and other comments there.

danielroe (approved):

pi0 (approved):
Thanks <3

PR #8190, opened 2022-10-14
feat(nuxt)!: `refresh` to override previous requests by default

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/pull/7864

### ❓ Type of change

### 📚 Description

This renames `override` to `dedupe` and enables it by default, which should be more the expected behaviour.

### 👉 Migration

If you were relying on the previous behaviour (not re-calling a fetch function if one was already running) then you can update your refresh calls like so:

```diff
- refresh()
+ refresh({ dedupe: false })
```

### 📝 Checklist


pi0 (approved):
I personally still believe `cancel: false` is easier to read than `dedup: true` but both are good....

PR #8198, opened 2022-10-14
fix(kit): log module id to the console when import fails

### LINKED ISSUE #8191: [DX] Improve error when added module is no module
### Describe the feature

When adding a npm package to `modules` that actually is not a Nuxt module (but e.g. exports CSS), the error message is rather cryptic:

e.g. ` ERROR  Cannot start nuxt:  expected expression, got ':'                                                 15:56:50`

It would be nice to provide a more readable warning and also hint that the module that is failing might not be a module.

[Repro](https://stackblitz.com/edit/https-github-comnuxtframeworkissues8191?file=nuxt.config.ts)

### Additional information

### Final checks

Araneline:
### 🔗 Linked issue

#8191 

### ❓ Type of change

### 📚 Description

Resolves #8191 - in case of error during module require/import it will be printed to the console in the form of 'Error while requiring module {module_name}: {error}'.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/internal/cjs.ts
+++ b/packages/kit/src/internal/cjs.ts
@@ -1,3 +1,4 @@
+import { relative } from 'node:path'
 import { pathToFileURL } from 'node:url'
 import { join, normalize } from 'pathe'
 import { interopDefault } from 'mlly'
@@ -124,7 +125,7 @@ export function requireModule (id: string, opts: RequireModuleOptions = {}) {
 
     return requiredModule
   } catch (error: unknown) {
-    console.error(`Error while requiring module \`${id.split('/').pop()}\`: ${error}`)
+    console.error(`Error while requiring module \`${relative(process.cwd(), resolvedPath)}\`: ${error}`)
     throw error
   }
 }
--- a/packages/kit/src/internal/cjs.ts
+++ b/packages/kit/src/internal/cjs.ts
@@ -118,10 +118,15 @@ export function requireModule (id: string, opts: RequireModuleOptions = {}) {
     clearRequireCache(resolvedPath)
   }
 
-  // Try to require
-  const requiredModule = _require(resolvedPath)
+  try {
+    // Try to require
+    const requiredModule = _require(resolvedPath)
 
-  return requiredModule
+    return requiredModule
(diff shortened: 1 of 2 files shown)


packages/kit/src/internal/cjs.ts
```diff
@@ -118,10 +118,15 @@ export function requireModule (id: string, opts: RequireModuleOptions = {}) {
...
-  // Try to require
-  const requiredModule = _require(resolvedPath)
+  try {
+    // Try to require
+    const requiredModule = _require(resolvedPath)
 
-  return requiredModule
+    return requiredModule
+  } catch (error: unknown) {
+    console.error(`Error while requiring module \`${id.split('/').pop()}\`: ${error}`)
```
pi0, line 127: Shall we preserve the original module id in the error log?

Araneline: Original id as I found out is a path to the module in case of not module-module at least. So I decided against printing the whole line, extracting only the last part of the path.

If you believe the whole path string would be better, I can commit that fix.

pi0: We can use this: `relative(process.cwd(), resolvedPath)` to make log more beatiful:

image

pi0:
Instead of `cjs.ts` util, we might add a try-catch directly to `module/install.ts:45` for better stack trace. The utility in cjs a confusing name it requires a _node module_ while in `install.ts` we have access to user input (not resolved) to show in custom error. It also handles ESM branch.

packages/kit/src/internal/cjs.ts
Araneline: Definitely. Thank you for pointing in the right direction. Made another commit.

Araneline:
> Instead of `cjs.ts` util, we might add a try-catch directly to `module/install.ts:45` for better stack trace. The utility in cjs a confusing name it requires a _node module_ while in `install.ts` we have access to user input (not resolved) to show in custom error. It also handles ESM branch.

Think I got it. Moved the try-catch block to install.ts and used an unresolved module name in the error.

pi0 (approved):

pi0:
Thanks!

PR #8205, opened 2022-10-14
fix(nuxt): fix broken external urls when baseURL is set in config

### LINKED ISSUE #7983: useFetch/$fetch and external URLs is broken when used in an app configured with a baseUrl
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.17.1`
- Nitro Version:    `0.5.4`
- Package Manager:  `npm@8.15.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/nuxt-starter-fsyggj?file=app.vue

Having a hard time reproducing the exact issue that I'm having. But this basic setup isn't working either so hopefully related.

Added two examples in App.vue (one of the commented out). 

### Describe the bug

When using baseUrl config (either by ENV or nuxt.config) useFetch / $fetch isn't working properly with external URLs.

Not entirely sure what is going on, basically what happens locally in my own project is that when using an external url (e.g. https://graphql.contentful.com/content/[...]) it gets called with a starting '/' which makes the url relative and of course resulting in a 404 (e.g. /https://graphql.contentful.com/content/[...]).

I can't reproduce this in stackblitz, what happens there instead is that the call is made over and over again, never returning any data. It is like it's stuck in a recursive loop. Removing the baseUrl configuration resolves the issue.

### Additional context

_No response_

### Logs

_No response_

Araneline:
### 🔗 Linked issue

#7983 

### ❓ Type of change

### 📚 Description

Resolves #7983 - set baseURL for ohmyfetch dynamically through interceptor. The urls were prefixed with baseURL even if they were external.

Using **onRequest** method of ohmyfetch we set a request interceptor which checks for:
- http(-s) at the beginning of request url
- not set baseURL in options

If both are true, it sets the baseURL to `''` (allowing external link calls). Otherwise if not already in request config - sets the baseURL

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/entry.ts
+++ b/packages/nuxt/src/app/entry.ts
@@ -19,7 +19,7 @@ if (!globalThis.$fetch) {
     onRequest ({ request, options }) {
       if (/^https?:\/\//gm.test(<string>request) && !options.baseURL) {
         options.baseURL = ''
-      } else {
+      } else if (!options.baseURL) {
         options.baseURL = baseURL()
       }
     }
--- a/packages/nuxt/src/app/entry.ts
+++ b/packages/nuxt/src/app/entry.ts
@@ -15,7 +15,14 @@ import AppComponent from '#build/app-component.mjs'
 if (!globalThis.$fetch) {
   // @ts-ignore
   globalThis.$fetch = $fetch.create({
-    baseURL: baseURL()
+    // @ts-ignore
+    onRequest ({ request, options }) {
+      if (/^https?:\/\//gm.test(<string>request) && !options.baseURL) {
+        options.baseURL = ''
+      } else {
+        options.baseURL = baseURL()
+      }
+    }
   })
 }
 


danielroe (commented):
This seems like an upstream bug to me, and we should likely resolve in ufo or ohmyfetch.

Previously we had a similar bug which was resolved in https://github.com/unjs/ufo/pull/68.

-> also posted in https://github.com/nuxt/framework/issues/7983#issuecomment-1279743702

Araneline:
> This seems like an upstream bug to me, and we should likely resolve in ufo or ohmyfetch.
> 
> 
> 
> Previously we had a similar bug which was resolved in https://github.com/unjs/ufo/pull/68.
> 
> 
> 
> -> also posted in https://github.com/nuxt/framework/issues/7983#issuecomment-1279743702

Ok. Would you mind if I had a look at ufo and try to fix it there? Alternatively I can close this PR.

danielroe:
That would be lovely 😊

Araneline:
Quick update/note: seems like the issue is boiled down to `withBase` method from _ufo_ which is used if `baseURL` is set in _ohmyfetch_.

Araneline:
@danielroe the fix to that problem was already committed 3 months ago to the [unjs/ufo](https://github.com/unjs/ufo). There was a patch which introduced check of `hasProtocol` alongside `isEmptyURL` in `withBase` function.

Looking at [unjs/ohmyfetch](https://github.com/unjs/ohmyfetch) repo after pnpm install, it's clear that the old version of the ufo is installed since there's no `hasProtocol` check.

**The solution:** update the version of unjs/ufo in unjs/fetch (probably releasing a new version first for ufo).

This branch can be safely disregarded after that.

danielroe:
Thanks for investigating. ❤️🙏

danielroe:
@pi0 shall i release new versions of  `ufo`  and `ohmyfetch`?

pi0:
`ufo@0.8.6` is released. The latest release of `ohmyfetch` has the compatible semver range dependency on ufo 0.8x. Please try recreating lockfile. It should fix the issue.

PR #8216, opened 2022-10-15
feat(nuxt): add `<NuxtClientFallback>` component

huang-julien:
### 🔗 Linked issue

resolve  https://github.com/nuxt/nuxt.js/issues/14885

### ❓ Type of change

### 📚 Description
Hi :wave: , this PR adds a `<ClientIfFail>` component to render its children even if it fails in SSR.
If it fails, the children will only be rendered `onMounted` hook. An `@ssr-error` event will also be triggered server-side.

This PR can be  improved and i'd like to have your opinion:
- [x] About the name of this component
- [x] Currently, this component needs a `uid` prop to know if it has failed to render in SSR. I think this can be improved by avoiding the users to set manually a random uid. Maybe by using a transform plugin to auto-generate a `uid="randomString"` on every `<CleintIfFail>` ? 
- [ ] https://github.com/nuxt/nuxt.js/issues/14885 @danielroe suggested enhancing `<NuxtErrorBoundary>`, do you think we can merge this component's behavior with `<NuxtErrorBoundary>` ? 
 

### 📝 Checklist


packages/nuxt/src/components/client-if-fail-auto-id.ts
```diff
@@ -0,0 +1,48 @@
...
+        return true
+      }
+      return isVueTemplate(id)
+    },
+    transform (code, id) {
+      const s = new MagicString(code)
+      const relativeID = isAbsolute(id) ? relative(options.rootDir, id) : id
+
+      s.replace(/<([cC]lient-?[iI]f-?[fF]ail)(.*)>/g, (full, name, attrs, offset) => {
+        return `<${name}${attrs} uid="${'$' + hash(`${relativeID}-${offset}`)}">`
```
pi0, line 35: Only depending on static component usage, has a limitation. Having `<BlogArticle :id>` component inside of it uses `<ClientIfFail>`, if it used twice and one of them fails, since static file/location of ClientIfFail is same both of them fail. 

We should at least document this as a limitation or we can also generate id at runtime based on `hash(uid, attrs)`

packages/nuxt/src/app/components/client-if-fail.mjs
```diff
@@ -0,0 +1,40 @@
...
+    if (process.server) {
+      const error = ref(false)
+
+      onErrorCaptured((_, instance) => {
+        error.value = true
+
+        useState(`${props.uid}`, () => true)
+        // modify ssr render to force render a simple div
+        instance._.ssrRender = (_ctx, _push, _parent, _attrs) => {
+          _push(`<div${ssrRenderAttrs(_attrs)}></div>`)
```
pi0, line 21: Can we make at least `div` tag customizable via a prop?

pi0:
LOVE THIS! 

For the naming component, I suggest using `<ClientFallback>`. I thought we could have more client-fallback strategies such as timeout and a more generic name could cover them.

Merging with `NuxtErrorBoundary` once it supported SSR error capturing, might be a really nice idea but we don't need to wait for it. Also same as ClientOnly (component vs `.client`) we might have more specific implementation that differs from `<NuxtErrorBoundary>` that is directly exposed to user-land, which might cause more logic into it.

packages/nuxt/src/components/client-if-fail-auto-id.ts
```diff
@@ -0,0 +1,48 @@
+import { createUnplugin } from 'unplugin'
+import { ComponentsOptions } from '@nuxt/schema'
```
pi0, line 2:
```suggestion
import type { ComponentsOptions } from '@nuxt/schema'
```

packages/nuxt/src/app/components/client-if-fail.mjs
huang-julien: yes :+1:

packages/nuxt/src/components/client-if-fail-auto-id.ts
huang-julien: Ah yes, I think that i see how this can potentially leads to this issue 🤔 .
If i understand well, : 
- App.vue
- BlogArticle.vue
- FailOrSuccess 

Having App.vue calling multiple BlogArticle with an `:id` (1, 2) that call a `FailOrSuccess` comp in `ClientOnly` -> if id 1 fails, id 2 would also be considered as failed since this is depending on the location of the component and the offset of `ClientFail`

pi0: Yes exactly.

huang-julien:
This PR has been reworked to have the same behavior with `ClientOnly` if at least one children fail to render.

At first, the PR was trying to handle only the components that has failed to render but handling it in client-side became a bit too complex. The new behavior is to render a `div` or its `fallbackTag` without any children if one child of the slot fails to render. As if we were using `ClientOnly`.

There's now a runtime uid for each `ClientFallback` based on:
- a hash of the relative ID of the file
- the attrs of the component 
- the number of `ClientFallback` in the component

huang-julien:
Added the `fallbackTag`|`placeholderTag `and the `fallback`|`placeholder` props + the `fallback` slot (just like `<ClientOnly>`).

There's a  `[Vue warn]: Component <Anonymous> is missing template or render function.` each time a component fails to render in ssr once build, i'm not sure if this should be documented or not.

huang-julien:
I'm thinking about splitting the component into a server and client version to avoid having the server-side logic into the client build (dev+prod).

I wanted to use `mode: 'client'` in `addComponent()` but this automatically convert it to `createClient()` so it doesn't work.

Using `const component = process.server ? ... : ...` then `export default defineComponent(component)` works + extracting `createBuffer` to a helper files removes around 10kb from the client build.

What do you all think about this (leave it like this, apply the new suggestion or any idea) ?

huang-julien:
@pi0 the module (and the PR) should be finally stable. Let me know what you think of it

packages/nuxt/src/components/client-fallback-auto-id.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (exclude.some(pattern => id.match(pattern))) {
+        return false
+      }
+      if (include.some(pattern => id.match(pattern))) {
+        return true
+      }
+      return isVueTemplate(id)
+    },
+    transform (code, id) {
+      if (!/[cC]lient-?[fF]allback/.test(code)) { return }
```
danielroe, line 30:
```suggestion
      if (!/(ClientFallback|client-fallback)/.test(code)) { return }
```

packages/nuxt/src/components/client-fallback-auto-id.ts
```diff
@@ -0,0 +1,52 @@
...
+import MagicString from 'magic-string'
+import { isAbsolute, relative } from 'pathe'
+import { hash } from 'ohash'
+import { isVueTemplate } from './helpers'
+interface LoaderOptions {
+  sourcemap?: boolean
+  transform?: ComponentsOptions['transform'],
+  rootDir: string
+}
+
```
danielroe, line 12: Maybe we could extract as a single regexp?

```suggestion
const CLIENT_FALLBACK_RE = /<(ClientFallback|client-fallback)( [^>]*)?>/g
```

packages/nuxt/src/components/client-fallback-auto-id.ts
```diff
@@ -0,0 +1,52 @@
...
+      return isVueTemplate(id)
+    },
+    transform (code, id) {
+      if (!/[cC]lient-?[fF]allback/.test(code)) { return }
+
+      const s = new MagicString(code)
+      const relativeID = isAbsolute(id) ? relative(options.rootDir, id) : id
+      let count = 0
+
+      s.replace(/<([cC]lient-?[fF]allback)( [^>]*)?>/g, (full, name, attrs) => {
```
danielroe, line 36:
```suggestion
      s.replace(CLIENT_FALLBACK_GLOBAL_RE, (full, name, attrs) => {
```

danielroe:
The bundle test is flagging up that importing from `vue/server-renderer` in the client app, there are side-effects.

If possible, we should avoid importing from there, or manually mark it as without side effects to rollup.

huang-julien:
I think splitting it into .server/.client is a bit cleaner, it would also avoid to include 'vue/server-renderer' in the client bundle.

PR #8225, opened 2022-10-15
perf(nuxt): improve link prefetching

danielroe:
### ❓ Type of change

### 📚 Description

This PR bundles a number of improvements to the link prefetching feature of `<NuxtLink>`:

1. It implements crossorigin prefetches using the [Speculation Rules API](https://github.com/WICG/nav-speculation/blob/main/triggers.md) - this is behind a flag for experimentation.
2. It disables prefetching for links with `target: _blank` as there is no benefit to doing so.
3. It throttles preload requests (as, the way we do it, these are high priority requests and could block other more urgent user interaction with the page). This implementation could likely be improved but I aimed to keep it lightweight.

### 📝 Checklist


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -329,21 +329,30 @@ function isSlowConnection () {
...
 
   if (!router._nuxtLinkPreloaded) { router._nuxtLinkPreloaded = new Set() }
   if (router._nuxtLinkPreloaded.has(to)) { return }
   router._nuxtLinkPreloaded.add(to)
 
+  const promises = router._preloadPromises ||= []
+
+  if (promises.length > 4) {
+    // Defer adding new preload requests until the existing ones have resolved
+    return Promise.all(promises).then(() => preloadRouteComponents(to, router))
```
pi0, line 343: Can we move this logic to preloadComponents?

packages/nuxt/src/app/plugins/cross-origin-prefetch.client.ts
```diff
@@ -0,0 +1,29 @@
...
+            }
+          ]
+        })
+      })
+    ]
+  })
+  nuxtApp.hook('link:prefetch', (url) => {
+    const { protocol } = parseURL(url)
+    if (!protocol || ['http:', 'https:'].includes(protocol)) {
+      externalURLs.value.add(url)
```
pi0, line 26: Why is `!protocol` assumed as external?

packages/nuxt/src/app/components/nuxt-link.ts
Aareksio: Composable would enable developers to easily code custom preload on user intent, instead of viewport intersection.

packages/nuxt/src/app/plugins/cross-origin-prefetch.client.ts
danielroe: it wasn't meant to be assuming as external; the speculation rules example gives [an example of local urls as well](https://github.com/WICG/nav-speculation/blob/main/triggers.md#requirements). Having said that, I think likely the browser behaviour isn't what we want in a SPA, so I'll remove.

packages/nuxt/src/app/composables/preload.ts
```diff
@@ -31,3 +33,31 @@ function _loadAsyncComponent (component: Component) {
...
+export async function preloadRouteComponents (to: string, router: Router & { _routePreloaded?: Set<string>; _preloadPromises?: Array<Promise<any>> } = useRouter()): Promise<void> {
+  if (process.server) { return }
+
+  if (!router._routePreloaded) { router._routePreloaded = new Set() }
+  if (router._routePreloaded.has(to)) { return }
+  router._routePreloaded.add(to)
+
+  const promises = router._preloadPromises ||= []
+
+  if (promises.length > 4) {
```
pi0, line 46: This is a nice refactor that `preloadRouteComponents` can be a standalone composable too. What i meant in https://github.com/nuxt/framework/pull/8225#discussion_r996413329 was that if we could move rate limit directly to `preloadComponents` composable (the one above) 🙈

danielroe: `preloadComponents` isn't used by link prefetching. Do you mean, implement the same functionality there as well, or extract to a shared utility?

pi0: Indeed then let's do it later

packages/nuxt/src/app/composables/index.ts
```diff
@@ -12,5 +12,5 @@ export type { CookieOptions, CookieRef } from './cookie'
 export { useRequestHeaders, useRequestEvent, setResponseStatus } from './ssr'
 export { abortNavigation, addRouteMiddleware, defineNuxtRouteMiddleware, setPageLayout, navigateTo, useRoute, useActiveRoute, useRouter } from './router'
 export type { AddRouteMiddlewareOptions, RouteMiddleware } from './router'
-export { preloadComponents, prefetchComponents } from './preload'
+export { preloadComponents, prefetchComponents, preloadRouteComponents } from './preload'
```
pi0, line 15: Shall we add it to auto imports too?

pi0 (approved):

PR #8226, opened 2022-10-15
fix(webpack): add global to new line

danielroe:
### ❓ Type of change

### 📚 Description

Previously we could end up with invalid JS and the following error:

```bash
[nuxt] [request error] [unhandled] [500] publicAssetsURL__webpack_public_path__ is not defined
```

### 📝 Checklist

### DIFF
--- a/packages/webpack/src/plugins/dynamic-base.ts
+++ b/packages/webpack/src/plugins/dynamic-base.ts
@@ -21,7 +21,7 @@ export const DynamicBasePlugin = createUnplugin((options: DynamicBasePluginOptio
         return
       }
       const s = new MagicString(code)
-      s.append(`${options.globalPublicPath} = buildAssetsURL();\n`)
+      s.append(`\n${options.globalPublicPath} = buildAssetsURL();\n`)
       return {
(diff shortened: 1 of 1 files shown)


pi0 (approved):

packages/webpack/src/plugins/dynamic-base.ts
```diff
@@ -21,7 +21,7 @@ export const DynamicBasePlugin = createUnplugin((options: DynamicBasePluginOptio
         return
       }
       const s = new MagicString(code)
-      s.append(`${options.globalPublicPath} = buildAssetsURL();\n`)
+      s.append(`\n${options.globalPublicPath} = buildAssetsURL();\n`)
```
pi0, line 24: @antfu What do you think about supporting `s.appendLine`?

antfu: We didn't have the concept of `line` in MagicString, as all operations are index based. I am not sure if this complexity is necessary, giving the fix here is straightforward, while on the other hand we might end up introduce the `xxxLine` version for every magic string methods.

PR #8245, opened 2022-10-17
refactor(nuxt): use h3 early hints util

danielroe:
### ❓ Type of change

### 📚 Description

This uses new h3 `writeEarlyHints` util: https://github.com/unjs/h3/pull/184

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -1,7 +1,7 @@
 import { createRenderer, renderResourceHeaders } from 'vue-bundle-renderer/runtime'
 import type { RenderResponse } from 'nitropack'
 import type { Manifest } from 'vite'
-import { appendHeader, getQuery } from 'h3'
+import { appendHeader, getQuery, writeEarlyHints } from 'h3'
 import devalue from '@nuxt/devalue'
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -160,7 +160,7 @@ export default defineRenderHandler(async (event) => {
   if (!isRenderingPayload && !process.env.prerender && event.res.socket) {
     const { link } = renderResourceHeaders({}, renderer.rendererContext)
     // TODO: use https://github.com/nodejs/node/pull/44180 when we drop support for node 16
```
pi0: Shall we drop TODO and `&& res.socket` checks now? (they are in h3)

packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -159,8 +159,7 @@ export default defineRenderHandler(async (event) => {
   // Render 103 Early Hints
   if (!isRenderingPayload && !process.env.prerender && event.res.socket) {
```
pi0, line 160: It should be checked in h3 already..

```suggestion
  if (!isRenderingPayload && !process.env.prerender) {
```

PR #8278, opened 2022-10-18
fix(nuxt): lazy-load entry CSS

### LINKED ISSUE #7619: prevent loading global style entry css file
We now have styles inlined in the page (https://github.com/nuxt/framework/pull/7160), and we've switched the global CSS file to a non-blocking prefetch. However, the vite preloader does still add these styles to the page, and we can probably avoid this (or have an option to avoid this).

**Note**: this may be unfeasible as `entry.css` may include _more_ CSS than is inlined in the page, and be necessary for subsequent navigation. See https://github.com/nuxt/framework/issues/7623 for example.

danielroe:
### 🔗 Linked issue

resolves #7619

### ❓ Type of change

### 📚 Description

With https://github.com/nuxt-contrib/vue-bundle-renderer/pull/39, we are now not blocking render with stylesheet, but we still need to load entry CSS on client side to include client-only CSS (among other things). This PR:

1. adds back support for rendering inline styles for `ssr: false` routes: https://github.com/nuxt/framework/pull/8106
2. adds in lazy css links which use onload media strategy to load CSS on client-side only

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -10,7 +10,12 @@ import { useRuntimeConfig, useNitroApp, defineRenderHandler, getRouteRules } fro
 import type { NuxtApp, NuxtSSRContext } from '#app'
 
 // @ts-ignore
-import { buildAssetsURL } from '#paths'
+import { buildAssetsURL, publicAssetsURL } from '#paths'
+
+// @ts-ignore
+globalThis.__buildAssetsURL = buildAssetsURL
+// @ts-ignore
+globalThis.__publicAssetsURL = publicAssetsURL
```
pi0, line 18: Wouldn't they register as side-effects of `#paths`? (and if they do, do we still need it from `#paths`?

danielroe: They aren't registered as globals [there](https://github.com/nuxt/framework/blob/256df28124611871949201326d327c7bf38d37e8/packages/nuxt/src/core/runtime/nitro/paths.ts#L1-L20).

And yes, we probably could wrap these lines in `if (process.client)` - will try:

https://github.com/nuxt/framework/blob/c404cb1be4bdd568cb77dcd3ae7628ec9068a730/packages/nuxt/src/core/templates.ts#L243-L244

pi0 (approved):

PR #8299, opened 2022-10-18
fix(nuxt): don't use or assignment

danielroe:
### ❓ Type of change

### 📚 Description

This is seemingly not compiled down 🤔 and is [not compatible](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment#browser_compatibility) with Node 14.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/preload.ts
+++ b/packages/nuxt/src/app/composables/preload.ts
@@ -41,7 +41,7 @@ export async function preloadRouteComponents (to: string, router: Router & { _ro
   if (router._routePreloaded.has(to)) { return }
   router._routePreloaded.add(to)
 
-  const promises = router._preloadPromises ||= []
+  const promises = router._preloadPromises = router._preloadPromises || []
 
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/preload.ts
```diff
@@ -41,7 +41,7 @@ export async function preloadRouteComponents (to: string, router: Router & { _ro
   if (router._routePreloaded.has(to)) { return }
   router._routePreloaded.add(to)
 
-  const promises = router._preloadPromises ||= []
+  const promises = router._preloadPromises = router._preloadPromises || []
```
DamianGlowala, line 44:
```suggestion
  const promises = router._preloadPromises || []
```

danielroe: We also want to share them across invocations of this function

pi0 (approved):

PR #8371, opened 2022-10-20
fix(nuxi): check for property to exist

danielroe:
### ❓ Type of change

### 📚 Description

In Bridge `nuxt.options.typescript` is not defined.

### 📝 Checklist

### DIFF
--- a/packages/nuxi/src/utils/prepare.ts
+++ b/packages/nuxi/src/utils/prepare.ts
@@ -28,7 +28,7 @@ export const writeTypes = async (nuxt: Nuxt) => {
       './nuxt.d.ts',
       join(relative(nuxt.options.buildDir, nuxt.options.rootDir), '**/*'),
       ...nuxt.options.srcDir !== nuxt.options.rootDir ? [join(relative(nuxt.options.buildDir, nuxt.options.srcDir), '**/*')] : [],
-      ...nuxt.options.typescript.includeWorkspace && nuxt.options.workspaceDir !== nuxt.options.rootDir ? [join(relative(nuxt.options.buildDir, nuxt.options.workspaceDir), '**/*')] : []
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/utils/prepare.ts
```diff
@@ -28,7 +28,7 @@ export const writeTypes = async (nuxt: Nuxt) => {
       './nuxt.d.ts',
       join(relative(nuxt.options.buildDir, nuxt.options.rootDir), '**/*'),
       ...nuxt.options.srcDir !== nuxt.options.rootDir ? [join(relative(nuxt.options.buildDir, nuxt.options.srcDir), '**/*')] : [],
-      ...nuxt.options.typescript.includeWorkspace && nuxt.options.workspaceDir !== nuxt.options.rootDir ? [join(relative(nuxt.options.buildDir, nuxt.options.workspaceDir), '**/*')] : []
+      ...nuxt.options.typescript?.includeWorkspace && nuxt.options.workspaceDir !== nuxt.options.rootDir ? [join(relative(nuxt.options.buildDir, nuxt.options.workspaceDir), '**/*')] : []
```
pi0, line 31: Can we normalize `typescript` in schema to have a default object value? (this can help to avoid similar issues)

danielroe: It is normalized in schema. The issue is that Bridge does not use schema.

pi0: Then maybe we can fix bridge instead to add polyfill for `typescript`?

pi0: Considering bridge is still with edge releases, i think it is faster if they migrate to latest versions of it instead of waiting for rc.13. I understand this is an easy fix here but over time we will be adding implicit patches to nuxt 3 codebase because of bridge.

PR #8383, opened 2022-10-21
fix(nuxt): routerBehavior comparison for hash block

jiblett1000:
### ❓ Type of change

- [ x] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

Unless I'm mistaken, the comparison made here should be checking if the routes are `===` and not `!==`. Before, the hash logic wasn't reachable.

### 📝 Checklist


pi0 (approved):
Nice spot!

PR #8421, opened 2022-10-23
fix(nuxt): avoid directly importing `vue-router` inside `<NuxtLayout>`

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/8415

### ❓ Type of change

### 📚 Description

As noted in https://github.com/nuxt/framework/pull/8133, this follows up to remove a `vue-router` dependency from all use of `<NuxtLayout>`.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -23,6 +23,11 @@ export default defineNuxtModule({
     // Disable module (and use universal router) if pages dir do not exists or user has disabled it
     if ((nuxt.options.pages === false || (nuxt.options.pages !== true && !pagesDirs.some(dir => existsSync(dir)))) && !isRouterOptionsPresent) {
       addPlugin(resolve(distDir, 'app/plugins/router'))
+      // Add vue-router import for `<NuxtLayout>` integration
+      addTemplate({
+        filename: 'vue-router-imports.mjs',
+        getContents: () => 'export { useRoute } from \'#app\''
+      })
       return
     }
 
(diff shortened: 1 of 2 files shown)


Atinux (approved):

packages/nuxt/src/pages/module.ts
```diff
@@ -23,6 +23,11 @@ export default defineNuxtModule({
     // Disable module (and use universal router) if pages dir do not exists or user has disabled it
     if ((nuxt.options.pages === false || (nuxt.options.pages !== true && !pagesDirs.some(dir => existsSync(dir)))) && !isRouterOptionsPresent) {
       addPlugin(resolve(distDir, 'app/plugins/router'))
+      // Add vue-router import for `<NuxtLayout>` integration
+      addTemplate({
+        filename: 'vue-router-imports.mjs',
+        getContents: () => 'export { useRoute } from \'#app\''
```
pi0, line 29: In this case, virtual name is not accurate. Maybe we can rename rename it to `pages.mjs` or something?

(as better refactor we shall provide a generic internal `#imports` layer but fine to not doing it now)

danielroe: `#imports` conflicts with our unimport integration.

danielroe: ... and just in case you're wondering, we can't import from it in virtual templates.

pi0: Both I'm aware of :) Suggesting to make a new internal _like_ `#imports` for this purpose...

pi0 (approved):

Issue #8414, opened 2022-10-22
Memory Leak on route navigation


warflash:
### Environment

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-rc.12`
- Nitro Version:    `0.6.0`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------

### Reproduction

https://v3.nuxtjs.org/getting-started/introduction

### Describe the bug

repeatedly switch tabs from e.g. `get started` to `guide` and watch the memory increase.
Fresh load
![image](https://user-images.githubusercontent.com/22006082/197361818-10e11af2-e8c4-4ca9-b356-05e40ae8eeda.png)
After 30 swaps
![image](https://user-images.githubusercontent.com/22006082/197361829-1373a5da-fb7e-4f99-8b79-4ae9a1d37504.png)

### Additional context

For bigger pages this results in 100s of leaked MBs within only a handful of route changes which in turn makes switching routes take seconds and is really bad user experience.

warflash:
@danielroe Just to make sure, this is not limited to the docs (since you added the docs label) and rather seems like common behaviour

danielroe:
@warflash Can you identify any commonalities between your site and the docs? (Are you using Nuxt content, for example?) Are you able to provide a minimal reproduction?

warflash:
Hey @danielroe we are using content indeed, so I tried removing it and the issue persisted. 
I do not have a minimal reproduction but the same behaviour can be see with nuxt movies which you guys recently released, which is somewhat minimal I guess? 
https://movies.nuxt.space/search
swapping back and fourth between any page takes up ram which even when force garbage collecting is not given back. It doesnt grow as quickly as in our app here, probably due to the app being simpler but the behaviour is the exact same.
Initial load:
![image](https://user-images.githubusercontent.com/22006082/197385357-cccf8c5a-dd18-4aaa-9f89-86e38bf04cac.png)
20 swaps or so between `/search` and `/tv`, force gc'ed afterwards:
![image](https://user-images.githubusercontent.com/22006082/197385382-1c8a0a26-801c-4926-8937-ac483f43755b.png)
Seems to be even worse on firefox, gotten up to 60mb on movies with just a few navigations.

warflash:
[...]

PR #8434, opened 2022-10-24
docs: disable transitions by default

danielroe:
### 🔗 Linked issue

hotfix to work around #8414 

### ❓ Type of change

### 📚 Description

I do not believe we are currently using transitions so we probably should disable these anyway. This also serves to seriously reduce a memory leak that we're looking into.

### 📝 Checklist

### DIFF
--- a/docs/nuxt.config.ts
+++ b/docs/nuxt.config.ts
@@ -1,6 +1,10 @@
 import { join } from 'pathe'
 
 export default defineNuxtConfig({
+  app: {
+    pageTransition: false,
+    layoutTransition: false
+  },
   extends: process.env.WEBSITE_THEME ? join(process.env.WEBSITE_THEME, 'theme') : '@nuxt-themes/website',
   build: {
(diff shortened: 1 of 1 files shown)


pi0 (approved):
LGTM. /cc @Atinux

Atinux:
Thanks @danielroe 

Shall we simply disable transitions by default in Nuxt 3? 🤔

danielroe:
There is merit in that. They cause issues like https://github.com/nuxt/framework/issues/3587 when many users may not be using them at all ...

wdyt @pi0?

Atinux:
It also brings Nuxt 3 as light as possible when starting

pi0:
Makese sense to disable transitions by default. We only need proper way to guide users to enable it back

manniL:
I think disabling them by default is a good idea to reduce friction, as long as docs cover how to use them (e.g. with an extra section).

Issue #7469, opened 2022-09-13
allow setting headers or returning a redirect from `error.vue`


KoenCa:
### Environment

- Nuxi 3.0.0-rc.9
- Node Version:     `v16.15.1`
- Nuxt Version:     `3.0.0-rc.9`
- Nitro Version:    `0.5.1`
- Builder:          `vite`

### Reproduction

Simple reproduction on Stackblitz: https://stackblitz.com/edit/github-2ty5wn?file=error.vue

How it is setup now is using `await navigateTo` which shows a 404 page that doesn't represent `error.vue`. I expect it to navigate away. When I use `return navigateTo` then I get a 500 page that says: `Vite Error`, but don't see an error anywhere.

### Describe the bug

It seems like using `navigateTo` in `error.vue` doesn't work for navigating the user away. I want to use this so that I can redirect a user when there is a 404 and a specific URL is being requested (redirects can be specified in a CMS). When I take a look at the docs this is supposed to work? https://v3.nuxtjs.org/api/utils/navigate-to#navigating-within-a-vue-component.

### Additional context

At the moment I can make it work if I put the call to `navigateTo` in the `onMounted` function, but this doesn't seem the right approach to me as I want to redirect as soon as possible on the server. Now you see a small flash of the 404 page before the redirection happens.

I also tried the following in an `error.ts` file in the `plugins` folder:
```
export default defineNuxtPlugin((nuxtApp) => {
  nuxtApp.hook("app:error", async (error) => {
    if (error.statusCode !== 404) return;
    const url = nuxtApp._route.path;

    if (url === "/en/testHook") {
      clearError();
      await navigateTo("/en/about");
    }
  });
});
```

That makes the URL change in the browser, but the app stays on the error page of `error.vue`.

aesyondu:
Hi, how about using a catch-all page? That way if no page matches it will render this page, not the `error.vue`

https://v3.nuxtjs.org/guide/directory-structure/pages/#catch-all-route

danielroe:
Currently this is not supported in the error page - only a HTML response is allowed. But I think it's worth considering this ability.

PR #8469, opened 2022-10-25
fix(nuxt): allow responding with custom headers from `error.vue`

danielroe:
### 🔗 Linked issue

follow-up to https://github.com/nuxt/framework/pull/7340, resolves https://github.com/nuxt/framework/issues/7469

### ❓ Type of change

### 📚 Description

This allows responding to an error within `error.vue` with more than just HTML and the original error code, such as modifying the code or redirecting to a different page.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/error.ts
+++ b/packages/nuxt/src/core/runtime/nitro/error.ts
@@ -1,7 +1,8 @@
 import { withQuery } from 'ufo'
 import type { NitroErrorHandler } from 'nitropack'
 import type { H3Error } from 'h3'
-import { getRequestHeaders } from 'h3'
+import { getRequestHeaders, H3Response } from 'h3'
+import { useNitroApp } from '#internal/nitro'
 import { normalizeError, isJsonRequest } from '#internal/nitro/utils'
 
 export default <NitroErrorHandler> async function errorhandler (error: H3Error, event) {
@@ -46,14 +47,15 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error, 
 
   // HTML response (via SSR)
   const isErrorPage = event.req.url?.startsWith('/__nuxt_error')
-  let html = !isErrorPage
-    ? await $fetch(withQuery('/__nuxt_error', errorObject), {
-      headers: getRequestHeaders(event) as HeadersInit
+  const res = !isErrorPage
+    ? await useNitroApp().localFetch(withQuery('/__nuxt_error', errorObject), {
+      headers: getRequestHeaders(event) as Record<string, string>,
+      redirect: 'manual'
     }).catch(() => null)
     : null
 
   // Fallback to static rendered error page
-  if (!html) {
+  if (!res) {
     const { template } = process.dev
       // @ts-ignore
       ? await import('@nuxt/ui-templates/templates/error-dev.mjs')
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -63,9 +65,13 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error,
       // TODO: Support `message` in template
       (errorObject as any).description = errorObject.message
     }
-    html = template(errorObject)
+    event.res.setHeader('Content-Type', 'text/html;charset=UTF-8')
+    return event.res.end(template(errorObject))
```
pi0, line 69: Can we please return string? H3 will add content-type too and also handles sending response string. Reducing req/res dependency

packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -63,9 +65,13 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error,
...
   }
 
-  event.res.setHeader('Content-Type', 'text/html;charset=UTF-8')
-  event.res.end(html)
+  event.respondWith(new H3Response(await res.text(), {
```
pi0, line 72: respond with API is not stable yet and might change. Let's use raw methods for now. I will have an idea for a way to simplify and standardized response format for nitro...

packages/nuxt/src/core/runtime/nitro/error.ts
```diff
@@ -46,14 +47,15 @@ export default <NitroErrorHandler> async function errorhandler (error: H3Error,
 
   // HTML response (via SSR)
   const isErrorPage = event.req.url?.startsWith('/__nuxt_error')
-  let html = !isErrorPage
-    ? await $fetch(withQuery('/__nuxt_error', errorObject), {
-      headers: getRequestHeaders(event) as HeadersInit
+  const res = !isErrorPage
+    ? await useNitroApp().localFetch(withQuery('/__nuxt_error', errorObject), {
```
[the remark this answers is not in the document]
danielroe: Yes. `$fetch.raw` returns `FetchError: 301` if a redirect is present. This is also why we use `fetch` rather than `$fetch.raw` in the test suite.

packages/nuxt/src/core/runtime/nitro/error.ts
danielroe: yes. though note I just moved this up - this wasn't a change in this pr

packages/nuxt/src/core/runtime/nitro/error.ts
danielroe: sounds good!

packages/nuxt/src/core/runtime/nitro/error.ts
danielroe: just tried - are you _sure_ an error handler supports returning value directly?

pi0: I can check. But seems strange because it is a nitro API route right?

packages/nuxt/src/core/runtime/nitro/error.ts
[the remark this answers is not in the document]
pi0: ~> https://github.com/unjs/ohmyfetch/issues/150

pi0 (approved):

thijsw:
I'd like to use this feature but I'm not sure how to use it? Does anybody might want to share an example of how I can redirect a user on the error page?

PR #8521, opened 2022-10-27
fix(nuxt): pass async-data errors through to client

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/2122

### ❓ Type of change

### 📚 Description

This PR passes error status/message through to client side. Together with https://github.com/unjs/ohmyfetch/pull/152 and https://github.com/unjs/h3/pull/213 this will enable accessing `error.statusCode` (which is useful for making decisions regarding 404s, etc.) and will pass a normalised error object through to client.

There will still be some differences between client + server errors because server errors are raw and not normalised with `createError`.

You can test with:

```vue
<template>
  <div>{{ error.message }}</div>
</template>
<script setup>
const { error } = await useFetch('https://jsonplaceholder.typicode.com/404')

onMounted(() => console.log(error.value))
</script>
```

### 📝 Checklist


packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -168,7 +169,22 @@ export function useAsyncData<
         asyncData.pending.value = false
         nuxt.payload.data[key] = asyncData.data.value
         if (asyncData.error.value) {
-          nuxt.payload._errors[key] = true
+          // Use `createError` to normalize the error
+          const _err = createError(asyncData.error.value)
+          const err = { ..._err }
```
pi0, line 174: Instead of creating an object with all props and removing some (which reshapes the object and has perf impact too) we can create an object with only expected keys.

pi0: We can also implement this as `H3Error.toObject` making support easier in the future as `createError` implementation changes, `toObject` changes consistantly

wobsoriano:
Finally 😅

pi0 (approved):

szulcus:
I'm very happy about this! ❤️

Issue #8499, opened 2022-10-26
useAsyncData with comma breaks


Tragio:
### Environment

- Operating System: `Darwin`
- Node Version:     `v14.19.1`
- Nuxt Version:     `3.0.0-rc.12`
- Nitro Version:    `0.6.0`
- Package Manager:  `yarn@1.22.19`
- Builder:          `vite`
- User Config:      `telemetry`, `ssr`, `app`, `modules`, `formkit`, `purgeCSS`, `pwa`, `css`, `vite`
- Runtime Modules:  `@pinia/nuxt@0.4.3`, `@kevinmarrec/nuxt-pwa@0.9.1`, `nuxt-font-metrics@0.2.1`, `nuxt-purgecss@2.0.0`, `@nuxtjs/html-validator@1.1.0`, `@vueuse/nuxt@9.3.1`, `@formkit/nuxt@1.0.0-beta.11`
- Build Modules:    `-`

### Reproduction

Not working:
```js
const data = await useAsyncData(() =>
  $fetch('https://api.nuxtjs.dev/mountains'),
)
```

Working:
```js
const data = await useAsyncData(() =>
  $fetch('https://api.nuxtjs.dev/mountains')
)
```

### Describe the bug

If I have a `,` at the end of the `$fetch` it does not work, if I remove, it works.

If I have the `,` I have this on the console `Uncaught SyntaxError: Unexpected token ',' (at edit.vue?macro=true:14:1)`

![CleanShot_2022-10-26_at_14 02 30](https://user-images.githubusercontent.com/5109593/198033645-56693139-e1e4-4324-aaa5-b6b40d6359cf.jpg)

antlionguard:
that's not an issue. you should not use extra comma if you don't pass extra params into function.

manniL:
> that's not an issue. you should not use extra comma if you don't pass extra params into function.

Actually it is because trailing commas, even for function parameters, are valid. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#function_calls

antlionguard:
> > that's not an issue. you should not use extra comma if you don't pass extra params into function.
> 
> Actually it is because trailing commas, even for function parameters, are valid. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#function_calls

[...]

PR #8529, opened 2022-10-27
fix(nuxt): check before appending comma in composable keys

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/8499

### ❓ Type of change

### 📚 Description

We were previously not confirming whether there was a trailing comma before appending our own.

### 📝 Checklist

### DIFF
--- a/packages/vite/src/plugins/composable-keys.ts
+++ b/packages/vite/src/plugins/composable-keys.ts
@@ -59,9 +59,11 @@ export const composableKeysPlugin = createUnplugin((options: ComposableKeysOptio
               break
           }
 
+          const endsWithComma = code.slice(codeIndex + (node as any).start, codeIndex + (node as any).end - 1).trim().endsWith(',')
+
           s.appendLeft(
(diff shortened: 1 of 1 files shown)


packages/vite/src/plugins/composable-keys.ts
```diff
@@ -59,9 +59,11 @@ export const composableKeysPlugin = createUnplugin((options: ComposableKeysOptio
               break
           }
 
+          const endsWithComma = code.slice(codeIndex + (node as any).start, codeIndex + (node as any).end - 1).trim().endsWith(',')
```
pi0, line 62: Perf: this is a heavy operation to run on each code block, copying the whole string, trimming (that also copies) to only check the last character is `,`. We could replace it with a look-behind loop.
```suggestion
          // TODO: Optimize me
          const endsWithComma = code.slice(codeIndex + (node as any).start, codeIndex + (node as any).end - 1).trim().endsWith(',')
```

pi0 (approved):

PR #8573, opened 2022-10-30
feat(nuxt): add warning `useState` init duplication

kottyan:
### ❓ Type of change

### 📚 Description

When duplicate same useState initialization, useState is silent.(no warning)
Wouldn't it be more developer-friendly to have warnings?

[example case]
```js
const counter = useState('counter', () => Math.round(Math.random() * 1000))
const sameCounter = useState('counter', () => Math.round(Math.random() * 2000)) // This 'init' is ignored. But nothing is reported.
```

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/state.ts
+++ b/packages/nuxt/src/app/composables/state.ts
@@ -25,7 +25,7 @@ export function useState <T> (...args: any): Ref<T> {
   const nuxt = useNuxtApp()
   const state = toRef(nuxt.payload.state.currentValue, key)
   const initState = nuxt.payload.state.init[key]
-  if (initState !== undefined && init !== undefined) {
+  if (process.dev && initState !== undefined && init !== undefined) {
     console.warn(`useState ${key} has already initialized.`)
   }
   if (initState === undefined && init !== undefined) {
--- a/packages/nuxt/src/app/composables/state.ts
+++ b/packages/nuxt/src/app/composables/state.ts
@@ -23,19 +23,15 @@ export function useState <T> (...args: any): Ref<T> {
   const key = '$s' + _key
 
   const nuxt = useNuxtApp()
-  const state = toRef(nuxt.payload.state.currentValue, key)
-  const initState = nuxt.payload.state.init[key]
-  if (process.dev && initState !== undefined && init !== undefined) {
-    console.warn(`useState ${key} has already been initialized.`)
-  }
-  if (initState === undefined && init !== undefined) {
-    nuxt.payload.state.init[key] = init
+  const state = toRef(nuxt.payload.state, key)
+  if (state.value !== undefined && init !== undefined) {
+    console.warn(`useState ${key} has already initialized.`)
   }
   if (state.value === undefined && init) {
     const initialValue = init()
     if (isRef(initialValue)) {
       // vue will unwrap the ref for us
-      nuxt.payload.state.currentValue[key] = initialValue
+      nuxt.payload.state[key] = initialValue
       return initialValue as Ref<T>
     }
     state.value = initialValue
--- a/packages/nuxt/src/app/composables/state.ts
+++ b/packages/nuxt/src/app/composables/state.ts
@@ -26,7 +26,7 @@ export function useState <T> (...args: any): Ref<T> {
   const state = toRef(nuxt.payload.state.currentValue, key)
   const initState = nuxt.payload.state.init[key]
   if (process.dev && initState !== undefined && init !== undefined) {
-    console.warn(`useState ${key} has already initialized.`)
+    console.warn(`useState ${key} has already been initialized.`)
   }
   if (initState === undefined && init !== undefined) {
     nuxt.payload.state.init[key] = init
--- a/packages/nuxt/src/app/composables/state.ts
+++ b/packages/nuxt/src/app/composables/state.ts
@@ -29,7 +29,7 @@ export function useState <T> (...args: any): Ref<T> {
     const defineInit = map.get(key)
 
     if (defineInit && init && defineInit !== init.toString()) {
-      console.warn(`useState ${key} has already initialized.`)
+      console.warn(`useState ${key} has already been initialized.`)
     }
     if (defineInit === undefined && init) {
       map.set(key, init.toString())
--- a/packages/nuxt/src/app/composables/state.ts
+++ b/packages/nuxt/src/app/composables/state.ts
@@ -24,6 +24,9 @@ export function useState <T> (...args: any): Ref<T> {
 
   const nuxt = useNuxtApp()
   const state = toRef(nuxt.payload.state, key)
+  if (state.value !== undefined && init !== undefined) {
+    console.warn(`useState ${key} has already initialized.`)
+  }
   if (state.value === undefined && init) {
     const initialValue = init()
     if (isRef(initialValue)) {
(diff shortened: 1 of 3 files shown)


danielroe (dismissed):
I think that there's a use case for calling `useState` more than once. For example:

```js
export const useUser = () => useState('user', () => ({}))
```

Now, this points to a single state across your app and will only run the init function once. We shouldn't warn the 2nd, 3rd, 4th times this function is called.

If we wanted to warn, then I think it would make sense only to warn when a different init function is provided (so we'd need to map key to init function), and only in development (so it can be tree-shaken out in build).

packages/nuxt/src/app/composables/state.ts
```diff
@@ -24,6 +24,9 @@ export function useState <T> (...args: any): Ref<T> {
 
   const nuxt = useNuxtApp()
   const state = toRef(nuxt.payload.state, key)
+  if (state.value !== undefined && init !== undefined) {
+    console.warn(`useState ${key} has already initialized.`)
```
Atinux, line 28:
```suggestion
    console.warn(`useState ${key} has already been initialized.`)
```

kottyan:
> I think that there's a use case for calling `useState` more than once. For example:
> 
> ```js
> export const useUser = () => useState('user', () => ({}))
> ```
> 
> Now, this points to a single state across your app and will only run the init function once. We shouldn't warn the 2nd, 3rd, 4th times this function is called.
> 
> If we wanted to warn, then I think it would make sense only to warn when a different init function is provided (so we'd need to map key to init function), and only in development (so it can be tree-shaken out in build).

It’s just as you said!!
I tried this.
Is it as expected?

- map key to init function
https://github.com/nuxt/framework/pull/8573/commits/82a2f206f4c4589daf13393078f97799ec07a179
- only in development
https://github.com/nuxt/framework/pull/8573/commits/752b7da9e58a92722d05f5bd7c3c44dd2744e60a

packages/nuxt/src/app/composables/state.ts
kottyan: Thank you!! I fixed it.
https://github.com/nuxt/framework/pull/8573/commits/752b7da9e58a92722d05f5bd7c3c44dd2744e60a

danielroe:
I wouldn't set this within `nuxtApp`, as this will persist into to the built version of the app.

Rather, you could use a map directly within `composables/state` to associate init functions (stringified) with the key. You would only reference this map within a check for `process.dev` so it could be fully tree-shaken out in build mode.

kottyan:
@danielroe 

> Rather, you could use a map directly within composables/state to associate init functions (stringified) with the key. You would only reference this map within a check for process.dev so it could be fully tree-shaken out in build mode.

Thank you.
I fixed it.
https://github.com/nuxt/framework/pull/8573/commits/24eeffd31938dd9250e9b50e5152c8143e0ff1fd

packages/nuxt/src/app/composables/state.ts
```diff
@@ -22,6 +22,18 @@ export function useState <T> (...args: any): Ref<T> {
   }
   const key = '$s' + _key
 
+  if (process.dev) {
+    const map = new Map<string, string>()
```
danielroe, line 26: this needs to be defined outside of `useState` if we want it to persist between calls. Approximately at L4 should be fine.

packages/nuxt/src/app/composables/state.ts
```diff
@@ -22,6 +22,18 @@ export function useState <T> (...args: any): Ref<T> {
...
+    const defineInit = map.get(key)
+
+    if (defineInit && init && defineInit !== init.toString()) {
+      console.warn(`useState ${key} has already been initialized.`)
```
danielroe, line 30:
```suggestion
      console.warn(`useState ${key} has already been initialized with a different initializer.`, defineInit, init.to.toString())
```

packages/nuxt/src/app/composables/state.ts
```diff
@@ -22,6 +22,18 @@ export function useState <T> (...args: any): Ref<T> {
   }
   const key = '$s' + _key
 
+  if (process.dev) {
+    const map = new Map<string, string>()
+    const defineInit = map.get(key)
```
danielroe, line 27: perf: you can likely avoid the multiple calls to `.toString()` below

```suggestion
    const defineInit = map.get(key)?.toString()
```

danielroe, line 27: minor perf point, but we can likely avoid the multiple calls to `init.toString()` below by doing it just once

```suggestion
    const defineInit = map.get(key)
    const newInit = init?.toString()
```

packages/nuxt/src/app/composables/state.ts
kottyan: oh..
It was my lack of understanding.
I fixed it. 
https://github.com/nuxt/framework/pull/8573/commits/44d8b782c54dd9d40f373be07f35ab65d2af7b40

packages/nuxt/src/app/composables/state.ts
kottyan: Thanks. I fixed it.
https://github.com/nuxt/framework/pull/8573/commits/b55c34b7d55a88bf3e298ac8391088c56202f1ac

packages/nuxt/src/app/composables/state.ts
kottyan: Thanks. I fixed it.
https://github.com/nuxt/framework/pull/8573/commits/b55c34b7d55a88bf3e298ac8391088c56202f1ac

packages/nuxt/src/app/composables/state.ts
```diff
@@ -22,6 +22,18 @@ export function useState <T> (...args: any): Ref<T> {
   }
   const key = '$s' + _key
 
+  if (process.dev) {
+    const defineInit = map.get(key)
+    const newInit = init?.toString()
```
pi0, line 27: I think we can simply check the references to be the same. Init function in two files might be exactly the same but has an external variable dependency that differs. Also makes perf slightly better

danielroe: Example:

```js
useState('thing', () => 'value')
```

Every time `useState` is called it receives a **new** factory function.

packages/nuxt/src/app/composables/state.ts
```diff
@@ -22,6 +22,18 @@ export function useState <T> (...args: any): Ref<T> {
...
+
+    if (defineInit && newInit && defineInit !== newInit) {
+      console.warn(`useState ${key} has already been initialized with a different initializer.`, defineInit, newInit)
```
danielroe, line 30:
```suggestion
      console.warn(`useState has been called for \`${key}\` with multiple initializers. This may be safe, or it may indicate an issue. Initializers:`, defineInit, newInit)
```

pi0:
Since `useState` (and it's default value) is isolated to each rendering context, there might be legitimate use cases that developers intentionally use different initializers and expect different/initial state based on first usage. I would a temporary redirect url as an example but could be more. Adding this forces those use-cases to avoid using the feature. 

Also with the stringification workaround, we still miss if the function depends on different context variable boundings (`const x = () => {}; const s = useState('key', () => { x() })` can be different in two pages and we cannot check)

Are there any real world stories/issues arising from misuse of `useState()`? If not i would suggest to improve documentation instead. An ESLint plugin can also help to globally analyze `useSate('key')` patterns and avoid duplicates via a rule...

danielroe:
First, my apologies for keeping you waiting on this PR. It's a nice implementation. But I agree with @pi0 here. I haven't seen many, if any, issues related to this possible issue.

Let's focus on improving docs. If you still think it's a good idea, would you open an issue on https://github.com/nuxt/nuxt/ so we can discuss first?

Issue #8099, opened 2022-10-10
globalName - change __nuxt id as it was in NUXT2


53rg0:
### Describe the feature

Please, return globalName param, its very important for customers and developers for a lot of reasons!

Thank you!

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

manniL:
> Its very important for customers and developers for a lot of reasons!

Can you provide some then? I always felt it more as a burden than something useful...

53rg0:
> > Its very important for customers and developers for a lot of reasons!
> 
> Can you provide some then? I, personally, always felt it more as a burden than something useful...

First of all - security reason. 
It is very important to protect the site from scripts that automatically scan technologies. Nuxt3 uses the Nitro server part and it is enough to find the __nuxt id in the code to already know what the site is running on. If any bugs are found in the future, scripts that scan websites will quickly find vulnerable ones.

Nuxt3 is great for me! I will definitely talk about this technology in my portfolio and will do my best to make as many people as possible learn about it.
Another thing is customers, especially very large ones, who very often have a requirement to remove any mention of other brands and their team includes those who check the project for such "Easter eggs". Already on the first demo with Nuxt 3, I received such a remark and knew that it could be removed, but later I realized that this was no longer possible.

Analysis of competitors. Often before development, an analysis of competitor solutions is carried out and we would not like to make it obvious to other developers what we use during the analysis. We don't mind talking about it in the portfolio, but we don't mind making it obvious.

Having this feature is very important, I don’t want to switch back to Nuxt 2 having a wonderful Nuxt 3 because of such a trifle, but still the last word is not with me, but with the clients.

manniL:
> First of all - security reason. It is very important to protect the site from scripts that automatically scan technologies. Nuxt3 uses the Nitro server part and it is enough to find the __nuxt id in the code to already know what the site is running on. If any bugs are found in the future, scripts that scan websites will quickly find vulnerable ones.

This sounds like security through obscurity. You will always be able to find out if a site was built with Nuxt, simply by scanning the `.js`/`.mjs` files and match the code vs. what Nuxt produces. Especially for built-in names.

manniL:
[...]

PR #8604, opened 2022-11-01
feat(nuxt): define a custom root element inside `app.config`

### LINKED ISSUE #8603: Customize app root element
### Describe the feature

Currently, the Nuxt app is mounted inside a `#__nuxt` container and this cannot be changed. Allowing for a custom root element would allow for:
- customizing the rendered HTML page using a Nitro plugin and mount the app inside that custom page;
- creating a no-SSR app and incorporating it into an existing HTML page, but with all the benefits that Nuxt brings along.

### Additional information

### Final checks

yassilah:
### ❓ Type of change

### 📚 Description

Resolves #8603 and #8099.
There's probably a better place to customize the root element selector than `app.config` but it allowed for a very simple implementation and for retrieving that root container from wherever we want inside the app using `useAppConfig()`, which might be useful in some cases 

### 📝 Checklist


HigherOrderLogic:
It would be nice if you also add an option to change the `window.__NUXT__` script to something like `window.__FOO__`.

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: 5.4
+lockfileVersion: 5.3
```
Aareksio, line 1: Shouldn't be a part of this PR

manniL:
> It would be nice if you also add an option to change the `window.__NUXT__` script to something like `window.__FOO__`.

I don't think this will be feasible to maintain, see https://github.com/nuxt/framework/issues/8099

mohsin (approved):
lgtm

pi0:
Thanks for PR @yassilah ❤️ I've made a slightly different implementation that we are also able to change renderer behavior based on configuration: #8883

yassilah:
that's much better, this way i don't even need a nitro plugin to render my custom div!

PR #8635, opened 2022-11-02
feat(kit): add `addServerPlugin` utility

danielroe:
### ❓ Type of change

### 📚 Description

This adds a helper utility for adding a nitro plugin.

### 📝 Checklist

### DIFF
--- a/packages/kit/src/nitro.ts
+++ b/packages/kit/src/nitro.ts
@@ -1,4 +1,5 @@
 import type { NitroEventHandler, NitroDevEventHandler, Nitro } from 'nitropack'
+import { normalize } from 'pathe'
 import { useNuxt } from './context'
 
 /**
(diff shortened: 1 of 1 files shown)


packages/kit/src/nitro.ts
```diff
@@ -30,6 +31,15 @@ export function addDevServerHandler (handler: NitroDevEventHandler) {
   useNuxt().options.devServerHandlers.push(handler)
 }
 
+/**
+ * Adds a nitro plugin
```
johannschopplich, line 35:
```diff
-* Adds a nitro plugin
+* Adds a Nitro plugin
```

Name of `Nitro` engine should be uppercased, right?

danielroe:
```suggestion
 * Adds a Nitro plugin
```

Atinux (approved):

packages/kit/src/nitro.ts
```diff
@@ -30,6 +31,15 @@ export function addDevServerHandler (handler: NitroDevEventHandler) {
   useNuxt().options.devServerHandlers.push(handler)
 }
 
+/**
+ * Adds a Nitro plugin
+ */
+export function addServerPlugin (plugin: string) {
+  const nuxt = useNuxt()
+  nuxt.options.nitro.plugins = nuxt.options.nitro.plugins || []
```
pi0, line 39: We might have to check lifecycle as this composable is not effective after nitro is being initialized.

Issue #5483, opened 2022-06-15
nuxt generate FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory 


safejace:
### Environment

Nuxt CLI v3.0.0-rc.4
Nuxt project info: 

------------------------------
- Operating System: `Darwin`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0-rc.3`
- Package Manager:  `yarn@1.22.17`
- Builder:          `vite`
- User Config:      `runtimeConfig`, `build`, `css`, `buildModules`, `modules`, `sanity`, `vite`
- Runtime Modules:  `@nuxtjs/sanity@1.2.2`, `@vueuse/nuxt@8.6.0`
- Build Modules:    `@pinia/nuxt@0.1.9`, `@safesoftware/safebase-vue/nuxt@0.4.0`
------------------------------

### Reproduction

yarn nuxt build

### Describe the bug

when trying to generate my site i get javascript heap allocation errors.

it hangs on `ℹ Initializing prerenderer` for a long time

my devops colleague said he had it running for the hour, and it finally failed (see logs). 

### Logs

```shell
yarn run v1.22.18
$ nuxt build
Nuxt CLI v3.0.0-rc.4                                                                                                                                       17:15:59
ℹ Running with Sanity project XXXXXX (production).                                                                                             @nuxtjs/sanity 17:16:14
ℹ Vite client warmed up in 77349ms                                                                                                                         17:17:46
 
 WARN                                                                                                                                                      17:18:46
(!) Some chunks are larger than 500 KiB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/guide/en/#outputmanualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
 
ℹ Client built in 137160ms                                                                                                                                 17:18:46
ℹ Building server...                                                                                                                                       17:18:46
 
 WARN  @fortawesome/fontawesome-pro doesn't appear to be written in CJS, but also doesn't appear to be a valid ES module (i.e. it doesn't have "type": "module" or an .mjs extension for the entry point). Please contact the package author to fix.
 
✔ Server built in 65079ms                                                                                                                                  17:19:51
✔ Generated public .output/public                                                                                                                    nitro 17:19:54
start Building server...                                                                                                                             nitro 17:19:54
 
[...]

PR #8639, opened 2022-11-02
fix(nuxt): don't force prerender `/` if user doesn't have that route

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/5483#issuecomment-1157960000

### ❓ Type of change

### 📚 Description

With pages module, we add routes to the default routes values. But if the user doesn't have a `~/pages/index.vue` file we can safely skip processing the default `/` route for prerender.routes.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/module.ts
+++ b/packages/nuxt/src/pages/module.ts
@@ -102,6 +102,8 @@ export default defineNuxtModule({
       })
       nuxt.hook('nitro:build:before', (nitro) => {
         for (const route of nitro.options.prerender.routes || []) {
+          // Skip default route value as we only generate it if it is already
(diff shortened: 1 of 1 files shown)


Atinux:
What about if the user has an app.vue but no pages? I think we should keep rendering the home page in that case

danielroe:
Yes, agreed. This change only affects the `pages` ad-hoc module. (So it only prevents force-rendering a 404.)

Atinux (approved):

pi0:
Would it still generate `index.html` for spa mode? (it is main fallback file)

pi0 (approved):

PR #8666, opened 2022-11-03
fix(nuxt): do not inline global styles in HTML response

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/issues/7619

### ❓ Type of change

### 📚 Description

This reverts https://github.com/nuxt/framework/pull/7386. Rather than attempt to lazy load the entry CSS, it's safer to rely on browser fetching it in the previous (blocking) manner. We still experience the benefit of inlining SFC styles and we can reconsider entry CSS inlining in future.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -289,18 +289,14 @@ function renderHTMLDocument (html: NuxtRenderHTMLContext) {
 }
 
 async function renderInlineStyles (usedModules: Set<string> | string[]) {
-  const { entryCSS } = await getClientManifest()
   const styleMap = await getSSRStyles()
   const inlinedStyles = new Set<string>()
-  for (const mod of ['entry', ...usedModules]) {
+  for (const mod of usedModules) {
     if (mod in styleMap) {
       for (const style of await styleMap[mod]()) {
         inlinedStyles.add(`<style>${style}</style>`)
       }
     }
-  }
-  for (const css of entryCSS?.css || []) {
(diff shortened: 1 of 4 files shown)


pi0 (approved):

daguitosama:
Hi there guys, the experimental  `inlineSSRStyles` is still present in `rc-13` `nuxt-config-schema`, even says it's `true` by default on hover, but it does not do the inline at all. This is way confusing. 

I think the inline is a really nice feature, and can be leaved here on the toolset as an opt-in feature.
I mean, if you have pretty much < 15 kb css in every page with Tailwindcss, why to block the render of the page awaiting the stylesheet it's fetch, when all that tiny css can be "easily" inlined in the Document Head.

danielroe:
@daguitosama At the moment, we do not inline _global_ CSS but CSS associated with individual Vue SFCs.

stuible:
> @daguitosama At the moment, we do not inline _global_ CSS but CSS associated with individual Vue SFCs.

Is this expected to change?  Some of the most critical CSS to a site (resets, grid systems, etc) is global CSS.

danielroe:
@stuible The global CSS is still loaded synchronously, so there will be no flash when your page loads.

Plinpod:
> Hi there guys, the experimental `inlineSSRStyles` is still present in `rc-13` `nuxt-config-schema`, even says it's `true` by default on hover, but it does not do the inline at all. This is way confusing.
> 
> I think the inline is a really nice feature, and can be leaved here on the toolset as an opt-in feature. I mean, if you have pretty much < 15 kb css in every page with Tailwindcss, why to block the render of the page awaiting the stylesheet it's fetch, when all that tiny css can be "easily" inlined in the Document Head.

I agree in some cases, it still makes sense to inline the global CSS. Is there a way to force this to happen in the current configuration?

PR #8674, opened 2022-11-03
fix(vite): enable fs strict mode

pi0:
### ❓ Type of change

### 📚 Description

Enable vite fs restriction by default(https://vitejs.dev/config/server-options.html#server-fs-strict). 

This ensures avoiding access to fs using `/_nuxt/@fs/` url.

This PR might introduce regressions with modules trying to access outside of `rootDir`, `srcDir`, `buildDir`, `workspaceDir` or either of `modulesDir` (modules are automatically added). In this case, can be manually added using `nuxt.options.vite.server.fs.allow[]`

```ts
  vite: {
    server: {
      fs: {
        allow: [
          '/xyz'
        ]
      }
    }
  }
})
```

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/vite.ts
+++ b/packages/schema/src/config/vite.ts
@@ -64,7 +64,6 @@ export default defineUntypedSchema({
     },
     server: {
       fs: {
-        strict: false,
         allow: {
           $resolve: async (val, get) => [
             await get('buildDir'),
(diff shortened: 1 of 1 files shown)


manniL:
This only has impact in dev mode, right?

OhB00:
> This only has impact in dev mode, right?

Correct, you can read the full report on huntr when it is published.

Issue #7965, opened 2022-10-03
Warn user when using a separate `nitro.config.ts`


manniL:
### Describe the feature

When using Nuxt, the nitro config should live inside the `nuxt.config.{js,ts}` instead of a separate `nitro.config.{js,ts}`. I'd suggest checking the existance of it during start and warn the user with an appropriate message that suggests to move settings into the Nuxt config.

### Additional information

- [X] Would you be willing to help implement this feature?
- [X] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

pi0:
As discussed in #8683 we can improve Nitro docs. 

It wouldn't hurt to also add a `Nitro Configuration` section in https://v3.nuxtjs.org/guide/directory-structure/server IMO to clarify this.

PR #8683, opened 2022-11-04
feat(nuxi): warn on standalone nitro config in dev

manniL:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/7965

### ❓ Type of change

### 📚 Description

When running `nuxi dev` and using a standalone `nitro.config.ts` (or `.js`), warn the user about it.

### 📝 Checklist


danielroe (approved):
Good idea.

darthf1:
Would this be the same for `vitest.config.ts` vs the `vite.test` config key in `nuxt.config.ts`?

danielroe:
Good question. What do you think @antfu?

antfu:
I am not very sure about it. It feels a little bit over-babysitting to me, as:

1. Extra performance cost, especially if later we need to support `{webpack,vite,vitest}.config.{js,ts,mjs,cjs,mts,cts}` etc.
2. In some cases users might configure those configs for other usages, like using Nuxt and Nitro in two different build steps
3. In some cases users might use standalone configs for better IDE support, and import dynamically, like:

```ts
import nitroConfig from './nitro.config'
import viteConfig from './vite.config'

  nitro: nitroConfig,
  vite: viteConfig
})
```

Then this warning could be false-alarmed.

manniL:
@antfu 

Very valid points! On the other hand, I see quite some users who use an additional `nitro.config` to their `nuxt.config` not knowing that they should not. 


Should we improve documentation then instead regarding these?

pi0:
Thanks for the PR @manniL. As @antfu mentioned, it might get tricky for cases that `nitro.config` is also present for another project or user's preferred convention to specify it (while not recommending this some might find it useful). 

In order to be consistent with other sub-packages such as vite that we ignore their config but don't warn, current behavior seems fine.

pi0:
> Should we improve documentation then instead regarding these?

Yes specially on nitro docs we can make it better for Nuxt users.

manniL:
Should we close #7965 then?

PR #8700, opened 2022-11-04
fix(nuxt): check if global transitions are activated for scroll behavior

### LINKED ISSUE #8699: Default scroll behavior without page transition is broken in rc.13
### Environment

- Operating System: `Darwin`
- Nuxt Version:     `3.0.0-rc.13`
- Nitro Version:    `0.6.1`
- Builder:          `vite`
- User Config:      `alias`, `app`, `build`, `modules`, `progress`, `plausible`, `runtimeConfig`, `vite`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.1.3`, `nuxt-progress@0.1.6`, `vue-plausible@1.3.2`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-ruxdek?file=pages/index.vue

### Describe the bug

Reproduction:
* Have two pages, each with more than 100vh width.
* Scroll down on page 1
* Navigate to page 2

Expected behavior:
* Page 2 should scroll to top

Actual behavior:
* Page 2 stays at the scroll position of page 1

The reason is that `router.options.ts` falsely thinks that an `undefined` page transition is `true`, meaning it HAS a transition.
https://github.com/nuxt/framework/blob/bde96947333501c22a5e981515bd27c24eb8b88f/packages/nuxt/src/pages/runtime/router.options.ts#L34

Whereas `page.mjs` correctly interprets an `undefined` transition as `false`.
https://github.com/nuxt/framework/blob/bde96947333501c22a5e981515bd27c24eb8b88f/packages/nuxt/src/pages/runtime/page.ts#L46

I will submit a PR to fix this issue.

### Additional context

_No response_

### Logs

_No response_

Atinux:
Resolves a regression from #8463

And resolves #8699

Learn more in the comments from #8327

### ❓ Type of change

### 📚 Description

Regression since we need to detect if there are global transitions activated in order to listen to `page:transition:finish` hook. Otherwise it's never called and the scrollBehavior is inactive.

Also, since the value of the next and previous page can be `undefined`, using `!== false` was also a wrong usage.

### DIFF
--- a/packages/nuxt/src/pages/runtime/router.options.ts
+++ b/packages/nuxt/src/pages/runtime/router.options.ts
@@ -2,6 +2,7 @@ import type { RouterConfig } from '@nuxt/schema'
 import type { RouterScrollBehavior, RouteLocationNormalized } from 'vue-router'
 import { nextTick } from 'vue'
 import { useNuxtApp } from '#app'
+import { appPageTransition as defaultPageTransition } from '#build/nuxt.config.mjs'
 
 type ScrollPosition = Awaited<ReturnType<RouterScrollBehavior>>
 
@@ -31,7 +32,7 @@ export default <RouterConfig> {
     }
 
     // Wait for `page:transition:finish` or `page:finish` depending on if transitions are enabled or not
-    const hasTransition = to.meta.pageTransition !== false && from.meta.pageTransition !== false
+    const hasTransition = defaultPageTransition || (to.meta.pageTransition && from.meta.pageTransition)
     const hookToWait = hasTransition ? 'page:transition:finish' : 'page:finish'
     return new Promise((resolve) => {
       nuxtApp.hooks.hookOnce(hookToWait, async () => {
--- a/packages/nuxt/src/pages/runtime/router.options.ts
+++ b/packages/nuxt/src/pages/runtime/router.options.ts
@@ -33,8 +33,8 @@ export default <RouterConfig> {
     }
 
     // Wait for `page:transition:finish` or `page:finish` depending on if transitions are enabled or not
-    const hasTransition = defaultPageTransition || (to.meta.pageTransition && from.meta.pageTransition)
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/pages/runtime/router.options.ts
```diff
@@ -31,7 +32,7 @@ export default <RouterConfig> {
     }
 
     // Wait for `page:transition:finish` or `page:finish` depending on if transitions are enabled or not
-    const hasTransition = to.meta.pageTransition !== false && from.meta.pageTransition !== false
+    const hasTransition = defaultPageTransition || (to.meta.pageTransition && from.meta.pageTransition)
```
pi0, line 35: This logic always waits for `page:transition:finish` hook. If the transition is disabled on page level but enabled globally, we should still wait for `page:finish`. We should use something like `... ?? defaultPageTransition ` when neither values are explicitly set.

Atinux: Not always, I tested and logged to see the behavior actually.

> If the transition is disabled on page level but enabled globally, we should still wait for page:finish

The best would be to share the same util which is used in
https://github.com/nuxt/framework/blob/fix/router-options-regression/packages/nuxt/src/pages/runtime/page.ts#L46 to avoid any mistake so we know what hook to listen.

pi0: (I meant always when `pageTransition: true` is set). Sharing utility would have been nice idea but I'm not sure if `NuxtPage` prop can be accessed here at least in same way.

pi0:
```suggestion
const hasTransition = (loc: RouteLocationNormalized) => !!(loc.meta.pageTransition ?? defaultPageTransition)
    const hookToWait = (hasTransition(from) && hasTransition(to)) ? 'page:transition:finish' : 'page:finish'
```

pi0 (approved):

PR #8704, opened 2022-11-04
refactor(nuxt)!: fix typo for `NuxtRenderHTMLContext.bodyPrepend`

MorevM:
### 🔗 Linked issue

I guess there is no need for issue since it's obvious.

### ❓ Type of change

### 📚 Description

Just worked on it and was surprised :)
Small mistake, but actually it is a breaking change since existed modules that uses `render:html` hook _may_ use the key with a typo. Good thing it's still in RC stage :)

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/core/runtime/nitro/renderer.ts
+++ b/packages/nuxt/src/core/runtime/nitro/renderer.ts
@@ -21,7 +21,7 @@ export interface NuxtRenderHTMLContext {
   htmlAttrs: string[]
   head: string[]
   bodyAttrs: string[]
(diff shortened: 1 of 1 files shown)


manniL:
Good catch!
Maybe we could use a compat layer + warning for now though, so people have time to upgrade until the version after (or until stable)?

MorevM:
@manniL I don't think a compat layer makes sense, every next RC might include breaking changes, and users of edge channel are ready for that as well :)

danielroe (approved):

pi0:
Thanks!

PR #8713, opened 2022-11-04
fix(nuxt): use parser to treeshake `<ClientOnly>`

### LINKED ISSUE #8397: Webpack builder with nuxt generate assigns incorrect scoped style attribute when using client-only
### Environment

Nuxi 3.0.0-rc.12
RootDir: /home/projects/github-xq94aj-dtefps
Nuxt project info: 

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-rc.12`
- Nitro Version:    `0.6.0`
- Builder:          `webpack`
- User Config:      `builder`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

 - Open [the reproduction](https://stackblitz.com/edit/github-xq94aj-dtefps?file=Component.vue)
 - Run `npm run generate`
 - Run `cd .output/public`
 - Run `npx serve`
 - The line on the bottom has no styles. When running `npm run dev` it does.

### Describe the bug

When:

 - Using the webpack builder
 - And using a <client-only> component somewhere in a component
 - And using scoped styles in that component
 - And using `npm run generate`

Then:
 - Only the elements inside of the <client-only> component have the scoped styles applied
 - It appears as if the data attributes that make scoped styles work are assigned incorrectly to all the elements that are not inside of the client-only component

I would expect:
 - The behaviour to be the same between `npm run dev` and `npm run generate`
 - Scoped styles to be applied for all elements in a component, no matter if <client-only> is used or not

### Additional context

It works as expected when using the default vite builder.

### Logs

_No response_

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/nuxt.js/issues/15227
resolve https://github.com/nuxt/nuxt.js/issues/15396
closes https://github.com/nuxt/framework/pull/8789 (by including it)

### ❓ Type of change

### 📚 Description

hi :wave: , i think this should fix nuxt/nuxt.js#15227.  
Instead of treeshaking the default slot before SFC compilation.
This PR:
 - move the module before the component loader plugin (to avoid having extra imports to remove)
 - enforce `post`, this way the scoped attributes will be already added when modifying the code
 - remove the default slot of `ClientOnly` components
 - remove components direct import (or Identifier)  based on if they are called only in the removed code.
 - it also remove the components from the setup return statement (mostly in dev)

Test:
 - This also adds a test to ensure that style are corretly applied once on client.

let me know what you think of this 🙂 

todo: 
- [x] change the regex to test `_component_` and `unref()` (resolve and direct import)

### 📝 Checklist


packages/nuxt/src/components/tree-shake.ts
```diff
@@ -1,66 +1,117 @@
...
-        const tags = clientOnlyComponents
-          .map(component => `<(${component})[^>]*>[\\s\\S]*?<\\/(${component})>`)
-
-        regexpMap.set(components, [new RegExp(`(${tags.join('|')})`, 'g'), clientOnlyComponents])
-      }
-
-      const [COMPONENTS_RE, clientOnlyComponents] = regexpMap.get(components)!
-      if (!COMPONENTS_RE.test(code)) { return }
+    transform (code, id) {
+      const CLIENTONLY_RE = /[c|C]lient[_|-]?[O|o]nly/
```
danielroe, line 23:
```suggestion
      const CLIENTONLY_RE = /(ClientOnly|client-only)/
```

we can also definite this outside the body of the transform function for a slight performance improvement.

packages/nuxt/src/components/tree-shake.ts
```diff
@@ -75,3 +126,33 @@ export const TreeShakeTemplatePlugin = createUnplugin((options: TreeShakeTemplat
...
+ * @param code code to test
+ * @param name component name
+ */
+function isRenderedInCode (code: string, name: string) {
+  return new RegExp(`ssrRenderComponent\\(${escapeRegExp(name)}`).test(code)
+}
+
+function escapeRegExp (string: string) {
+  return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+}
```
danielroe, line 158: we use `escape-string-regexp` most other places

packages/nuxt/src/components/tree-shake.ts
```diff
@@ -75,3 +126,33 @@ export const TreeShakeTemplatePlugin = createUnplugin((options: TreeShakeTemplat
...
+ * @param {string} importName - name of the import
+ */
+function findImportDeclaration (declarations: AcornNode<ImportDeclaration>[], importName: string): AcornNode<ImportDeclaration> | undefined {
+  const declaration = declarations.find((d) => {
+    const specifier = d.specifiers.find(s => s.local.name === importName)
+    if (specifier) { return true }
+    return false
+  })
+
+  if (declaration) { return declaration }
```
danielroe, line 143:
```suggestion
  return declaration
```

packages/nuxt/src/components/tree-shake.ts
```diff
@@ -1,66 +1,117 @@
...
+    enforce: 'post',
     transformInclude (id) {
       const { pathname } = parseURL(decodeURIComponent(pathToFileURL(id).href))
       return pathname.endsWith('.vue')
     },
-    async transform (code, id) {
-      const template = code.match(/<template>([\s\S]*)<\/template>/)
-      if (!template) { return }
-
-      const components = options.getComponents()
```
danielroe, line 28: we also want to tree-shake out the non-fallback slots of any client-only components, not just `<ClientOnly>`.

danielroe (commented):
Thanks for working on this!

packages/nuxt/src/components/tree-shake.ts
huang-julien: About this, [`createClientOnly()`](https://github.com/nuxt/framework/blob/fe433a4222dd9248ec04bd17b366ffa1aa00f94e/packages/nuxt/src/app/components/client-only.mjs#L23-L-74) does not have any fallback slot. Is this a missing feature (something to add ?) or is it done on purpose ?

pi0:
@danielroe What is status of this PR?

packages/nuxt/src/components/tree-shake.ts
```diff
@@ -1,66 +1,141 @@
...
+    enforce: 'post',
     transformInclude (id) {
       const { pathname } = parseURL(decodeURIComponent(pathToFileURL(id).href))
       return pathname.endsWith('.vue')
     },
-    async transform (code, id) {
-      const template = code.match(/<template>([\s\S]*)<\/template>/)
-      if (!template) { return }
+    transform (code, id) {
+      const regexpMap = new WeakMap<Component[], [RegExp, RegExp, string[]]>()
```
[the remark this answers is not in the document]
huang-julien: indeed, thanks !

huang-julien:
@danielroe  @pi0 it should be stable now :)

warflash:
Hey, just wanted to check in on this PR and ask if there's plans to get it merged any time soon? The issues it's set to resolve prevent us from going from rc.11 to finally 3.0.0 so really looking forward to that :)
Thank you!

huang-julien:
I'm also a bit worried about this, but could'nt find another way to fix it. I hope i didn't miss anything

packages/nuxt/src/components/tree-shake.ts
```diff
@@ -1,67 +1,140 @@
...
 
-      const [COMPONENTS_RE, clientOnlyComponents] = regexpMap.get(components)!
+      const s = new MagicString(code)
+      const importDeclarations: AcornNode<ImportDeclaration>[] = []
+
+      const [COMPONENTS_RE, COMPONENTS_IDENTIFIERS_RE] = regexpMap.get(components)!
       if (!COMPONENTS_RE.test(code)) { return }
 
-      const s = new MagicString(code)
+      walk(this.parse(code, { sourceType: 'module', ecmaVersion: 'latest' }), {
```
danielroe, line 49: I think we're nearly ready to merge this, though it would be nice to add some unit/snapshot tests for this transform as well.

huang-julien: i'll add it :+1:

huang-julien: Unit test added 🙂, i used the vite plugin to get a compiled component. Should i add a snapshot ?

danielroe: Fantastic! And yes, that would be great. ❤️

huang-julien: hmm... sadly i think we might have to remove the snapshots. Generated hashes are differents on the pipeline

danielroe: can you use a regexp to normalise the hashes to a consistent pattern?

huang-julien: i'm not sure about it, there seems to be somthing internal to `vue/compiler-sfc` we can't mock or pass into the plugin. I'll try to mock what is being hashed

huang-julien: well... i had to mock `node:crypto` module

danielroe (commented):
The latest snapshot test failure is likely due to a vue or vite plugin vue upgrade - it's replicable with latest `node_modules` even undoing the merge commit.

Would you be able to have a look? 🙏

huang-julien:
@danielroe Did the CI failed due to 2 timeout ?
Couldn't reproduce it in local

PR #8754, opened 2022-11-06
Fixed cookie value type

magyarb:
### Discussed in https://github.com/nuxt/framework/discussions/2576

The cookie.value is typed as string, so // @ts-ignore must be used to do this. I have fixed the types, so the cookie can be deleted without a ts error.

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #8769, opened 2022-11-07
fix(nuxt): allow cookies to be set to `null` to unset them

danielroe:
### ❓ Type of change

### 📚 Description

remake of https://github.com/nuxt/framework/pull/8754

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/cookie.ts
+++ b/packages/nuxt/src/app/composables/cookie.ts
@@ -15,7 +15,9 @@ export interface CookieOptions<T = any> extends _CookieOptions {
   default?: () => T | Ref<T>
 }
 
-export interface CookieRef<T> extends Ref<T> {}
+export interface CookieRef<T> extends Ref<T> {
+  value: T | null | undefined;
+}
 
 const CookieDefaults: CookieOptions<any> = {
   path: '/',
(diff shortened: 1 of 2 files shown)


pi0 (approved):

Aareksio:
@danielroe This caused `useCookie<Type[]>('name', { default: () => [] })` to be `Ref<Type[] | null>` which forces us to write null checks while in reality the type is strictly never `null`.

danielroe:
@Aareksio I appreciate it; the issue is that it can be null if you set it to null. You could cast it `as Ref<Type[]>` in your case.

Aareksio:
After this change, all code must be explicitly null-checked or `as`-patched. 
I would argue that neither of the following is good DX:

```ts
const languages = useCookie<string[]>('languages', { default: () => ['en'] }) as Ref<string[]>
// or
for (const language of (languages as Ref<string[]>).value) {}
```

danielroe:
I wonder if we could allow overriding with generic - so by default it infers a type with null, but you can override by setting your own value? And if you need to delete you can add `| null` to the generic type?

Aareksio:
Before the change, you should be able to do `useCookie<string | null>()` and the type would be `Ref<string | null>`. Rather painless opt-in.

PR #8805, opened 2022-11-08
fix(ssr): ensure useRequestHeaders are case-insensitive

manniL:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/8803

### ❓ Type of change

### 📚 Description

Makes headers case-insensitive.

Would be good to double-check if headers always come lower-cased from the server. If no, maybe add a transform step there too.
https://github.com/unjs/h3/blob/6beee463d0e95207480acfd21badc2c2fa46f6c3/src/utils/request.ts#L58-L65 does not handle casing. Should I create a PR or an issue there too?

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/composables/ssr.ts
+++ b/packages/nuxt/src/app/composables/ssr.ts
@@ -2,7 +2,7 @@
 import type { H3Event } from 'h3'
 import { useNuxtApp, NuxtApp } from '../nuxt'
 
-export function useRequestHeaders<K extends string = string> (include: K[]): Record<K, string | undefined>
+export function useRequestHeaders<K extends string = string> (include: K[]): Record<Lowercase<K>, string | undefined>
 export function useRequestHeaders (): Readonly<Record<string, string | undefined>>
 export function useRequestHeaders (include?: any[]) {
   if (process.client) { return {} }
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/composables/ssr.ts
```diff
@@ -8,7 +8,7 @@ export function useRequestHeaders (include?: any[]) {
   if (process.client) { return {} }
   const headers = useNuxtApp().ssrContext?.event.req.headers ?? {}
   if (!include) { return headers }
-  return Object.fromEntries(include.filter(key => headers[key]).map(key => [key, headers[key]]))
+  return Object.fromEntries(include.map(key => key.toLowerCase()).filter(key => headers[key]).map(key => [key, headers[key]]))
```
aaharu, line 11: useRequestHeaders's return type is `Record<K, string | undefined>`.

In this case, object key is not `K`.
```ts
useRequestHeaders<'Cookie'>(['Cookie'])
// returns { cookie: '...' }
```

manniL: Good point! Do you have a suggestion how to avoid this?

danielroe: pushed lowercase type helper.

manniL:
Oh well, that was easy. Thanks @danielroe ☺️

pi0 (approved):
LGTM. We might also update this utility with new `h3.useRequestHeaders` that does the normalization

PR #8851, opened 2022-11-09
Added an auth example with JWT and Prisma

jeky1950:
### ❓ Type of change

### 📚 Description

Added an Auth example that uses Prisma ORM and JsonWebToken. The example comes with a Login Page, and Registration Page and a Home Page where the user will land to when they have successfully logged in

There is no holistic implementation of Prisma and JWT in Nuxt 3 that utilizes the best practices as determined by Prisma and Nuxt. This is also a very basic an simple implementation of JWT Auth in Nuxt that can used to scaffold a Nuxt 3 project.

### 📝 Checklist


pi0:
Hi @jeky1950. Thanks so much for providing example but at this point, `examples/` dir in framework repository is mainly to showcase each feature of framework instead of examples.

It would be nice if you can make it a starter template like `jeky1950/nuxt-prisma` for Nuxt 3 so people can fork it using `nuxi init -t jeky1950/nuxt-prisma`

PR #8885, opened 2022-11-10
fix(nuxt)!: remove `initialCache` option

pi0:
### ❓ Type of change

### 📚 Description

Race condition issue in #3985 seems to be resolved. Removing initialCache option and only use utility to preserve payload data **while hydrating**.

BREAKING CHANGE: Payload cache won't be used when doing client-side navigation back to same page and fresh data will be fetched.

### 📝 Checklist

### DIFF
--- a/docs/content/3.api/1.composables/use-async-data.md
+++ b/docs/content/3.api/1.composables/use-async-data.md
@@ -25,7 +25,6 @@ type AsyncDataOptions<DataT> = {
   transform?: (input: DataT) => DataT
   pick?: string[]
   watch?: WatchSource[]
-  initialCache?: boolean
   immediate?: boolean
 }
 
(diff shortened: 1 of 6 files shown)


danielroe (approved):

danielroe:
I was too quick - looks like the test fixture is rightly finding an issue.

pi0:
It is a breaking change coming with this PR indeed. When navigating on client-side, the initial payload cache won't be used as option is dropped.

danielroe:
**Note:** now that I think about, I think we should disable payload extraction by default _anyway_ as I think there's some more refinement/discussion to do, particularly around `useData`, etc. that we probably can't adequately handle before the stable release.

danielroe (approved):

PR #8890, opened 2022-11-10
fix(schema)!: disable payload extraction by default

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/pull/8885#issuecomment-1310248154

### ❓ Type of change

### 📚 Description

I think payload extraction is a killer feature but I think there are still some outstanding issues which we can likely fix shortly after stable release. Until that point, I think it makes sense to disable this by default so users are aware that it is experimental and the API and behaviour may change.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/experimental.ts
+++ b/packages/schema/src/config/experimental.ts
@@ -60,7 +60,7 @@ export default defineUntypedSchema({
      * @type {boolean | ((id?: string) => boolean)}
      */
     inlineSSRStyles: {
-      async $resolve(val, get) {
+      async $resolve (val, get) {
         if (val === false || (await get('dev')) || (await get('ssr')) === false || (await get('builder')) === '@nuxt/webpack-builder') {
           return false
         }
@@ -75,9 +75,9 @@ export default defineUntypedSchema({
     noScripts: false,
 
     /**
-     * When this option is enabled (by default) payload of pages generated with `nuxt generate` are extracted
+     * When this option is enabled, payloads of pages generated with `nuxt generate` are extracted
      */
-    payloadExtraction: true,
+    payloadExtraction: false,
 
     /** Enable cross-origin prefetch using the Speculation Rules API. */
     crossOriginPrefetch: false,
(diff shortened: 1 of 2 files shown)


pi0:
Enabling it back in a semver-minor by default can be a more breaking change. and we need this feature as a baseline for Nuxt 3 release. We shall instead document possibilities of behavior changes with pre-rendered pages and asyncData behavior.

danielroe:
I think disabling it is a better default. There are some issues at the moment (for example, something like https://github.com/nuxt/framework/pull/7805 needs to be merged - though not necessarily that solution).

And I think there are some issues with the implementation that I have described in https://github.com/nuxt/framework/pull/7569#issuecomment-1249172667. I would rather not have people relying on this behaviour in the stable release.

pi0:
Supporting hybrid prerendering is a good idea but we are not aiming for it in the initial v3 release.

`nuxt generate`, replicating Nuxt 2 behavior to crawl and prerender all possible links prerendering is however necessary, and even if not complete it works today.

danielroe:
The issue I linked is not with hybrid prerendering but for sites with dynamic content where a payload wasn't generated, which triggers 404 warnings. There are also issues being reported by users with payloads being wrongly cached and incorrect data being loaded on a subsequent deployment.

pi0:
I'm aware. Hope to fix it before release. But payload cache issues between deployments are not unique to Nuxt3 really nor judge to disable the feature completely. 

Would be happy to discuss this later in team meeting on this :)

danielroe:
I guess the point I'm trying to make is that this feature and its API are not 'stable', and trying to think of the best way to communicate that to people. Happy to defer further discussion to our meeting.

pi0:
Surely we need to communicate in final release notes and docs about the stability of each feature and any known issues.

My point is, even if having issues, it worth to be enabled. Because enabling a feature like payload extraction by default is a real breaking change. It changes caching strategy per navigation and disabled live fetches from server. We really need it and can use leftover time to resolve as much as known issues possible but disabling is not an option for me unless we decide to make it opt-in for Nuxt 3 always.

PR #8901, opened 2022-11-10
refactor(nuxt)!: move `head` option support into `defineNuxtComponent`

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/framework/discussions/8874

### ❓ Type of change

### 📚 Description

This PR moves support for `head` option into `defineNuxtComponent` to allow proper tree-shaking of used composables and avoid registering a global mixin that runs on every single component.

This PR also removes legacy support for `useMeta` which has been deprecated for some time. `useHead` is a drop-in replacement.

### 👉 Migration

Anywhere you are using the Options API `head` hook, you can instead wrap the component with `defineNuxtComponent`. (Or alternatively you may decide to use a component like `<Meta>`, `<Link>`, etc.)

Anywhere you are using `useMeta`, you should replace `useHead`.

### 📝 Checklist


harlan-zw:
Oh nice, this is a good idea.

I'll make the mixin from https://github.com/nuxt/framework/pull/8908 opt-in

pi0:
Any updates on this? 🙏🏼 What is a blocker?

harlan-zw:
lgtm

danielroe:
The issue at the moment is that this PR produces a separate, tiny `_nuxt/composables.js` chunk.

danielroe:
I've looked into this a bit further and it seems inevitable from a rollup point of view, even if we were to move `useHead` into `#app` directly. The issue is that the error components (dynamically imported) are importing _only_ useHead from app and rollup therefore produces a tiny shared chunk. We may be able to refine behaviour with manualChunks in future but I think it should be okay for now, and not replicable if a user is using their own `error.vue`.

pi0 (approved):

PR #8923, opened 2022-11-11
docs: add a bit more detail in the `definePageMeta` warning to specify it needs to be in a page

acidjazz:
This addresses #3175 - a bit more detail on having to be in a page

### DIFF
--- a/packages/nuxt/src/pages/runtime/composables.ts
+++ b/packages/nuxt/src/pages/runtime/composables.ts
@@ -38,7 +38,7 @@ export interface PageMeta {
 }
 
 declare module 'vue-router' {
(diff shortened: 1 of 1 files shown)


danielroe (approved):

Atinux (approved):

vedmant:
Ho do I update it globally for all pages? I tried to run it in app.vue, and getting this error:

definePageMeta() is a compiler-hint helper that is only usable inside the script block of a single file component which is also a page. Its arguments should be compiled away and passing it at runtime has no effect.

danielroe:
You can't use this macro to do that. Instead, you could use `extendPages` or something similar. If you have other questions, please do raise a discussion or ask a question in Discord instead of commenting on a closed PR. 🙏

PR #8941, opened 2022-11-13
feat(pages): routes HMR

antfu:
### ❓ Type of change

### 📚 Description

This make adding/removing routes HMR, without requiring a full reload.

### 📝 Checklist


packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -67,10 +67,23 @@ export default defineNuxtPlugin(async (nuxtApp) => {
   const router = createRouter({
     ...routerOptions,
     history,
-    routes
+    routes: []
   })
+  let clearRoutes = router.addRoute({ path: '', children: routes })
```
posva, line 72: shouldn't this be
```suggestion
  let clearRoutes = router.addRoute({ path: '/', children: routes })
```

did you find any trouble with it? The router is more preared to take paths that start with `/` when they have no parent

packages/nuxt/src/pages/module.ts
```diff
@@ -159,7 +159,17 @@ export default defineNuxtModule({
...
+        return [
+          ...imports,
+          `export default ${routes}`,
+          'export function onUpdate(fn) { onUpdate._fn = fn }',
+          'if (import.meta.hot) {',
+          '  import.meta.hot.accept((mod) => {',
+          '    onUpdate._fn?.(mod.default)',
+          '    mod.onUpdate._fn = onUpdate._fn',
+          '  })',
+          '}'
```
danielroe, line 171: do we need to add an equivalent for webpack, like the following?

https://github.com/nuxt/framework/blob/1700bf822e31ffe2c9ec66b61d6f6562d45dcf2b/packages/nuxt/src/app/config.ts#L73-L77

packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -67,10 +67,23 @@ export default defineNuxtPlugin(async (nuxtApp) => {
...
+  if (import.meta.hot) {
+    // @ts-expect-error
+    import('#build/routes').then(({ onUpdate }) => {
+      onUpdate((newRoutes: any) => {
+        const routes = routerOptions.routes?.(newRoutes) ?? newRoutes
+        clearRoutes()
+        clearRoutes = router.addRoute({ path: '', children: routes })
+      })
+    })
+  }
```
danielroe, line 85: as above - do we need to add a webpack equivalent?

packages/nuxt/src/pages/module.ts
```diff
@@ -159,7 +159,17 @@ export default defineNuxtModule({
         const pages = await resolvePagesRoutes()
         await nuxt.callHook('pages:extend', pages)
         const { routes, imports } = normalizeRoutes(pages)
-        return [...imports, `export default ${routes}`].join('\n')
+        return [
+          ...imports,
+          `export default ${routes}`,
+          'export function onUpdate(fn) { onUpdate._fn = fn }',
+          'if (import.meta.hot) {',
```
danielroe, line 166: It might be worth also adding a test for `process.dev`

packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -67,10 +67,23 @@ export default defineNuxtPlugin(async (nuxtApp) => {
...
     ...routerOptions,
     history,
-    routes
+    routes: []
   })
+  let clearRoutes = router.addRoute({ path: '', children: routes })
   nuxtApp.vueApp.use(router)
 
+  // Routes HMR
+  if (import.meta.hot) {
```
danielroe, line 76: as above, it might be worth also adding a test for `process.dev`

packages/nuxt/src/pages/runtime/router.ts
danielroe, line 72: is there any runtime overhead to doing this this way? if so, I guess we can use `process.dev` to only do this in dev mode?

danielroe:
Thank you! This will be amazing 🚀

packages/nuxt/src/pages/runtime/router.ts
posva: isn't this automatically false in production too?

packages/nuxt/src/pages/runtime/router.ts
posva: It shouldn't have too much of an overhead but I haven't benchmarked it. This is the code that runs with this: https://github.com/vuejs/router/blob/main/packages/router/src/matcher/index.ts#L161-L170

If it does end up having a bad impact on performance, then another more verbose version that inlines the routes at router creation in production and adds each route individually in dev would work too:

```ts
let clearRoutes = routes.map(route => router.addRoute(route))

clearRoutes.forEach(removeRoute => removeRoute())
```

packages/nuxt/src/pages/module.ts
antfu: I know nothing about Webpack HMR 😅, feel free to push code you think necessary.

danielroe: ok, will have a look later!

packages/nuxt/src/pages/runtime/router.ts
danielroe: yes, it is but `process.dev` is needed for compatibility with both vite + webpack tree shaking (e.g. vite defines `import.meta.hot` == 'false' but IIRC not `import.meta.webpackHot`)

danielroe:
@antfu I was checking this out but I think something must have changed elsewhere (maybe in vite or vite-node?) as the `onUpdate` function isn't getting called when a new page is added or changed. Do you have any idea what the issue might be?

PR #8953, opened 2022-11-13
docs: clarify augmenting input/output app config

xtoolkit:
### ❓ Type of change

### 📚 Description

useAppConfig type is `AppConfig` and not extend to AppConfigInput. so typescript can't see user define

### 📝 Checklist


danielroe (commented):
It might be worth moving the previous example to the module author guide (see todo item just before this section). There are two use cases:

1. Users typing the _output_ of app.config.
2. Module authors and users typing what is valid _input_ for app.config.

Ideally use case #1 should never be required as app config should be fully typed based on the _actual_ values.

danielroe (approved):

danielroe:
Thank you for raising this 😊

PR #8959, opened 2022-11-14
docs: match the open graph protocol markup

IsraelOrtuno:
### ❓ Type of change

### 📚 Description

https://ogp.me/ suggests to use the attribute `property` instead of `name` when defining OpenGraph metadata.

### 📝 Checklist

### DIFF
--- a/docs/content/1.getting-started/5.seo-meta.md
+++ b/docs/content/1.getting-started/5.seo-meta.md
@@ -234,7 +234,7 @@ And then in your layout file, you might use the route's metadata you have previo
 const route = useRoute()
 
 useHead({
-  meta: [{ name: 'og:title', content: `App Name - ${route.meta.title}` }]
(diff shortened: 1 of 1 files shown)


danielroe (approved):

manniL:
@harlan-zw Maybe `unhead` could allow both, `name` *or* `property` and automatically choose the right one (e.g. property for `og:` tags, name otherwise?)

harlan-zw:
> @harlan-zw Maybe `unhead` could allow both, `name` _or_ `property` and automatically choose the right one (e.g. property for `og:` tags, name otherwise?)

The logic for determining that is actually quite complex (like 1kb of scripts), so instead, we implement this type of logic in a opt-in composable.

https://unhead.harlanzw.com/guide/guides/useseometa

PR should be up very shortly

PR #8967, opened 2022-11-14
docs(api): add spaces to avoid breaking mobile layout

clemcode:
### ❓ Type of change

### 📚 Description
This PR add spaces in long inline code example to avoid breaking mobile layout.

### DIFF
--- a/docs/content/3.api/5.commands/add.md
+++ b/docs/content/3.api/5.commands/add.md
@@ -89,7 +89,7 @@ npx nuxi add middleware auth
 
 ## `nuxi add api`
 
-* Modifier flags: `--method=connect|delete|get|head|options|patch|post|put|trace` or `--get`, `--post`, etc.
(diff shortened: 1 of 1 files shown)


docs/content/3.api/5.commands/add.md
```diff
@@ -89,7 +89,7 @@ npx nuxi add middleware auth
 
 ## `nuxi add api`
 
-* Modifier flags: `--method=connect|delete|get|head|options|patch|post|put|trace` or `--get`, `--post`, etc.
+* Modifier flags: `--method=connect | delete | get | head | options | patch | post | put | trace` or `--get`, `--post`, etc.
```
danielroe, line 92: what about something like:

```suggestion
* Modifier flags: `--method` (can accept `connect`, `delete`, `get`, `head`, `options`, `patch`, `post`, `put` or `trace`) or alternatively you can directly use `--get`, `--post`, etc.
```

PR #8908, opened 2022-11-11
feat: `@vueuse/head` v1

harlan-zw:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Read the v1 @vueuse/head post for context https://harlanzw.com/blog/vue-use-head-v1

### Nuxt & @vueuse/head

I have pushed this PR up to try and get it in before the v3 stable.

The DOM patching logic is more complex than the previous implementation and it's quite possible there are issues that won't be uncovered until wide testing. So while my personal testing has gone quite well, I'll leave this as a draft PR until I have a playground ready and can do some more testing.

### Next Steps

In terms of development, once this is merged then I can look at implementing the rest of the new features unlocked by this:
- add shortcut composables (includes useTagMetaFlat, solves the Meta SEO of this [discussion](https://github.com/nuxt/framework/discussions/1823))
- add `useServerHead`, opens the door for 0kb runtime head
- expose hooks, this will be powerful for module authors

## Migration Guide

(copied from the release notes)

⚠️ Breaking changes are minimal, but there are some changes to be aware of.

Please report any issues you find and they will fixed be promptly.

You may consider using [@unhead/vue](https://unhead.harlanzw.com/integrations/vue/setup) directly if you [don't need @vueuse/head](https://unhead.harlanzw.com/integrations/vue/vueuse-head).

### Verify your tags

The new DOM patching algorithm has not been tested in all possible scenarios, it's possible that there are unforeseen edge cases.  

### `htmlAttrs` and `bodyAttrs` merge strategy

If you had built your code around with the assumption that setting `htmlAttrs` or `bodyAttrs` would clear the old tags, this is now different.

```ts
// old
useHead({
  htmlAttrs: {
    class: 'my-class',
  },
})

useHead({
  htmlAttrs: {
    class: 'new-class',
  },
})

// <html class="new-class">
```

```ts
// new
useHead({
  htmlAttrs: {
    class: 'my-class',
  },
})

useHead({
  htmlAttrs: {
    class: 'new-class',
  },
})

// <html class="my-class new-class">
```

[...]

PR #8975, opened 2022-11-14
feat(nuxt, schema): official @vueuse/head v1 support

harlan-zw:
### 🔗 Linked issue

Remake of PR https://github.com/nuxt/framework/pull/8908#issuecomment-1313377541

- https://github.com/nuxt/framework/issues/5492

### ❓ Type of change

### 📚 Description

New Nuxt v3 projects are already using v1 @vueuse/head outside of this PR. There are no open issues for the v1 release currently. 

This PR aims to:
- not using the deprecated functions
- fix tag generation in SPA mode https://github.com/nuxt/framework/issues/5492

### 📝 Checklist

Basic documentation, will need continued work over the coming weeks. I've kept the two new composable undocumented for the initial release, shouldn't be an issue for now?


test/bundle.test.ts
```diff
@@ -33,6 +33,7 @@ describe.skipIf(isWindows)('minimal nuxt application', () => {
         "_nuxt/error-404.js",
         "_nuxt/error-500.js",
         "_nuxt/error-component.js",
+        "_nuxt/index.js",
```
harlan-zw, line 36: No idea why this is being generated

harlan-zw:
Ready for review, I'll write a couple of tests soon

packages/nuxt/src/head/module.ts
```diff
@@ -29,6 +29,17 @@ export default defineNuxtModule({
       })
     }
 
+    // add non useHead composables
+    nuxt.hooks.hook('imports:sources', (sources) => {
+      sources.push({
+        from: '@vueuse/head',
+        imports: [
+          'useSeoMeta',
+          'injectHead'
```
pi0, line 38: Since we are close to v3 release, it would be nice to add auto imports behind an experimental flag and use direct import for internal usage.

packages/nuxt/src/head/runtime/lib/vueuse-head.plugin.ts
```diff
@@ -1,63 +1,48 @@
...
 
-  nuxtApp._useHead = (_meta: MaybeComputedRef<MetaObject>, options: HeadEntryOptions) => {
-    if (process.server) {
-      head.addEntry(_meta, options)
-      return
-    }
-
-    const cleanUp = head.addReactiveEntry(_meta, options)
+  // basic support for users to modify tags before render
+  head.internalHooks.hook('tags:resolve', ctx => nuxtApp.hooks.callHook('head:tags:resolve', ctx))
```
pi0, line 34: Is this hook necessary? If not can we please add it later?

pi0:
Thanks so much for working on this @harlan-zw 💯 I will have a local look soon but overall unhead migration seems already in good direction.

Since we are so close to latest RC and v3 release, i think it is good idea to consider including this change and release fixes in upstream if any. Only, i would appreciate if we can minimize new feature additions to reduce chances of breaking changes for end-users.

harlan-zw:
> Thanks so much for working on this @harlan-zw 100 I will have a local look soon but overall unhead migration seems already in good direction.
> 
> Since we are so close to latest RC and v3 release, i think it is good idea to consider including this change and release fixes in upstream if any. Only, i would appreciate if we can minimize new feature additions to reduce chances of breaking changes for end-users.

Good idea, will remove anything non-essential

pi0:
Overall looks good to me! Let's try changes on Nuxt edge.

nathanchase:
It doesn't seem that useHead updates on computed properties. 

Example:
```
const pageTitle = computed(() =>
  route.params.query
    ? `🔍 Search results for "${route.params.query}"`
    : '🔍 Search',
);

useHead({
  title: pageTitle.value,
});
```
Even if `pageTitle` correctly includes the variable, the page title never updates.

tylerforesthauser:
Alternatively, it is suggested to use a getter. 
```js
useHead({
  title: () => pageTitle.value,
})
```

PR #9011, opened 2022-11-15
refactor: update unjs dependencies to stable v1

pi0:
### ❓ Type of change

### 📚 Description

Update uns packages with semver major to v1.

Also updating nitropack temporarily to edge channel to finish update in one step.

### 📝 Checklist


manniL:
@pi0 Could you briefly explain why was `ohmyfetch` renamed to `ofetch`? I'm a little afraid people have to memorize "yet another library name" after getting used to `ohmyfetch` + using the package in projects 🙈

pi0:
> Could you briefly explain why was ohmyfetch renamed to ofetch? 

it is a shorter name. I initially wanted to release as shorter name but took long to get publish access to it. As for Nuxt users, interface name is still `$fetch`. Perhaps can add some references to ofetch readme to explain renaming.

manniL:
> it is a shorter name

Do you think the shorter name really worth all the migration for all packages using `ohmyfetch` now explicitly?

pi0:
Hmm will probably keep `ohmyfetch` as placeholder in a semver-major. Making it easier to migrate other packages. But for renaming really had to do this before v3 as you notice, it gets harder to break API or make renames as we have more users.

manniL:
> Hmm will probably keep `ohmyfetch` as placeholder in a semver-major. 

That'd be good at least 👍 

> But for renaming really had to do this before v3

Or not at all 🙈

PR #9018, opened 2022-11-15
fix(nuxt)!: enable payload extraction only for `nuxi generate`

### LINKED ISSUE #8914: Disable payload extraction for `nuxi build`
Followup https://github.com/nuxt/framework/pull/8890 by @danielroe

There are currently some uncertain decisions about the final API and behavior with payload extraction behavior like how we apply payload updates on CSR route change (should the key of static payload be prefixed with routes as well or remain global same as asyncData keys). (Tracker issue: #6411)

Making the decision to enable this feature or not is a tricky one because on one hand, we are really close to v3.0.0 release deadline and we don't want to mark a functionality as stable and change it in a semver-minor release and on the other hand, payload extraction is an essential feature for the release with full static generation support.

I suggest two behaviors:

- For `nuxt build` when we have a server and the possibility of having hybrid prerendering, disable an option by default. **Payload extraction will be always opt-in in Nuxt 3 for server mode**.
  - Docs: Deploy>Node.js Server https://v3.nuxtjs.org/getting-started/deployment#nodejs-server
- For `nuxt generate` when we are expected to output a "full static" output, enable the future
  - Update the docs and warn about the stability of `nuxt generate` 
  - Explain how payload extraction works and what are consequences of disabling it (needing a server)
  - Docs: Deploy> Static Hosting https://v3.nuxtjs.org/getting-started/deployment#static-hosting

An alternative is to always disable feature until we are really sure it is stable but there are few downsides:
- Enabling it by default for `nuxt generate`, even if super stable, is still a breaking change considering we essentially change caching strategy
- Enabling it with a new flag (nuxt2 was `target: 'static'`) is less intuitive for starter users and even more confusing to explain
- Differing to enable, means also less usage of the feature and harder to ever stabilize it

pi0:
### 🔗 Linked issue

Resolves #8914

### ❓ Type of change

### 📚 Description

Only enable full-static support with payload extraction when using `nuxi generate`. It will display a warning always with how to opting out.

With this PR landing for v3, payload extraction support for `nuxi build` will be **always opt-in**.

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/experimental.ts
+++ b/packages/schema/src/config/experimental.ts
@@ -77,7 +77,15 @@ export default defineUntypedSchema({
     /**
      * When this option is enabled (by default) payload of pages generated with `nuxt generate` are extracted
      */
-    payloadExtraction: true,
+    payloadExtraction: {
+      async $resolve(enabled, get) {
+        enabled = enabled ?? false
+        if (enabled) {
+          console.warn('Using experimental payload extraction for full-static output. You can opt-out by setting `experimental.payloadExtraction` to `false`.')
+        }
+        return enabled
+      }
+    },
 
     /** Enable cross-origin prefetch using the Speculation Rules API. */
     crossOriginPrefetch: false,
--- a/packages/nuxi/src/commands/build.ts
+++ b/packages/nuxi/src/commands/build.ts
@@ -27,6 +27,11 @@ export default defineNuxtCommand({
         cwd: rootDir,
         fileName: args.dotenv
       },
+      defaults: {
+        experimental: {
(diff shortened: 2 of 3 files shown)


packages/schema/src/config/experimental.ts
```diff
@@ -77,7 +77,15 @@ export default defineUntypedSchema({
     /**
      * When this option is enabled (by default) payload of pages generated with `nuxt generate` are extracted
      */
-    payloadExtraction: true,
+    payloadExtraction: {
+      async $resolve(enabled, get) {
```
pi0, line 81: Note: `get('_generate')` cannot be accessed from here. Seems an issue with untyped

PR #9111, opened 2022-11-17
docs(seo): add robots file

clemcode:
### ❓ Type of change

### 📚 Description
This PR adds a `robots.txt` file with default rules

### DIFF
--- a/docs/public/robots.txt
+++ b/docs/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Allow: /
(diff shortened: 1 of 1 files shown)


docs/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Allow: /
```
danielroe, line 2: I think this is the same as an empty file.

```suggestion
```

clemcode: Go for an empty file then :)

Atinux (approved):

Issue #8962, opened 2022-11-14
How to use "extendRoutes" in nuxt3?


alterhu2020:
### Discussed in https://github.com/nuxt/framework/discussions/1523

<div type='discussions-op-text'>

<sup>Originally posted by **crotanite** October 27, 2021</sup>
nuxt 2 documentation for modules says there is an "extendRoutes" function, but this doesn't seem to be importable at all through nuxt3

Is it currently possible to use this in nuxt3?</div>

IsraelOrtuno:
I think this should be moved to Discussions.

Not sure what you are trying to accomplish but redirections can be configured via `routeRules` property from `nuxt.config.ts`
https://v3.nuxtjs.org/guide/concepts/rendering#route-rules

Check also this example about how to extend your pages from a module
https://v3.nuxtjs.org/examples/advanced/module-extend-pages#module-extend-pages

alterhu2020:
> I think this should be moved to Discussions.
> 
> Not sure what you are trying to accomplish but redirections can be configured via `routeRules` property from `nuxt.config.ts` https://v3.nuxtjs.org/guide/concepts/rendering#route-rules
> 
> Check also this example about how to extend your pages from a module https://v3.nuxtjs.org/examples/advanced/module-extend-pages#module-extend-pages

Thanks @IsraelOrtuno , the example page: https://v3.nuxtjs.org/examples/advanced/module-extend-pages/#module-extend-pages  shows that we can use `extendPages` to add more routes, but the problem here is that `pages` parameter's type is `NuxtPage`, and it has the `redirect` property as well, but `redirect` property only supports `string` type. which is different as vue router standard api:

<img width="401" alt="image" src="https://user-images.githubusercontent.com/5029046/201798652-80ccc6dc-def3-4666-a107-f5ef4f8af58d.png">

<img width="854" alt="image" src="https://user-images.githubusercontent.com/5029046/201799060-d795ca1f-952d-4c42-973e-7ee7cbde3986.png">

If i want the have more custom function in `redirect` property, how can i achieve it? thanks for help

PR #9120, opened 2022-11-17
feat: support redirect function for routes in `pages:extend`

ms-fadaei:
### 🔗 Linked issue

#8962 

### ❓ Type of change

### 📚 Description

Hi ✋ 
As mentioned in the linked issue, the `NuxtPage` type doesn't support the redirect option as a function.
This PR will fix that.

### 📝 Checklist


danielroe (commented):
I don't think we want to support stringifying functions within `pages:extend`. Instead, a custom `app/router.options` would likely cover this use case.

ms-fadaei:
> I don't think we want to support stringifying functions within `pages:extend`. Instead, a custom `app/router.options` would likely cover this use case.

This is a bad idea from my perspective too, but since the Nuxt 2 supported it and we used it I think we should have it if you agree with this approach. If you aren't, feel free to close the PR.

danielroe:
I appreciate you raising the PR, but I will go ahead and close it in that case :+1:

PR #9161, opened 2022-11-18
fix(nuxt): don't try to override computed layouts in `definePageMeta`

### LINKED ISSUE #8678: `definePageMeta`'s layout property errors when computed
### Environment

- Operating System: `Linux`
- Nuxt Version:     `3.0.0-rc.13-27789756.491d02f`
- Nitro Version:    `0.6.1-27791756.086802d`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://stackblitz.com/edit/github-s6twqp?file=pages%2Findex.vue

### Describe the bug

It [appears to me](https://v3.nuxtjs.org/api/utils/define-page-meta#parameters) that one can use a computed layout in `definePageMeta`. But apparently (see minimal reproduction) that does not work and prints `proxy set handler returned false for property '"layout"'` instead.

That error only appears on client side though, the initial ssr page displays correctly. Try reloading the page.

### Additional context

I receive the same error on RC12.

ms-fadaei:
### 🔗 Linked issue

resolves #8678 

### ❓ Type of change

### 📚 Description
Hey
As mentioned in the linked issues, there is a bug with the computed layout when using `definePageMeta`.
Also, there is a workaround for that as @danielroe mentioned it.
But after a bit of research I figured out what's going on and it can be fixed with this PR.
I also changed the example to be clearer.

### 📝 Checklist


ms-fadaei:
I have a question too: as the document says we can use ref for the reactive layout but it throws an error (compiler macros I think) when using just a simple ref (without computed wrapper). Do you have any idea about that?

examples/routing/layouts/pages/dynamic.vue
```diff
@@ -20,13 +20,9 @@
   </div>
 </template>
 
-<script>
+<script setup>
+const layout = ref('custom')
 definePageMeta({
-  layout: false
+  layout: computed(() => layout.value)
```
danielroe, line 26: I think we should keep this example as is, but maybe add this usage to the basic/ test fixture.

ms-fadaei: Done! but it has an error related to the macro (same as the ref that mentioned above). I think this problem is related to the compiler macro process.

ms-fadaei: This is the processed macro string:

```
const __nuxt_page_meta = {                                                                                                  16:15:43
  layout: computed(() => test.value)
}
export default __nuxt_page_meta
// Vite
if (import.meta.hot) {
  import.meta.hot.accept(mod => {
    Object.assign(__nuxt_page_meta, mod)
  })
}
// Webpack
if (import.meta.webpackHot) {
  import.meta.webpackHot.accept((err) => {
    if (err) { window.location = window.location.href }
  })
}
```

danielroe: yes. you can reference imports but not other variables within the setup function.

ms-fadaei: So this must be fixed too or not? This is a serious bug and means we can't use reactive variables with the `definePageMeta`!

danielroe: no, that should not have worked (sharing context between route metadata and the component instance) - instead, you can just directly set `useRoute().meta.someReactiveThing...

ms-fadaei: Soooo...
I think this PR is not necessary because we can't use reactive property (ref or computed) in the `definePageMeta` at all and the provided example (`computed(() => 'custom')`) is not a real-world example.

Maybe should we update the [documentation](https://nuxt.com/docs/api/utils/define-page-meta#type)? 
What do you think @danielroe @pi0 ?

packages/nuxt/src/app/plugins/router.ts
```diff
@@ -222,8 +222,8 @@ export default defineNuxtPlugin<{ route: Route, router: Router }>((nuxtApp) => {
   nuxtApp.hooks.hookOnce('app:created', async () => {
     router.beforeEach(async (to, from) => {
       to.meta = reactive(to.meta || {})
-      if (nuxtApp.isHydrating) {
-        to.meta.layout = initialLayout.value ?? to.meta.layout
+      if (nuxtApp.isHydrating && initialLayout.value !== undefined && initialLayout.value !== null) {
```
danielroe, line 225: nice find.

would we be better off checking if `to.meta.layout` is writeable with the `isReadonly` helper from vue?

```suggestion
      if (nuxtApp.isHydrating && !isReadonly(to.meta.layout) {
```

packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -117,8 +117,8 @@ export default defineNuxtPlugin(async (nuxtApp) => {
   const initialLayout = useState('_layout')
   router.beforeEach(async (to, from) => {
     to.meta = reactive(to.meta)
-    if (nuxtApp.isHydrating) {
-      to.meta.layout = initialLayout.value ?? to.meta.layout
+    if (nuxtApp.isHydrating && initialLayout.value !== undefined && initialLayout.value !== null) {
```
danielroe, line 120:
```suggestion
    if (nuxtApp.isHydrating && !isReadonly(to.meta.layout) {
```

packages/nuxt/src/pages/runtime/router.ts
```diff
@@ -117,8 +117,8 @@ export default defineNuxtPlugin(async (nuxtApp) => {
...
+      to.meta.layout = initialLayout.value
```
danielroe, line 121:
```suggestion
      to.meta.layout = initialLayout.value ?? to.meta.layout
```

packages/nuxt/src/app/plugins/router.ts
```diff
@@ -222,8 +222,8 @@ export default defineNuxtPlugin<{ route: Route, router: Router }>((nuxtApp) => {
...
+        to.meta.layout = initialLayout.value
```
danielroe, line 226:
```suggestion
        to.meta.layout = initialLayout.value ?? to.meta.layout
```

examples/routing/layouts/pages/dynamic.vue
danielroe: I think it's still a good PR ❤️  Sorry for the delay in getting this merged.

danielroe (approved):

PR #9181, opened 2022-11-19
fix: remove `postcss.config` from schema

nathanchase:
### 🔗 Linked issue
resolves https://github.com/nuxt/framework/issues/5368

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/packages/schema/src/config/postcss.ts
+++ b/packages/schema/src/config/postcss.ts
@@ -4,10 +4,6 @@ import { defineUntypedSchema } from 'untyped'
 
 export default defineUntypedSchema({
   postcss: {
-    /** Path to postcss config file. */
-    /** @type string | false */
-    config: false,
-
     /**
      * Options for configuring PostCSS plugins.
      *
--- a/packages/webpack/src/utils/postcss.ts
+++ b/packages/webpack/src/utils/postcss.ts
@@ -59,17 +59,7 @@ export const getPostcssConfig = (nuxt: Nuxt) => {
   if (!nuxt.options.webpack.postcss || !nuxt.options.postcss) {
     return false
   }
-
-  const configFile = nuxt.options.postcss.config
-  if (configFile) {
-    return {
-      postcssOptions: {
-        config: configFile
-      },
-      sourceMap: nuxt.options.webpack.cssSourceMap
-    }
-  }
-
+ 
   let postcssOptions = cloneDeep(nuxt.options.postcss)
   // Apply default plugins
   if (isPureObject(postcssOptions)) {
--- a/packages/webpack/src/utils/postcss.ts
+++ b/packages/webpack/src/utils/postcss.ts
@@ -59,7 +59,7 @@ export const getPostcssConfig = (nuxt: Nuxt) => {
   if (!nuxt.options.webpack.postcss || !nuxt.options.postcss) {
     return false
   }
- 
+
   let postcssOptions = cloneDeep(nuxt.options.postcss)
   // Apply default plugins
   if (isPureObject(postcssOptions)) {


nathanchase:
I brought up the reason for this PR on the Nuxt Discord: https://discord.com/channels/473401852243869706/1043174741135069184

Essentially, the docs seemed to indicate that this was a boolean, even though it seems like it should have been a string (path to a postcss.config.js). https://github.com/nuxt/framework/issues/5368 seems to indicate this would be a moot point anyway if it is to be deprecated.

danielroe:
We don't want to support an external `postcss.config.js` as other modules may well depend on the user-set values and need to be able to read/set them from the nuxt config file.

See for example https://github.com/nuxt/framework/issues/5368.

nathanchase:
> We don't want to support an external `postcss.config.js` as other modules may well depend on the user-set values and need to be able to read/set them from the nuxt config file.

Understood. That makes sense in context of Nuxt modules. We may want to be explicit in the Nuxt docs somewhere then that Nuxt can only configure postcss via the nuxt.config - since postcss's docs explicitly say to include a postcss.config as the recommended method of configuring it.

pi0:
Thanks for your pull request @nathanchase ❤️ Webpack and Schema were left over i forgot to finally include in 3.0.0.

> since postcss's docs explicitly say to include a postcss.config as the recommended method of configuring it.

Indeed it is a tricky situation when comes to external config files. Technically it is not a limitation. We can load postcss config and merge it with Nuxt config and also allow modules extend the merged result but more a consistency aspect since we don't also load external `vite.config`, `nitro.config`, `babel.config` etc. Again tricky because some modules (like tailwind) actually do support external configs!

> We may want to be explicit in the Nuxt docs somewhere then that Nuxt can only configure postcss via the nuxt.config

Seems good idea. We can clarify it with JSDocs for configuration that we don't load it. But if you have an idea of somewhere more accessible that we talk about generic Nuxt and External config files that would be awesome 👍🏼

pi0 (approved):

nathanchase:
> Thanks for your pull request @nathanchase ❤️ Webpack and Schema were left over i forgot to finally include in 3.0.0.
> 
> > since postcss's docs explicitly say to include a postcss.config as the recommended method of configuring it.
> 
> Indeed it is a tricky situation when comes to external config files. Technically it is not a limitation. We can load postcss config and merge it with Nuxt config and also allow modules extend the merged result but more a consistency aspect since we don't also load external `vite.config`, `nitro.config`, `babel.config` etc. Again tricky because some modules (like tailwind) actually do support external configs! PS: They have usually more than one file to configure. We can try to _mimic_ config loading but it is not precise and can be conflicting with another project in same root dir.
> 
> > We may want to be explicit in the Nuxt docs somewhere then that Nuxt can only configure postcss via the nuxt.config
> 
> Seems good idea. We can clarify it with JSDocs for configuration that we don't load it. But if you have an idea of somewhere more accessible that we talk about generic Nuxt and External config files that would be awesome 👍🏼

I'd say https://nuxt.com/docs/getting-started/configuration seems like a good candidate for where to describe how Nuxt does or doesn't support external config files. I'd say for sure to include the ones you mentioned - `vite.config`, `nitro.config`, `babel.config`, `postcss.config`. and how they are configured in Nuxt via nuxt.config instead. 

As you said, it is confusing since there are configuration files like `.eslintrc.js` or `tsconfig.json` or `.editorconfig` that do work for the respective use cases and are generally necessary external config files.

I'll work on another PR and try to come up with a list of common config files that you might have in a Nuxt 3 project, and perhaps we can just create a side-by-side table saying how those configurations are integrated with Nuxt.

nathanchase:
PR added: https://github.com/nuxt/framework/pull/9189

HigherOrderLogic:
Isn't this a breaking change?

danielroe:
Yes, this is a breaking change. And technically vite still has the ability to use an external `postcss.config` file unless we explicitly disable it (which we are not doing, but could)...

@pi0 shall we instead deprecate this config and then remove it entirely in 3.1?

Issue #9185, opened 2022-11-19
Template refs in .client components not available at `onMounted()`


fkammer:
### Environment

```
❯ npx nuxi info
Nuxi 3.0.0

RootDir: /home/projects/github-dpyktf-owfxvw
Nuxt project info: 

------------------------------
- Operating System: `Linux`
- Node Version:     `v16.14.2`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `npm@7.17.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`
------------------------------
```

### Reproduction

https://stackblitz.com/edit/github-dpyktf-owfxvw?file=app.vue

### Describe the bug

[Template refs](https://vuejs.org/guide/essentials/template-refs.html) are not available on [client-side only components](https://nuxt.com/docs/guide/directory-structure/components#client-components) defined in `.client` files.

It works as expected on components not having `.client` in their file name. It makes no difference if SSR is enabled or not. As long as a components file name ends with `.client` the ref is not set as expected on `onMounted`.

### Logs

Console logs from the linked StackBlitz example:

```
{component: 'FocusInput', ref: input}
{component: 'FocusInputBroken', ref: null}
```

huang-julien:
Hi 👋 thanks for reporting this ! 
Currently .client components await for the component to be mounted before rendering it so you have to `await nextTick()` .
it's like this : 

````
- vue render .server or div
-- vue mount your component and call `onMounte()` hook, but the template defined in your component in not rendered yet
-- next tick
- vue render your component
````

changing your `onMounted` to: 
````ts
onMounted(async () => {
    await nextTick()
  console.log({
    component: 'FocusInputBroken',
    ref: input.value,
  });
  // input.value?.focus();
  refValue.value = input.value ? 'HTMLInputElement' : 'null';
});
````
should work. 

@danielroe I'm not sure if we can delay the `onMounted()` hook of a `.client` component at runtime. What do you think about adding a `await nextTick()` in all `.client` `onMounted()` through a plugin ?

fkammer:
Hi, thanks for having a look at this! Awaiting the next tick works well.

I would expect `onMounted()` to behave the same on a `.client` component as it does on a "regular" component. If that would cause other side-effects, we should at least add a warning to the documentation, that a `.client` component isn't rendered when `onMounted()` is called and mention the `await nextTick()` trick.

[...]

PR #9195, opened 2022-11-20
feat(nuxt): add `await nextTick()` to all .client components onMounted hook

huang-julien:
### 🔗 Linked issue
resolve https://github.com/nuxt/framework/issues/9185

### ❓ Type of change

### 📚 Description
Hi :wave: ! 

This PR transforms all `onMounted` hooks in `.client` components to add an `await nextTick()`.

This is due to the result of `createClientOnly` that renders the component only after being mounted. 

If an user try to access the DOM or a ref in `onMounted`, this would fail since the component is only rendering the `.server` part or a simple `<div>` at the time the component is being rendered.

Note that it only transforms `onMounted` hooks `.client` files, this issue still can be reproduced if `onMounted` is called from a composable within a `.client` component.

@fkammer also mentionned adding `await nextTick()` in the docs for `.client` components.

 Let me know what you think about it.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/components/client.ts
+++ b/packages/nuxt/src/components/client.ts
@@ -1,0 +1,88 @@
+import { pathToFileURL } from 'node:url'
+import MagicString from 'magic-string'
+import { walk } from 'estree-walker'
+import { createUnplugin } from 'unplugin'
+import { parseURL } from 'ufo'
+import type { ArrowFunctionExpression, CallExpression, ExpressionStatement, Identifier, Property, FunctionDeclaration, BlockStatement } from 'estree'
+import { genImport } from 'knitwork'
+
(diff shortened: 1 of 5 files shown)


tylerforesthauser:
I'm no expert so I'm just spitballing here but is there any way to use the onBeforeMount hook instead of onMount in the client only component itself?

huang-julien:
> I'm no expert so I'm just spitballing here but is there any way to use the onBeforeMount hook instead of onMount in the client only component itself?

The issue with using `onBeforeMount` is that the component will render before the hydration is completed, thus causing a hydration mismatch for vue

huang-julien:
closing, prefer document this instead. See https://github.com/nuxt/framework/issues/9185#issuecomment-1323481929

PR #9204, opened 2022-11-20
fix(nuxt): include missing `<NuxtPage>` component props

DamianGlowala:
### ❓ Type of change

### 📚 Description

Adds missing props to the `<NuxtPage>` component's type cast. Are default `undefined`/`null` values of the props serving any purpose? Can they be removed?

### 📝 Checklist


danielroe:
> Are default undefined/null values of the props serving any purpose? Can they be removed?

Yes, they are serving a purpose - we want to distinguish between default (undefined, in which case we want to fall back to app-wide defaults) and deliberately false values.

danielroe (approved):
Thank you ❤️

PR #9262, opened 2022-11-22
feat: `useNuxtData` composable

### LINKED ISSUE #7568: `useData` composable

### LINKED ISSUE #9260: Accessing async data cache object with a new composable
### Describe the feature

The cache for the data fetching mechanism is already accessible via `useNuxtApp().payload.data`, I'm suggesting exposing it with a separate composable (`useNuxtData` or `getNuxtData`). it can introduce some interesting data-fetching patterns such as optimistic updates.

### Examples
- Showing stale data instead of a loading spinner (reading from cache): 

```ts
interface Post {
  userId: number;
  title: string;
  body: string;
  id: number
}

const id = useRoute().params.id

const { data } = useLazyFetch(`/api/post/${id}`, {
  key: `post-${id}`,
  server: false,
  default: () => {
    return (useNuxtData().posts as Post[]).find(d => d.id === +id)
  }
})
```

- Update the UI while the query is being invalidated in the background (writing to cache): 

```ts
// in List.vue
const { data } = useLazyFetch('/api/notes', {
  key: 'notes'
})

// In Form.vue
const text = ref('')

const add = () => {
  (useNuxtData().notes as string[]).push(text.value)
  // ... Some mutation
  refreshNuxtData('notes')
}
```

### Additional information

- [X] Could this feature be implemented as a module?

### Final checks

mahdiboomeri:
### 🔗 Linked issue

resolves #9260 
resolves #7568

### ❓ Type of change

### 📚 Description
It introduces a new `useNuxtData` composable.

### 📝 Checklist


Atinux:
I really like this idea of optimistic UI.

I would personally rename it to `getNuxtData()`.

Happy to get feedback also from other team member

pi0:
Related PR to access data: #7569

I think `useNuxtData` makes sense (comparing to `useData()` in other PR and `getNuxtData()`). We use `use*` prefix for composables that use implicit context/CAPI. And also `nuxt` in the name makes sense.

Comparing to #7569:
- We could support `key` param to get one data
- We should return a ref IMO to allow changing value by key

mahdiboomeri:
Thanks for looking into this! 
I've updated the PR base on @pi0's suggestion

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -268,6 +268,13 @@ export function useLazyAsyncData<
   return useAsyncData(key, handler, { ...options, lazy: true }, null)
 }
 
+export function useNuxtData (key: string): Ref<any>
+export function useNuxtData (): Record<string, any>
+export function useNuxtData (key?: string): unknown {
+  const nuxt = useNuxtApp()
+  return key ? ref(nuxt.payload.data[key]) : nuxt.payload.data
```
pi0, line 275: - We should use [`toRef](https://vuejs.org/api/reactivity-utilities.html#toref) in order to convert and keep reactivity
- For consistency, we may always return a ref (using `toRefs` when no key is provided).
- (please check) I think we might need to initialize value on first use call when key is not already set

mahdiboomeri: I've implemented the first the third point, about the second one are we aiming for `Ref<Record<string, any>>`? in that case `toRefs` only converts the properties and not the base object. base on my test wrapping `nuxt.payload.data` with `ref` still keeps the reactivity.

pi0: Using toRefs allows destructuring indeed. So we can have `const { todos } useNuxtData()` and `todos` remains reactive.

docs/content/1.docs/3.api/1.composables/use-nuxt-data.md
```diff
@@ -6,7 +6,7 @@
 
 ```ts
 useNuxtData<T = any> (key: string): Ref<T>
-useNuxtData():‌ Recod<string, any>
+useNuxtData():‌ Ref<Recod<string, any>>
```
danielroe, line 9:
```suggestion
useNuxtData():‌ Ref<Record<string, any>>
```

pi0 (approved):
💯

packages/nuxt/src/app/composables/asyncData.ts
pi0: (let's go with explicit key support only)

docs/content/1.docs/3.api/1.composables/use-nuxt-data.md
```diff
@@ -0,0 +1,69 @@
...
+const previousTodos = ref([])
+
+const { data } = await useFetch('/api/addTodo', {
+  key: 'addTodo',
+  method: 'post',
+  body: {
+    todo: newTodo.value
+  },
+  onRequest () {
+    previousTodos.value = useNuxtData('todos').value // Store the previously cached value to restore if fetch fails.
```
danielroe, line 58: we shouldn't call this here or it will throw an error - it should be called outside the scope of this function

packages/nuxt/src/app/composables/asyncData.ts
```diff
@@ -268,6 +268,19 @@ export function useLazyAsyncData<
   return useAsyncData(key, handler, { ...options, lazy: true }, null)
 }
 
+export function useNuxtData<DataT = any> (key: string): { data: Ref<DataT | null> } {
+  const nuxt = useNuxtApp()
+
+  // Initialize value when key is not already set
+  if (nuxt.payload.data[key] === undefined) {
```
danielroe, line 275:
```suggestion
  if (!(key in nuxt.payload.data)) {
```

danielroe (approved):
Thanks ❤️

pi0:
Thanks again dear @mahdiboomeri <3 A short explainer of recent changes:

- `useNuxtData` requires an explicit key. This makes usage easier and also explicit for keys (excluding auto generated keys). Also allows more flexibility to change the implementation in the future.
- Return value of `useNuxtData` is `{ data: Ref }`. This makes it consistent with other data composables and also ensures to have upward compatibility to add any other fields related to data 
- Docs updated as daniel suggested to cache value of `useNuxtData()` outside of handler. However we restore context and it is always available on client-side, it cannot always work in async conditions where nuxt context is lost. It also makes examples easier to read as we call it once.

PR #9322, opened 2022-11-25
docs: use `layout` in example of `definePageMeta`

falko-apheris:
The main example showed the usage of the `title` parameter which is not part of the `definePageMeta` interface. The example should reflect real values since in the current example it easy to confuse it with `useHead` which must actually be used to set the page title using the `title` parameter.

### ❓ Type of change

### 📝 Checklist


danielroe (approved):
good suggestion 👍

PR #9472, opened 2022-12-02
fix(nuxt): update head entry when speculation rules are updated

danielroe:
### ❓ Type of change

### 📚 Description

Previously new crossorigin prefetches were not being injected into the head as it was not reactive outside a vue component. This PR fixes that by manually calling `patch` when necessary.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/plugins/cross-origin-prefetch.client.ts
+++ b/packages/nuxt/src/app/plugins/cross-origin-prefetch.client.ts
@@ -4,26 +4,31 @@ import { defineNuxtPlugin, useHead } from '#app'
 
 export default defineNuxtPlugin((nuxtApp) => {
   const externalURLs = ref(new Set<string>())
-  useHead({
-    script: [
-      () => ({
-        type: 'speculationrules',
-        innerHTML: JSON.stringify({
-          prefetch: [
-            {
-              source: 'list',
-              urls: [...externalURLs.value],
-              requires: ['anonymous-client-ip-when-cross-origin']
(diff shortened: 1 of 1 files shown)


packages/nuxt/src/app/plugins/cross-origin-prefetch.client.ts
```diff
@@ -4,26 +4,31 @@ import { defineNuxtPlugin, useHead } from '#app'
...
-              source: 'list',
-              urls: [...externalURLs.value],
-              requires: ['anonymous-client-ip-when-cross-origin']
-            }
-          ]
-        })
+  function generateRules () {
+    return {
+      type: 'speculationrules',
+      key: 'speculationrules',
```
harlan-zw, line 10: in theory, the `key` shouldn't be needed if this is client only, but it doesn't hurt

packages/nuxt/src/app/plugins/cross-origin-prefetch.client.ts
```diff
@@ -4,26 +4,31 @@ import { defineNuxtPlugin, useHead } from '#app'
...
+      innerHTML: JSON.stringify({
```
harlan-zw, line 11: should probably add support for objects to be stringified automatically :thinking:

harlan-zw (approved):

harlan-zw:
TIL nav speculation API is a thing, looks quite interesting

pi0 (approved):

PR #9475, opened 2022-12-02
fix(nuxt): defer render-blocking prefetches until after load

danielroe:
### ❓ Type of change

### 📚 Description

Two related performance bugs:

1. When loading a page with links, these pages are prefetched immediately, which will delay the DOM Content Loaded event. This PR defers initialising observer until component is mounted, and additionally waits until the entire app is hydrated.

2. If we are rendering a page with inlined styles, vite will immediately inject a preload for the very same CSS chunk that is inlined; we can use control of the chunk preloader to skip loading those chunks.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/app/components/nuxt-link.ts
+++ b/packages/nuxt/src/app/components/nuxt-link.ts
@@ -213,7 +213,7 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
               })
             }
             if (nuxtApp.isHydrating) {
-              nuxtApp.hook('app:suspense:resolve', registerCallback)
+              nuxtApp.hooks.hookOnce('app:suspense:resolve', registerCallback)
             } else {
               registerCallback()
             }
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/app/components/nuxt-link.ts
```diff
@@ -193,23 +193,30 @@ export function defineNuxtLink (options: NuxtLinkOptions) {
...
+                      nuxtApp.hooks.callHook('link:prefetch', to.value as string).catch(() => {}),
+                      !isExternal.value && preloadRouteComponents(to.value as string, router).catch(() => {})
+                    ])
+                    prefetched.value = true
+                  })
+                }
+              })
+            }
+            if (nuxtApp.isHydrating) {
+              nuxtApp.hook('app:suspense:resolve', registerCallback)
```
pi0, line 216: Can it be hookOnce?

packages/vite/src/client.ts
```diff
@@ -77,8 +88,8 @@ export async function buildClient (ctx: ViteBuildContext) {
   // We want to respect users' own rollup output options
   clientConfig.build!.rollupOptions = defu(clientConfig.build!.rollupOptions!, {
     output: {
-      chunkFileNames: ctx.nuxt.options.dev ? undefined : withoutLeadingSlash(join(ctx.nuxt.options.app.buildAssetsDir, '[name].[hash].js')),
-      entryFileNames: ctx.nuxt.options.dev ? 'entry.js' : withoutLeadingSlash(join(ctx.nuxt.options.app.buildAssetsDir, '[name].[hash].js'))
+      chunkFileNames: ctx.nuxt.options.dev ? undefined : withoutLeadingSlash(join(buildAssetsDir, '[name].[hash].js')),
```
pi0, line 91: Since we are already normalizing leading slash, does using other var makes a difference? (i prefer ctx.* usage as it keeps code more portable and tracable)

pi0 (approved):

PR #9617, opened 2022-12-09
feat(head): useSeoMeta composable

harlan-zw:
### ❓ Type of change

### 📚 Description

With the migration to @vueuse/head v1, there are new composables available that may be useful for Nuxt users. Specifically the `useSeoMeta` composable has been requested.

You can see details on it here: https://unhead.harlanzw.com/guide/guides/useseometa

<details>
 <summary>No longer relevant</summary>
Because the meta that is injected is only needed for robots, we can safely remove the function from the production client build using AST parsing to save around ~2kb from the client build.

You may want to review this AST parsing, maybe there is an easier way to handle this in Nuxt? Ideally, we'd be able to tell unimport to just remove a composable in certain environments. See https://github.com/unjs/unhead/blob/main/packages/addons/src/unplugin/TreeshakeServerComposables.ts

The same AST parsing is used to get rid of the useServerHead composable, which I've included. I'll need to document this separately in a new PR.
</details>

### 📝 Checklist


packages/nuxt/src/head/module.ts
```diff
@@ -17,6 +18,21 @@ export default defineNuxtModule({
...
+      from: '@vueuse/head',
+      imports: [
+        'useSeoMeta',
+        'injectHead',
+        'useServerHead'
+      ]
+    })
+
+    // removes useSeoMeta, useServerHead in production client builds
+    addVitePlugin(UnheadVite({ root: nuxt.options.rootDir }), {
```
pi0, line 31: We might need to support options._layers[].rootDir

harlan-zw:
Thanks for the review! Will update this over the next couple of days

harlan-zw:
I've simplified this PR:
- Removed the Vite plugin, I don't think the AST parsing is worth the cost.
- Removed useServerHead / injectHead auto-imports, will introduce in a separate PR at a later point

Having had users use `useSeoMeta` with the [nuxt-unhead](https://github.com/harlan-zw/nuxt-unhead) module, it seems they get confused when meta doesn't change between pages. So I've re-considered having it as server only, I've documented how they can optimise the client bundle themselves.

Ideally, `useSeoMeta` would act as a macro, similar to https://github.com/danielroe/magic-regexp. Just swapping out the code with the equivalent `useHead` code. But I'll need to work on that and introduce it later.

harlan-zw:
Have added support for providing `title` and `titleTemplate` within this composable which avoids the user having to use `useHead` just to set a title.

PR #9627, opened 2022-12-09
fix(nuxt): avoid injecting url helpers into `globalThis`

### LINKED ISSUE #6090: Running 2 Nuxt handlers (different projects) in the same environment mixes static urls
### Environment

- Package Manager:  `npm@8.6.0`
- Builder:          `vite`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

I've determined where the issue stems from. A reproduction would need deployments to AWS Lambda, but I could provide a Serverless Framework project if required.

### Describe the bug

I found myself trying to run 2 Nuxt handlers in the same AWS Lambda environment:

```node
// My custom AWS lambda handler.js

import tldjs from 'tldjs';
import { handler as nuxtHandlerA } from '../clients/A/.output/server/index.mjs';
import { handler as nuxtHandlerB } from '../clients/B/.output/server/index.mjs';

export async function handler(event, context) {

  const subdomain = tldjs.getSubdomain(event.requestContext?.domainName);
  const clientType = await clientTypeFromSubdomain(subdomain) // Arbitrary function that maps a subdomain to a clientType

  if (clientType === 'A') return await nuxtHandlerA(event, context);
  if (clientType === 'B') return await nuxtHandlerB(event, context);

  return {
    statusCode: 404,
  };
}
```

Both `.output` folders were built with the `'aws-lambda'` nitro preset and `serveStatic: true`.

The problem I'm facing is that the static path of `clientA` is getting overwritten by the static path of `clientB`, **at runtime**, resulting in requests to assets in `clientA` failing due to files not found (because they're being looked up in  the wrong static path).

**I've found that this is due to the static path being derived from `globalThis._importMeta_.url`**, where the `server/index.mjs` path for a project is stored. Being a global state, it gets overwritten up to the last imported Nuxt project.

This is where the project's `server/index.mjs` path is set globally:
```node
// .output/server/index.mjs
...
globalThis._importMeta_={url:import.meta.url,env:process.env}
export { h as handler } from './chunks/aws-lambda.mjs';
...
```

This is where the static path is derived from global state to access an asset:
```node
// .output/server/chunks/aws-lambda.mjs
...
function readAsset (id) {
  const serverDir = dirname(fileURLToPath(globalThis._importMeta_.url));
  return promises.readFile(resolve(serverDir, assets[id].path))
}
...
```
---------------------

### Additional context

Honestly I don't know if this is a bug or a feature. In the latter case then I wouldn't know what is the best way to achieve this setup.

- Manually overwriting `globalThis._importMeta_` before each handler call sounds like it would break sooner than later.
- Hosting statics in a CDN would be a good specific solution, but I have no assurance that the rest of the projects don't depend on this global state now or in the future.
- Deploying to separate lambdas is not desired since I still need one as a middleware to decide what project to run based on subdomain. This would add the latency of calling a Nuxt project in another lambda.

Also I might be incurring into the XY problem. [The underlying problem I'm trying to solve is detailed in this discussion](https://github.com/nuxt/framework/discussions/6057)

Thanks for your time :)!

### Logs

_No response_

danielroe:
### 🔗 Linked issue

resolves #6090

### ❓ Type of change

### 📚 Description

On server-side, rather than declaring `__buildAssetsURL` and `__publicAssetsURL` on `globalThis`, this PR relies on nitro auto-imports to ensure they are injected where needed in the server app.

**Note**: We would need to ensure that the nitro auto-imports functionality cannot be turned off if we are relying on it in this context. See also https://github.com/unjs/nitro/pull/742.

### 📝 Checklist


packages/nuxt/src/core/runtime/nitro/renderer.ts
```diff
@@ -13,12 +13,7 @@ import type { NuxtApp, NuxtSSRContext } from '#app'
 // @ts-ignore
 import { appRootId, appRootTag } from '#internal/nuxt.config.mjs'
 // @ts-ignore
-import { buildAssetsURL, publicAssetsURL } from '#paths'
-
-// @ts-ignore
-globalThis.__buildAssetsURL = buildAssetsURL
-// @ts-ignore
-globalThis.__publicAssetsURL = publicAssetsURL
```
pi0, line 21: I think we can keep injecting to globalThis as well

packages/vite/src/server.ts
```diff
@@ -24,11 +24,11 @@ export async function buildServer (ctx: ViteBuildContext) {
           return { relative: true }
         }
         if (type === 'public') {
-          return { runtime: `globalThis.__publicAssetsURL(${JSON.stringify(filename)})` }
+          return { runtime: `__publicAssetsURL(${JSON.stringify(filename)})` }
```
pi0, line 27: Then here we can conditionally render either with globalThis or without based on auto imports being enabled.

pi0:
Although this will heal the issues similar to #6090, both server and client instances are supposed to run in an isolated environment. I'm okay with such change since also helps with vercel-edge but I'm afraid it won't solve the multi-instance issue. Other things including hookable core depend on unique globalThis instance and it is shared.

PR #9630, opened 2022-12-09
refactor(nuxt): do not add all composable auto-imports to `globalThis`

danielroe:
### ❓ Type of change

### 📚 Description

`globals` is turned off by default, and I think we can keep the same behaviour as without it. Otherwise there's extra work done on initial build/warmup for all composables and I'm not sure of the benefit. (I think it is a hold-over from previous default behaviour.)

### 📝 Checklist


pi0:
`global` option was actually added to remove the parsing overhead (to detect usage in dev) but i think transform overhead of loading composable files should be more at least in a small sized project (we never benchmarked AFAIK)

Removing this option makes sense 👍🏼 Only i think we can consider this PR a refactor since is not changing any behavior?

antfu (approved):

PR #9634, opened 2022-12-09
feat(nuxt): allow customising bundle renderer options

danielroe:
### 🔗 Linked issue

https://github.com/nuxt/nuxt.js/issues/14584

### ❓ Type of change

### 📚 Description

This allows an escape hatch for end users to access `shouldPrefetch` and `shouldPreload` hooks in `vue-bundle-renderer`.

### 📝 Checklist


pi0:
I think we can directly support build-time flags in bundle-renderer and use build time hook to customize manifest. (this gives both workaround with current hooks and more consistent solution to what we do by handling most of manifest on build-time)  https://github.com/nuxt-contrib/vue-bundle-renderer/issues/45.

Atinux:
I understand the idea of having a build-time flag for optimisation but still believe this hooks is worth adding for on-demand customisation.

danielroe:
~> https://github.com/nuxt/nuxt/pull/18420

PR #9648, opened 2022-12-10
fix(schema): add hookable dependency

sxzz:
fix the phantom dependency issue

### 🔗 Linked issue

relate #5411

### ❓ Type of change

### 📚 Description

Add `hookable`. It's only added in `nuxt` package. I only need `defineNuxtModule` from `@nuxt/kit`. Then `hookable` pkg is missing.

### 📝 Checklist

### DIFF
--- a/packages/schema/package.json
+++ b/packages/schema/package.json
@@ -16,8 +16,8 @@
   "devDependencies": {
     "@types/lodash.template": "^4",
     "@types/semver": "^7",
-    "@vitejs/plugin-vue": "^3.2.0",
     "@unhead/schema": "^1.0.13",
+    "@vitejs/plugin-vue": "^3.2.0",
     "nitropack": "^1.0.0",
     "unbuild": "latest",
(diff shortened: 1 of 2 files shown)


danielroe:
What is the issue? I'm assuming it's about the types provided by `hookable`, as I think there is no runtime dependency on it?

sxzz:
See here https://github.com/sxzz/unplugin-vue-macros/blob/rollup-3/packages/nuxt/src/index.ts#L20 

I only use `defineNuxtModule` to define a Nuxt module, and `nuxt.hook` referenced `Hookable` finally.

```
	hooks: Hookable<NuxtHooks>;
    hook: Nuxt['hooks']['hook']; // definition of `nuxt.hook`
```

pi0 (approved):
Thanks for PR!

Ideally we should be able to inline external types for schema but since we already added other deps, this looks good to me as fix.

Issue #9653, opened 2022-12-11
nuxt use absolute path when build and generate


flihub:
### Environment

- Operating System: `Windows_NT`
- Node Version:     `v18.12.0`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `pnpm@7.18.1`
- Builder:          `vite`
- User Config:      `css`, `vite`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/nuxt/framework/discussions/9584

### Describe the bug

no one answers in disccusions.

### Additional context

want to hide : "C:/Laragon/www/nuxt/"

PR #9655, opened 2022-12-11
fix(nuxt): remove absolute paths from routes objects

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/9653

### ❓ Type of change

### 📚 Description

Absolute file paths were being embedded in the build by the pages integration in two places - in the meta variable as well as in the `.file` attribute (which is not used by `vue-router` as far as I can tell). We can safely relativise the former and remove the latter.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/pages/utils.ts
+++ b/packages/nuxt/src/pages/utils.ts
@@ -223,11 +223,12 @@ function prepareRoutes (routes: NuxtPage[], parent?: NuxtPage) {
 }
 
 export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> = new Set()): { imports: Set<string>, routes: string } {
+  const nuxt = useNuxt()
   return {
     imports: metaImports,
     routes: genArrayFromRaw(routes.map((route) => {
       const file = normalize(route.file)
-      const metaImportName = genSafeVariableName(file) + 'Meta'
+      const metaImportName = genSafeVariableName(relative(nuxt.options.rootDir, file)) + 'Meta'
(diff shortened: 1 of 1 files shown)


manniL (approved):

packages/nuxt/src/pages/utils.ts
```diff
@@ -223,28 +223,34 @@ function prepareRoutes (routes: NuxtPage[], parent?: NuxtPage) {
...
 export function normalizeRoutes (routes: NuxtPage[], metaImports: Set<string> = new Set()): { imports: Set<string>, routes: string } {
+  const nuxt = useNuxt()
   return {
     imports: metaImports,
-    routes: genArrayFromRaw(routes.map((route) => {
-      const file = normalize(route.file)
-      const metaImportName = genSafeVariableName(file) + 'Meta'
+    routes: genArrayFromRaw(routes.map((page) => {
+      const file = normalize(page.file)
+      const metaImportName = genSafeVariableName(relative(nuxt.options.rootDir, file)) + 'Meta'
```
pi0, line 231: Any reason not using `srcDir`?

pi0: It is mainly used to generate a variable name (and `src` would be extra and change output behavior). Are we generating var names with rootDir relative elsewhere?

danielroe: Yes, we are doing the same for plugins.

pi0 (approved):
LGTM. (resolving from srcDir to be fixed later together with other places)

PR #9664, opened 2022-12-12
feat(nuxt): deep watch `useCookie` ref value by default

xtoolkit:
### ❓ Type of change

### 📚 Description

`useCookie` only change when `.value` update. we can use deep flag in watch and track all change.

### 📝 Checklist

### DIFF
--- a/docs/content/1.docs/3.api/1.composables/use-cookie.md
+++ b/docs/content/1.docs/3.api/1.composables/use-cookie.md
@@ -140,11 +140,37 @@ Specifies the `boolean` or `string` value for [watch](https://vuejs.org/api/reac
 - `shallow` - Will watch cookie ref data changes for only top level properties
 - `false` Will not watch cookie ref data changes.
 
+**Example 1:**
+
+```vue
+<template>
+  <div>User score: {{ user?.score }}</div>
+</template>
+
+<script setup>
+const user = useCookie(
+  'userInfo',
+  {
+    default: () => ({ score: -1 }),
+    watch: false
+  }
+)
+
+if (user.value && user.value !== null) {
+  user.value.score++; // userInfo cookie not update with this change
+}
+</script>
+```
+
+**Example 2:**
+
 ```vue
 <template>
   <div>
-    <h1>List: {{ list }}</h1>
+    <h1>List</h1>
+    <pre>{{ list }}</pre>
     <button @click="add">Add</button>
+    <button @click="save">Save</button>
   </div>
 </template>
 
@@ -159,6 +185,14 @@ const list = useCookie(
 
 function add() {
   list.value?.push(Math.round(Math.random() * 1000))
+  // list cookie not update with this change
+}
+
+function save() {
+  if (list.value && list.value !== null) {
+    list.value = [...list.value]
+    // list cookie update with this change
+  }
 }
 </script>
 ```
--- a/docs/content/1.docs/3.api/1.composables/use-cookie.md
+++ b/docs/content/1.docs/3.api/1.composables/use-cookie.md
@@ -138,10 +138,13 @@ be returned as the cookie's value.
 
 Specifies a function that returns the cookie's default value. The function can also return a `Ref`.
 
-### `deepWatch`
+### `watch`
 
-Specifies the `boolean` value for the deep watch Cookie ref. When truthy,
-Cookie update on deep data changed; otherwise it is not. By default, you need reassign your data.
+Specifies the `boolean` or `string` value for [watch](https://vuejs.org/api/reactivity-core.html#watch) cookie ref data.
+
+- `true` or `shallow` will watch cookie ref data.
+- `false` will not watch cookie data.
+- `deep` will watch cookie ref data and data sources change.
 
 ```vue
 <template>
--- a/docs/content/1.docs/3.api/1.composables/use-cookie.md
+++ b/docs/content/1.docs/3.api/1.composables/use-cookie.md
@@ -25,16 +25,10 @@ The example below creates a cookie called `counter`. If the cookie doesn't exist
 ```vue
 <template>
   <div>
-    <h1> Counter: {{ counter || '-' }}</h1>
-    <button @click="counter = null">
-      reset
-    </button>
-    <button @click="counter--">
-      -
-    </button>
-    <button @click="counter++">
-      +
-    </button>
+    <h1>Counter: {{ counter || '-' }}</h1>
+    <button @click="counter = null">reset</button>
+    <button @click="counter--">-</button>
+    <button @click="counter++">+</button>
   </div>
 </template>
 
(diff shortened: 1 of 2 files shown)


packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -32,7 +33,9 @@ export function useCookie <T = string | null> (name: string, _opts?: CookieOptio
   const cookie = ref<T | undefined>(cookies[name] as any ?? opts.default?.())
 
   if (process.client) {
-    watch(cookie, () => { writeClientCookie(name, cookie.value, opts as CookieSerializeOptions) })
+    watch(cookie, () => { writeClientCookie(name, cookie.value, opts as CookieSerializeOptions) }, {
+      deep: opts.deepWatch
```
pi0, line 37: I think it makes sense to be default behavior as a fix. We can support `watch: true | false | 'deep' | 'shallow'` too.

xtoolkit: Did you mean `useCookie` options? If yes, what's different between `true` and `shallow` in this case?

| watch | mean |
| :-----: | :-----: |
| false | don't watch |
| true | shallow watch |
| deep | deep watch |
| shallow | ? |

pi0: They can be aliases or we can simply support `false | deep | shallow`.

xtoolkit:
Sorry for my bad English. please rewrite document ...

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -32,7 +33,12 @@ export function useCookie <T = string | null> (name: string, _opts?: CookieOptio
   const cookie = ref<T | undefined>(cookies[name] as any ?? opts.default?.())
 
   if (process.client) {
-    watch(cookie, () => { writeClientCookie(name, cookie.value, opts as CookieSerializeOptions) })
+    const callback = () => { writeClientCookie(name, cookie.value, opts as CookieSerializeOptions) }
+    if (opts.watch) {
```
pi0, line 37: to avoid breaking change. keeping watch enabled by default.

```suggestion
    if (opts.watch !== false) {
```

packages/nuxt/src/app/composables/cookie.ts
```diff
@@ -15,6 +15,7 @@ export interface CookieOptions<T = any> extends _CookieOptions {
   decode?(value: string): T
   encode?(value: T): string
   default?: () => T | Ref<T>
+  watch?: boolean | 'deep' | 'shallow'
```
pi0, line 18:
```suggestion
  watch?: false | 'deep' | 'shallow'
```

docs/content/1.docs/3.api/1.composables/use-cookie.md
```diff
@@ -138,6 +138,35 @@ be returned as the cookie's value.
 
 Specifies a function that returns the cookie's default value. The function can also return a `Ref`.
 
+### `watch`
+
+Specifies the `boolean` or `string` value for [watch](https://vuejs.org/api/reactivity-core.html#watch) cookie ref data.
+
+- `true` or `shallow` will watch cookie ref data.
```
pi0, line 145:
```suggestion
- `shallow` will watch cookie ref data.
```

pi0:
@danielroe Do you agree to making default behavior deep watching cookie value?

danielroe:
Agreed - I think deep behaviour is more intuitive. 👍

packages/nuxt/src/app/composables/cookie.ts
xtoolkit: Why not set in `CookieDefaults`?
```ts
const CookieDefaults: CookieOptions<any> = {
  path: '/',
  decode: val => destr(decodeURIComponent(val)),
  encode: val => encodeURIComponent(typeof val === 'string' ? val : JSON.stringify(val)),
  watch: true
}
```

packages/nuxt/src/app/composables/cookie.ts
xtoolkit: Why not set in CookieDefaults?

const CookieDefaults: CookieOptions<any> = {
  path: '/',
  decode: val => destr(decodeURIComponent(val)),
  encode: val => encodeURIComponent(typeof val === 'string' ? val : JSON.stringify(val)),
  watch: true
}

docs/content/1.docs/3.api/1.composables/use-cookie.md
```diff
@@ -138,6 +132,37 @@ be returned as the cookie's value.
...
+  </div>
+</template>
+
+<script setup>
+const list = useCookie(
+  'list',
+  { default: () => [] },
+  {
+    watch: 'shallow'
+  }
```
danielroe, line 157:
```suggestion
  {
    default: () => [],
    watch: 'shallow'
  }
```

docs/content/1.docs/3.api/1.composables/use-cookie.md
```diff
@@ -138,6 +132,37 @@ be returned as the cookie's value.
 
 Specifies a function that returns the cookie's default value. The function can also return a `Ref`.
 
+### `watch`
+
+Specifies the `boolean` or `string` value for [watch](https://vuejs.org/api/reactivity-core.html#watch) cookie ref data.
+
+- `true` - Will watch cookie ref data changes and it's nested properties. (default)
```
danielroe, line 139:
```suggestion
- `true` - Will watch cookie ref data changes and its nested properties. (default)
```

docs/content/1.docs/3.api/1.composables/use-cookie.md
```diff
@@ -138,6 +132,37 @@ be returned as the cookie's value.
...
+    <button @click="add">Add</button>
+  </div>
+</template>
+
+<script setup>
+const list = useCookie(
+  'list',
+  {
+    default: () => [],
+    watch: 'shallow'
```
xtoolkit, line 156: In this example I was looking to show how enable deep watch in `useCookie`.
If `useCookie` watch deep by default, we don't need change `watch` field. so i think this example not good.

danielroe: yes, that makes sense given the example! 😆 can you think of a better one?

pi0: An example would be reading cookie value using `useCookie` but not wanting to watch it for performance reasons. Otherwise probably are not much examples. `shallow` is mainly to have an escape hatch.

xtoolkit: Hello again
Do you think this example is good?

pi0 (approved):
Thanks! LGTM ❤️

PR #9759, opened 2022-12-16
refactor(vite): show log when client build is starting

antfu:
### ❓ Type of change

### 📚 Description

As the app getting big, the building on the client does not show any state and is stuck on this line, which could be confusing. We have `Building server...` indicator, I think we should do the same for client build.

image

### 📝 Checklist


pi0 (approved):
👍🏼

Atinux (approved):

danielroe (approved):

PR #9771, opened 2022-12-17
feat(kit): add `extendRouteRules` method

xtoolkit:
### ❓ Type of change

### 📚 Description

In nuxt module develop with nuxt/kit, `extendRouteRules` method can be help add custom routeRules config.

### 📝 Checklist


packages/kit/src/pages.ts
```diff
@@ -1,4 +1,5 @@
 import type { NuxtHooks } from '@nuxt/schema'
+import { NitroRouteConfig } from 'nitropack'
```
danielroe, line 2:
```suggestion
import type { NitroRouteConfig } from 'nitropack'
```

packages/kit/src/pages.ts
```diff
@@ -11,3 +12,15 @@ export function extendPages (cb: NuxtHooks['pages:extend']) {
...
+export function extendRouteRules (route: string, rule: NitroRouteConfig) {
+  const nuxt = useNuxt()
+  if (!nuxt.options.routeRules) {
+    nuxt.options.routeRules = {}
+  }
+  nuxt.options.routeRules[route] = rule
+  if (!nuxt.options.nitro.routeRules) {
+    nuxt.options.nitro.routeRules = {}
+  }
+  nuxt.options.nitro.routeRules[route] = rule
```
danielroe, line 25: should we merge rather than entirely overriding the rule?

manniL: I'd prefer merging too. 
But maybe we could get the current route rule object (if existing) and alter it e.g. via function?

xtoolkit: We can add option to merge rules like this:
```ts
function extendRouteRules (route: string, rule: NitroRouteConfig, merge: boolean = true)
```
or
```ts
function extendRouteRules (route: string, rule: NitroRouteConfig, override: boolean = true)
```

pi0: `{ ovrride: true }` as 3rd param makes sense. When disabled (default) user config has priority for merging object. When enabled, it can override user config.

danielroe: Agreed 👍

xtoolkit: hi
check again please

danielroe (changes requested):
can we support, instead of a boolean flag as third argument, an options object with an `override` property?

danielroe (approved):
Thanks! Sorry for the back-and-forth. This looks good. ❤️

packages/kit/src/pages.ts
```diff
@@ -11,3 +13,28 @@ export function extendPages (cb: NuxtHooks['pages:extend']) {
...
+  override?: boolean
+}
+
+export function extendRouteRules (route: string, rule: NitroRouteConfig, options: ExtendRouteRulesOptions = {}) {
+  const nuxt = useNuxt()
+  if (!nuxt.options.routeRules) {
+    nuxt.options.routeRules = {}
+  }
+  nuxt.options.routeRules[route] = options.override
+    ? rule
```
pi0, line 32: I belive in override mode, we still can keep existing extra rules but give the `rule` more priority than current rules

xtoolkit: Can you write a example? I did't understand!

packages/kit/src/pages.ts
```diff
@@ -11,3 +13,28 @@ export function extendPages (cb: NuxtHooks['pages:extend']) {
...
+    : defu(rule, nuxt.options.routeRules[route] || {})
```
danielroe, line 33: I think this is what @pi0 is suggesting:

```suggestion
  nuxt.options.routeRules[route] = options.override
    ? defu(rule, nuxt.options.routeRules[route])
    : defu(nuxt.options.routeRules[route], rule)
```

xtoolkit: I thing in this example, override doesn't mean.
I thing remove override option is better! Or I misunderstood :(

packages/kit/src/pages.ts
```diff
@@ -11,3 +13,28 @@ export function extendPages (cb: NuxtHooks['pages:extend']) {
...
+    nuxt.options.routeRules = {}
+  }
+  nuxt.options.routeRules[route] = options.override
+    ? defu(rule, nuxt.options.routeRules[route])
+    : defu(nuxt.options.routeRules[route], rule)
+  if (!nuxt.options.nitro.routeRules) {
+    nuxt.options.nitro.routeRules = {}
+  }
+  nuxt.options.nitro.routeRules[route] = options.override
+    ? rule
```
danielroe, line 38:
```suggestion
    ? defu(rule, nuxt.options.routeRules[route])
    : defu(nuxt.options.routeRules[route], rule)
```

packages/kit/src/pages.ts
```diff
@@ -11,3 +13,28 @@ export function extendPages (cb: NuxtHooks['pages:extend']) {
...
+  }
+  nuxt.options.routeRules[route] = options.override
+    ? defu(rule, nuxt.options.routeRules[route])
+    : defu(nuxt.options.routeRules[route], rule)
+  if (!nuxt.options.nitro.routeRules) {
+    nuxt.options.nitro.routeRules = {}
+  }
+  nuxt.options.nitro.routeRules[route] = options.override
+    ? defu(rule, nuxt.options.nitro.routeRules[route])
+    : defu(rule, nuxt.options.nitro.routeRules[route] || {})
```
danielroe, line 39:
```suggestion
    : defu(nuxt.options.nitro.routeRules[route], rule)
```

packages/kit/src/pages.ts
```diff
@@ -11,3 +13,28 @@ export function extendPages (cb: NuxtHooks['pages:extend']) {
...
+  }
+  nuxt.options.routeRules[route] = options.override
+    ? defu(rule, nuxt.options.routeRules[route])
+    : defu(nuxt.options.routeRules[route], rule)
+  if (!nuxt.options.nitro.routeRules) {
+    nuxt.options.nitro.routeRules = {}
+  }
+  nuxt.options.nitro.routeRules[route] = options.override
+    ? defu(rule, nuxt.options.nitro.routeRules[route])
+    : defu(nuxt.options.nitro.routeRules[route], rule)
```
danielroe, line 39: feels like we could optimise slightly:

```suggestion
  for (const options of [nuxt.options, nuxt.options.nitro]) {
	  if (!options.routeRules) {
	    options.routeRules = {}
	  }
	  options.routeRules[route] = options.override
	    ? defu(rule, options.routeRules[route])
	    : defu(options.routeRules[route], rule)
  }
```

pi0 (approved):
Thanks 💯

PR #9775, opened 2022-12-18
docs(getting-started): improve usage of `pnpm`

AuroraTea:
### ❓ Type of change

#### changes
1, Add `--shamefully-hoist` for `pnpm install`.
2, Omit `run` for pnpm script.

#### why
1, For new project, `--shamefully-hoist` is more comfortable than `.npmrc` with `shamefully-hoist=true`. And `#` does not affect the copying of the entire shell run.
2, In terms of reading order, --shamefullu-hoist is written after the shell command, and although alert is visually obvious, there is a chance that it will be missed.
3, Since yarn omits run, pnpm can actually omit run as well😜.

### 📝 Checklist

### DIFF
--- a/docs/content/1.docs/1.getting-started/2.installation.md
+++ b/docs/content/1.docs/1.getting-started/2.installation.md
@@ -78,7 +78,7 @@ npm install
 ```
 
 ```bash [pnpm]
-pnpm install #Notice the note below
+pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
 ```
 
 ::
--- a/docs/content/1.docs/1.getting-started/2.installation.md
+++ b/docs/content/1.docs/1.getting-started/2.installation.md
@@ -78,13 +78,13 @@ npm install
 ```
 
 ```bash [pnpm]
-pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
+pnpm install #Notice the note below
 ```
 
 ::
 
 ::alert
-**Note:** If using **pnpm**, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`. Or add `--shamefully-hoist` directly when `pnpm i`.
+**Note:** If using **pnpm**, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`. Otherwise, you have to add `--shamefully-hoist` when running `pnpm i`.
 ::
 
 ## Development Server
--- a/docs/content/1.docs/1.getting-started/2.installation.md
+++ b/docs/content/1.docs/1.getting-started/2.installation.md
@@ -78,7 +78,8 @@ npm install
 ```
 
 ```bash [pnpm]
-pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
+# Make sure you have `shamefully-hoist=true` in `.npmrc` before running pnpm install
+pnpm install
 ```
 
 ::
--- a/docs/content/1.docs/1.getting-started/2.installation.md
+++ b/docs/content/1.docs/1.getting-started/2.installation.md
@@ -78,13 +78,13 @@ npm install
 ```
 
 ```bash [pnpm]
-pnpm install
+pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
 ```
 
 ::
 
 ::alert
-**Note:** If using **pnpm**, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`.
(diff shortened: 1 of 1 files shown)


docs/content/1.docs/1.getting-started/2.installation.md
```diff
@@ -78,13 +78,13 @@ npm install
...
 ```bash [pnpm]
-pnpm install
+pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
 ```
 
 ::
 
 ::alert
-**Note:** If using **pnpm**, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`.
+**Note:** If using **pnpm**, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`. Or add `--shamefully-hoist` directly when `pnpm i`.
```
manniL, line 87:
```suggestion
**Note:** If using **pnpm**, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`. Otherwise, you have to add `--shamefully-hoist` when running `pnpm i`.
```

AuroraTea: Thank you for your better English!~ But I have some additional thoughts here.

Would  _Otherwise, you have to add `--shamefully-hoist` when running `pnpm i`._  be better?

Because it may feel easier to add `--shamefully-hoist` when running `pnpm i` than `.npmrc` with `shamefully-hoist=true` from the text before and after this part of the documentation.

Maybe _If using **pnpm**, you have to add `--shamefully-hoist` when running `pnpm i`. Otherwise, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`._ is more better?

HigherOrderLogic: > Because it may feel easier to add `--shamefully-hoist` when running `pnpm i` than `.npmrc` with `shamefully-hoist=true` from the text before and after this part of the documentation.

I disagree. Having `shamefully-hoist=true` in `.npmrc` is a lots better than having to include `--shamefully-hoist` every time you run `pnpm i`.

manniL: I'd also prefer a `.npmrc` as the "preferred" method because it is less error prone (forgetting the parameter / consistency throughout platforms etc. etc.).

AuroraTea: You are right! Now I deeply agree with both of you.

docs/content/1.docs/1.getting-started/2.installation.md
```diff
@@ -78,13 +78,13 @@ npm install
 ```
 
 ```bash [pnpm]
-pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
+pnpm install #Notice the note below
```
manniL, line 81: I'd leave the info that was here before though. ☺️ 
```suggestion
pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
```

docs/content/1.docs/1.getting-started/2.installation.md
```diff
@@ -78,13 +78,13 @@ npm install
...
 ```bash [pnpm]
-pnpm install
+pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
 ```
 
 ::
 
 ::alert
-**Note:** If using **pnpm**, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`.
+**Note:** If using **pnpm**, make sure to have `.npmrc` with `shamefully-hoist=true` inside it before `pnpm install`. Otherwise, you have to add `--shamefully-hoist` when running `pnpm i`.
```
pi0, line 87: Since usage of flag won't remain on fresh lock files, and also makes detection harder, i think docs shall remain mentioning npmrc method only.

docs/content/1.docs/1.getting-started/2.installation.md
```diff
@@ -78,13 +78,13 @@ npm install
 ```
 
 ```bash [pnpm]
-pnpm install
+pnpm install --shamefully-hoist # or shamefully-hoist=true in .npmrc
```
pi0, line 81:
```suggestion
# Make sure you have `shamefully-hoist=true` in `.npmrc` before running pnpm install
pnpm install
```

docs/content/1.docs/1.getting-started/2.installation.md
AuroraTea: I think you have a point. 

Also, I was wondering whether or not to introduce the thing that `pnpm dlx nuxi init` comes with `.npmrc` with `shamefully-hoist=true` compared to `npx nuxi init`

pi0 (approved):
Thanks <3

PR #9832, opened 2022-12-22
fix(nuxt): set is loading state for  `<NuxtLoadingIndicator>` after throttle

Arecsu:
### ❓ Type of change

### 📚 Description

Previously, animation will always trigger despite throttle being declared due to `isLoading.value` being `true` outside of the `setTimeout` function. Now the loading indicator only shows up if the page takes more than the configured throttle time, which is the intended purpose of it. Default is `200ms`. Setting it to 0 will always trigger the loading animation.


antfu (approved):

pi0 (approved):

pi0:
Thanks ❤️

PR #9854, opened 2022-12-25
Use `NuxtLayout`

datamadsen:
After introducing `layouts/default.vue`, `pages/index.vue` and `pages/about.vue` can use `<NuxtLayout>...</NuxtLayout>` to take advantage of the layout which was not in the documentation.


manniL (changes requested):
If not "escaping" the chosen layout, one should use `NuxtLayout` in the `app.vue` instead of repeating the component in each subpage ☺️

datamadsen:
> If not "escaping" the chosen layout, one should use `NuxtLayout` in the `app.vue` instead of repeating the component in each subpage ☺️

Thanks - I removed the pages files and added a modified app.vue listing.

manniL (approved):

docs/content/1.docs/1.getting-started/3.views.md
```diff
@@ -110,25 +110,16 @@ If you only have a single layout in your application, we recommend using app.vue
...
+```vue [app.vue]
 <template>
   <div>
-    <h1>Welcome to the homepage</h1>
-    <AppAlert>
-      This is an auto-imported component
-    </AppAlert>
+    <NuxtLayout>
+      <NuxtPage />
+    </NuxtLayout>
```
danielroe, line 118: if no `app.vue` is present, then the default Nuxt `app.vue` is used, which is:

```
<template>
  <NuxtLayout>
    <NuxtPage />
  </NuxtLayout>
</template>
```

That is why it is omitted in this example.

datamadsen: No problem. As a newcomer to nuxt and vue I was just following along in the examples and had a stumble when trying out examples in the layout section because the app.vue from an earlier example didn’t include layouts. The only reason for the pr is to try and minimize the search to an answer for why it didn’t work/how to get it to work, but it could be that it’s not really a problem so it’s fine with me to just close the pr unmerged

PR #9937, opened 2023-01-03
docs(configuration): clarify the location of `app.config.ts` in the source directory

### LINKED ISSUE #9914: [Docs] `app.config.ts` must be in the source directory, which is not necessarily the root
Hi, I see that there is a [closed issue](https://github.com/nuxt/framework/issues/8012) about that, but I still think the documentation is misleading:

The [documentation](https://nuxt.com/docs/getting-started/configuration#app-configuration) clearly says that `app.config.ts` must be located at the root of a project. But as Daniel said in the previous issue, it must be in the `srcDir`, so if we change it in `nuxt.config.ts`, it is no more the root.

Here is what I suggest:
- reword slightly the paragraph in `Get Started > Configuration > App Configuration` (or add a little "info panel" ?) to mention the source directory instead of root
- update the folder structure in the [nuxt config reference](https://nuxt.com/docs/api/configuration/nuxt-config#srcdir) to show the `app.config.ts` in the `src` folder (and maybe `app.vue` too ?)

I would be happy to do the PR if you want

jsulpis:
### 🔗 Linked issue

Resolve #9914 

### ❓ Type of change

### 📚 Description

The current [documentation](https://nuxt.com/docs/getting-started/configuration#app-configuration) says that the `app.config.ts` is in the root directory, but in reality it is the source directory. I slightly reworded the sentence to clarify this.

I also removed a link to a non-existing page in the following section (hope this is not too unrelated).

### 📝 Checklist

### DIFF
--- a/docs/content/1.docs/1.getting-started/3.configuration.md
+++ b/docs/content/1.docs/1.getting-started/3.configuration.md
@@ -69,7 +69,7 @@ const runtimeConfig = useRuntimeConfig()
 
 ## App Configuration
 
-The `app.config.ts` file, also located at the root of a Nuxt project, is used to expose public variables that can be determined at build time. Contrary to the `runtimeConfig` option, these can not be overridden using environment variables.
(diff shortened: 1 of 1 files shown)


pi0 (approved):
💯 Thanks

PR #9946, opened 2023-01-04
fix(nuxi): show an error if no value is supplied for the `--template` flag

jamesgeorge007:
### 🔗 Linked issue
https://github.com/nuxt/framework/issues/1408#issuecomment-1367535701

### ❓ Type of change

- ~[ ] 📖 Documentation (updates to the documentation or readme)~
- ~[ ] 👌 Enhancement (improving an existing functionality like performance)~
- ~[ ] ✨ New feature (a non-breaking change that adds functionality)~
- ~[ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)~

### 📚 Description
Previously not supplying a value with the `--template` flag results in the following error.

image

This PR aims at showing an error message and exiting with a code of `1` in this scenario.

Screenshot 2023-01-04 at 11 18 36 PM

### 📝 Checklist

- ~[ ] I have updated the documentation accordingly.~

### DIFF
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -4,15 +4,6 @@ import { relative } from 'pathe'
 import consola from 'consola'
 import { defineNuxtCommand } from './index'
 
-const resolveTemplate = (template: string | boolean) => {
-  if (typeof template === 'boolean') {
-    consola.error('Please specify a template!')
-    process.exit(1)
-  }
-
-  return template || 'v3'
-}
-
 const rpath = (p: string) => relative(process.cwd(), p)
 
 const DEFAULT_REGISTRY = 'https://raw.githubusercontent.com/nuxt/starter/templates/templates'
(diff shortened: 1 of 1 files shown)


packages/nuxi/src/commands/init.ts
```diff
@@ -16,7 +25,7 @@ export default defineNuxtCommand({
   },
   async invoke (args) {
     // Clone template
-    const template = args.template || args.t || 'v3'
+    const template = resolveTemplate(args.template || args.t)
```
pi0, line 28: can we maybe inline it using `let template` and two ifs? It can keep code portable and also easier to refactor later on cli.

jamesgeorge007: Sure thing, just an extra check for the boolean value would do. I've made the changes.

pi0 (approved):

pi0:
Thanks

jamesgeorge007:
@pi0, I noticed that a failed attempt to download the template results in the stack trace being logged in to the console. May I know if it's intentional? Else, we could handle the error within a `try-catch` block.

https://github.com/nuxt/framework/blob/da38c620005596d71f3ae22f33fb2653312515d0/packages/nuxi/src/commands/init.ts#L26-L32

PR #9952, opened 2023-01-05
fix(nuxt): don't exclude pnpm layers from nitro esbuild/imports

### LINKED ISSUE #9951: Extending nitro server using pnpm fails
### Environment

- Operating System: `Darwin`
- Node Version:     `v16.16.0`
- Nuxt Version:     `3.0.0`
- Nitro Version:    `1.0.0`
- Package Manager:  `pnpm@7.16.1`
- Builder:          `vite`
- User Config:      `extends`, `api`, `hooks`, `sentry`, `tailwindcss`, `runtimeConfig`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

I tried to make a reproduction extending from a simple Github repo but I get a `socket hang up` error...

### Describe the bug

When extending a Nuxt config using an NPM package and due to the way `node_modules` are stored when using Pnpm, the current RegExp used for Nitro's `esbuild.options.exclude` and for `imports.exclude` options won't include/transform the layer's files.

### Additional context

To fix the issue, I had to change the RegExp to `/(?<=\/)(?:.+\.pnpm\/.+\/)?node_modules\/(.+)$/` in: https://github.com/nuxt/framework/blob/da38c620005596d71f3ae22f33fb2653312515d0/packages/nuxt/src/core/nitro.ts#L30 
and override the default `/[\\/]node_modules[\\/]/` exclude pattern of `unimport` to the same RegExp.

### Logs

_No response_

yassilah:
### 🔗 Linked issue
fixes nuxt/nuxt.js#12394 

### ❓ Type of change

### 📝 Checklist


packages/nuxt/src/core/nitro.ts
```diff
@@ -17,6 +17,20 @@ import { ImportProtectionPlugin } from './plugins/import-protection'
...
   // Resolve config
   const _nitroConfig = ((nuxt.options as any).nitro || {}) as NitroConfig
+
+  const excludePattern = new RegExp(
+                `node_modules\\/(?!${nuxt.options._layers
+                    .flatMap((l) => {
+                        return [
+                            l.cwd.match(/(?<=\/)node_modules\/(.+)$/)?.[1],
+                            l.cwd.match(/\.pnpm\/.+\/node_modules\/(.+)$/)?.[1]
+                        ]
```
manniL, line 27: nitpick: you could omit the return here

manniL (approved):

packages/nuxt/src/core/nitro.ts
```diff
@@ -17,6 +17,18 @@ import { ImportProtectionPlugin } from './plugins/import-protection'
 export async function initNitro (nuxt: Nuxt & { _nitro?: Nitro }) {
   // Resolve config
   const _nitroConfig = ((nuxt.options as any).nitro || {}) as NitroConfig
+
+  const excludePattern = new RegExp(
+    `node_modules\\/(?!${nuxt.options._layers
+      .flatMap(l => [
+        l.cwd.match(/(?<=\/)node_modules\/(.+)$/)?.[1],
+        l.cwd.match(/\.pnpm\/.+\/node_modules\/(.+)$/)?.[1]
```
danielroe, line 25: It looks like anything this regexp matches should be matched by the one above as well - what is the difference this brings?

danielroe:
Would you be able to provide a reproduction or more info on what was not working previously? 🙏

packages/nuxt/src/core/nitro.ts
yassilah: yes but the matched name isn't the name as pnpm will have both a `node_modules/xxx/...` and a `.pnpm/xxx@0.0.0_yyy@0.0.0/...` folder

yassilah:
Of course, i'll try to make one now

yassilah:
Here it is: the exact same app extending one layer, which adds a Nitro plugin from a module.
Using npm: https://stackblitz.com/edit/nuxt-starter-gwcz9v
Using pnpm: https://stackblitz.com/edit/nuxt-starter-z6srch

As you should see, the pnpm example throws an error saying that "defineNitroPlugin" is not defined, while the npm example works fine

danielroe:
I think the bit that's changing things is the change to `nitro.options.imports.exclude` not to the regexp.

yassilah:
Actually you are right as well, the first
```
nitro.options.esbuild.options.exclude = [excludePattern];
```
doesn't seem to be useful in this example (though i'm almost certain I had an issue when not setting it in my case), but wouldn't those "pnpm node_modules files" be excluded as well from any further esbuild transformation/plugin?

danielroe (approved):

yassilah:
Thanks a lot for reviewing 😊

PR #9971, opened 2023-01-06
feat(imports): `imports:context` hook for unimport context

antfu:
### ❓ Type of change

### 📚 Description

Mainly for devtools to get the imports infomation. It could also be useful for other modules to do custom import injection if needed.

### 📝 Checklist


pi0 (approved):

packages/nuxt/src/imports/module.ts
```diff
@@ -34,14 +35,16 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
 
     // Create a context to share state between module internals
     const ctx = createUnimport({
-      presets,
-      imports: options.imports,
-      virtualImports: ['#imports'],
       addons: {
```
danielroe, line 38:
```suggestion
      ...options,
      addons: {
```

packages/nuxt/src/imports/module.ts
```diff
@@ -34,14 +35,16 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
...
-        vueTemplate: options.autoImport
-      }
+        vueTemplate: options.autoImport,
+        ...options.addons
```
danielroe, line 40: as we are overriding addons with `...options` at the end, it looks like this line will never be applied.

packages/nuxt/src/imports/module.ts
```diff
@@ -34,14 +35,16 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
...
+      },
+      ...options,
```
danielroe, line 42: as we are overriding addons with `...options` at the end, it looks like this line will never be applied.

```suggestion
        ...options.addons
      },
```

danielroe (approved):

PR #9983, opened 2023-01-06
feat(nuxt): support fragment identifier with `<NuxtLink>`

lihbr:
### 🔗 Linked issue

#5561

### ❓ Type of change

### 📚 Description

This PR adds support for fragment identifier (`<a href="#foo">`) to `<NuxtLink>` by treating links starting with a hash as external.
Vue Router and the scroll behavior are handling those for other pages but not for the current one (out of scope I guess).

### 📝 Checklist

   - _Not sure if there's a specific need, I updated the example_

Let me know if anything ☺️


danielroe (approved):

pi0:
I'm wondering if we should router type to not be hashed router (opt-in feature checking from router options). Also any other custom behavior RouterLinks adds for links on same route with #

lihbr:
Actually, I'm confused, looks like Vue Router now supports fragment identifier(?) https://stackblitz.com/edit/nuxt-framework-wjrvwo?file=pages/index.vue (hash link at the end, make the window narrow to get scroll) But I can't find any mention of it in the changelog 🤔 So maybe this is just unneeded now? (or maybe I'm looking at the wrong thing)

---

_Original thoughts_

> I'm wondering if we should router type to not be hashed router (opt-in feature checking from router options).

I thought about this too, but when you use hashed router, you're still making links with `to="/foo"` (not hash-prefixed), so using this feature with a hashed router makes no real sense(?) Happy to explicitly disable it though, let me know.

> Also any other custom behavior RouterLinks adds for links on same route with #

I don't think so? Afaik it's not doing anything today with

pi0:
> you're still making links with to="/foo" (not hash-prefixed), 

Ideally yes. I still saw many projects (in nuxt2) they use identical (hash prefixed) `to` prop and it should be supported via RouterLink. 

> Afaik it's not doing anything today with

Was wondering if emits an event or causes **reactivity** to be triggered for `useRoute()` otherwise should be good :)

lihbr:
OK, makes sense for hash router, will support! But do we still want to more forward with it considering ✨ *recent* ✨ findings 🤔

> Actually, I'm confused, looks like Vue Router now supports fragment identifier(?) [stackblitz.com/edit/nuxt-framework-wjrvwo?file=pages/index.vue](https://stackblitz.com/edit/nuxt-framework-wjrvwo?file=pages/index.vue) (hash link at the end, make the window narrow to get scroll) But I can't find any mention of it in the changelog 🤔 So maybe this is just unneeded now? (or maybe I'm looking at the wrong thing)

lihbr:
As per Eduardo verification (https://github.com/nuxt/insiders/discussions/39#discussioncomment-4676689), this is indeed not needed anymore, therefore I'll be closing this PR.

Thanks for the prompt review Daniel, Pooya, and sorry for putting up something that wasn't needed anymore~

Onwards o/

PR #10017, opened 2023-01-09
fix(nuxt): allow overriding lower layer composables

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/9871

### ❓ Type of change

### 📚 Description

This PR adds a priority to composables scanned from directories so that higher priority layers can override lower priority layers.

### 📝 Checklist

### DIFF
--- a/packages/nuxt/src/imports/module.ts
+++ b/packages/nuxt/src/imports/module.ts
@@ -72,7 +72,7 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
     addVitePlugin(TransformPlugin.vite({ ctx, options, sourcemap: nuxt.options.sourcemap.server || nuxt.options.sourcemap.client }))
     addWebpackPlugin(TransformPlugin.webpack({ ctx, options, sourcemap: nuxt.options.sourcemap.server || nuxt.options.sourcemap.client }))
 
-    const priorities = [...nuxt.options._layers].reverse().map((layer, i) => [layer.config.srcDir, i] as const).sort(([a], [b]) => b.length - a.length)
+    const priorities = nuxt.options._layers.map((layer, i) => [layer.config.srcDir, -i] as const).sort(([a], [b]) => b.length - a.length)
 
     const regenerateImports = async () => {
       ctx.clearDynamicImports()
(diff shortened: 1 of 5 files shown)


packages/nuxt/src/imports/module.ts
```diff
@@ -72,11 +72,17 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
     addVitePlugin(TransformPlugin.vite({ ctx, options, sourcemap: nuxt.options.sourcemap.server || nuxt.options.sourcemap.client }))
     addWebpackPlugin(TransformPlugin.webpack({ ctx, options, sourcemap: nuxt.options.sourcemap.server || nuxt.options.sourcemap.client }))
 
+    const priorities = [...nuxt.options._layers].reverse().map((layer, i) => [layer.config.srcDir, i] as const).sort(([a], [b]) => b.length - a.length)
```
pi0, line 75: Sorting by `srcDir` length can cause a layer in development and shorter path (`/workpace/layer`) has more priority than main project (`/workspace/my/project`). Also giget _might_ clone remote layers into `/tmp` for example that has similar length issues.

What about checking with includes before compare by length or simply use second [,i] ?

packages/nuxt/src/imports/module.ts
```diff
@@ -72,11 +72,17 @@ export default defineNuxtModule<Partial<ImportsOptions>>({
...
+
     const regenerateImports = async () => {
       ctx.clearDynamicImports()
       await ctx.modifyDynamicImports(async (imports) => {
         // Scan composables/
-        imports.push(...await scanDirExports(composablesDirs))
+        const composableImports = await scanDirExports(composablesDirs)
+        for (const i of composableImports) {
+          i.priority = i.priority || priorities.find(([dir]) => i.from.startsWith(dir))?.[1]
```
pi0, line 83: What do you think to assign a negative peririty (most important -1, more layers -2, -3, etc). This way it won't conflict with user assigned peririties

packages/nuxt/src/imports/module.ts
danielroe: the _priority_ isn't set by length but by layer order.

PR #10046, opened 2023-01-11
fix(nuxt): include components `runtime` dir in build output

danielroe:
### ❓ Type of change

### 📚 Description

`server-component` in the `runtime` dir was not being generated/included in the final bundle

### 📝 Checklist

### DIFF
--- a/packages/nuxt/build.config.ts
+++ b/packages/nuxt/build.config.ts
@@ -12,6 +12,7 @@ export default defineBuildConfig({
     ...[
       'core',
       'head',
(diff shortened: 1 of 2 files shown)


SirMishaa:
Hello @danielroe, thank you for this fix, we're going to use Nuxt at work and this is important for us. 

Do you have an estimate on when it will be live on the next version of Nuxt? We have a pretty big mob party in two weeks and having this particular fix would be super cool. 

Thanks

PR #10083, opened 2023-01-13
chore: link package readmes to central framework readme

danielroe:
### ❓ Type of change

### 📝 Checklist


pi0:
We could have it as a sandalone script also in `scripts/` it can use a list of mono-repo packages to copy the readme. However i think we would need to only copy it to main `nuxt` published package otherwise searching in NPM/Google shows all sub-packages (such as vite-builder) with a readme that looks like usage of nuxt .

.gitignore
```diff
@@ -3,6 +3,8 @@ node_modules
 jspm_packages
 
 package-lock.json
+packages/*/README.md
```
pi0, line 6: Part of the purpose of this README file is also to help contributors find out about the purpose of each directory in monorepo. By removing it, we might at least need to put it into the documentation (which is somehow less accessible for someone directly diving into the code)

danielroe: I agree, it should be in the documentation.

danielroe:
~> https://github.com/nuxt/nuxt/pull/18421

PR #10087, opened 2023-01-13
fix(nuxt): support deep assign on empty object for app config

Atinux:
### 🔗 Linked issue

No linked issue, but a reproduction: https://stackblitz.com/edit/github-mgmeto?file=app.config.ts,app.vue

Tested locally in playground with my fix and works.

### ❓ Type of change

### DIFF
--- a/packages/nuxt/src/app/config.ts
+++ b/packages/nuxt/src/app/config.ts
@@ -26,6 +26,7 @@ function deepAssign (obj: any, newObj: any) {
   for (const key in newObj) {
     const val = newObj[key]
     if (val !== null && typeof val === 'object') {
+      obj[key] = obj[key] || {}
       deepAssign(obj[key], val)
     } else {
       obj[key] = val


pi0 (approved):
💯

Tahul:
I have a small suggestion about `updateAppConfig` and the current update strategy.

We currently use `deepAssign` and `deepDelete` in sequence so when calling `updateAppConfig(content)`, the `content` becomes the new `appConfig`, without mutating the original object.

The issue with that is that `deepDelete` will inevitably delete all the missing keys from `content`, avoiding the possibility to make partial update of the `app.config` by sending only parts of it.

While this works very good, I think it would be amazing being able to send only partial updates to the app.config, without implying `deepDelete`.

Also, I don't think **deleting** keys from the `app.config` is safe or needed, as if the key is defined it's inevitable that features in the project somehow depend on these keys, and then removing them could lead to runtime throws.

pi0:
@Tahul Can you please open an issue with steps to reproduce? Indeed update shouldn't delete missing sub-keys too. BTW we initially did it because it was making strange behavior when a user removes an optional-option from main `app.config` and HMR was not reflecting that.

Atinux:
The `deepDelete` is only used in development actually for HMR: https://github.com/nuxt/framework/blob/fix/update-app-config/packages/nuxt/src/app/config.ts#L61

What are the steps that brings a bad DX?

Tahul:
> The `deepDelete` is only used in development actually for HMR: [`fix`/update-app-config/packages/nuxt/src/app/config.ts#L61](https://github.com/nuxt/framework/blob/fix/update-app-config/packages/nuxt/src/app/config.ts?rgh-link-date=2023-01-13T17%3A04%3A00Z#L61)
> 
> What are the steps that brings a bad DX?

Wow, my bad, sorry for not double-checking that.

I thought the same sequence occurred in HMR and in runtime call.

I was mentioning this lately while working on studio because we allow partial updates for tokens.config and I thought that same partial updates would break with `app.config`. We are going to switch between holding the **complete** version of the app.config file in `json` files to holding only the keys that has been modified from `studio` for performances purpose.

You can consider this closed, sorry 😓

danielroe (approved):

PR #9972, opened 2023-01-06
feat(nuxt): server-only components

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/framework/issues/648
closes #5688

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

#### Summary

This PR builds on https://github.com/nuxt/framework/pull/5689 to add support for registering and rendering server components with `.server.vue` suffix (implementing https://github.com/nuxt/framework/issues/648).

#### Usage

Standalone server components will always be rendered on the server. When their props update, this will result in a network request that will update the rendered HTML in-place.

Server components are currently experimental and in order to use them, you need to enable the 'component islands' feature in your nuxt.config:

```ts [nuxt.config.ts]
export default defineNuxtConfig({
  experimental: {
    componentIslands: true
  }
})
```

Now you can register server-only components with the `.server` suffix and use them anywhere in your application automatically.

```bash
| components/
--| HighlightedMarkdown.server.vue
```

```html
<template>
  <div>
    <!--
      this will automatically be rendered on the server, meaning your markdown parsing + highlighting
      libraries are not included in your client bundle.
     -->
    <HighlightedMarkdown markdown="# Headline" />
  </div>
</template>
```

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #10113, opened 2023-01-16
feat(nuxt): support server components with extracted payloads

danielroe:
### ❓ Type of change

### 📚 Description

This builds on https://github.com/nuxt/framework/pull/9972 to support server components in fully static websites. It does two things:
1. it emits standalone data for a server component, e.g. `~/.output/public/__nuxt_island/PureComponent:6LYrSuODRS`. This means that server components aren't necessarily tied to the page in which they are used.
2. it also embeds any server components within a page's payload. This means that when loading a page, the payload can be prefetched without needing to make further network calls when switching to the page.

There is scope for improving this, for example, it's not necessarily to include the HTML in the payload if we could prefetch it. But I don't think this PR binds us to that - it's just an implementation detail at the moment and I would suggest we switch to fully extracted server components once we have a way of prefetching them - maybe through something like https://github.com/nuxt/framework/pull/7805.

### 📝 Checklist


pi0:
This is a good idea to use asyncData to enable payload extraction but on the other hand, we would be unable to implement and support hybrid islands (with remote sources) in static websites.

As far as I know, we currently use dynamic `import()` and with unique URLs for each island props, the browser/ESM resolver is supposed to cache them for second navigation or rendering. Isn't it happening as expected?

danielroe:
Nice thought! Created an issue so we can track: https://github.com/nuxt/framework/issues/10119. I'll push a fix to this PR so we support both payload extraction _and_ runtime fetching when props are changed.

danielroe:
The issue is the waterfall - they aren't fetched until the page starts being loaded, at which point the delay the transition, whereas if they are in the payload then they can be prefetched ahead of time and the page transition is instant.

pi0:
This is kinda by design for islands in general islands to be individually loaded assets (like images). However they are expected to be pre-rendered if not it is probably a bug.

I guess we can have a separate implementation `nuxt-server-island` component for the specific purpose of server components then that are inlined into page payloads and optimized without the need to support remote. Reusing code and server logic as much as possible but optimize for their purpose.

danielroe:
Pushed a couple of improvements (and fixed the bug of not emitting island output during prerender) and updated the issue description. Let me know what you think 😊

pi0:
Thanks for the updates. However still, in full static mode with payload extraction, when two pages use the same islands/props, duplicate data would be downloaded (and prefetched) in the payload of each page. Copy `islands.vue` to `islands2.vue` in the fixture for example.

Is it possible that we enable `useAsyncData` behavior only for server-only `.server` components to leveraging full-static? While keeping `<NuxtIsland>` to network fetch and rely on browser as is and cache with individual chunks we can experiment with hybrid remote islands easier also. (Since `useAsyncData` is always cached in the payload.)

Regarding prerender chunks, I'm wondering why [this](https://github.com/nuxt/framework/blob/8892cccb06c594b61423a1e2e602cd2deec15af9/packages/nuxt/src/core/runtime/nitro/renderer.ts#L249) wasn't working?

pi0 (approved):

PR #10155, opened 2023-01-18
feat(nuxt): prefetch middleware/layouts + await layout loading

Atinux:
### ❓ Type of change

### 📚 Description

Hooked into `link:prefetch` to prefetch the layouts and middleware :rocket:

- Without: https://nuxt3-no-prefetch.surge.sh
- With: https://nuxt3-prefetch.surge.sh

I also forced thge prefetching of layout on route navigation (not needed to depend on Suspense) like I did on Nuxt 2, this way we don't change the page until the layout if fully resolved.

**Before (slow 3g)**:

https://user-images.githubusercontent.com/904724/213266458-6774d029-1e55-4e2d-9dd8-e72fbfd5ae26.mp4

**After (slow 3g)**:

https://user-images.githubusercontent.com/904724/213266476-e2e5486d-9a67-4de5-a7d4-1f4c861200f7.mp4


packages/nuxt/src/core/nuxt.ts
```diff
@@ -196,6 +196,9 @@ async function initNuxt (nuxt: Nuxt) {
     addPlugin(resolve(nuxt.options.appDir, 'plugins/debug'))
   }
 
+  // Add prefetching support for middleware & layouts
+  addPlugin(resolve(nuxt.options.appDir, 'plugins/prefetch.client'))
```
danielroe, line 200: This should probably only be added in the pages module route meta is not present with the universal router.

packages/nuxt/src/app/plugins/prefetch.client.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    for (const name of middleware) {
+      if (typeof namedMiddleware[name] === 'function') {
+        namedMiddleware[name]()
+      }
+    }
+
+    if (layout && typeof layouts[layout] === 'function') {
+      layouts[layout]()
+    }
```
danielroe, line 38: At the moment, this is only imported in `<NuxtLayout>` which means layout files are not bundled unless a layout is actually used. I wonder if we can move this elsewhere (maybe first use of `<NuxtLayout>` triggers hook registration?)

If not, I think it's a reasonable trade-off given the very nice improvement in this PR.

packages/nuxt/src/core/nuxt.ts
Atinux: Makes sense to move it in the pages modules

packages/nuxt/src/app/plugins/prefetch.client.ts
Atinux: You mean importing the build/layouts will make the creation of the chunks?

i think it’s fine if they are always created IMO

we could also check if we have layouts defined but NuxtLayout is never used (I experienced this, layouts but forgot to use the NuxtLayout component in app.vue and got no warnings)

danielroe: It's a nice idea! Only scenario I can think of where it would be a pain is when you have a theme that provides layouts but the end project doesn't want to use any of them. Or if `<NuxtLayout>` is only being used within a page and only on some of the site. (But still I think the warning would be fine.)

(I don't think it has to be in this PR though.)

danielroe (approved):

Luffyyy:
I don't know if it's related, but this looks similar https://github.com/nuxt/framework/issues/7388 just with nested routes instead of layouts and it's been like this for a while. Could you check if this pull fixes it or address it in general?

danielroe:
Testing on edge would be very welcome @Luffyyy 🙏

PR #10178, opened 2023-01-21
Update 5.transitions.md

jjaimealeman:
When I was learning to add Page Transitions, I failed to notice that each page requires a single containing root element. So I thought it would be wise to remind users to add that 👍


danielroe:
Would you mind remaking this PR against https://github.com/nuxt/nuxt ? I'm sorry, I think it must have crept in just as we were migrating the repositories...

